PR #18, opened 2017-01-23
fix dirname import

ofek:
`dirname` is still in [os.path](https://docs.python.org/3/library/os.path.html#os.path.dirname) in Python 3.


kennethreitz:
✨🍰✨

Issue #26, opened 2017-01-23
Run as a module with python -m pipenv


illume:
This is something you can do with virtualenv and pip. pipenv is just missing the __main__.py file to do so.

```
python3.6 -m venv anenv
. ./anenv/bin/activate
python -m pip install pygame
```

However pipenv can't run as a module.
```
$ python3.6 -m pipenv install
/usr/local/opt/python3/bin/python3.6: No module named pipenv.__main__; 'pipenv' is a package and cannot be directly executed
```

This is useful for running the script with different python interpreters in build scripts and such which use different python versions, or on different platforms where things like "pip3.6" do not exist.

kennethreitz:
i like this idea.

kennethreitz:
open to a PR!

illume:
Cooking something up then...

PR #28, opened 2017-01-23
python -m pipenv

illume:
This is for issue #26 

It should make `python -m pipenv` work.


illume:
Not sure how it should be documented.

kennethreitz:
✨🍰✨

PR #39, opened 2017-01-23
Better output for where command when no Pipfile

rajalokan:
Presently when there is no Pipfile inside a directory, running
`pip --where` gives
`Pipfile found at None. Considering this to be the project home.`.
This seems ambiguous and I've changed it to
```
$ pipenv --where
No Pipfile present at project home. Consider running `pipenv install`
first to automatically generate a Pipfile for you.
$
```


rajalokan:
This should fix https://github.com/kennethreitz/pipenv/issues/38

PR #45, opened 2017-01-23
Fixes terminal size not being set properly

spiliopoulos:
See at http://pexpect.sourceforge.net/pexpect.html the documentation
of interact. I include here for easy reference

  | Note that if you change the window size of the parent the SIGWINCH
  | signal will not be passed through to the child.

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -3,6 +3,10 @@ import json
 import os
 import sys
 import distutils.spawn
+if sys.version_info < (3, 3):
+    from backports.shutil_get_terminal_size import shutil_get_terminal_size
+else:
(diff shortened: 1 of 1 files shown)


kennethreitz:
this looks good!

spiliopoulos:
Although I have still not used this package extensively, it seems it will be really helpful.
Thank you for this excellent contribution.

kennethreitz:
did some cleanup

kennethreitz:
https://github.com/kennethreitz/pipenv/commit/e6f09f8c3bd143ee22e6eeed7ecc07eb5c4884b1

kennethreitz:
released!

PR #60, opened 2017-01-24
Use uncompiled version of pep508checker

### LINKED ISSUE #59: "pipenv check" checks against the host Python instance rather than the venv's
For example, if I am running `pipenv` under Python 2.7, and have a Pipfile with:
```
[requires]
python_version = "3.5"
```
`pipenv check` will fail even if the project's Python version is correct.

thirtyseven:
This ensures that it will work when running across different Python versions.

Fixes #59


kennethreitz:
thanks!

Issue #88, opened 2017-01-25
pipenv install causes TypeError


Jackevansevo:
I get the following issue attempting to install packages using Pipenv in a fresh environment on my machine. Using Python version 3.4.2

```
❯ pipenv --three
Creating a Pipfile for this project...
Creating a virtualenv for this project...
Running virtualenv with interpreter /usr/bin/python3
Using base prefix '/usr'
New python executable in /home/jwe/Documents/DjangoLibrary/.venv/bin/python3
Also creating executable in /home/jwe/Documents/DjangoLibrary/.venv/bin/python
Installing setuptools, pip, wheel...done.

Virtualenv location: /home/jwe/Documents/DjangoLibrary/.venv

DjangoLibrary git/master*
❯ pipenv shell

...

DjangoLibrary git/master*
(DjangoLibrary)❯ pipenv install django
Traceback (most recent call last):
  File "/home/jwe/.local/bin/pipenv", line 11, in <module>
    sys.exit(cli())
  File "/home/jwe/.local/lib/python3.4/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/home/jwe/.local/lib/python3.4/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/home/jwe/.local/lib/python3.4/site-packages/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/jwe/.local/lib/python3.4/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/jwe/.local/lib/python3.4/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/home/jwe/.local/lib/python3.4/site-packages/pipenv/cli.py", line 531, in install
    new_name = proper_case(old_name)
  File "/home/jwe/.local/lib/python3.4/site-packages/pipenv/cli.py", line 421, in proper_case
    parser.feed(r.content)
  File "/usr/lib/python3.4/html/parser.py", line 164, in feed
    self.rawdata = self.rawdata + data
TypeError: Can't convert 'bytes' object to str implicitly
```

mjalas:
I am having the same issue when trying to install nose and tox, when using python 3.6.0.

nateprewitt:
This is an issue in Python 3 because `HTMLParser` expects strings but `content` from Requests will always return bytes.

[...]

PR #96, opened 2017-01-25
pass unicode to parser instead of bytes

nateprewitt:
Quick fix for #88, passing `str`/`unicode` instead of `bytes` to the parser because Python 3 can't process `bytes`.

I chose Requests' `text` property instead of `decode('utf-8')` in the event we get an unexpected page encoding, but that may be overkill. There isn't a significant increase in overhead using this since `chardet` is already being loaded with Requests, and we're parsing the whole page.


kennethreitz:
thanks!

PR #99, opened 2017-01-26
Fix virtualenv activation failing with spaces in path

jessebraham:
Fixes issue kennethreitz/pipenv#51


nateprewitt:
Thanks @jessebraham!

PR #100, opened 2017-01-26
Remove unneeded assignment

FooBarQuaxx:


nateprewitt:
Good catch, thanks @FooBarQuaxx!

PR #102, opened 2017-01-26
Use list of arguments in calls to `pexpect.spawn`

### LINKED ISSUE #81: pexpect.spawn should use list of arguments
`pexpect.spawn` should use list of arguments instead of `str.format`.

```python
import pexpect

path = '/tmp/dir-without-spaces'
child = pexpect.spawn('ls {0}'.format(path))
print(child.read())
child.close()

path = '/tmp/dir with spaces'
child = pexpect.spawn('ls {0}'.format(path))
print(child.read())
child.close()

path = '/tmp/dir with spaces'
child = pexpect.spawn('ls', [path])
print(child.read())
child.close()
```

```
In [1]: run test.py
file1.txt  file2.txt

ls: cannot access '/tmp/dir': No such file or directory
ls: cannot access 'with': No such file or directory
ls: cannot access 'spaces': No such file or directory

file1.txt  file2.txt
```

https://github.com/kennethreitz/pipenv/blob/master/pipenv/cli.py#L626
https://github.com/kennethreitz/pipenv/blob/master/pipenv/cli.py#L667

hellerve:
This PR fixes #81 by using a list of arguments for calls to `pexpect.spawn` instead of a formateed string.

Cheers

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -700,9 +700,8 @@ def shell(three=None):
     terminal_dimensions = get_terminal_size()
 
     c = pexpect.spawn(
-            "{0} -i'".format(
-                shell
-            ),
+            shell,
+            ["-i"],
             dimensions=(
                 terminal_dimensions.lines,
                 terminal_dimensions.columns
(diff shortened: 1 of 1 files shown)


hellerve:
Will look into Travis fails.

hellerve:
All should be swell now.

nateprewitt:
It does indeed look to be swell :) Thanks @hellerve!

PR #105, opened 2017-01-26
Support uninstalling of dev packages. Closes #103

### LINKED ISSUE #103: uninstall --dev

timofurrer:
Use:

```sh
pipenv uninstall pytest --dev
```

to uninstall `pytest` from the `[dev-packages]` section as described in #103

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -92,6 +92,7 @@ def ensure_project(three=None):
     ensure_pipfile()
     ensure_virtualenv(three=three)
 
+
 def ensure_proper_casing():
     """Ensures proper casing of Pipfile packages, writes to disk."""
     p = project.parsed_pipfile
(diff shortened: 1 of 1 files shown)


nateprewitt:
Thanks for taking this on @timofurrer!

One thing that may need to be addressed here is how we handle `pipenv uninstall --dev`. I think it's reasonable to expect that syntax will only remove packages from dev-packages. With this patch though, `pipenv uninstall --dev` removes **all** packages, identically to `pipenv uninstall`.

We should be able to access the full `dev-packages` list from `project.parsed_pipfile` and only uninstall those.

Something along the lines of this perhaps?

```python
    # Un-install all dependencies, if none was provided.
    if package_name is False:
        if not dev:
            click.echo(crayons.yellow('No package provided, un-installing all packages.'))
            do_purge(allow_global=system)
            sys.exit(1)
        elif 'dev-packages' in project.parsed_pipfile:
            package_names = project.parsed_pipfile['dev-packages']
        else:
            click.echo(crayons.yellow('No dev-packages to uninstall.'))
            sys.exit(1)
```

timofurrer:
Thanks for your feedback. I've just added a commit.

kennethreitz:
✨🍰✨

PR #110, opened 2017-01-26
Use clicks feature for terminal colors

timofurrer:
Don't get me wrong @kennethreitz I really do like crayons. It has a much better API than click has for the colors. 

However, my situation is the following: I work with Python in a regulated environment which basically means every third-party dependency has to be documented, validated and maintained (mirroring) (for up to 15 years). Especially for a project like `pipenv` (which might become very essential to the Python ecosystem, I hope) I'd love to see it with as few dependencies as possible but with as much that no code has to be duplicated. In the `crayons` case ... `click` already comes with support for terminal colors so it would be very cheap to use those.

And as I said, I'd much rather use crayons if it would only be about the API - but it isn't. Think about it ;)

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -1,12 +1,11 @@
+[packages]
+"backports.shutil_get_terminal_size" = "*"
+toml = "*"
+requirements-parser = "*"
+click = "*"
+"delegator.py" = ">=0.0.6"
+parse = "*"
+requests = "*"
+
 [dev-packages]
 pytest = "*"
-
-[packages]
-crayons = "*"
-requirements-parser = "*"
-"delegator.py" = ">=0.0.6"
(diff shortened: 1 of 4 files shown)


kennethreitz:
sticking to crayons, but i appreciate the work!

timofurrer:
For what reason if I may ask?

kennethreitz:
because i published the library just for this project :)

kennethreitz:
I'd much rather see us vendor it in if needed.

kennethreitz:
As we would then do with all dependencies.

timofurrer:
> because i published the library just for this project :)

Well, yeah, that would be too bad to remove it.

Vendor in sounds look a feasible plan ;) Thanks!

Issue #98, opened 2017-01-25
ModuleNotFoundError when running installed packages


jacebrowning:
I am seeing this issue with lots of packages that have a command-line entry point (at least `pytest` and `honcho`):

```
$ pipenv --version
pipenv, version 3.2.0

$ pipenv install honcho
Installing honcho...
Collecting honcho
  Using cached honcho-0.7.1-py2.py3-none-any.whl
Installing collected packages: honcho
Successfully installed honcho-0.7.1

Adding honcho to Pipfile's [packages]...

$ pipenv run honcho
Traceback (most recent call last):
  File "<root>/.venv/bin/honcho", line 7, in <module>
    from honcho.command import main
ModuleNotFoundError: No module named 'honcho'
```

nateprewitt:
Hmm, I'm actually unable to reproduce this on macOS 10.12.2 with a standard bash shell. @jacebrowning What version of python and shell are you using? Could you also provide what `pipenv run pip list` returns for me? Thanks!

jacebrowning:
I'm running ZSH and Python 3.6.0 on macOS 10.12.2.

```
$ pipenv run pip list
DEPRECATION: The default format will switch to columns in the future. You can use --format=(legacy|columns) (or define a format=(legacy|columns) in your pip.conf under the [list] section) to disable this warning.
appdirs (1.4.0)
honcho (0.7.1)
httpie (0.9.9)
packaging (16.8)
pip (9.0.1)
py (1.4.32)
Pygments (2.2.0)
pyparsing (2.1.10)
pytest (3.0.6)
requests (2.13.0)
setuptools (34.0.2)
six (1.10.0)
wheel (0.29.0)
```

nateprewitt:
So I've tried to recreate this with the same dependencies both locally and on Travis without success. I think there may either be a conflict on your path that's causing problems, or this may be something specific to ZSH.

Have you tried running `pipenv uninstall` to purge your pipenv and starting over? Are you getting an identical `ModuleNotFoundError` for pytest?

Without a consistent repro, I'm not sure much more can be done to debug this right now.

jacebrowning:
Yes, I get the same error with `pytest`:

```
$ pipenv uninstall
No package provided, un-installing all packages.
Found 10 installed package(s), purging...
Uninstalling appdirs-1.4.0:
  Successfully uninstalled appdirs-1.4.0
Uninstalling honcho-0.7.1:
  Successfully uninstalled honcho-0.7.1
[...]

PR #113, opened 2017-01-26
Prevent invalid shebangs with Homebrew Python

jacebrowning:
Fixes https://github.com/kennethreitz/pipenv/issues/98

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -440,6 +440,8 @@ def do_init(dev=False, requirements=False, skip_virtualenv=False, allow_global=F
 
 
 def pip_install(package_name=None, r=None, allow_global=False):
+    # Prevent invalid shebangs with Homebrew-installed Python: https://bugs.python.org/issue22490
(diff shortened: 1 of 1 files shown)


nateprewitt:
Glad you got to the bottom of this @jacebrowning! One thing that may be important here is to make this toggleable rather than always forcing the stripping. Is it possible someone is depending on this variable and `pipenv` will still work, or will `pipenv` always fail when this is present?

jacebrowning:
@nateprewitt From my testing, the only behavior this variable seems to control is the contents of the shebang line in installed scripts. I believe that pipenv's logic is already correctly locating the Python binary in the virtual environment, thus the value of this variable can be ignored.

jacebrowning:
It looks like tox did the exact same thing: https://bitbucket.org/hpk42/tox/pull-requests/118/fix-__pyvenv_launcher__-problem/diff

volkanunsal:
Hello, there may be a reversion. I'm seeing this exact issue in version 0.5.0. The same one described in #98.

PR #115, opened 2017-01-26
fixing OrderedDict

nateprewitt:
This should fix the current issues with the OrderdDict changes in the build. There's a certain amount of complexity in using OrderedDict in older versions of Python (specifically 2.6) that's already been handled in Requests. Since that is a hard dependency I leveraged it's OrderedDict implementation to spare us the headache. Hopefully that's OK.


kennethreitz:
great idea!

PR #121, opened 2017-01-27
Add pipfile dependency to Pipfile

timofurrer:
As a reminder: the `Pipfile.lock` has to be ordered as well. I'll have a look at it later.


nateprewitt:
The issue is that Pipfile.lock is copied from the Pipfile back into a dictionary. That functionality is in the Pipfile repo, rather than pipenv, so it'll need to be fixed over there I think.

nateprewitt:
Also thanks for this @timofurrer!

timofurrer:
> The issue is that Pipfile.lock is copied from the Pipfile back into a dictionary. That functionality is in the Pipfile repo, rather than pipenv, so it'll need to be fixed over there I think.

Okay, I'll have a look at it there.

PR #123, opened 2017-01-27
Use tox for multi version testing

FooBarQuaxx:


Makefile
```diff
@@ -1,5 +1,11 @@
 run-tests:
 	pipenv run pytest test_pipenv.py
+run-tox:
+	pipenv run tox
```
timofurrer: I'm not sure if it's clever to use the module-under-test to actually run the tests. However, on the other side who is going to ensure that `tox` is actually installed ...

Makefile
```diff
@@ -1,5 +1,11 @@
...
 init:
 	pip install pipenv
 	pipenv install --dev
+clean:
```
timofurrer: What about `__pycache__`?

Makefile
FooBarQuaxx: The same thing is done on the previous makefile target (run-tests). I thought if it's good enough to run pytest, then it should be ok for running tox.

Makefile
FooBarQuaxx: I can get this part fixed easily.

kennethreitz:
We're currently relying on travis-ci for this.

PR #134, opened 2017-01-28
only write file when changed

nateprewitt:
We write the Pipfile at the beginning of every run regardless of if it changed or not. Let's only record changes.


kennethreitz:
✨🍰✨

PR #159, opened 2017-01-30
Verify if PIPENV_COLORBLIND env variable is set to 1

timofurrer:


pipenv/cli.py
```diff
@@ -37,7 +37,7 @@
 click_completion.init()
 
 # No color mode, for unfun people.
-if 'PIPENV_COLORBLIND' in os.environ:
+if os.environ.get('PIPENV_COLORBLIND') == '1':
```
FooBarQuaxx: For sake of being complete maybe support other known 1-ish values ( on, yes, 1, true).

nateprewitt: While this it's possible the user could set the value to these, there's a whole host of options that *could* be used if we open it up to that. What about y, n, TrUE, 0x01? It's not really worth the time trying to guess what people *might* use here. `1` is the standard "yes" flag and should be sufficient.

nateprewitt:
Thanks for this @timofurrer!

kennethreitz:
i think we should just check if it's set at all

nateprewitt:
My thinking was it's unintuitive to specify `PIPENV_COLORBLIND=0` and have that activate colorblind mode. We also set PIPENV_ACTIVE to 1 but don't check that it's 1 or 0 though. It's probably best to be consistent across how we used variables. If you'd rather use the existence as a flag, then let's do it :)

kennethreitz:
already done :)

timofurrer:
> i think we should just check if it's set at all

What's the reason for that?

It's very cumbersome to temporarily unset a variable. If the value has to be "1" you can just set the variable for this single call like:

```
export PIPENV_COLORBLIND=1
PIPENV_COLORBLIND=0 pipenv
```

And it's **very** confusing if things like this work: `PIPENV_COLORBLIND=0`. Isn't it?

We should do the same for `PIPENV_ACTIVE` as well.

kennethreitz:
this is a dumb feature anyway, i don't want to give it more thought than it deserves.

timofurrer:
> this is a dumb feature anyway, i don't want to give it more thought than it deserves.

It's definitely not dumb at all .. What about piping the output to a log file and read it in an editor? What about parsing the output for what ever reason? Or piping/redirecting it to other tools?

(That's my use-cases .. and I'd happy too if `pipenv` would check if the output is being redirected or not.)

timofurrer:
And yeah, if it's actually being handled by `crayons` then it's indeed dumb..

kennethreitz:
it is handled by crayons for piping :)

kennethreitz:
this is just for people who don't want colors in their cli

PR #164, opened 2017-01-30
use specified python version over --three/--two

nateprewitt:
With the introduction of `--python` commands in the form `pipenv install --three --python=3.6.0` became possible. With the current ordering, `python3` will be chosen rather than the more specific `python3.6.0` command.

While users **should** only use one of these parameters, we want to give precedence to the more specific parameter. This way a `python3` binding that doesn't match 3.6.0 isn't used, creating a possibly unintended environment.


kennethreitz:
✨🍰✨

PR #166, opened 2017-01-30
Support all version control systems that pip does

vilhalmer:
Ran into needing SVN immediately upon trying this out, so I figured I'd finish that TODO for you. :)

### DIFF
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -90,19 +90,18 @@ def convert_deps_to_pip(deps, r=True):
         if vcs:
             extra = '{0}+{1}'.format(vcs, deps[dep][vcs])
 
-            if extra:
-                # Support for @refs.
-                if 'ref' in deps[dep]:
-                    extra += '@{0}'.format(deps[dep]['ref'])
+            # Support for @refs.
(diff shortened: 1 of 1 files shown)


pipenv/utils.py
```diff
@@ -83,23 +83,26 @@ def convert_deps_to_pip(deps, r=True):
...
-            else:
-                dep = ''
+        # Support for version control
+        maybe_vcs = [vcs for vcs in ('git', 'svn', 'hg', 'bzr') if vcs in deps[dep]]
+        vcs = maybe_vcs[0] if maybe_vcs else None
+
+        if vcs:
+            extra = '{0}+{1}'.format(vcs, deps[dep][vcs])
+
+            if extra:
```
nateprewitt: Is there a reason for adding this conditional? Given `extra = '{0}+{1}'.format(vcs, deps[dep][vcs])`, when can `extra` be a non-truthy value?

kennethreitz:
looks good!

nateprewitt:
Thanks for the work @vilhalmer!

vilhalmer:
Glad to be of help!

PR #175, opened 2017-02-01
Improve dashed version parsing

natpen:
Hi! Hope you don't mind that I wrote some quick code for this. It was breaking on the dash in version 3.2.3-2 of functools32, which is preventing me from sharing some pipenv code with colleagues.

With this PR, instead of truncating the ends of versions containing dashes, it will keep them intact. Also I added some tests for the parse_download_fname function.

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -2,6 +2,7 @@
 import codecs
 import json
 import os
+import re
 import sys
 import distutils.spawn
 import shutil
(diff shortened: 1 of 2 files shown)


natpen:
Hi! Hope you don't mind that I wrote some quick code for this. It was breaking on the dash in version 3.2.3-2 of functools32, which is preventing me from sharing some pipenv code with colleagues.

kennethreitz:
✨🍰✨

PR #177, opened 2017-02-01
Improve parse_download_fname performance

natpen:
This is a small performance-oriented followup to #175 which should ~halve the amount of regex being run!

Once the beginning of the version string has been found, the flag is no longer modified, and thus there's no reason to keep running the regex.


kennethreitz:
thanks!

PR #192, opened 2017-02-02
Add PIPENV_NOSPIN option for cleaner build logs

jacebrowning:
I would like to be able to clean up this noise on CircleCI:

screen shot 2017-02-02 at 1 02 59 pm

### DIFF
--- a/HISTORY.txt
+++ b/HISTORY.txt
@@ -1,3 +1,5 @@
+unreleased:
+- Disable spinner by setting PIPENV_NOSPIN=1 environment variable.
 3.3.4:
  - Fix PIPENV_VENV_IN_PROJECT mode.
  - Fix PIPENV_SHELL_COMPAT mode.
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -91,6 +91,8 @@ will detect it.
 
     - ``PIPENV_COLORBLIND`` — Disable terminal colors, for some reason.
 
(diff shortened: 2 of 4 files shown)


jacebrowning:
@kennethreitz Thoughts on this?

It probably makes more sense to add a `spinner.disable()` API in blindspin, but this is start...

nateprewitt:
I can see the desire to have cleaner logs, but I don't think we need another variable specifically for the spinner. It's more surface area for the user to have to worry about for a pretty specific edge case.

If Kenneth wants to entertain this, I think there are two possibly simpler options:
 * a `PIPENV_CLEAN` (or `PIPENV_NOFUN` 😉)  that replaces and encapsulates both `PIPENV_COLORBLIND` and `PIPENV_NOSPIN`
* placing the spinner inside of a `if not bare` check to actually leverage our `--bare` argument.

kennethreitz:
colorblind can disable the spinnner too.

jacebrowning:
@kennethreitz But I love the colors! 😄

I'll push a change that uses the same environment variable for you to consider.

kennethreitz:
this change is sane, i'll keep it.

kennethreitz:
✨🍰✨

kennethreitz:
released!

kennethreitz:
might be refactored in the future

PR #205, opened 2017-02-06
Cleanup: IO and formatting

imanhodjaev:
Since `open` by default in read mode I removed `'r'` arguments, also made some minor cleanups.

Thanks.

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -209,7 +209,7 @@ def do_install_dependencies(dev=False, only=False, bare=False, requirements=Fals
     else:
         if not bare:
             click.echo(crayons.yellow('Installing dependencies from Pipfile.lock...'))
-        with open(project.lockfile_location, 'r') as f:
+        with open(project.lockfile_location) as f:
             lockfile = json.load(f)
 
     # Install default dependencies, always.
@@ -224,7 +224,7 @@ def do_install_dependencies(dev=False, only=False, bare=False, requirements=Fals
 
     # --requirements was passed.
     if requirements:
-        with open(deps_path, 'r') as f:
+        with open(deps_path) as f:
             click.echo(f.read())
             sys.exit(0)
 
@@ -536,7 +536,7 @@ def do_init(dev=False, requirements=False, skip_virtualenv=False, allow_global=F
     if project.lockfile_exists:
 
         # Open the lockfile.
-        with codecs.open(project.lockfile_location, 'r') as f:
+        with codecs.open(project.lockfile_location) as f:
             lockfile = json.load(f)
 
         # Update the lockfile if it is out-of-date.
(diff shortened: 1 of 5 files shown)


pipenv/environments.py
```diff
@@ -20,7 +20,7 @@
 
 # User-configuraable max-depth for Pipfile searching.
 # Note: +1 because of a temporary bug in Pipenv.
-PIPENV_MAX_DEPTH = int(os.environ.get('PIPENV_MAX_DEPTH', '3')) + 1
+PIPENV_MAX_DEPTH = int(os.environ.get('PIPENV_MAX_DEPTH', 3)) + 1
```
nateprewitt: The default type of `os.environ` values should be `str` and we're always casting the value to `int`. I don't think there's any particular advantage to having varying types here.

pipenv/cli.py
```diff
@@ -209,7 +209,7 @@ def do_install_dependencies(dev=False, only=False, bare=False, requirements=Fals
     else:
         if not bare:
             click.echo(crayons.yellow('Installing dependencies from Pipfile.lock...'))
-        with open(project.lockfile_location, 'r') as f:
+        with open(project.lockfile_location) as f:
```
nateprewitt: It looks like this, and possibly other areas are being read in as bytes instead of strings now. This is breaking the build for Python 3.3 and 3.4.

pipenv/cli.py
```diff
@@ -678,7 +678,7 @@ def gen(out):
 
 def easter_egg(package_name):
     if package_name in ['requests', 'maya', 'crayons', 'delegator.py' 'records', 'tablib']:
-        click.echo('P.S. You have excellent taste! ✨ 🍰 ✨')
+        click.echo(u'P.S. You have excellent taste! ✨ 🍰 ✨')
```
nateprewitt: Can we verify that this works as expected in both Python 2 and 3?

pipenv/environments.py
imanhodjaev: @nateprewitt thank you for clarification I will put it back, I just thought that since it is intended to be a number we could have as a number for default value.

pipenv/cli.py
imanhodjaev: @nateprewitt I think I found why it failed. The reason is that I removed `'r'` argument from `codecs.open` at line `540` of `cli.py`

imanhodjaev: Also according to Python documentation at https://docs.python.org/3/library/functions.html#open `It defaults to 'r' which means open for reading in text mode.`

pipenv/cli.py
imanhodjaev: @nateprewitt Yes it works on 2.6, 2.7 and 3.6

kennethreitz:
✨🍰✨

imanhodjaev:
✨Yuhuuu thank you✨

PR #206, opened 2017-02-06
Enable basic functionality for Windows OS

AleksanderGondek:
This pull request aims to bring basic compatibility with Windows OS (as I love pipenv but need it to work across windows/linux).
Currently I believe I have made all commands except for `shell` to work. 

I know the approach taken is not ideal (comments are welcomed), however this unblocks any workflow that requires projects to be created with venv both on Windows and Linux. 

Caveats: Pipfile available from pypi needs to be upgraded, otherwise it will return invalid path on Windows ([relevant issue]( https://github.com/pypa/pipfile/issues/67))

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -14,6 +14,7 @@ import click_completion
 import crayons
 import delegator
 import parse
+from pexpect.popen_spawn import PopenSpawn
 import pexpect
 import requests
 import pipfile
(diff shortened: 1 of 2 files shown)


kennethreitz:
i think i'm going to pass on this for now.

AleksanderGondek:
:(

Issue #203, opened 2017-02-06
Compatibility for packages with "irregular" version numbers


adrianliaw:
Currently, there is a restriction that the version numbers of installed libraries have to be matching `^[\d]+\.[\d]+.*` (starting with this regex)

I'm trying to install pdfminer.six in my project, but due to the "irregular" version number that looks something like `20160614`, pipenv seemed to be failing to catch the version number and failed to write it properly into the lock file.
And it ended up looking like this:
```
"pdfminer.six": {
    "hash": "sha256:8897b7ca5e94329aef5b862e501834ea2c0460c9940c904fdbd07dba297c2553",
    "version": "=="
},
```
Which causes an installation failure while running `pipenv install`, that shows `Could not find a version that satisfies the requirement pdfminer.six==`

A way to fix this is to simply use another way to determine the version number when installing a package. That is, instead of looping through every component inside the downloaded distribution's file name separated by "-" and detect the version number using regex, we simply exclude the package name itself and file extension from the download fname, the leaving middle part is just the version number.

Will this approach be more suitable? Or is there a special reason for using the original approach?

nateprewitt:
Thanks for raising this issue @adrianliaw. I agree the regex isn't tolerant of non-standard version numbers, but has this far been the most universal solution. We originally had a solution similar to your suggestion, but packages aren't guaranteed to start with their own name and they often use - and _ interchangeably in projects, even from one version to the next.

I do think this is an area that could use improvement but we need a solution that is more permissive of different version number patterns without mangling the majority of cases pipenv covers now.

adrianliaw:
@nateprewitt Thanks for the explanation! But it seems like the file names of downloaded distributions such as tarballs and wheels, always start with the form of PKGNAME-VERSION, even though the module name you're going to use to import may not exactly be PKGNAME, but PKGNAME is exactly what you type in when you're trying to install. And according to [PEP426](https://www.python.org/dev/peps/pep-0426/#name), the package name MUST be case insensitive and MUST have no distinguishing between hyphens and underscores, so it wouldn't affect that much.

In addition, I also think that it might be better to use the official registered name of packages inside Pipfile, instead of using what you've typed in when you use `pipenv install PKGNAME`. e.g. If you used `pipenv install jinja2` to install Jinja2, pipenv should write `Jinja2` into Pipfile instead of `jinja2`. Running `pipenv install youtube-dl` should write `youtube_dl` instead of `youtube-dl` inside Pipfile.

[...]

PR #215, opened 2017-02-09
Use safe_name from pkg_resources for proper_case

adrianliaw:
As far as I understand, the purpose of `proper_case` is to find a somehow "normalized" name of a package. The current way for doing this is to hit the PyPI's simple package index page, and fetch the title text. Now, if we can figure out how exactly PyPI produces that "normalized" package name, we can simply apply the same rule inside pipenv, without exhaustively making requests and caching the results. So I've changed the `proper_case` function to exactly how [PyPI's simple package index produces the package name](https://github.com/pypa/pypi-legacy/blob/master/webui.py#L1082).

Please let me know if I misunderstood anything about the purpose of `proper_case`, or about the reasons behind checking proper names by asking PyPI.

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -19,6 +19,7 @@ import requests
 import pipfile
 from blindspin import spinner
 from requests.packages.urllib3.exceptions import InsecureRequestWarning
+from pkg_resources import safe_name
 
 from .project import Project
 from .utils import convert_deps_from_pip, convert_deps_to_pip, is_required_version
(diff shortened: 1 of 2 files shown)


kennethreitz:
I'm going to defer to @nateprewitt on this one.

nateprewitt:
Hey @adrianliaw, thanks for taking the time to put this together. While you are correct this *appears* to be the code generating the output for https://pypi.python.org/simple and https://pypi.org/simple, this isn't what we're actually getting. An example is https://pypi.python.org/simple/twisted. You'll note the uppercased `Twisted` which isn't what `pkg_resources.safe_name(name).lower()` will return.

The purpose of `proper_case` is to return a properly cased and formatted name for the project. PyPI supplies this in the title tag which is why we're accessing PyPI to get it. This proposed solution doesn't support that, but rather strips characters, replaces underscores with dashes and forces all names to be lower case. While this normalization may make comparisons slightly easier, it defeats the purpose of using `proper_case`.

`proper_case` also accounts for retrieving packages from outside of PyPI, such as a version control system which is not required to adhere to PEP426. This means this solution will strip possibly required characters from the package name and no longer be able to distinguish between the two methods of file retrieval.

For these reasons, I don't think we're going to be able to merge this patch. Fortunately, I pushed a solution for this last night to our refactor branch which should handle your version problem. #203 will hopefully be addressed in an official release soon. Thanks again for taking the time to try fixing this.

PR #217, opened 2017-02-10
Fix PyPI simple API URL

adrianliaw:
Follow up as a comment for #215, if the purpose of `proper_case` is to get the "official registered" name of the package on PyPI, then fetching the text on https://pypi.python.org/simple wouldn't help anything, it'll turn out to be the "normalized" package name.

@nateprewitt also referenced an example of [Twisted](https://pypi.org/simple/twisted/), which supposed to be "Twisted" instead of "twisted", but since pipenv is currently referencing pypi.python.org instead of pypi.org, it actually shows ["twisted" instead of "Twisted"](https://pypi.python.org/simple/twisted/), so it wouldn't fix the problem of "proper casing". I think hitting pypi.org/simple gives our expected result, right?

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -186,7 +186,7 @@
     "_meta": {
         "sources": [
             {
-                "url": "https://pypi.python.org/simple",
+                "url": "https://pypi.org/simple",
                 "verify_ssl": true
             }
         ],
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -588,7 +588,7 @@ def proper_case(package_name):
                 collected.append(data)
 
     # Hit the simple API.
-    r = requests.get('https://pypi.org/simple/{0}'.format(package_name))
+    r = requests.get('{0}/{1}/'.format(project.source['url'], package_name))
     if not r.ok:
         raise IOError('Unable to find package {0} in PyPI repository.'.format(crayons.green(package_name)))
 
--- a/pipenv/project.py
+++ b/pipenv/project.py
@@ -117,7 +117,7 @@ class Project(object):
             return json.load(lock)
 
     def create_pipfile(self):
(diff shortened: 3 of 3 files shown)


nateprewitt:
You're absolutely correct @adrianliaw, I realized this was a consistent conflict after I left that comment. I patched it immediately afterwards with a slightly different approach. 8ebcd1a changes the source we hit in `proper_case` to always be https://pypi.org/simple.

While I usually am not a fan of hardcoding things like this, the solution of changing our default doesn't actually solve the problem for anyone who previously created a Pipfile. It also doesn't handle people who actively choose to use the pypi.python.org/simple endpoint. We've already baked specifics of the the pypi endpoint pages into `proper_case`, so I don't think we're losing any more functionality by binding to a specific endpoint. It won't affect installation and fails gracefully since this is only a cosmetic feature at this point.

I'm only +0.5 on the approach I implemented though, and I can see only doing a best effort of trying to point people at the correct endpoint with defaults. I'd like to leave this open for now and see if @kennethreitz has opinions on direction here.

kennethreitz:
can you do a rebase?

adrianliaw:
@kennethreitz I've done a rebase, but since @nateprewitt also solved this problem in #218, here it's just the difference between hard-coding the PyPI simple index URL and changing the source url in Pipfile.lock

nateprewitt:
Since #218 is now merged, I think we're going to go with that approach so I'm going to close this for now. If you feel strongly there's value we're missing here @adrianliaw please let us know and we can discuss it further.

Thanks again for taking the time to put this together! We really do appreciate the effort.

adrianliaw:
@nateprewitt Awesome! I also don't think there's anything missing here, thanks for all the work!

PR #218, opened 2017-02-10
[WIP] Better package name handling, location caching and testing

nateprewitt:
This is an initial proposal to simply and fix our longest running issue of package name conflicts. Also included is an increase to test coverage and some minor refactoring to separate cli-specific code from functions to make testing easier.

As the title says, this is still a bit of a WIP, but I'd like to open it up for comment if anyone has thoughts here.

Changes:

 - Only use https://pypi.org/simple to resolve package names since pypi.python.org doesn't provide them.
 - Don't try to validate Pipfiles when the virtualenv hasn't been initialized (only relevant to new projects)
 - Reliably handle version string in filenames per PEP426
 - Avoid doing multiple file location lookups when trying to access a location
 - Avoid loading proper names list each time we check a package name
 - Tests (more would be nice)


kennethreitz:
@nateprewitt what's stopping us from merging this now?

nateprewitt:
I still have a few things to push up which aren't blockers.

I was waiting for your input on #217, which would require some rework here if merged and just wanted a second pair of eyes before merging a chunk this big.

kennethreitz:
everything looks good to me!

nateprewitt:
Let's do this then!

PR #222, opened 2017-02-11
Add extra-index-url support

arruda:
I've done the following changes:
Making `cli.proper_case` use a new method that runs through all sources in project.sources, and return the Request's return for the first one that hit OK, or neither did ok, then raise an exception.

Also doing something similar to `cli.pip_install` so that it will try to pip install for all the sources and return the first cmd that worked or the last that didn't work.

And changed `project.source` to `project.sources` and making it return a list as default instead of just the first source present.

And added some tests

### DIFF
--- a/tests/test_pipenv.py
+++ b/tests/test_pipenv.py
@@ -135,15 +135,14 @@ class TestPipenv():
 
         assert command == '{0}/bin/activate'.format(venv)
 
-    @patch('pipenv.project.Project.virtualenv_location', new_callable=PropertyMock(return_value='foo'))
     @patch('pipenv.project.Project.sources', new_callable=PropertyMock)
     @patch('delegator.run')
-    def test_pip_install_should_try_every_possible_source(self, mocked_delegator, mockec_sources, mocked_venv_location):
+    def test_pip_install_should_try_every_possible_source(self, mocked_delegator, mocked_sources):
         sources = [
             {'url': 'http://dontexistis.in.pypi/simple'},
             {'url': 'http://existis.in.pypi/simple'}
         ]
-        mockec_sources.return_value = sources
+        mocked_sources.return_value = sources
         first_cmd_return = Mock()
         first_cmd_return.return_code = 1
         second_cmd_return = Mock()
@@ -152,15 +151,14 @@ class TestPipenv():
         c = pip_install('package')
         assert c.return_code == 0
 
-    @patch('pipenv.project.Project.virtualenv_location', new_callable=PropertyMock(return_value='foo'))
     @patch('pipenv.project.Project.sources', new_callable=PropertyMock)
     @patch('delegator.run')
-    def test_pip_install_should_return_the_last_error_if_no_cmd_worked(self, mocked_delegator, mockec_sources, mocked_venv_location):
+    def test_pip_install_should_return_the_last_error_if_no_cmd_worked(self, mocked_delegator, mocked_sources):
         sources = [
             {'url': 'http://dontexistis.in.pypi/simple'},
             {'url': 'http://dontexistis.in.pypi/simple'}
         ]
-        mockec_sources.return_value = sources
+        mocked_sources.return_value = sources
         first_cmd_return = Mock()
         first_cmd_return.return_code = 1
         second_cmd_return = Mock()
@@ -170,15 +168,14 @@ class TestPipenv():
         assert c.return_code == 1
         assert c == second_cmd_return
 
-    @patch('pipenv.project.Project.virtualenv_location', new_callable=PropertyMock(return_value='foo'))
     @patch('pipenv.project.Project.sources', new_callable=PropertyMock)
     @patch('delegator.run')
-    def test_pip_install_should_return_the_first_cmd_that_worked(self, mocked_delegator, mockec_sources, mocked_venv_location):
+    def test_pip_install_should_return_the_first_cmd_that_worked(self, mocked_delegator, mocked_sources):
         sources = [
-            {'url': 'http://dontexistis.in.pypi/simple'},
-            {'url': 'http://dontexistis.in.pypi/simple'}
+            {'url': 'http://existis.in.pypi/simple'},
+            {'url': 'http://existis.in.pypi/simple'}
         ]
-        mockec_sources.return_value = sources
+        mocked_sources.return_value = sources
         first_cmd_return = Mock()
         first_cmd_return.return_code = 0
         second_cmd_return = Mock()
--- a/tests/test_pipenv.py
+++ b/tests/test_pipenv.py
@@ -1,12 +1,15 @@
 import os
 
+from mock import patch, Mock, PropertyMock
+
 import pytest
 import delegator
 import toml
 
 from pipenv.cli import (activate_virtualenv, ensure_proper_casing,
-    parse_download_fname, parse_install_output)
+    parse_download_fname, parse_install_output, pip_install)
 from pipenv.project import Project
+
 
 class TestPipenv():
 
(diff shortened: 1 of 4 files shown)


nateprewitt:
@arruda, it looks like there's some conflicts with recent merges. Could you rebase on top of those, there have been a few changes in how we handle `proper_case` and the tests have been moved to their own directory.

nateprewitt:
Actually, while we're here. Could you clarify for me how your private index is set up? Are you running a copy of [pypi-legacy](https://github.com/pypa/pypi-legacy). I don't know if we should be making assumptions about non-standard endpoints in `proper-names`. It may be best to leave it how it is and have it pass over names that can't be resolved from the official PyPI endpoint.

arruda:
Yeah, I just saw it, I was going to ask, how are you running the tests? Just to make sure I'm running them correctly

nateprewitt:
`py.test tests` should run the test suite. I'll update the Makefile to reflect that now.

arruda:
I just updated my master branch, and runned the tests (`pytest tests`) and got this error:
```
___________________________________________________________________ TestProject.test_pew_by_default ____________________________________________________________________

self = <tests.test_project.TestProject instance at 0x7ff309b53878>

    def test_pew_by_default(self):
        proj = pipenv.project.Project()
>       assert proj.virtualenv_location.endswith('.local/share/virtualenvs/pipenv')
E       assert False
E        +  where False = <built-in method endswith of str object at 0x7ff309b531e0>('.local/share/virtualenvs/pipenv')
E        +    where <built-in method endswith of str object at 0x7ff309b531e0> = '/home/myuser/.virtualenvs/pipenv'.endswith
E        +      where '/home/myuser/.virtualenvs/pipenv' = <pipenv.project.Project object at 0x7ff309ba4310>.virtualenv_location

tests/test_project.py:15: AssertionError
================================================================= 1 failed, 32 passed in 11.97 seconds =================================================================
```
I'm not sure what's the problem here, this was run in the master branch after updating it, not in my branch, or anything like that.

I think this might be because of how I'm running my dev env:
* created using pipenv --two
* activated using virtualenvwrapper: `workon pipenv`
* run tests

Maybe I'm missing something or this is not compatible with virtualenvwrapper?

nateprewitt:
You know what, I'm just going to pull that test because it doesn't provide a ton of value and will likely cause issues with different system configurations.

arruda:
Ok, I'll fix the rest that are breaking because of my changes

arruda:
@nateprewitt I had to change the behavior of some tests because of the difference in behavior from https://pypi.org/simple/ and how others urls deal with the package name, ex: https://pypi.python.org/simple/django/

arruda:
Sorry, I just saw your other message:

> Actually, while we're here. Could you clarify for me how your private index is set up? Are you running a copy of pypi-legacy?
> 
> I don't know if we should be making assumptions about non-standard endpoints in proper_case. Particularly with how we do caching. It may be best to leave it how it is and have it pass over names that can't be resolved from the official PyPI endpoint.

Actually I don't know if my private index is setup, I would need to check, but it has the same behavior as https://pypi.python.org/simple/. That is, it doesn't treat the names like https://pypi.org/simple/ does.

nateprewitt:
Sorry, @arruda, I probably should have clarified my previous comment further.

>I don't know if we should be making assumptions about non-standard endpoints in proper_case. Particularly with how we do caching. It may be best to leave it how it is and have it pass over names that can't be resolved from the official PyPI endpoint.

I don't know if we actually want to check every source for the proper name. The reason it was specified to be https://pypi.org/simple is for the exact issue you reverted in the test.

We can get the exact package name from pypi.org/simple but other endpoints (including pypi.python.org/simple) may return varied naming. `proper_case` is purely a best effort attempt to give a nice name, it doesn't have to succeed for installation to work. If the `proper_name` chosen varies based on the order of sources, I'd consider that undesired behaviour.

For that reason, and the fact we have no guarantee a private PyPI endpoint will return anything useful, I think it's unnecessary, and probably unhelpful overhead, to check every possible endpoint.

I think it's a good idea to check all provided sources, but it's probably best to limit the patch to that.

arruda:
Hummm... I see, I'll change that back then. In my head it would stop the installation if it couldn't find the proper_case. =)

nateprewitt:
>Actually I don't know if my private index is setup, I would need to check, but it has the same behavior as https://pypi.python.org/simple/. That is, it doesn't treat the names like https://pypi.org/simple/ does.

Yeah, that's what I thought, I believe you're using the pypi-legacy endpoint setup. Let's just implement the `pip_install` and `sources` changes for now :)

arruda:
Ok, now I think it's done. And sorry for not paying much attention in your previews comment =S

tests/test_pipenv.py
```diff
@@ -131,3 +134,56 @@ def test_activate_virtualenv_no_source(self):
         venv = Project().virtualenv_location
 
         assert command == '{0}/bin/activate'.format(venv)
+
+    @patch('pipenv.project.Project.virtualenv_location', new_callable=PropertyMock(return_value='foo'))
+    @patch('pipenv.project.Project.sources', new_callable=PropertyMock)
+    @patch('delegator.run')
+    def test_pip_install_should_try_every_possible_source(self, mocked_delegator, mockec_sources, mocked_venv_location):
```
nateprewitt: I think this should be `mocked_sources`.

tests/test_pipenv.py
```diff
@@ -131,3 +134,56 @@ def test_activate_virtualenv_no_source(self):
         venv = Project().virtualenv_location
 
         assert command == '{0}/bin/activate'.format(venv)
+
+    @patch('pipenv.project.Project.virtualenv_location', new_callable=PropertyMock(return_value='foo'))
```
nateprewitt: What purpose is the `mocked_venv_location` serving here?

tests/test_pipenv.py
```diff
@@ -131,3 +134,56 @@ def test_activate_virtualenv_no_source(self):
...
+    @patch('pipenv.project.Project.sources', new_callable=PropertyMock)
+    @patch('delegator.run')
+    def test_pip_install_should_try_every_possible_source(self, mocked_delegator, mockec_sources, mocked_venv_location):
+        sources = [
+            {'url': 'http://dontexistis.in.pypi/simple'},
+            {'url': 'http://existis.in.pypi/simple'}
+        ]
+        mockec_sources.return_value = sources
```
nateprewitt: mocked

tests/test_pipenv.py
```diff
@@ -131,3 +134,56 @@ def test_activate_virtualenv_no_source(self):
...
+        first_cmd_return = Mock()
+        first_cmd_return.return_code = 1
+        second_cmd_return = Mock()
+        second_cmd_return.return_code = 0
+        mocked_delegator.side_effect = [first_cmd_return, second_cmd_return]
+        c = pip_install('package')
+        assert c.return_code == 0
+
+    @patch('pipenv.project.Project.virtualenv_location', new_callable=PropertyMock(return_value='foo'))
```
nateprewitt: Same question as above.

tests/test_pipenv.py
```diff
@@ -131,3 +134,56 @@ def test_activate_virtualenv_no_source(self):
...
+    @patch('pipenv.project.Project.sources', new_callable=PropertyMock)
+    @patch('delegator.run')
+    def test_pip_install_should_return_the_last_error_if_no_cmd_worked(self, mocked_delegator, mockec_sources, mocked_venv_location):
```
nateprewitt: `mocked_sources`

tests/test_pipenv.py
```diff
@@ -131,3 +134,56 @@ def test_activate_virtualenv_no_source(self):
...
+        sources = [
+            {'url': 'http://dontexistis.in.pypi/simple'},
+            {'url': 'http://dontexistis.in.pypi/simple'}
+        ]
+        mockec_sources.return_value = sources
```
nateprewitt: mocked

tests/test_pipenv.py
```diff
@@ -131,3 +134,56 @@ def test_activate_virtualenv_no_source(self):
...
+        c = pip_install('package')
+        assert c.return_code == 1
+        assert c == second_cmd_return
+
+    @patch('pipenv.project.Project.virtualenv_location', new_callable=PropertyMock(return_value='foo'))
+    @patch('pipenv.project.Project.sources', new_callable=PropertyMock)
+    @patch('delegator.run')
+    def test_pip_install_should_return_the_first_cmd_that_worked(self, mocked_delegator, mockec_sources, mocked_venv_location):
+        sources = [
+            {'url': 'http://dontexistis.in.pypi/simple'},
```
nateprewitt: nit: let's use the "successful" url (http://existis.in.pypi/simple) if we're going to create urls to denote that.

tests/test_pipenv.py
```diff
@@ -131,3 +134,56 @@ def test_activate_virtualenv_no_source(self):
...
+            {'url': 'http://dontexistis.in.pypi/simple'}
```
nateprewitt: Same here.

tests/test_pipenv.py
```diff
@@ -131,3 +134,56 @@ def test_activate_virtualenv_no_source(self):
...
+        second_cmd_return.return_code = 1
+        mocked_delegator.side_effect = [first_cmd_return, second_cmd_return]
+        c = pip_install('package')
+        assert c.return_code == 1
+        assert c == second_cmd_return
+
+    @patch('pipenv.project.Project.virtualenv_location', new_callable=PropertyMock(return_value='foo'))
+    @patch('pipenv.project.Project.sources', new_callable=PropertyMock)
+    @patch('delegator.run')
+    def test_pip_install_should_return_the_first_cmd_that_worked(self, mocked_delegator, mockec_sources, mocked_venv_location):
```
nateprewitt: `mocked_sources` again, both places.

tests/test_pipenv.py
arruda: Yeah, I replicated this typo in other places :sweat_smile: too

tests/test_pipenv.py
arruda: So... this one...
The problem was: I patched `delegator.run`, and it's was used in `which_pip` and in turn in the `which` function, when project.virtualenv_location was called it would also call `delegator.run` and I would get a error in `which` because of `Mock` object and `Str` concatenation.

One thing that I would suggest (but in another PR) is to move all the helper functions from cli.py to another place (maybe `utils.py`), because click does something do whatever is in `cli.py` that makes any function present in it inaccessible to Mock (that's why I couldn't mock `which_pip` for instance in this tests).

But I saw that `project.virtualenv_location` is no longer calling the  `delegator.run` by default (only if a given env var is present), so now I can remove this mocks for this `virtualenv_location` too =)

tests/test_pipenv.py
arruda: thanks, missed out this ones.

arruda:
I'll commit the fixes in the typos, urls names and remove the `project.virtualenv_location` mock. Thanks for the heads up

nateprewitt:
This looks good to me, thanks @arruda!

arruda:
:tada: =D

bretth:
I have one package I get from a private pypi and if I don't want all the other packages to fail on first install attempt I have to put the private one first which is counterintuitive:

```
[[source]]

url = "https://pypi.viewflow.io/xxxxxxxx/simple"
verify_ssl = true
name = "viewflowpro"

[[source]]

verify_ssl = true
name = "pypi"

[packages]
...
django-viewflow-pro = {version="*", index="viewflowpro"}
```

PR #227, opened 2017-02-15
Add options for setting pypi url

sharkspeed:
I want to do `pipenv install --pypi 'https://pypi.tuna.tsinghua.edu.cn/simple'` to easily change my pypi url in my Pipfile. So I just did it.

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -89,9 +89,15 @@ def ensure_pipfile(validate=True, pypi=PYPI_URL):
         project.create_pipfile(pypi=pypi)
 
     # Validate the Pipfile's contents.
-    if validate:
+    if validate and project.virtualenv_exists:
         # Ensure that Pipfile is using proper casing.
-        ensure_proper_casing()
+        p = project.parsed_pipfile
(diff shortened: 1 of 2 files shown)


nateprewitt:
Hmm, the problem you're trying to solve here is manually adding a source to your Pipfile? You already will be able to supply as many sources as you'd like in the next release of pipenv and it will iterate through those for you. Would that be sufficient?

sharkspeed:
Actually, my group uses a pypi mirror which is on our private server. When other guys try pipenv, I had to remind them to modify their Pipfile. So it will be better if there is a way to set pypi url when Pipfile is created. Or if there is a config file like "$HOME/.pip/pip.conf", it will spend less time to avoid changing source url one project by another.

kennethreitz:
edit your pipfile :)

wonderbeyond:
@kennethreitz "edit your pipfile" is a challenging thing when a project needs to run in different locations (e.g. China and America).

I need a way to use different index-url in different environment.
I think `$HOME/.pip/pip.conf` is already the best solution, Why not keep to it?

Issue #51, opened 2017-01-23
`pipenv shell` fails with path including spaces.


tsbertalan:
```
tsbertalan@sindri:~$ sudo pip install pipenv
... snip ...
tsbertalan@sindri:~$ cd dirname\ with\ spaces/
tsbertalan@sindri:~/dirname with spaces$ pipenv shell
Creating a Pipfile for this project...
Creating a virtualenv for this project...
New python executable in /home/tsbertalan/dirname with spaces/.venv/bin/python
Installing setuptools, pip, wheel...done.

Virtualenv location: /home/tsbertalan/dirname with spaces/.venv
Spawning environment shell (/bin/bash).
/bin/bash: /home/tsbertalan/dirname: No such file or directory
tsbertalan@sindri:~$ cd dirname\ with\ spaces/
tsbertalan@sindri:~/dirname with spaces$ pipenv shell
Creating a Pipfile for this project...
Creating a virtualenv for this project...
New python executable in /home/tsbertalan/dirname with spaces/.venv/bin/python
Installing setuptools, pip, wheel...done.

Virtualenv location: /home/tsbertalan/dirname with spaces/.venv
Spawning environment shell (/bin/bash).
/bin/bash: /home/tsbertalan/dirname: No such file or directory
```

jessebraham:
I believe the directory name in this instance needs to be quoted prior to creating the virtual environment, although I could be mistaken.

Adding quotes to [this line](https://github.com/kennethreitz/pipenv/blob/master/pipenv/cli.py#L142) should be all that's required I believe.

jessebraham:
Okay, so I've developed a naive solution, but I'm just looking for input on this. In `cli.py`, on lines 276-279 (at the time of writing), we have this:

```python
if source:
    return 'source {0}/bin/activate{1}'.format(project.virtualenv_location, suffix)
else:
    return '{0}/bin/activate'.format(project.virtualenv_location)
```

Simply replacing the block with the following enables proper functionality by escaping the spaces:

```python
venv_location = project.virtualenv_location.replace(' ', '\ ')

if source:
    return 'source {0}/bin/activate{1}'.format(venv_location, suffix)
else:
    return '{0}/bin/activate'.format(venv_location)
```

[...]

PR #230, opened 2017-02-16
Handle the directory name with whitespaces properly (#228)

### LINKED ISSUE #228: Cannot install a wheel package in a directory with whitespaces
### Summary

- If the name of current working directory contains the whitespaces
- You won't be able to install a package with [the wheel format](https://www.python.org/dev/peps/pep-0427/).

This bug occurs regardless of the value of `PIPENV_VENV_IN_PROJECT` environment variable.

### How to reproduce this error
Please comment if you cannot reproduce the error.

```console
$ mkdir 'directory with spaces'
$ cd 'directory with spaces/'


$ pipenv --three
⠋Running virtualenv with interpreter /usr/bin/python3
Using base prefix '/usr'
New python executable in /home/simnalamburt/.local/share/virtualenvs/directory with spaces/bin/python3
Also creating executable in /home/simnalamburt/.local/share/virtualenvs/directory with spaces/bin/python

Virtualenv location:


$ pipenv install numpy
⠋Using base prefix '/usr'
New python executable in /home/simnalamburt/.local/share/virtualenvs/directory with spaces/bin/python
  File "/usr/bin/virtualenv", line 11, in <module>
    sys.exit(main())
  File "/usr/lib/python3.6/site-packages/virtualenv.py", line 713, in main
    symlink=options.symlink)
  File "/usr/lib/python3.6/site-packages/virtualenv.py", line 925, in create_environment
    site_packages=site_packages, clear=clear, symlink=symlink))
  File "/usr/lib/python3.6/site-packages/virtualenv.py", line 1370, in install_python
    os.symlink(py_executable_base, full_pth)
FileExistsError: [Errno 17] File exists: 'python' -> '/home/simnalamburt/.local/share/virtualenvs/directory with spaces/bin/python3.6'

Virtualenv location:
Installing numpy...
Collecting numpy
  Using cached numpy-1.12.0-cp36-cp36m-manylinux1_x86_64.whl
Installing collected packages: numpy

Error:  An error occurred while installing numpy!
Exception:
  File "/usr/lib/python3.6/site-packages/pip/basecommand.py", line 215, in main
    status = self.run(options, args)
  File "/usr/lib/python3.6/site-packages/pip/commands/install.py", line 342, in run
    prefix=options.prefix_path,
  File "/usr/lib/python3.6/site-packages/pip/req/req_set.py", line 784, in install
    **kwargs
  File "/usr/lib/python3.6/site-packages/pip/req/req_install.py", line 851, in install
    self.move_wheel_files(self.source_dir, root=root, prefix=prefix)
  File "/usr/lib/python3.6/site-packages/pip/req/req_install.py", line 1064, in move_wheel_files
    isolated=self.isolated,
  File "/usr/lib/python3.6/site-packages/pip/wheel.py", line 345, in move_wheel_files
    clobber(source, lib_dir, True)
  File "/usr/lib/python3.6/site-packages/pip/wheel.py", line 316, in clobber
    ensure_dir(destdir)
  File "/usr/lib/python3.6/site-packages/pip/utils/__init__.py", line 83, in ensure_dir
    os.makedirs(path)
  File "/usr/lib/python3.6/os.py", line 220, in makedirs
    mkdir(name, mode)
PermissionError: [Errno 13] Permission denied: '/usr/lib/python3.6/site-packages/numpy-1.12.0.dist-info'
```
screenshot
*(Click to see the screenshot)*

### My Environment
I could reproduce the error in multiple environments. Please comment if you want more information.

```console
$ uname -a
Linux 4.9.6-1-ARCH #1 SMP PREEMPT Thu Jan 26 09:22:26 CET 2017 x86_64 GNU/Linux
$ python3 --version
Python 3.6.0
$ pipenv --version
pipenv, version 3.4.1
```
```console
$ uname -a
Darwin 16.4.0 Darwin Kernel Version 16.4.0: Thu Dec 22 22:53:21 PST 2016; root:xnu-3789.41.3~3/RELEASE_X86_64 x86_64
$ python3 --version
Python 3.6.0
$ pipenv --version
pipenv, version 3.4.1
```

simnalamburt:
Closes #228. Please review and comment. Thanks!

Use the command below to try patched version of pipenv.
```
pip install git+https://github.com/simnalamburt/pipenv.git@plan-a
```

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -17,6 +17,7 @@ psutil = "*"
 "backports.shutil-get-terminal-size" = "*"
 pew = ">=0.1.26"
 blindspin = "*"
+python-slugify = ">=1.2.1,<2"
 
 [requires]
 python_version = "2.7"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,196 +1,4 @@
 {
-    "default": {
-        "crayons": {
-            "version": "==0.1.2",
-            "hash": "sha256:5e17691605e564d63482067eb6327d01a584bbaf870beffd4456a3391bd8809d"
-        },
-        "requirements-parser": {
-            "version": "==0.1.0",
-            "hash": "sha256:fee2380a469ffe4067bc7f0096a6fcfb27539da7496fae12b74b8d5d0f33a4ee"
-        },
-        "backports.shutil-get-terminal-size": {
-            "version": "==1.0.0",
-            "hash": "sha256:0975ba55054c15e346944b38956a4c9cbee9009391e41b86c68990effb8c1f64"
-        },
-        "blindspin": {
-            "version": "==2.0.1",
-            "hash": "sha256:b30ef7a26ef1637c1d047667b279e8c15dc2a78fcfaad6b3027d7217e752bdba"
-        },
-        "parse": {
-            "version": "==1.6.6",
-            "hash": "sha256:a4862be306f334c36ae7adc73af028c56ca0139b8e39435e935bde8d481dd99e"
-        },
-        "MarkupSafe": {
-            "version": "==0.23",
-            "hash": "sha256:a4ec1aff59b95a14b45eb2e23761a0179e98319da5a7eb76b56ea8cdc7b871c3"
-        },
-        "colorama": {
-            "version": "==0.3.7",
-            "hash": "sha256:a4c0f5bc358a62849653471e309dcc991223cf86abafbec17cd8f41327279e89"
-        },
-        "pyparsing": {
-            "version": "==2.1.10",
-            "hash": "sha256:67101d7acee692962f33dd30b5dce079ff532dd9aa99ff48d52a3dad51d2fe84"
-        },
-        "click-completion": {
-            "version": "==0.2.1",
-            "hash": "sha256:079fb138887d4de12a0b7fbebf8d92d396b7c1a9c49f63475d9f3909d2588976"
-        },
-        "pipfile": {
-            "version": "==0.0.1",
-            "hash": "sha256:a0bf9924d9bd25fa5e216d643fde49ff67f48727cb711a02ad806bf7c4722ea2"
-        },
-        "appdirs": {
-            "version": "==1.4.0",
-            "hash": "sha256:85e58578db8f29538f3109c11250c2a5514a2fcdc9890d9b2fe777eb55517736"
-        },
-        "virtualenv": {
-            "version": "==15.1.0",
-            "hash": "sha256:39d88b533b422825d644087a21e78c45cf5af0ef7a99a1fc9fbb7b481e5c85b0"
-        },
-        "packaging": {
-            "version": "==16.8",
-            "hash": "sha256:99276dc6e3a7851f32027a68f1095cd3f77c148091b092ea867a351811cfe388"
-        },
-        "pexpect": {
-            "version": "==4.2.1",
-            "hash": "sha256:f853b52afaf3b064d29854771e2db509ef80392509bde2dd7a6ecf2dfc3f0018"
-        },
-        "resumable-urlretrieve": {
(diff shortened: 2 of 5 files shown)


nateprewitt:
Hey @simnalamburt, thanks for taking the time to put this together. This issue was researched fairly extensively when it originally cropped up in #51. As the issue you linked to in the other thread (pypa/pip#923) notes, this is a problem with virtualenv.

As I noted [here](https://github.com/kennethreitz/pipenv/issues/51#issuecomment-275316816), trying to run an executable with spaces in the path will not work because the linux kernel parses the first space as the end of the command, and everything that follows is processed as a parameter. This issue (pypa/virtualenv/#53) has been open in virtualenv for nearly 6 years.

As for this patch, If start a fresh pipenv project with spaces, and then run `pipenv install numpy` as you showed in your example, I still receive `/bin/sh: /bin/pip: No such file or directory`. This is because the shebangs in the files created by virtualenv are all in the form `#!"/Users/pipenv/.local/share/virtualenvs/test with spaces/bin/python"`. When you run a command like `pip install` it calls a bash script in the /bin/ directory (`pip`) which has a shebang bound to the absolute path of the python version is needs to use. Without being able to modify all shebangs in the directory every time we make a change, we can't really guarantee any form of stability.

Due to the fact that both pip and virtualenv have had issues opened for multiple years about this, without resolution, I'm hesitant to introduce a patch that doesn't seem to work universally and makes breaking API changes. For those reasons, I don't think we're properly equipped to address this right now. If you find more information on a fix upstream in both pip and virtualenv, I'll gladly look at it. Thanks again for spending time researching this.

simnalamburt:
@nateprewitt Did you really tried the patched version of pipenv? With this pull request being merged, the python executable won't contain whitespaces in its file path. It'll be like this: `/Users/username/.local/share/virtualenvs/test-with-spaces/bin/python`

Please try again with the command below:
```bash
pip install --upgrade git+https://github.com/simnalamburt/pipenv.git@plan-a
```

nateprewitt:
@simnalamburt, my apologies, you're correct. There was something about my environment earlier that was experiencing conflict.

So that puts us at whether we're ready to consider a 4.0 jump. This will break previously created virtualenvs that were casing specific or used anything other than dashes as a separator. `Test With Spaces`, `test_with spaces`, `Test.With.Spaces` and `test-with-spaces` will now all be the same virtualenv. We'll need to make sure we provide documentation about that since it seems like it could be a stumbling block.

This will also provide an expectation for directories with spaces to work properly, however with `PIPENV_VENV_IN_PROJECT` this is still broken. It would be nice if we could make this solution work uniformly but that requires changing the project directory. I'd like @kennethreitz to sign off on this before we move forward, and we'll likely need to open a PR against a future 4.0.0 branch while we round up some of the other breaking changes that need to be queued up.

simnalamburt:
@nateprewitt Hi! Thank you for reviewing my PR

> `Test With Spaces`, `test_with spaces`, `Test.With.Spaces` and `test-with-spaces` will now all be the same virtualenv.

I'll come back with the new solution which solves the problem that you said. If this PR breaks the backward compatibility, I'll solve as many problem as possible with a single breakage.

ChristopherRabotin:
When will this be on PyPi? I encountered this very frustrating issue today. Regardless, thanks a lot for fixing this.

nateprewitt:
Hey @ChristopherRabotin, I'm hoping to get a release (3.5.0) out in the next week, but can't make any promises. There's a few other features/bug fixes that are close to finished but not ready to merge.

In the meantime, if this is a critical bug fix for you, you can always install pipenv with `pip install git+https://github.com/kennethreitz/pipenv.git#egg=pipenv` to get the current HEAD of our master branch. I wouldn't recommend this as a permanent solution, but it will keep you going for the next few days.

PR #233, opened 2017-02-18
Make pip_download respect project.sources

### LINKED ISSUE #231: Pipenv lock don't respect multiple sources
Just saw now that the lock uses `pip_download` that don't use anything other than pip's default configuration on index-url.

One approach that seems reasonable would be to use pip_download like pip_install function: with `-i` and trying for each project.source available.

arruda:
As stated in #231, `pip_download` (function used by the pipenv lock command) ins't using the project.sources to download the pypackages.
I changed it so that `pip_download` would work like `pip_install` and try downloading using each of the defined `project.sources` as the `index-url` in the `pip` md.

Also added some tests fo check for this behavior.

fix #231

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -565,7 +565,18 @@ def pip_install(package_name=None, r=None, allow_global=False):
 
 
 def pip_download(package_name):
-    c = delegator.run('{0} download "{1}" -d {2}'.format(which_pip(), package_name, project.download_location))
+    for source in project.sources:
+        c = delegator.run(
+            '{0} download "{1}"  -i {2} -d {3}'.format(
+                which_pip(),
+                package_name,
+                source['url'],
+                project.download_location
+            )
+        )
+        if c.return_code == 0:
+            break
+
     return c
 
 
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -566,14 +566,13 @@ def pip_install(package_name=None, r=None, allow_global=False):
 
 def pip_download(package_name):
     for source in project.sources:
-        c = delegator.run(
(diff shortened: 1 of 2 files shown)


pipenv/cli.py
```diff
@@ -565,7 +565,18 @@ def pip_install(package_name=None, r=None, allow_global=False):
 
 
 def pip_download(package_name):
-    c = delegator.run('{0} download "{1}" -d {2}'.format(which_pip(), package_name, project.download_location))
+    for source in project.sources:
+        c = delegator.run(
```
nateprewitt: I think I'd rather see this broken into two lines, one for the command string and one for the `delegator.run` call.

arruda: Fair enough, I'll change it then.

nateprewitt:
Sorry, I guess I should have been clearer 😅 I was suggesting having the command occupy a single line rather than 5, it's more of a stylistic quirk that this project and most of the kennethreitz projects follow. I'm just going to merge this now and I'll tidy things on our end. Thanks so much for the work @arruda!

arruda:
@nateprewitt sorry, my mistake :man_facepalming:

nateprewitt:
No need to apologies @arruda, I could have been clearer 😊

mikejohn23:
Dear

My name is Ashman Duke , a partner and head of dispute resolution at Duke Law

Chambers Benin Republic, we want to know if late Deniz who came to Benin in 1939

was related to you.

We contacted you because he has the same last name with you also from the same

Country with you. Our search shows that he is your extended relative. Please do

you know late Deniz?

We need your urgent reply.

Respectfully.

Barrister Ashman Duke
Of Counsel Benin
Mob. +22998955626


________________________________

PR #238, opened 2017-02-20
Hash the project path and use it as a virtualenv's name

### LINKED ISSUE #228: Cannot install a wheel package in a directory with whitespaces

simnalamburt:
This PR closes #228. See https://github.com/kennethreitz/pipenv/issues/228 for the former discussion.

Use the command below to try patched version of pipenv.
```bash
pip install --upgrade git+https://github.com/simnalamburt/pipenv.git@issue-228
```

Please review and comment. Thanks!

### DIFF
--- a/pipenv/project.py
+++ b/pipenv/project.py
@@ -46,7 +46,7 @@ class Project(object):
 
         # The user wants the virtualenv in the project.
         if not PIPENV_VENV_IN_PROJECT:
-            c = delegator.run('pew dir {0}'.format(self.name))
+            c = delegator.run('pew dir "{0}"'.format(self.name))
             loc = c.out.strip()
         # Default mode.
         else:
--- a/pipenv/project.py
+++ b/pipenv/project.py
@@ -1,6 +1,9 @@
 # -*- coding: utf-8 -*-
 import json
 import os
+import re
+import base64
+import hashlib
 
 import pipfile
 import toml
@@ -38,6 +41,23 @@ class Project(object):
         return os.path.isdir(self.virtualenv_location)
 
     @property
+    def virtualenv_name(self):
+        # Replace dangerous characters into '_'
+        #
+        # References:
+        #   https://www.gnu.org/software/bash/manual/html_node/Double-Quotes.html
+        #   http://www.tldp.org/LDP/abs/html/special-chars.html#FIELDREF
+        sanitized = re.sub(r'[ $`!*@"\\\r\n\t]', '_', self.name)
+
+        # Hash the full path of the pipfile
+        hash = hashlib.sha256(self.pipfile_location.encode()).digest()[:6]
+        encoded_hash = base64.urlsafe_b64encode(hash).decode()
+
+        # If the pipfile was located at '/home/user/MY_PROJECT/Pipfile',
+        # the name of its virtualenv will be 'my-project-wyUfYPqE'
+        return sanitized + '-' + encoded_hash
+
+    @property
     def virtualenv_location(self):
 
         # Use cached version, if available.
@@ -46,7 +66,7 @@ class Project(object):
 
         # The user wants the virtualenv in the project.
         if not PIPENV_VENV_IN_PROJECT:
-            c = delegator.run('pew dir "{0}"'.format(self.name))
+            c = delegator.run('pew dir "{0}"'.format(self.virtualenv_name))
             loc = c.out.strip()
         # Default mode.
         else:
--- a/pipenv/project.py
+++ b/pipenv/project.py
@@ -42,12 +42,19 @@ class Project(object):
 
     @property
     def virtualenv_name(self):
-        # Replace dangerous characters into '_'
+        # Replace dangerous characters into '_'. The length of the sanitized
+        # project name is limited as 42 because of the limit of linux kernel
+        #
+        # 42 = 127 - len('/home//.local/share/virtualenvs//bin/python2') - 32 - len('-HASHHASH')
+        #
+        #      127 : BINPRM_BUF_SIZE - 1
+        #       32 : Maxmimum length of username
         #
         # References:
         #   https://www.gnu.org/software/bash/manual/html_node/Double-Quotes.html
         #   http://www.tldp.org/LDP/abs/html/special-chars.html#FIELDREF
-        sanitized = re.sub(r'[ $`!*@"\\\r\n\t]', '_', self.name)
+        #   https://github.com/torvalds/linux/blob/2bfe01ef/include/uapi/linux/binfmts.h#L18
+        sanitized = re.sub(r'[ $`!*@"\\\r\n\t]', '_', self.name)[0:42]
 
         # Hash the full path of the pipfile
         hash = hashlib.sha256(self.pipfile_location.encode()).digest()[:6]
--- a/tests/test_project.py
+++ b/tests/test_project.py
@@ -1,3 +1,6 @@
+from hashlib import sha256
+from base64 import urlsafe_b64encode
+
 import delegator
 
 import pipenv.project
@@ -6,9 +9,13 @@ class TestProject():
 
     def test_project(self):
         proj = pipenv.project.Project()
+        hash = urlsafe_b64encode(
+            sha256(proj.pipfile_location.encode()).digest()[:6]).decode()
+
         assert proj.name == 'pipenv'
         assert proj.pipfile_exists
         assert proj.virtualenv_exists
+        assert proj.virtualenv_name == 'pipenv-' + hash
 
     def test_proper_names(self):
         proj = pipenv.project.Project()
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -321,7 +321,7 @@ def do_create_virtualenv(three=None, python=None):
         cmd = ['virtualenv', project.virtualenv_location, '--prompt=({0})'.format(project.name)]
     else:
         # Default: use pew.
-        cmd = ['pew', 'new', project.name, '-d']
+        cmd = ['pew', 'new', project.virtualenv_name, '-d']
 
     # Pass a Python version to virtualenv, if needed.
     if python:
@@ -891,7 +891,7 @@ def shell(three=None, python=False, compat=False, shell_args=None):
     # Standard (properly configured shell) mode:
     else:
         cmd = 'pew'
-        args = ["workon", project.name]
+        args = ["workon", project.virtualenv_name]
 
     # Grab current terminal dimensions to replace the hardcoded default
     # dimensions of pexpect
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -66,13 +66,13 @@ def ensure_latest_pip():
     """Updates pip to the latest version."""
 
     # Ensure that pip is installed.
-    c = delegator.run('{0} install pip'.format(which_pip()))
+    c = delegator.run('"{0}" install pip'.format(which_pip()))
 
     # Check if version is out of date.
     if 'however' in c.err:
         # If version is out of date, update.
         click.echo(crayons.yellow('Pip is out of date... updating to latest.'))
-        c = delegator.run('{0} install pip --upgrade'.format(which_pip()), block=False)
+        c = delegator.run('"{0}" install pip --upgrade'.format(which_pip()), block=False)
         click.echo(crayons.blue(c.out))
 
 
@@ -374,7 +374,7 @@ def get_downloads_info(names_map, section):
         version = parse_download_fname(fname, name)
 
         # Get the hash of each file.
-        c = delegator.run('{0} hash {1}'.format(which_pip(), os.sep.join([project.download_location, fname])))
+        c = delegator.run('"{0}" hash "{1}"'.format(which_pip(), os.sep.join([project.download_location, fname])))
         hash = c.out.split('--hash=')[1].strip()
 
         # Verify we're adding the correct version from Pipfile
@@ -491,7 +491,7 @@ def do_purge(bare=False, downloads=False, allow_global=False):
         shutil.rmtree(project.download_location)
         return
 
-    freeze = delegator.run('{0} freeze'.format(which_pip(allow_global=allow_global))).out
+    freeze = delegator.run('"{0}" freeze'.format(which_pip(allow_global=allow_global))).out
     installed = freeze.split()
 
     # Remove setuptools and friends from installed, if present.
@@ -502,7 +502,7 @@ def do_purge(bare=False, downloads=False, allow_global=False):
 
     if not bare:
         click.echo('Found {0} installed package(s), purging...'.format(len(installed)))
-    command = '{0} uninstall {1} -y'.format(which_pip(allow_global=allow_global), ' '.join(installed))
+    command = '"{0}" uninstall {1} -y'.format(which_pip(allow_global=allow_global), ' '.join(installed))
     c = delegator.run(command)
 
     if not bare:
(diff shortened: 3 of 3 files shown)


kennethreitz:
let's only use the hash for the project name if it includes a space! that way 90% of our users get the intended usage.

simnalamburt:
@kennethreitz I also prefer readable virtualenv name more than a random hash, but I wanted to solve the "same project name" issue like this

- `/home/user/my-project`
- `/home/user/workspace/my-project`
- `/home/user/another workspace/my-project`

Shall we use the both slugified project name and hash, like this?

- `/home/user/My,project` &rarr; `my-project-WRR-`
- `/home/user/MY_PROJECT` &rarr; `my-project-hpxK`

kennethreitz:
i like that!

simnalamburt:
@kennethreitz OK thanks! I'll do both update the PR and fix the unit tests

simnalamburt:
Don't merge it yet.

Updated the PR to match the new design scheme, but have not fixed the unit tests.

kennethreitz:
@simnalamburt ping me with the tests are ready!

pipenv/project.py
```diff
@@ -26,7 +33,13 @@ def __init__(self):
     @property
     def name(self):
         if self._name is None:
-            self._name = self.pipfile_location.split(os.sep)[-2]
+            slug = slugify(self.pipfile_location.split(os.sep)[-2])
+            hash = blake2b(self.pipfile_location.encode(), digest_size=3)
```
nateprewitt: So two notes here, the first is are we sure we want to have the hash in the project name? This is will change if the directory is ever moved, while the actual project name stays the same. This seems like this dirty laundry doesn't need to be exposed to the user in a public property, especially since it won't function consistently with `PIPENV_VENV_IN_PROJECT`. I'd be in favor of having this functionality moved to a separate function and only used in `virtualenv_location`.

The second issue is, while having a 4 char hash is nicer for readability, I'm consistently encountering a collision with this approach in ~1/5500 project names. As names get longer (10-15 characters), I'm encountering collisions in less than 1/1000. Those aren't great odds for guaranteed uniqueness.

tests/test_project.py
```diff
@@ -6,7 +6,7 @@ class TestProject():
 
     def test_project(self):
         proj = pipenv.project.Project()
-        assert proj.name == 'pipenv'
+        assert proj.name == 'pipenv-uoEO'
```
nateprewitt: Are we sure this is true for all platforms? If we run the tests in `/home/my_user/test_space/pipenv` v.s. `/home/alfred/test_space/pipenv`, won't we get different values for `proj.name`?

kennethreitz:
I don't disagree with @nateprewitt

simnalamburt:
I'll send a PR again somewhere else `master` branch when the review is done.

#### About stating "don't use spaces in virtualenv project folders" to the users...
If you really want not to support project folders with whitespaces, I don't have any power to stop it. But for me and my colleagues who are mostly CJK users, it was quite a big entry barrier. So I must keep maintaining a fork of pipenv which supports whitespaces or give up using pipenv in production.

If you don't like the result, I may try harder to support it. But I really want pipenv to support project name with whitespaces someday.

#### Why BLAKE2?
I wanted to change the size of the digest in a reliable way. Since the result length of SHA256 is fixed, there's no choice but to just cut the result of it.

Also, SHA256 is still used in production but I didn't want to use if for the future since SHA256 is semi-deprecated by NSA (they has stopped using it).

#### Why base64?
It is the shortest and reliable way to encode random binary into an URL-safe string.

```python
digest = bytes([211, 144, 10, 1, 53, 210, 45, 118, 0])

digest.hex()                                # 'd3900a0135d22d7600'
base64.urlsafe_b64encode(digest).decode()   # '05AKATXSLXYA'
```

pipenv/project.py
simnalamburt: There's no good way for pipenv to distinguish the next two:

- It's the same project with the different location
- It's the different project with the same name

So we must make a choice in here.

1. Assume that all projects with the same name as same projects. (Current implementation, the "same project name" issue occurs)
2. Assume that all projects with the same name as different projects. (This PR, the "reinstall required after moving project directory" issue occurs)

I thought that the "same project name" issue is more severe than the "reinstall required after moving project directory" issue since the latter one is just an inconvenience while the former one is an issue without a proper way to workaround.

About `virtualenv_location`, good point. I'll move codes at there.

The hash collision probability is 0.03%, and I though that it's practical enough since users won't create ~3000 projects with the same name. Even if it may collide, it's better than the collision with 100% probability without a hash. But I'm OK to use a longer hash if I can sacrifice some beauty. Shall we?

And the hash collision probability has no relation with the name of the virtualenv, and the hash is OK to collide with a project which has a different name. Because it uses `<name>-<hash>` as a key.

tests/test_project.py
simnalamburt: I'm working on it but not now (it's currently 03:00 AM in my timezone)

nateprewitt: There's no rush here, @simnalamburt :) I probably won't have a moment to respond to the other comments for a few hours, so don't feel like you need to stay up working on this.

nateprewitt:
>If you really want not to support project folders with whitespaces, I don't have any power to stop it. But for me and my colleagues who are mostly CJK users, it was quite a big entry barrier. So I must keep maintaining a fork of pipenv which supports whitespaces or give up using pipenv in production.
If you don't like the result, I may try harder to support it. But I really want pipenv to support     
project name with whitespaces someday.

Forgive me for not having more background here, a cursory google search isn't providing many answers. Could you help me understand what about the CJK character-set creates (or requires?) spaces? I'm definitely sympathetic to this, and am much more willing to work to solve the whitespace issue if it's a current limitation for CJK.

>I wanted to change the size of the digest in a reliable way. Since the result length of SHA256 is fixed, there's no choice but to just cut the result of it.

So, while the byte pool is bigger with Blake2, we're restricting it to 3 bytes which defeats a lot of the benefit especially when we encode those bytes down to 4 characters of a base64 encoded string. So at that point, is there a reason not to use the first 4 characters from a sha256 value? It's the same 4^64 possibilities as the first 4 characters of a base64 string, no? I'm just trying to keep our dependency footprint small if there's no significant value add.

>Also, SHA256 is still used in production but I didn't want to use if for the future since SHA256 is semi-deprecated by NSA (they has stopped using it).

I think this may be conflating cryptographic integrity and hash uniqueness. We don't care about whether someone can reverse engineer the hash here, they already know what the base value is. The only thing we're trying to accomplish with the hash function is to create a uniquely identifiable value which I believe sha256 is sufficient for. We're already restricting the value to 4 characters so I believe it's an identical collision space. 


---

SO, with all of this, if spaces are truly a byproduct of file paths with CJK characters let's get this merged! Unless you have further thoughts on the virtualenv name and hashes, I'd still like to see the external dependencies removed. Let's replace blake2b with sha256 and instead of running slugify, it should be sufficient to replace whitespace with a `-`, correct? Is `slugify` doing something else to modify CJK characters? Once that's done and moved out of the `name` property, I think we'll be set.

Thanks for working through this with me @simnalamburt, I think we're getting close to an agreeable solution :) Let me know if you have any more feedback on my comments.

simnalamburt:
I just update the PR, but don't merge it yet. I didn't try to remove the slugify dependency yet.

1. **New property: `virtualenv_name`**
    - If a project is located in `/home/user/MY_PROJECT/Pipfile`
    - `project.name` will be `MY_PROJECT`
    - `project.virtualenv_name` will be `my-project-wyUfYPqE`.
1. Use sha256 of stdlib instead of blake2 as pointed in https://github.com/kennethreitz/pipenv/pull/238#issuecomment-281192121. No more additional pyblake2 dependency.
1. Use a 48bit hash instead of 24bit hash as pointed in https://github.com/kennethreitz/pipenv/pull/238#discussion_r102049269. Now there's 0.18% chance of hash collision when you make one million projects.

Slugify changes all non-ASCII characters and special characters into [portable filename character set](http://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap03.html#tag_03_276), so it complete removes the unsupported character in the path issue. But I'll give a try to solve it without slugify dependency.

nateprewitt:
So it looks like `unidecode` is primarily what they're using under the hood which would work but only exists in the standard lib in Python 2.7 and 3.5+. Given that, I don't think it's worth the time trying to brew our own thing. If @kennethreitz is alright with the slugify dependency, I think I'm convinced it's needed here.

simnalamburt:
I'm currently digging the bash spec document, and I found that we need to replace all those characters below:

```
$  `  \  !  "  *  @  (space)  (newline)  (tab)
```

... and maybe there can be more. I'll give a try so please hang on for an hour. Thanks!

###### References
- https://www.gnu.org/software/bash/manual/html_node/Double-Quotes.html
- http://www.tldp.org/LDP/abs/html/special-chars.html#FIELDREF

simnalamburt:
**Don't merge it yet**

I blacklisted certain characters, and it works great. I tested it with the directories below:

```
mkdir \!@\#\$\%\^\&\*\(\)\)\)\)\)\)__+\{\}\[\]-=\;\',.\<\>\?한글\ \ ㅇㅅㅇ\ \>ㅅㅇ
mkdir "Ω≈ç√∫˜µ≤≥÷"
mkdir "åß∂ƒ©˙∆˚¬…æ"
mkdir "œ∑´®†¥¨ˆøπ“‘"
mkdir "¡™£¢∞§¶•ªº–≠"
mkdir "¸˛Ç◊ı˜Â¯˘¿"
mkdir "ÅÍÎÏ˝ÓÔÒÚÆ☃"
mkdir "Œ„´‰ˇÁ¨ˆØ∏”’"
mkdir "`⁄€‹›ﬁﬂ‡°·‚—±"
mkdir "⅛⅜⅝⅞"
mkdir "٠١٢٣٤٥٦٧٨٩"
```

But soon I found that there's 128 byte limit in the hashbang executable, so I'm fixing it. Hang on

### References
- https://github.com/minimaxir/big-list-of-naughty-strings

simnalamburt:
Finally, I have fixed the whole issues mentioned above. *This is it! please review this!*

Even though I just found a new issue (😱...) that virtualenv with python2 crashes with CJK characters, but that's an issue of virtualenv not pipenv. Python2 users could not user CJK characters in the project directory in the first place, so they won't complain about this PR. (It took me a long time to find this since I do not use the python2)

I may fix the virtualenv to support python2 but the fact that support of python2 will be ended in 3 years discourages me. I'll leave that issue for another day.

### What this PR fixed
- "Cannot install a wheel package in a directory with whitespaces" #228 
- "Projects with the same names in different location" issue
- "Special characters in a project name" issue
- "Too long project name" issue

### What this PR does not fixes
- With python2, https://github.com/pypa/virtualenv/issues/457 will still occur.

nateprewitt:
Ok, I think I'm happy with this! I'll leave it up for a little bit for @kennethreitz to peek at if he'd like, otherwise I'll merge it later today. Thanks so much for all the work, @simnalamburt! 🎉

nateprewitt:
Alright, I'm going to go ahead and merge this in. Thanks again @simnalamburt!

kennethreitz:
does this change the name of all existing virtualenvs?

nateprewitt:
It's going to modify the directory name for virtualenvs generated for pew, this is the only way to address multiple projects with the same name. It will install packages to a different directory than pipenv currently does in 3.4.2.

abulte:
First: thanks for the great work on this awesome project, I really do not want to sound ungrateful below.

BUT having the name of existing virtualenv change is quite a pain as an end-user. This means reinstalling potentially a lot of packages when we update `pipenv`, and on multiple machines in my case, some of them quite slow. This happened twice already since I use pipenv. Could you maybe consider adding a test for an existing virtualenv when the designated path changes in a new feature?

nateprewitt:
Hey @abulte, we recognize that this was a inconvenient change in 3.5.0 but was fundamentally necessary. The issue here that needed to be resolved was virtualenv naming conflicts and the fact the user was capable of accidentally deleting their entire project by creating a new one in a separate, but identically named directory. The fact you previously could clone a project into a new directory and silently delete your entire production environment without realizing it was pretty significant.

I'm definitely sympathetic to having to recreate your existing venvs, and we didn't make this change lightly. I went back and forth on doing a MAJOR version bump here, but ended up deciding against it. Prior to 3.5.0 we didn't actually have a way to properly determine whether a project already exists or not. We're asking users who didn't opt into `PIPENV_VENV_IN_PROJECT` to do a one time setup for their existing environments. This saves most users having to do this reinstall every time there's a naming conflict or they're attempting to test a project in another directory.

abulte:
Thanks for the feedback @nateprewitt. I totally understand the need for this feature, I was just complaining selfishly about its drawbacks ;-) Glad you thought of the inconvenience before hand.

On a related: shouldn't there be a CHANGELOG for this project? I couldn't find one (maybe overlooked) and ended up doing a Google search to understand why my new virtualenv had a strange name after upgrading pipenv :-/ (and land here).

nateprewitt:
We use the naming convention [`HISTORY.txt`](https://github.com/kennethreitz/pipenv/blob/master/HISTORY.txt) in this project, because that's been the precedent in other kennethreitz projects.

abulte:
Got it thanks!

Issue #240, opened 2017-02-24
Comparison of package names must be case insensitive


jdno:
Hi,

when trying to uninstall a package, I noticed that it stayed in the `Pipfile` even though the command output told we it would be removed. I started to dig into this thinking I discovered a bug in the uninstall command, but it seems it's a more general problem related to the way `pipenv` compares package names.

See this example:

     $ pipenv install pyyaml
     $ pipenv uninstall PyYAML

These commands successfully install and uninstall the Python package. But, the package won't be removed from the `Pipfile`, because the comparison in [project.remove_package_from_pipfile](https://github.com/kennethreitz/pipenv/blob/ed6b73c84f16e44488473c5d461b2ddfe3df6eb3/pipenv/project.py#L179) is case sensitive.

This seems to be a violation of [PEP-426](https://www.python.org/dev/peps/pep-0426/#name), which states that

> All comparisons of distribution names MUST be case insensitive, and MUST consider hyphens and underscores to be equivalent.

The easiest way to handle this seems to be to transform the package name with something like `package_name.lower()` and `package_name.replace('-', '_')` into a "known" format, but it might also be worth it to look how `pip` deals with this. Sadly, my time is limited right now, otherwise I would have already looked into this myself.

Let me know how you want to solve this, and I'll see if I can prepare a pull request. :)

nateprewitt:
Hey @jdno, thanks for opening this ticket! We've been battling with getting this working consistently for the last few minor versions and have open issues regarding how this affects the code in other places in #229 and #206. I'm currently in the process of redoing how we handle both Pipfile and the lockfile internally which will give us a model consistent with pep426 so we're only doing one transform before writing. We're hoping to see this land in the next week in 3.5.0.

jdno:
Damn it, I didn't check the existing issues well enough. But I'm glad to hear you're spending more time on it than I would have, to properly fix the issue... 😉 

Should we close this then?

nateprewitt:
@jdno, let's leave it open for now as a reminder to make sure we have proper testing in place for this kind of case. I'll wrap it into the issues we'll close out with 3.5.0.

nateprewitt:
This should be resolved in #242 and will be released in 3.5.0.

Issue #229, opened 2017-02-16
Duplicate requirement in Pipfile.lock when case differs 


jacebrowning:
With 3.4.1, `$ pipenv install --lock mkdocs pdoc` produces a `Pipfile.lock` containing:

```
        "markdown": {
            "hash": "sha256:812ec5249f45edc31330b7fb06e52aaf6ab2d83aa27047df7cb6837ef2d269b6",
            "version": "==2.4.1"
        },
        "Markdown": {
            "hash": "sha256:0ac8a81e658167da95d063a9279c9c1b2699f37c7c4153256a458b3a43860e33",
            "version": "==2.6.8"
        },
```

which fails on reinstall:

```
$ pipenv install
No package provided, installing all dependencies.
Pipfile found at /private/tmp/Pipfile. Considering this to be the project home.
Installing dependencies from Pipfile.lock...
An error occured while installing!
Double requirement given: Markdown==2.6.8 (from -r /var/folders/ly/0k1p93rj1q56hk9rnk5j774h0000gn/T/tmpfc2fvv8a-requirements.txt (line 9)) (already in markdown==2.4.1 (from -r /var/folders/ly/0k1p93rj1q56hk9rnk5j774h0000gn/T/tmpfc2fvv8a-requirements.txt (line 8)), name='Markdown')
```

nateprewitt:
Yep, this is definitely unintended @jacebrowning. The casing issue is easy to fix here, the thing that may need a bit more thought is versions. `mkdocs` requires Markdown>=2.3.1, where `pdoc` requires Markdown<=2.5.0. Fortunately 2.4.1 meets both of these requirements and works, but I'm concerned about a case that we choose the last downloaded file and end up with an incompatible dependency.

I'm going to have to read up on whether pip auto-detects this sort of things, and what our options are from third-part or pypa libraries. If anyone has background here and would like to share or propose a solution, that would be helpful.

jacebrowning:
If pip doesn't detect this kind of behavior, maybe https://github.com/rbarrois/python-semanticversion (or another semver library) could help do something like this pseudocode:

```python
spec = Spec('>=2.3.1', '<=2.5.0')
for version in available_pip_downloads:
    if Version(version) in spec:
        return version
```

nateprewitt:
This should be resolved with #242 and will be released in 3.5.0.

Issue #204, opened 2017-02-06
installing packages from git broken


evertlammerts:
Fresh install:

```
$ pipenv --version
pipenv, version 3.3.6
```

Simple Pipfile:

```
$ cat Pipfile
[[source]]
url = "https://pypi.python.org/simple"
verify_ssl = true

[packages]
crayons = { git="https://github.com/kennethreitz/crayons.git", ref="v0.1.2" }
```

Creates what seems to me a valid lock file:

```
$ cat Pipfile.lock
{
    "_meta": {
        "hash": {
            "sha256": "39b2447b7dd6976804de490321577a2db0f4fc4327e0795a93436ea69afad9a8"
        },
        "requires": {},
        "sources": [
            {
                "verify_ssl": true,
                "url": "https://pypi.python.org/simple"
            }
        ]
    },
    "default": {
        "crayons from git+https://github.com/kennethreitz/crayons.git@v0.1.2#egg=crayons": {
            "hash": "sha256:a292b7ef29131345fd329ff5a94d1e2e59569be4dbd9fa8b99ff7114d72102a6",
            "version": "==0.1.2"
        },
        "colorama": {
            "hash": "sha256:a4c0f5bc358a62849653471e309dcc991223cf86abafbec17cd8f41327279e89",
            "version": "==0.3.7"
        }
    },
    "develop": {}
}
```

And now installing breaks:

```
$ pipenv install
No package provided, installing all dependencies.
Pipfile found at /[...]/Pipfile. Considering this to be the project home.
Installing dependencies from Pipfile.lock...
An error occured while installing!
Invalid requirement: 'crayons from git+https://github.com/kennethreitz/crayons.git@v0.1.2#egg=crayons==0.1.2'
It looks like a path. Does it exist ?


To activate this project's virtualenv, run the following:
 $ pipenv shell
```

These are the versions in my env:

```
$ pip freeze
appdirs==1.4.0
blindspin==2.0.0
click==6.7
click-completion==0.2.1
colorama==0.3.7
crayons==0.1.2
delegator.py==0.0.8
Jinja2==2.9.5
MarkupSafe==0.23
packaging==16.8
parse==1.6.6
pew==0.1.26
pexpect==4.2.1
pipenv==3.3.6
pipfile==0.0.1
ptyprocess==0.5.1
pyparsing==2.1.10
pythonz-bd==1.11.4
requests==2.13.0
requirements-parser==0.1.0
resumable-urlretrieve==0.1.4
[...]

PR #242, opened 2017-02-25
Fix VCS dependencies

nateprewitt:
~This is **DO NOT MERGE** until at least later today. Still need to find some time to clean up but want to allow some time for review from anyone who's interested.~

This is an attempt to address #204, #229 and #240 which all have the same underlying issue. When dealing with Pipfile and Pipfile.lock entries we've had a lot of issues with casing and name formatting in comparisons. We'd fixed this previously for package names specified by us, or pipenv users, but missed the case with mis-cased names in other packages' `install_requires` lists.

In order to prevent this from occurring in the future, I've transitioned to using a normalized naming scheme derived from pep426 where all names will be lower-cased and use dashes in place of underscores. This should allow uniform comparison and then we will rewrite the file after passing it through `proper_case` on the way out to ensure "pretty" package names.

This fixes part of the VCS problem but exposed another. We currently require hashes for ALL dependencies which pip can't support for packages that aren't hosted on PyPI. Due to this requirement, we will now split vcs dependencies off from PyPI packages before install to keep things working correctly.

This still has a few cosmetic/organization changes that need to be made, along with some testing for the new functions, but this "works" right now. If anyone has comments or wants to test this out you can clone the repository, run `git fetch origin pull/242/head:fix_vcs_dependencies`, and checkout the branch `fix_vcs_dependencies`.

### DIFF
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -166,6 +166,7 @@ def is_vcs(pipfile_entry):
         return any(key for key in pipfile_entry.keys() if key in VCS_LIST)
     return False
 
+
 def pep426_name(name):
     """Normalize package name to pep426 style standard."""
     return name.lower().replace('_','-')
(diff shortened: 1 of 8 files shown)


simnalamburt:
Thank you for making a big change! You did something which others could not. I have tested this branch with the command below, and I would like to make a proposal.

```bash
pip install --upgrade git+https://github.com/nateprewitt/pipenv@fix_vcs_dependencies
```

#### A. Let's make an error when users did not give the `#egg=` parameter.
The right way to install VCS dependencies is this:

```bash
pipenv install git+https://github.com/kennethreitz/requests.git#egg=requests

# [packages.requests]
# git = "https://github.com/kennethreitz/requests.git"
```

But users are able to forget this and use it like this:

```bash
pipenv install git+https://github.com/kennethreitz/requests.git

# [packages.none]
# git = "https://github.com/kennethreitz/requests.git"
```

With current implementation, pipenv just succeeds and installs the dependency as "none". I think this may cause confusion to some users. How do you think about fast-failing this and giving some error message to the users?

It'll be best if we can parse the setup.py and automatically decide the package name, but I have no idea how to implement it.

nateprewitt:
Great catch @simnalamburt, I'll look at getting that added in the morning. Thanks!

simnalamburt:
`crayons` module have never been imported into utils.py, but it was used. Please merge https://github.com/simnalamburt/pipenv/commit/2e840676 this commit before accepting the PR

nateprewitt:
Yep, that was definitely an oversight. Thanks @simnalamburt! Instead of importing crayons, I think we'll just pull the coloring use off the exception message because it's no longer surfaced in the cli.

pipenv/utils.py
```diff
@@ -144,3 +153,35 @@ def is_required_version(version, specified_version):
...
+        def handle_data(self, data):
+            # Remove extra blank data from https://pypi.org/simple
+            data = data.strip()
+            if len(data) > 2:
+                collected.append(data)
+
+    # Hit the simple API.
+    r = requests.get('https://pypi.org/simple/{0}'.format(package_name))
+    if not r.ok:
+        raise IOError('Unable to find package {0} in PyPI repository.'.format(crayons.green(package_name)))
```
nateprewitt: There's no longer a reason to import crayons as of 47b5ded :)

simnalamburt:
Looks good to me now!

nateprewitt:
Alright, I'm planning on leaving this up for a few more days for @kennethreitz to look over if he feels like it. After that I'll merge this in either Friday or Monday and try to push a 3.5.0 release shortly after.

simnalamburt:
No problem! 👍

kennethreitz:
:sparkles: :cake: :sparkles:

Issue #239, opened 2017-02-22
Recursive use of pipenv shell


OrkoHunter:
We would have to use `pipenv shell` while building a CLI project. So, once we switch to the new shell and suppose if we forgot and executed `pipenv shell` again it then creates a new shell once again and we have to exit two times to get back to the default shell.

It'd be great if we have already used `pipenv shell`, the next time it maybe gives a message saying "You're already in the pipenv shell".

nateprewitt:
Hey @OrkoHunter, could you tell me which shell you're using?

I currently get a warning in standard bash saying `Be aware that this environment will be nested on top of 'your-project-name'` when I try to activate the shell more than once. I believe most prompts are modified to also include the virtualenv name that pipenv is using after `pipenv shell` is used once. Is this not behaviour you're seeing?

OrkoHunter:
Hey @nateprewitt

I am using oh-my-zsh shell and I am not seeing any messages. :-/

nateprewitt:
Ok, we can look into trying to detect if the shell has been activated and raising a message. If we go down that route I'm almost wondering if we should just prohibit running pipenv shell from inside the shell. I'm not sure if there's a real use case for allowing that in the first place.

PR #243, opened 2017-03-01
Only spawn shell once

nateprewitt:
This addresses #239, and will place a restriction on shell creation with a maximum depth of 1 in the CLI. There doesn't seem to be a compelling reason for why a user would want to activate the virtualenv inside of itself, and this behaviour is causing issues for some users. pipenv will now simply print a message saying the shell is already activated and take no action.


kennethreitz:
:sparkles: :cake: :sparkles:

PR #255, opened 2017-03-04
Add sort_keys param to sort lockfile

KhasanovBI:
When I run pip update, a lot of requirements change position in lock file. 
It hard to understood, which packages updates and in git comes lot of unnecessary changes.


nateprewitt:
Hey @KhasanovBI, thanks for taking the time to put this together!

PR #266, opened 2017-03-12
add --ignore-hashes support

nateprewitt:
This will add the `--ignore-hashes` option to `pipenv install` which will allow the user to ignore package hash checking.

The implementation is a little funkier than I wanted, but if you include hashes anywhere in a requirements.txt file, it will require them for everything. This now requires us to purge all hashes read in from the Pipfile.lock before trying to install.


kennethreitz:
✨🍰✨

PR #268, opened 2017-03-13
Fix redundant '--python' warning on Windows

jacebrowning:
The warning about using `--python` on Windows is shown even when `--python` is used:

screen shot 2017-03-12 at 8 29 17 pm


nateprewitt:
This seems reasonable to me. Thanks @jacebrowning!

kennethreitz:
thanks!!

PR #292, opened 2017-03-27
Enables more environment variables for scriptability

isaacsanders:
- Adds environment variables for passing in executables (virtualenv,
  pew)
- Adds project root to list of environment variables
- Changes cwd when passing in a root path

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -24,7 +24,8 @@ from .utils import (convert_deps_from_pip, convert_deps_to_pip, is_required_vers
     proper_case, pep426_name, split_vcs, recase_file)
 from .__version__ import __version__
 from . import pep508checker
-from .environments import PIPENV_COLORBLIND, PIPENV_NOSPIN, PIPENV_SHELL_COMPAT, PIPENV_VENV_IN_PROJECT
+from .environments import PIPENV_COLORBLIND, PIPENV_NOSPIN, PIPENV_SHELL_COMPAT, \
+PIPENV_VENV_IN_PROJECT, PIPENV_VIRTUALENV_EXECUTABLE, PIPENV_PEW_EXECUTABLE, PIPENV_ROOT
 
 # Backport required for earlier versions of Python.
 if sys.version_info < (3, 3):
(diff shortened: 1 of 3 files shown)


kennethreitz:
i'm generally in favor of having fewer configuration variables, not more. i'll have to think about this. @nateprewitt any thoughts?

isaacsanders:
I wouldn't mind command line switches, but `PIPENV_PIPFILE` matches `BUNDLE_GEMFILE` for example.

The executables are there because I didn't seem to have a way to add them to the path via ssh-oneliner. I could just be wrong.

nateprewitt:
Hey @isaacsanders, thanks for taking the time to put this together. So, I'm a little confused as to why these are needed. Could you describe your scenario a bit more in depth? My current thought process is in line with @kennethreitz, that we probably don't want to expand environment params here.

The executables should be installed on the path if your remote server has python and pipenv installed properly. What is the goal of adding the custom paths?

As for `PIPENV_ROOT`, I'm not sure I understand this one either. This would limit pipenv to only being usable on one project. That, in most cases, probably isn't what a user would want. Especially if you run a non-idempotent command expecting it in your current directory, but your forgotten environment variable navigates to another directory.

I'm definitely open to hearing more detail here, but that's my initial reaction. We really do appreciate you taking the time to put this together though :)

isaacsanders:
Ok, removed the executable stuff. After trying some more things with different configurations, I realized I didn't need these to do what I was looking for.

As for the project root, it allows me to write crontab entries or ssh scripts that either run a script in the environment or install the environment on deploy. I am a Python neophyte, and don't know the intricacies of these other tools, which is why I have chosen to use pipenv. Being able to declare the environment I am using statically is something that is important to my use case.

nateprewitt:
So I would say that `cd $PROJECT_PATH && pipenv COMMAND <xyz>` is an equally as easy since you're modifying the bash environment already. This seems like an edge case which also has some possibly subtle "gotchas". That's keeping me leaning towards passing on this.

isaacsanders:
Ok. That works. I appreciate the help in illuminating my ignorance or <whatever the word is for not seeing a solution because you are focused on a certain level of technology>.

Issue #293, opened 2017-03-27
Install from Lockfile


isaacsanders:
Hey all,

I am working on deploying an app that is managing its dependencies with pipenv, and I was wondering if there was a way to install directly from the lockfile, as opposed to the lockfile being checked against and updated to match the pipfile at every `pipenv install`.

nateprewitt:
Hey again @isaacsanders :) I don't know if we have a way to force install from Pipfile.lock right now. It will always try to find a difference between the Pipfile and Pipfile.lock before installing. I do agree there are cases when you'd possibly want to skip this part and install directly from the lockfile. I think we can look into that, but I'd like to get @kennethreitz's thoughts here first.

kennethreitz:
hmm, and `--ignore-pipfile` option would be nice.

isaacsanders:
:D

nateprewitt:
Great! Alright @isaacsanders, if you'd like to put together a pull request that hands a `--ignore-pipfile` option from `def install(...)` to `def do_init(...)` and skip over [this conditional](https://github.com/kennethreitz/pipenv/blob/master/pipenv/cli.py#L542) when `ignore_pipfile` is passed. Otherwise, I'll toss this together later this afternoon.

isaacsanders:
Have you already gotten started? If not I can do it.

nateprewitt:
Go for it 😊

nateprewitt:
Resolved in #294

PR #294, opened 2017-03-27
Adds a flag for ignoring the Pipfile in favor of the Pipfile.lock

isaacsanders:
As discussed in #293.


nateprewitt:
Alright this looks good to me! Thanks @isaacsanders ✨

Issue #298, opened 2017-03-29
Dependency resolution order difference between pip and pipenv


c17r:
Took me awhile to track down what I was seeing as well as come up with the smallest test example.

So if you have a requirement of `pytest-django<2.10` which in it's setup.py has an `install_requires` entry of `pytest>=2.5`.  You also have a requirement of `pytest<2.10`.

Pip on it's own handles this just fine.  Pipenv ends up with the pytest at the current release version.

```shell
[test_3-9pMWYqV9][christian@WorkBookPro:~/tmp/test_3]
[18:51:42] $ ll
.
[-rw-r--r-- christian staff     121 Mar 29 18:29]  Pipfile
[-rw-r--r-- christian staff      31 Mar 29 18:51]  req.txt

0 directories, 2 files

[test_3-9pMWYqV9][christian@WorkBookPro:~/tmp/test_3]
[18:51:43] $ cat Pipfile 
[[source]]
url = "https://pypi.python.org/simple"
verify_ssl = true

[packages]
pytest = "<2.10"
pytest-django = "<2.10"

[test_3-9pMWYqV9][christian@WorkBookPro:~/tmp/test_3]
[18:51:44] $ cat req.txt 
pytest<2.10
pytest-django<2.10

[test_3-9pMWYqV9][christian@WorkBookPro:~/tmp/test_3]
[18:51:46] $ pip freeze

[test_3-9pMWYqV9][christian@WorkBookPro:~/tmp/test_3]
[18:52:27] $ pip install -r req.txt
Collecting pytest<2.10 (from -r req.txt (line 1))
  Using cached pytest-2.9.2-py2.py3-none-any.whl
Collecting pytest-django<2.10 (from -r req.txt (line 2))
  Using cached pytest_django-2.9.1-py2.py3-none-any.whl
Collecting py>=1.4.29 (from pytest<2.10->-r req.txt (line 1))
  Using cached py-1.4.33-py2.py3-none-any.whl
Installing collected packages: py, pytest, pytest-django
Successfully installed py-1.4.33 pytest-2.9.2 pytest-django-2.9.1

[test_3-9pMWYqV9][christian@WorkBookPro:~/tmp/test_3]
[18:52:35] $ pip freeze
py==1.4.33
pytest==2.9.2
pytest-django==2.9.1

[test_3-9pMWYqV9][christian@WorkBookPro:~/tmp/test_3]
[18:52:39] $ pip freeze

[...]

PR #305, opened 2017-04-05
enforce explicit version specifications in Pipfile

nateprewitt:
This should address #298 with a more robust version checking system. If a user specifies a requirement directly in the Pipfile (`Django<1.11.0`), Pipenv will now only add requirements to the Pipfile.lock that meet that spec. This will prevent dependencies of other packages from overwriting the explicit version in the Pipfile.

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -398,7 +398,7 @@ def parse_download_fname(fname, name):
 def get_downloads_info(names_map, section):
     info = []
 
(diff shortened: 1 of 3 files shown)


kennethreitz:
:sparkles: :cake: :sparkles:

nateprewitt:
This is unfortunately still not a complete solution as noted in the further discussion in #298. Still stewing on better options before we merge this.

PR #312, opened 2017-04-12
Adjusts proper casing to account for each source

### LINKED ISSUE #311: Non-public package not installing
Hey all,

I am working with a package that isn't in `https://pypi.org/simple` and when the CLI tries to `proper_case(package_name)` the package, my installation fails erroneously, as the package is in a different source that I have listed in my Pipfile. The error below:

```
Error:  Unable to find package <package-name> in PyPI repository.
```

Led me to the culprit. The `proper_case` function in the `util` module only statically checks the `https://pypi.org/simple` repo. It is not immediately obvious to me how to fix this, but it is not correct behavior.

isaacsanders:
Fixes #311

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -1,3 +1,7 @@
+[[source]]
+url = "https://pypi.org/simple"
+verify_ssl = true
+
 [dev-packages]
 pytest = "*"
 mock = "*"
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -99,7 +99,7 @@ def ensure_pipfile(validate=True):
     if validate and project.virtualenv_exists:
         # Ensure that Pipfile is using proper casing.
         p = project._pipfile
-        changed = ensure_proper_casing(pfile=p)
+        changed = ensure_proper_casing(pfile=p, urls=project.source_urls)
 
         # Write changes out to disk.
         if changed:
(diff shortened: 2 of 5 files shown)


kennethreitz:
Looks good to me! I'll wait for @nateprewitt to review though.

isaacsanders:
Still working on getting the CI tests to run clean.

nateprewitt:
Hey @isaacsanders, so as I said in your original issue, https://pypi.org/simple is hardcoded for a very specific reason. pypi.org is running pypa's new packaging service `warehouse` which returns compliant names. `https://pypi.python.org/simple` is running the legacy endpoint as are most third party endpoints. There's a bug in that code which returns arbitrarily formatted names, so depending on the package release, the name may be formatted differently.

We limited this process to only the one working endpoint since it's only for cosmetic purposes but has functionality implications. I'm -1 on changing this unless we have rigorous testing that we won't reintroduce the naming bugs we had for the first two months of the project. I'd like to investigate your issue further but I'm not convinced this is the correct solution.

isaacsanders:
I don't understand the need for proper casing. If the user can't install their package from a repository due to casing issues, they can fix the casing in their install. By gatekeeping with this endpoint, you keep anyone from using a repo that isn't pypi.org.

isaacsanders:
Rather than increase our flexibility on one side (accepting improperly cased names of packages), but stunting it on another (not accepting other sources than pypi.org), why not make it more rigid in the inputs we accept that are negotiable (the names of packages), rather than the non-negotiable ones, (the sources of packages).

nateprewitt:
@isaacsanders, it doesn't enforce proper casing. It does this as a nicety, and if it can't find the correct casing, it should gracefully move on without making any changes. I can't find a place in the current codebase where this exception can be raised, so I'm not sure why you're encountering this. I'd like to get to the bottom of why you're encountering this before we start patching things.

isaacsanders:
It appears that an old version of pipenv was causing this for me.

nateprewitt:
Alright, that makes sense. To confirm, the latest release 3.5.6 is working for you? If so I'd like to close this and #311 out as already fixed.

isaacsanders:
It's working.

Issue #270, opened 2017-03-13
Pipfile.lock from pipenv lock missing hash


jamessha:
I have a Pipfile that looks like
`[packages]
jupyter = "==1.0.0"
numpy = "==1.12.0"
pandas = "==0.19.2"
requests = "==2.13.0"
scikit-learn = "==0.18.1"
scipy = "==0.18.1"
alembic = "==0.8.10"
psycopg2 = "==2.6.2"
python-gflags = "==3.1.1"
nose2 = "==0.6.5"
h5py = "==2.6.0"
pydotplus = "==2.0.2"
tensorflow-gpu = "==1.0.1"
hyperopt = "==0.1"
lime = "==0.1.1.19"
SQLAlchemy = "==1.1.5"
Keras = "==1.2.2"

[requires]
python_version = "3.5"`

pipenv lock outputs a Pipefile.lock that looks like
`...
"future": {
            "hash": "sha256:e39ced1ab767b5936646cedba8bcce58239
8233d6a627067d4c6a454c90cfedb",
            "version": "==0.16.0"
        },
        "h5py": "==2.6.0",
        "html5lib": {
            "hash": "sha256:b8934484cf22f1db684c0fae27569a0db40
4d0208d20163fbf51cc537245d008",
            "version": "==0.999999999"
        },
...
`

which looks like it's missing the hash for h5py. However a hash for 2.6.0 seems to exist hash=sha256:cbb004160ddcfa5d75bc3f8894dbfe453
7913941fcba126c40e91be92a2fcf5a. I can workaround by manually adding the hash, however is this expected behavior?

Thanks!

nateprewitt:
Hey @jamessha, this is not expected behavior. I'm not at a computer right now, so I can't do any deep diving until later this evening. Could you confirm for me you're using pipenv 3.5.3? If you aren't, would you please upgrade and verify this problem still exists on the latest version on pipenv? Thanks!

jamessha:
Yes, 
`pipenv --version
pipenv, version 3.5.3`

jamessha:
I noticed that for packages this occurs there's a more recent version of the package of the same version but with "-1" appended, eg h5py-2.6.0-1 or matplotlib-2.0.0-1. I don't know enough about python package versioning to know what -1 means but pipenv seems to create hashes for these versions when a version isn't specified.

nateprewitt:
[...]

PR #314, opened 2017-04-13
strip implicit post releases from version numbers in Pipfile.lock

nateprewitt:
Alright, so after sitting on this for a month, I think I'm ready to pull the trigger on this. This PR addresses #270 by stripping "[implicit post release](https://www.python.org/dev/peps/pep-0440/#implicit-post-releases)" data out of the version number. This information, while important, seems to be handled under the hood by `pip`. Passing a version number with the implicit post release directly to `pip` doesn't work, so the reasonable option here seems to be removing it and allowing `pip` to give us the most recent post release.

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -392,6 +392,10 @@ def parse_download_fname(fname, name):
     # Substring out package name (plus dash) from file name to get version.
     version = fname[len(name)+1:]
 
+    # Ignore implicit post releases in version number.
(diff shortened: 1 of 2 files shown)


nateprewitt:
I'm going to merge this since it does fix the problem specified in #270 and we haven't been able to find a counter example of a valid case that `-` is needed in the version number. We'll move forward with this approach for 3.6.0 and reevaluate from there if more information arises.

PR #325, opened 2017-04-24
remove package from all locations in Pipfile

nateprewitt:
This is a follow up to https://github.com/kennethreitz/pipenv/issues/208#issuecomment-296668612, which aptly points out we've had a broken uninstall command for `dev-packages` for some time.

The semantics that were defined for `--dev` is that it uninstalls every package listed in `[dev-packages]` without removing them from the Pipfile. In order to remove a package the user *should* have been able to use `pipenv uninstall package_name` but that hasn't been working.

Since we don't have a way to specify where the package exists in the Pipfile, `pipenv uninstall package_name` should probably remove `package_name` from both `[packages]` and `[dev-packages]` for consistency. Having a different/conflicting dependency for `packages` vs `dev-packages` doesn't really make sense, so in theory it should only appear in one section.

I'm not convinced this is the best way to handle this, but it's consistent with what we've defined in the past. @kennethreitz, any further thoughts on current uninstall design?

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -849,10 +849,9 @@ def uninstall(package_name=False, more_packages=False, three=None, python=False,
 
     # Un-install all dependencies, if --all was provided.
     if all is True:
-        if not dev:
-            click.echo(crayons.yellow('Un-installing all packages from virtualenv...'))
-            do_purge(allow_global=system)
-            sys.exit(0)
+        click.echo(crayons.yellow('Un-installing all packages from virtualenv...'))
+        do_purge(allow_global=system)
+        sys.exit(0)
 
     # Uninstall [dev-packages], if --dev was provided.
     if dev:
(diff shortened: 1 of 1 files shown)


mhubig:
I was expecting the following behavior:

```
$ pipenv install django         # install it and add it to [packages]
$ pipenv uninstall --dev django
ERROR: no package named django in section [dev-packages]
$ pipenv uninstall django       # uninstall it and remove it from [packages]

$ pipenv install --dev green    # install it and add it to [dev-packages]
$ pipenv uninstall green
ERROR: no package named green in section [packages]
$ pipenv uninstall --dev green  # uninstall it and remove it from [dev-packages]

$ pipenv uninstall --all        # uninstall all packages, don't change Pipfile
$ pipenv uninstall --dev        # uninstall dev packages, don't change Pipfile
```

But if there is no way to specify where or if the package exists in the Pipfile, I think the following behavior would be OK:

```
$ pipenv install django         # install it and add it to [packages]
$ pipenv uninstall --dev django # don't uninstall nothing
$ pipenv uninstall django       # uninstall it and remove it from [packages]

$ pipenv install --dev green    # install it and add it to [dev-packages]
$ pipenv uninstall green        # don't uninstall nothing
$ pipenv uninstall --dev green  # uninstall it and remove it from [dev-packages]

$ pipenv uninstall --all        # uninstall all packages, don't change Pipfile
$ pipenv uninstall --dev        # uninstall dev packages, don't change Pipfile
```

nateprewitt:
@mhubig, I think we should clarify the thought process in how this is currently implemented. While I understand your desire to use `--dev` with uninstall it really isn't necessary.

Given the nature of `packages` being requirements for the application to function, they **MUST** be installed with every install. The `dev-packages` are then supplementary to add functionality only need for some development purpose. If you specify a dependency under `packages`, it shouldn't exist in `dev-packages` because that would overwrite a core requirement for the program. We don't explicitly forbid this right now, but it's not intentionally supported either. When you uninstall a dependency, it should only be in one place, so specifying `--dev` is irrelevant. We'll remove it from where it is, and section shouldn't matter (for uninstalls at least).

`--dev` very explicitly describes itself as "Uninstalls ALL dependencies from [dev-packages]", so modifying that functionality is a fundamental change which I don't think we're looking to do at this point.

kennethreitz:
yeah i don't want to change the current behavior.

nateprewitt:
@kennethreitz, this doesn't change the current behaviour. It maintains what we currently do, but does fix a missing edge case where we can't uninstall packages from dev. This will enable that without modifying what `--dev` does.

nateprewitt:
So if we're agreed that we want to keep `--dev` the same but have `pipenv uninstall package` remove the package from the Pipfile, then we should be ready to merge this.

PR #337, opened 2017-04-27
get python version from Pipfile's requires

urbaniak:
It's a first draft of feature described here https://github.com/kennethreitz/pipenv/issues/286.

When there's no python version (or --two/--three) specified it tries to figure out the python version based on python_version from [requires] section (pep508).

The match is simple, gets only first 3 chars (like 2.7 or 3.5) from python version and tries to find python executable for that version.

If it's not found - the behaviour is exactly the same as previous.

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -386,6 +386,16 @@ def parse_install_output(output):
     return names
 
 
+def get_python_from_requires():
+    python_version = project.parsed_pipfile.get('requires', {}).get('python_version')
+
+    if python_version:
+        for path in os.environ["PATH"].split(os.pathsep):
+            executable = os.path.join(path, "python%s" % python_version[:3])
+            if os.path.exists(executable) and os.access(executable, os.X_OK):
+                return executable
+
+
 def do_create_virtualenv(three=None, python=None):
     """Creates a virtualenv."""
     click.echo(crayons.yellow('Creating a virtualenv for this project...'), err=True)
@@ -407,6 +417,8 @@ def do_create_virtualenv(three=None, python=None):
             python = 'python2'
         elif three is True:
             python = 'python3'
+        else:
+            python = get_python_from_requires()
     if python:
         cmd = cmd + ['-p', python]
 


pipenv/cli.py
```diff
@@ -407,6 +417,8 @@ def do_create_virtualenv(three=None, python=None):
             python = 'python2'
         elif three is True:
             python = 'python3'
+        else:
+            python = get_python_from_requires()
```
nateprewitt: This is only usable with `pipenv install` which encourages a pattern that's been problematic in the past. It's usually a much better idea to initialize your environment before running the install command.

The unpredictability of the outcome here is also a negative in my eyes. If the user has to run `pipenv run python --version` to confirm that this didn't silently fail over to the system default python, then we're not saving much time. It may work well in highly controlled images but I don't think that's the average users setup. I'd lean back on `explicit is better than implicit` here and say it's just as easy, and more predictable, to pass `--python pythonX.X` with your install command.

pipenv/cli.py
```diff
@@ -386,6 +386,16 @@ def parse_install_output(output):
     return names
 
 
+def get_python_from_requires():
+    python_version = project.parsed_pipfile.get('requires', {}).get('python_version')
+
+    if python_version:
+        for path in os.environ["PATH"].split(os.pathsep):
+            executable = os.path.join(path, "python%s" % python_version[:3])
```
nateprewitt: This also doesn't fail gracefully for poorly formatted/bad values in `python_version`. We should probably handle that rather than blowing up on `python_version[:3]` with an less helpful traceback.

pipenv/cli.py
nateprewitt: We also are going to need testing built around this. It likely won't catch a lot of the edge cases we're concerned about, but it'll be good to have the expected best case behaviour guaranteed.

kennethreitz:
I had thought about implementing this myself, but I suspected that people would have problems with 2 vs 3 being available on their systems. We're going to have to stick to one (and only one) PEP 503 (I think) selector to base this behavior on, and document it, and also document that pipenv works best when you have both versions of python installed and available on your system. 

When a version of Python isn't available, we are going to have to give some very nice error messaging around this, and point them to python-guide.org with installation directions for making it available.

pipenv/cli.py
```diff
@@ -386,6 +386,16 @@ def parse_install_output(output):
     return names
 
 
+def get_python_from_requires():
+    python_version = project.parsed_pipfile.get('requires', {}).get('python_version')
```
kennethreitz: `python_version` is the right thing to base this on. I like that.

pipenv/cli.py
kennethreitz: +1 @nateprewitt. I think these are good changes though, and they are welcome. We'll just not want to jump the gun on them. Take it slow :)

pipenv/cli.py
```diff
@@ -386,6 +386,16 @@ def parse_install_output(output):
...
+
+    if python_version:
+        for path in os.environ["PATH"].split(os.pathsep):
+            executable = os.path.join(path, "python%s" % python_version[:3])
+            if os.path.exists(executable) and os.access(executable, os.X_OK):
```
kennethreitz: I like this!

pipenv/cli.py
kennethreitz: Let's get this updated!

kennethreitz:
add to whitelist

kennethreitz:
passing on this for now

PR #338, opened 2017-04-27
hash Pipfile and Pipfile.lock content instead of path

urbaniak:
I think it's better to include the Pipfile and Pipfile.lock hash instead of path hash in the virtualenv name.

It will prevent of re-creating the virtualenv when just moved the project and it will also "cache" the virtualenv for different Pipfiles.

It's quite useful when having multiple branches with different dependencies - don't need to re-install the virtualenv everytime and it will use correct virtualenv just after switch.

Another use case is having multiple projects with the same dependencies (for example multiple projects which require just ansible), they can safely share the same virtualenv when Pipfile.lock is the same.

It will also massively improve build time on CI runs when WORKON_HOME is shared across builds.

### DIFF
--- a/pipenv/project.py
+++ b/pipenv/project.py
@@ -63,10 +63,10 @@ class Project(object):
         #   http://www.tldp.org/LDP/abs/html/special-chars.html#FIELDREF
         #   https://github.com/torvalds/linux/blob/2bfe01ef/include/uapi/linux/binfmts.h#L18
         sanitized = re.sub(r'[ $`!*@"\\\r\n\t]', '_', self.name)[0:42]
+        encoded_hash = hashlib.sha256(open(self.pipfile_location).read().encode()).hexdigest()[:6]
 
-        # Hash the full path of the pipfile
-        hash = hashlib.sha256(self.pipfile_location.encode()).digest()[:6]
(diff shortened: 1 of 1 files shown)


nateprewitt:
Hey @urbaniak, thanks for taking the time to put this together!

Unfortunately, I think we're going to have to pass here for a few reasons. While this paradigm could in theory reduce the number of virtualenvs used, it optimizes for identical projects which I think is an edge case rather than the norm. It also has the problem of creating a lot of orphaned and unidentifiable virtualenvs that will pile up over time.

You'll also need to build and install the entire virtualenv everytime you perform a simple add or delete for a dependency.  At least until you have a sufficiently dense collection of "abandoned" virtualenvs that fill these permutations.

We also have the issue of two pipenv projects using the same virtualenv. While this seems like it may be helpful, it has downsides that I'd consider regressions of the original reason we put the hash in place. If you have a long running web app using the same dependencies as a project you're tinkering on, you may run `pipenv uninstall --all` and remove all the dependencies for your web app without realizing it. The subtleties here outweigh the benefits from what I can see. If we go back to allowing these kinds of accidental clobberings, we should just remove the hash all together.

Thanks again though for putting forth the effort here, we do appreciate the initiative! Let me know if you have any further thoughts.

Issue #343, opened 2017-05-03
Public/Private Dependency issues


isaacsanders:
Hey all,

I am back with a bug, and I have already checked that I am on the latest version of Pipenv ;)

__The Problem__

We have a couple of private dependencies and plenty of public dependencies.

When trying to install with `pipenv install` with no package, we always fail to install in a clean environment.

The issue is that no matter which source is used, the underlying `pip install -r <requirements>` command fails. Either the public dependencies don't exist in the private repo, or the private dependencies don't exist in the public repo, causing either command to fail.

I have found the part of the code that causes this. In the `pip_install` function in the `cli` module, the sources are tried until they succeed, and if all sources fail, only the last is reported. For awhile, it just looked like my installation couldn't find `Flask` and `gunicorn`, an error that I thought was strange, but could deal with by scripting in anger.

__A Solution__

One way of fixing this might be to store the source used for a package in the lockfile (or somewhere) and during install create separate `requirements.txt` files for each source to install from.

If that isn't possible/simple or there is an obviously better way to do this, I would be happy to help/write code for this.

__A Note__

I would also prefer it if every unsuccessful command could be logged, preferably in the format of command-followed-by-error. While it might be more information than others want, I might have figured out this bug weeks ago if I had seen one more error message, rather than just the last one. (This may be part of a PR, if I happen to be the one who writes it)

– Isaac

nateprewitt:
Hey @isaacsanders, thanks for opening this. Could you clarify a few things for me really quickly. Are you installing from a `Pipfile` or `requirements.txt`? Could you provide the contents of that file, and if you need to remove any sensitive data just make sure the structure of the requirement is preserved?

I'm not entirely sure how this would fail to find those packages since it *should* check every source before throwing an error. I'm particularly interested in the values of your sources again this time.

isaacsanders:
I am installing from a `Pipfile`, but `pipenv` appears to internally generate a `requirements.txt` tempfile.

nateprewitt:
@isaacsanders, it does create one, installs from it, and then deletes the file. Could you provide the contents of the Pipfile?

isaacsanders:
It does check every source, but either repo doesn't mirror every package in the `requirements.txt`, so it fails for both, each for missing a package that the other source has.

isaacsanders:
This is in the pattern of our `Pipfile`

```
[[source]]
url = "https://pypi.python.org/simple"
verify_ssl = true

[[source]]
[...]

PR #344, opened 2017-05-03
Adds all sources as extra indices to pip install and download commands

isaacsanders:
A fix for #343

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -591,40 +591,31 @@ def do_init(dev=False, requirements=False, allow_global=False, ignore_hashes=Fal
 
 
 def pip_install(package_name=None, r=None, allow_global=False, ignore_hashes=False):
-    # try installing for each source in project.sources
-    for source in project.sources:
-        if r:
-            install_reqs = ' -r {0}'.format(r)
-        else:
-            install_reqs = ' "{0}"'.format(package_name)
+    if r:
+        install_reqs = ' -r {0}'.format(r)
+    else:
+        install_reqs = ' "{0}"'.format(package_name)
 
-        # Skip hash-checking mode, when appropriate.
-        if r:
-            with open(r) as f:
-                if '--hash' not in f.read():
-                    ignore_hashes = True
-        else:
-            if '--hash' not in install_reqs:
+    # Skip hash-checking mode, when appropriate.
+    if r:
+        with open(r) as f:
+            if '--hash' not in f.read():
                 ignore_hashes = True
+    else:
+        if '--hash' not in install_reqs:
+            ignore_hashes = True
 
-        if not ignore_hashes:
-            install_reqs += ' --require-hashes'
+    if not ignore_hashes:
+        install_reqs += ' --require-hashes'
 
-        c = delegator.run('"{0}" install {1} -i {2}'.format(which_pip(allow_global=allow_global), install_reqs, source['url']))
-
-        if c.return_code == 0:
-            break
-    # return the result of the first one that runs ok or the last one that didn't work
-    return c
+    source_options = ' '.join(map(lambda surl: '--extra-index-url {0}'.format(surl), project.source_urls))
+    return delegator.run('"{0}" install {1} {2}'.format(which_pip(allow_global=allow_global), install_reqs, source_options))
 
 
 def pip_download(package_name):
-    for source in project.sources:
-        cmd = '"{0}" download "{1}" -i {2} -d {3}'.format(which_pip(), package_name, source['url'], project.download_location)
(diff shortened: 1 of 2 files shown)


pipenv/project.py
```diff
@@ -199,6 +199,10 @@ def sources(self):
         else:
             return [{u'url': u'https://pypi.python.org/simple', u'verify_ssl': True}]
 
+    @property
+    def source_urls(self):
```
nateprewitt: I'm not sure I understand why we need this instead of simply using `sources`. Why isn't iterating sources and calling the `url` key not sufficient?

pipenv/cli.py
```diff
@@ -591,40 +591,31 @@ def do_init(dev=False, requirements=False, allow_global=False, ignore_hashes=Fal
...
 
 
 def pip_download(package_name):
-    for source in project.sources:
-        cmd = '"{0}" download "{1}" -i {2} -d {3}'.format(which_pip(), package_name, source['url'], project.download_location)
-        c = delegator.run(cmd)
-        if c.return_code == 0:
-            break
-    return c
+    source_options = ' '.join(map(lambda surl: '--extra-index-url {0}'.format(surl), project.source_urls))
```
nateprewitt: Perhaps a list comprehension would be a bit easier to read? That will be more accessible to the broader python community and map doesn't provide any particular benefit here.

```python
source_options = ' '.join(['--extra-index-url {0}'.format(s['url']) for s in project.sources])
```

pipenv/cli.py
```diff
@@ -591,40 +591,31 @@ def do_init(dev=False, requirements=False, allow_global=False, ignore_hashes=Fal
...
+    cmd = '"{0}" download "{1}" {2} -d {3}'.format(which_pip(), package_name, source_options, project.download_location)
```
nateprewitt: With the removal of `-i` here, does this command always check PyPI first now? I think that may be a regression if that's the case.

For compatibility, it may be worth having the first entry in `sources` passed as the "primary" source and then add in additional ones via `--extra-index-url`. This may change pending discussion in the original issue though.

pipenv/cli.py
isaacsanders: I generally try to prefer functions over syntax for list iteration, so comprehensions aren't the tool I reach for.

pipenv/project.py
isaacsanders: I wish I could find one place to remove the duplication, but none of the choices seemed particularly good to me.

pipenv/cli.py
isaacsanders: The option drastically changes the way the code works. It would take a lot of time and effort to document the behavior, let alone determine the proper way to handle this.

nateprewitt: Could you clarify what you're referring to by option? Do you mean using `-i`?

This code, in its current state, is a breaking change. We currently pass the first source into `-i` which will be the primary repo we pull from. This is typically https://pypi.python.org/simple which is also pip's default but not guaranteed to be the first source a user provides.

If you only pass `--extra-index-url`, pip will default to checking https://pypi.python.org/simple first every time. So in order to maintain backwards compatibility here, we can't remove -i at this time unless we want to push this fix off to a 4.0.0 release.

pipenv/cli.py
nateprewitt: While I appreciate personal coding style choices, we'd like to keep the codebase's overall style as consistent as possible. Comprehensions are typically more accessible to the average python programmer than the map(lambda function) approach.

pipenv/project.py
nateprewitt: Could you clarify the duplication you're referring to? We're now iterating the sources content twice and creating a second data structure to represent the same data.

If there isn't a particular benefit of the list, it would be better to simply use the `sources` property.

pipenv/cli.py
isaacsanders: It was more of an explanation than an excuse.

pipenv/project.py
isaacsanders: The duplication still exists. It also prevents a breakage of encapsulation.

By the `cli` module needing to understand the structure of the source, the encapsulation of the `project` module is broken, where `source` is a concept. I suppose that might a be a better explanation for the method's existence.

kennethreitz:
Builds are failing.

kennethreitz:
add to whitelist

kennethreitz:
add to whitelist

pipenv/cli.py
kennethreitz: yes, let's not lambda, almost ever.

Issue #342, opened 2017-05-03
requirements.txt are not installed


kakulukia:
Hi! 

Some package(s) in my requirements obviously confuse pipenv so that nothing gets installed.
I end up with an empty Pipfile when executing _pipenv install_ with this requiremtens file present:

    appdirs==1.4.3
	appnope==0.1.0
	asn1crypto==0.22.0
	autopep8==1.3.1
	backports.shutil-get-terminal-size==1.0.0
	beautifulsoup4==4.5.3
	cffi==1.10.0
	cryptography==1.8.1
	decorator==4.0.11
	Django==1.10.7
	django-appconf==1.0.2
	django-bootstrap3==8.2.2
	django-compressor==2.1.1
	django-crispy-forms==1.6.1
	django-csp==3.0
	django-debug-toolbar==1.7
	django-extensions==1.7.8
	django-gravatar2==1.4.0
	django-imagekit==4.0
	django-libsass==0.7
	django-livereload-server==0.2.3
	django-modelcluster==3.1
	django-modeltranslation==0.12.1
	django-redis==4.8.0
	django-rosetta==0.7.13
	Django-Select2==5.10.0
	django-silk==1.0.0
	django-taggit==0.22.1
	django-treebeard==4.1.0
	django-undeletable==0.4.2
	django-webpack-loader==0.4.1
	djangorestframework==3.6.2
	easy-thumbnails==2.4.1
	Fabric3==1.13.1.post1
	flake8==3.3.0
	gnureadline==6.3.3
	google-api-python-client==1.6.2
	html5lib==0.999999999
	httplib2==0.10.3
	idna==2.5
	ipdb==0.10.3
	ipython==6.0.0
	ipython-genutils==0.2.0
	jedi==0.10.2
	Jinja2==2.9.6
	libsass==0.12.3
	MarkupSafe==1.0
	mccabe==0.6.1
	microsofttranslator==0.8
	oauth2client==4.0.0
	olefile==0.44
	packaging==16.8
	paramiko==2.1.2
	path.py==10.3.1
	pep8==1.7.0
	pexpect==4.2.1
	pickleshare==0.7.4
	pilkit==2.0
	Pillow==4.1.0
	polib==1.0.8
	prompt-toolkit==1.0.14
	psycopg2==2.7.1
	ptyprocess==0.5.1
	pyasn1==0.2.3
	pyasn1-modules==0.0.8
	pycodestyle==2.3.1
	pycparser==2.17
[...]

PR #347, opened 2017-05-07
create Pipfile from requirements.txt without installing

nateprewitt:
This PR is to address the issue pointed out in #342.

The install process is dependent upon the virtualenv existing which isn't always the case for fresh projects. This is the most common case this code will execute in, so we need to be able to create the Pipfile before the virtualenv is created. This will read the contents of the requirements.txt file and write them directly into a new Pipfile.

This approach required a little reworking of your initial implementation @kennethreitz, so I'd like you to sign off on this.

I removed the install portion completely because not only was it broken, but I don't know if it makes sense to install with *any* command passed to pipenv. Things like `pipenv shell` and `pipenv check` currently install the requirements.txt into the new environment which probably isn't what the user would expect.

I also removed the `BAD_PACKAGES` check because it *seems* to have been put in place to handle the fact those packages are always returned by `pip freeze`. If the user explicitly defines something like `appdirs` at a specific version in their requirements.txt file, I feel like we should move that to the Pipfile. They felt strongly enough to make it a requirement, so let's not silently delete it.

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -17,6 +17,7 @@ import pexpect
 import requests
 import pipfile
 from blindspin import spinner
+from pip.req.req_file import parse_requirements
 from requests.packages.urllib3.exceptions import InsecureRequestWarning
 
 from .project import Project
@@ -39,7 +40,7 @@ else:
 #         |_|
 
 # Packages that should be ignored later. 
-BAD_PACKAGES = ['setuptools', 'pip', 'wheel', 'six', 'packaging', 'pyparsing', 'appdirs']
+BAD_PACKAGES = ('setuptools', 'pip', 'wheel', 'six', 'packaging', 'pyparsing', 'appdirs')
 
 # Enable shell completion.
 click_completion.init()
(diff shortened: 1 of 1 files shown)


kennethreitz:
That's fine, but we need to strip comments from the requirements files. Installing avoids all of that.

nateprewitt:
@kennethreitz I think we should be handling that [here](https://github.com/kennethreitz/pipenv/pull/347/files#diff-e1e68886e05fef92bb74ed1699e5179eR110). I'm reading over the requirements.txt spec now and I may need to expand that check slightly. Pending that, I think this will be as permissive as the install option.

kennethreitz:
Have to be more careful of lines that have `#` after the requirement.

nateprewitt:
@kennethreitz Yep, we need expand the check to be stripping inline comments. I'll work on reintegrating the `BAD_PACKAGES` stripping too.

kennethreitz:
\o/

kennethreitz:
Basically, I want to err on the side of removing them, then revisiting that if people have problems.

nateprewitt:
Alright, this should be good to go. Feel free to merge if you don't have any further feedback or I will this afternoon.

kennethreitz:
✨🍰✨

PR #350, opened 2017-05-12
Fix for key error when uninstalling [dev-]packages

varunagrawal:
When a user tries to uninstall a package from `packages`, the `dev-packages` key throws a `KeyError` in the `_pipfile` OrderedDict if they have never installed a dev package. This could be the case for new users of Pipenv who are just trying out PoCs rather than full-fledged projects. This PR fixes this.

Similarly, if a user tries to `pipenv uninstall <package>` without ever installing that package, the `packages` key throws a `KeyError`. While this is unintended usage, the user would still see a stack trace and this manifests as poor workmanship, hence fixing that too.

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -906,7 +906,7 @@ def uninstall(package_name=False, more_packages=False, three=None, python=False,
 
         if pipfile_remove:
             norm_name = pep423_name(package_name)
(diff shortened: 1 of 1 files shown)


kennethreitz:
thanks! I ran into this myself.

PR #351, opened 2017-05-12
No hashes, default

kennethreitz:


nateprewitt:
@kennethreitz this seems perfectly reasonable to me. I do think this represents a breaking change though which *should* be a major bump. I had started a v4.0.0 branch to put this work on if that seems reasonable to you. It already has similar work on it with a `--require-hashes` option, but we can remove that in favor of this if you'd like.

If you'd rather skip the major bump then we can probably merge this straight in.

kennethreitz:
@nateprewitt this will be a major version bump. Let's do another major version bump for the next set of changes.

PR #364, opened 2017-05-17
handle package[extra]==version properly

c17r:
Previously doing `pipenv install "django-rest-auth[with_social]==0.8.2` would
lose the extra information and if you edited the Pipfile by hand to
```
[[source]]
verify_ssl = true

[packages.django-rest-auth]
version = "==0.8.2"
extras = [ "with_social",]
```

`pipenv update` would break with a malformed pip command.


nateprewitt:
Hey @c17r, thanks for this! I think this looks good, so assuming the tests come back green, we should be set to merge this.

PR #395, opened 2017-06-03
use virtualenvs when they're available

kennethreitz:


jstasiak:
I'd like to provide a data point here. I routinely operate with a user-wide virtualenv activated (to have some Python tools available). After upgrading pipenv it removed packages from that virtualenv instead of manipulating project-specific virtualenv (which would be expected by me).

nateprewitt:
Hey @jstasiak, this seems to be a side effect of automatically selecting an activated virtualenv instead of using the current project. If you'd like to revert to the old functionality, I believe you're going to have to set the `PIPENV_IGNORE_VIRTUALENVS` environment variable.

PR #398, opened 2017-06-07
run: Activate virtualenv before os.exec

nonylene:
Activate virtualenv before os.exec in `pipenv run`. 

This will resolve [the problem around PATH](https://github.com/kennethreitz/pipenv/issues/378#issuecomment-306855801) mentioned by @jacebrowning .

- About `activate_this.py` : https://virtualenv.pypa.io/en/stable/userguide/#using-virtualenv-without-bin-python
- About `exec` : https://stackoverflow.com/a/437857/4205654

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -651,9 +651,9 @@ def pip_download(package_name):
     return c
 
 
-def which(command):
+def which(command, exe=True):
     if os.name == 'nt':
-        return os.sep.join([project.virtualenv_location] + ['Scripts\{0}.exe'.format(command)])
+        return os.sep.join([project.virtualenv_location] + ['Scripts\{0}{1}'.format(command, '.exe' if exe else '')])
     else:
         return os.sep.join([project.virtualenv_location] + ['bin/{0}'.format(command)])
 
@@ -1042,6 +1042,12 @@ def shell(three=None, python=False, compat=False, shell_args=None):
 def run(command, args, three=None, python=False):
     # Ensure that virtualenv is available.
     ensure_project(three=three, python=python, validate=False)
+
+    # Activate virtualenv under the current interpreter's environment
+    activate_this = which('activate_this.py', False)
+    with open(activate_this) as f:
+        code = compile(f.read(), activate_this, 'exec')
+        exec(code, dict(__file__=activate_this))
 
     command_path = which(command)
 


nonylene:
I will re-run test with #399 :)

nateprewitt:
Hey @nonylene, sorry I haven't had the chance to review this yet. One thing I'd definitely need before we merge this is passing tests on Windows. We'll need to get @kennethreitz to reenable those before we can move forward here.

It would also expedite the process if you wouldn't mind throwing together a quick rundown of what this change is accomplishing and why it's needed. I realize the links are included, but wading through those takes additional time that I haven't had a chance to set aside yet.

I'm also really not excited about executing arbitrary code from a directory, so if there's another approach to solve this issue, I'd love to see that. That way we can properly weigh the need for this, and what the tradeoffs of this approach is to another.

Thanks!

nonylene:
Thank you for response!

> It would also expedite the process if you wouldn't mind throwing together a quick rundown of what this change is accomplishing and why it's needed. 

[`activate_this.py`](https://github.com/pypa/virtualenv/blob/master/virtualenv_embedded/activate_this.py) modifies `os.path` , `sys.path` and `sys.prefix` to bootstrap virtualenv. 

`(virtualenv home)/bin` will be included in `$PATH` environment variable after executing.

### PATH Examples

```sh
# With activated virtualenv, `(virtualenv home)/bin` directory is inlcuded in $PATH
$ pipenv shell
Launching subshell in virtual environment. Type 'exit' or 'Ctrl+D' to return.

$ echo $PATH
/home/nonylene/.local/share/virtualenvs/pipenv-CnA1puTH/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin
$ exit

# $PATH is not changed with current released pipenv
# Sometimes this `$PATH` difference cause problems like https://github.com/kennethreitz/pipenv/issues/378#issuecomment-306855801.
$ pipenv run python3 -c "import os; print (os.environ['PATH'])"
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin


# Correct $PATH will be set with modified pipenv
$ python -m pipenv run python3 -c "import os; print (os.environ['PATH'])"
/home/nonylene/.local/share/virtualenvs/pipenv-CnA1puTH/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin
```

`sys.path` or `sys.prefix` modifications are not related here, because pipenv runs `os.exec` just after `activate_this.py` is executed.

> I'm also really not excited about executing arbitrary code from a directory, so if there's another approach to solve this issue, I'd love to see that.

- I think running `source (virtualenv home)/bin/activate` also can be said as "executing arbitrary code". Both scripts are given by virtualenv and can be modified by users.
- We can set environment variables manually and _activate virtualenv_.
  - `$PATH` and `$VIRTUAL_ENV` should be set to emulate `bin/activate`.

>  That way we can properly weigh the need for this, and what the tradeoffs of this approach is to another.

pros: `activate_this.py` is created by virtualenv, so we can delegate activating process. 
cons: This executes arbitrary code inside running python interpreter for little effect. 

Either approaches will be fine for me 😃.

jacebrowning:
I haven't tested this, but it sounds like this might fix https://github.com/kennethreitz/pipenv/issues/249 as well.

pipenv/cli.py
```diff
@@ -651,9 +651,9 @@ def pip_download(package_name):
     return c
 
 
-def which(command):
+def which(command, exe=True):
```
kennethreitz: `exe` should not default to `True`.

kennethreitz:
I think this problem is caused by problems with your shell setup, not problems with pipenv.

kennethreitz:
e.g. You should only be setting `PATH` during a login session. The [pew docs](https://github.com/berdario/pew) have lots of information on how to setup your shell's config properly.

nonylene:
I think `PATH` should be changed among environments (original `virtualenv` changes `PATH`).

For example, the `in` command of `pew` modifies `PATH` and `VIRTUAL_ENV` mannualy before `subprocess.check_call`. 

https://github.com/berdario/pew/blob/f10e80f85fddf5c3246d85224fe72af87cc92c92/pew/pew.py#L596

kennethreitz:
merged manually

kennethreitz:
✨🍰✨

nonylene:
Thanks!

nonylene:
Hi, @kennethreitz, 4e5f59a seems to modify wrong function?

 `shell` function already activates virtualenv by `'source {0}/bin/activate{1}'`.
https://github.com/kennethreitz/pipenv/blob/4e5f59a5122801bac0f184a6fdc28e41c9b98b6e/pipenv/cli.py#L1101

In `run` function, `activate_this.py` is still not called. 

### Test script

This will cause error with `Error: [Errno 2] No such file or directory: 'py.test'`, because `$PATH` is not changed.

```shell
$ pipenv install pytest-watch
$ pipenv run ptw
```
Below commands works well, because `source {0}/bin/activate` is called and `py.test`exists in `$PATH`.

```shell
$ pipenv shell
$ ptw
```

kennethreitz:
@nonylene to be clear, this is an issue with your shell configuration.

kennethreitz:
v5.4.2 released, which adds this to run.

PR #399, opened 2017-06-07
Add test for pipenv run

nonylene:
Add small `pipenv run` test ( added in order to ensure no regression by #398 ) !


nateprewitt:
Thanks @nonylene!

PR #405, opened 2017-06-10
Vendor delagator.py update

erinxocon:
In the process of splitting up commits now that work is over.  This one updates the delegator.py to the latest version, v0.0.12, which will allow for custom timeouts to be implemented. I also realize I spelled delegator incorrectly in the branch name :(


nateprewitt:
Cool, this looks good to me. Thanks @erinxocon! ✨

PR #403, opened 2017-06-09
Added --timeout to allow custom timeouts for command execution

erinxocon:
This pull requests addresses issues #65 and #401. On slower machines, or machines connected to a slow connection, virtualenv may have problems completing within the default 30 seconds.   These changes allow a user to specify something like `pipenv --two --timeout 60` will allow 60 seconds for the virtualenv to be completely created instead of the 30.  I've tested this on my Raspberry Pi 0 which takes 57 seconds to create a virtualenv.  Setting `--timeout 80` let it complete and I'm happily using pipenv to deploy code to the pi zero!  I updated `delegator.py` in the vendor release, and I put a pull request on on the actual delegator.py repo containing the version I'm using here!

PR #406, opened 2017-06-10
Timeouts test

erinxocon:
Need to sqaush some of these commits, but it should be merged after PR #403 is pulled in


erinxocon:
@kennethreitz This will fail until #403 is merged.

nateprewitt:
Sorry, I didn't notice this sooner. It would be great to have these in with #403 so we're certain they work as expected. Would you mind cherry-picking these commits into that PR?

erinxocon:
no problem, was unsure if I should split up the tests into their own PR!

PR #416, opened 2017-06-15
Added logic to parse VCS links from a requirements file

erinxocon:
into it's proper Pipfile format

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -110,7 +110,9 @@ def ensure_pipfile(validate=True):
             for package in reqs:
                 if package.name not in BAD_PACKAGES:
                     if package.original_link is not None:
-                        project.add_package_to_pipfile(str(package.original_link))
+                        pack_link = str(package.original_link)
+                        pack = '-e {0}'.format(pack_link) if package.editable else pack_link
+                        project.add_package_to_pipfile(pack)
                     else:
                         project.add_package_to_pipfile(str(package.req))
 
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -109,10 +109,9 @@ def ensure_pipfile(validate=True):
 
             for package in reqs:
                 if package.name not in BAD_PACKAGES:
-                    if package.original_link is not None:
-                        pack_link = str(package.original_link)
-                        pack = '-e {0}'.format(pack_link) if package.editable else pack_link
-                        project.add_package_to_pipfile(pack)
+                    if package.link is not None:
+                        package_string = '-e {0}'.format(package.link) if package.editable else str(package.link)
+                        project.add_package_to_pipfile(package_string)
                     else:
                         project.add_package_to_pipfile(str(package.req))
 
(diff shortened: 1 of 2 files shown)


pipenv/cli.py
```diff
@@ -105,11 +105,16 @@ def ensure_pipfile(validate=True):
             # Pip requires a `PipSession` which is a subclass of requests.Session.
             # Since we're not making any network calls, it's initialized to nothing.
             from pip.req.req_file import parse_requirements
-            reqs = [r.req for r in parse_requirements(project.requirements_location, session='')]
+            reqs = [r for r in parse_requirements(project.requirements_location, session='')]
 
             for package in reqs:
                 if package.name not in BAD_PACKAGES:
-                    project.add_package_to_pipfile(str(package))
+                    if package.original_link is not None:
```
nateprewitt: Can `package.original_link` actually be `None`? It doesn't look like it from the constructor.

pipenv/cli.py
```diff
@@ -105,11 +105,16 @@ def ensure_pipfile(validate=True):
...
+                        pack_link = str(package.original_link)
```
nateprewitt: What type are we expecting link to be here? Is this to call `__str__` on a custom pip object?

pipenv/cli.py
erinxocon: When I read in a requirements file manually, origial_link was None for all but my custom repositories, for multiple different configurations of a requirements file.

pipenv/cli.py
erinxocon: Yes, the intention was to call a str magic method on a requirement object, although original link might always be a string, I put it in there with some other changes and it was working, so this could be removed, I'll test.

erinxocon:
Sure thing, I'l throw together some new tests and commit them up here

pipenv/cli.py
jjangsangy: It mind be better to call  text_type from https://github.com/kennethreitz/pipenv/blob/22b9fb47a62fabf5dc0a23c9ffd98fb85d5188b7/pipenv/vendor/markupsafe/_compat.py#L22 for Python 2/3 compat.

erinxocon: Looked into it.  `original_link`  and `link` is of type `pip.index.Link`.  I changed up the logic a bit to remove making it a string, since .format will take care of it in one of the scenarios.

jjangsangy:
👍 Sweet, ran this over a couple projects and was able to convert all my `requirements.txt` to lock files!

nateprewitt:
Hey @erinxocon, it looks like the test you added is currently failing. I'm not immediately sure why that is, but if you can work that out, I think we're ready to merge.

erinxocon:
@nateprewitt I'm gonna squash some of these commits, but I got the test working.  Since this test doesn't involve `touch Pipfile` it was searching up the tree.  Setting PIPENV_MAX_DEPTH just for that test prevents that and forces it to only look in the current directory (to my understanding).  Anyways let me know if you need anything else!

erinxocon:
@nateprewitt Happy to help! ✨ 🍰 ✨ Sorry for the extended delay, I've been rebuilding a NAS all week and this got put on the back burner for a bit!

PR #425, opened 2017-06-29
Adding --pipfile flag to 'pipenv install'

JackDanger:
For building a project from a directory other than root we'll need to be able to specify the Pipfile location. Does not conflict with --ignore-pipfile

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -793,16 +793,21 @@ def cli(ctx, where=False, venv=False, rm=False, bare=False, three=False, python=
 @click.option('--dev', '-d', is_flag=True, default=False, help="Install package(s) in [dev-packages].")
 @click.option('--three/--two', is_flag=True, default=None, help="Use Python 3/2 when creating virtualenv.")
 @click.option('--python', default=False, nargs=1, help="Specify which version of Python virtualenv should use.")
+@click.option('--pipfile', '-p', default='Pipfile', nargs=1, help="Select which Pipfile to use")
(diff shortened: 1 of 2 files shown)


nateprewitt:
Hey @JackDanger, thanks for taking the time to contribute to Pipenv!

I've got a couple comments on the general implementation here. The first being while we've added this for install, we don't have a symmetric option for uninstall. That means we won't necessarily be able to remove packages from our specified Pipfile. We would need to make sure this works uniformly before merging. The second question is whether this is really needed. Do we gain anything here other than avoiding changing directories when running a command? Is this any different than just activating `pipenv shell`?

While I can see some cases that may get better in adding this, it encourages a paradigm of multiple Pipfiles for a project, which I believe @kennethreitz wanted to avoid. I'm leaning towards passing on this feature, but would like to get Kenneth's thoughts before we make a decision here. Also if I'm missing something, please feel free to correct any of my assumptions, @JackDanger. Thanks again for the effort here!

JackDanger:
Thank you for the great response. I suspect you're right that this isn't necessary. It felt very useful when I was copying a Pipfile into a Docker container at build time because I wanted to do something like this to ensure the dependencies were installed early in the build process (and therefore cached so later builds would be fast):
```
COPY ./Pipfile /tmp/Pipfile
RUN pipenv install --pipfile /tmp/Pipfile
```
But you're right that changing directories is trivial and solves my problem.

Because all code is technical debt I am closing this PR and I'll jump in when I find something that I think can contribute clearer value to this project. Thanks!

Issue #396, opened 2017-06-06
OpenSSL.SSL.Error: Missing cacert.pem in vendorized `requests` 2.13.0?


msegado:
When I run `pipenv install` I get the a stack trace ending in
```
OpenSSL.SSL.Error: [('system library', 'fopen', 'No such file or directory'), ('BIO routines', 'BIO_new_file', 'no such file'), ('x509 certificate routines', 'X509_load_cert_crl_file', 'system lib')]
```

This may be caused by the bundled `requests` module (v2.13) missing a `cacert.pem` file... see requests/requests#4019 ("OpenSSL issue when vendoring requests"). I can confirm that trying `import requests; requests.get(...)` with `pipenv/vendor/requests` on my path causes the error above while the same command using my system-wide `requests` (v2.14.2) does not.

Related: #387 ("Strange error after upgrade on ubuntu 16.04")

aeleos:
I had the same issue on ubuntu 16.04. It started when I installed certbot (Lets Encrypt certificate managment tool) and fixed itself when I uninstalled it with ```sudo apt-get --purge certbot && sudo apt-get autoremove```  

I am not sure which package specifically causes the issue, it could be the bundled requests package in their [repo](https://launchpad.net/~certbot/+archive/ubuntu/certbot) or it could be that it requires python-openssl to be installed.

Hopefully this helps someone figure out the actual issue as this error is really annoying.

msegado:
From what I remember when troubleshooting a few weeks ago, `cacert.pem` wasn't present in my installed version of pipenv; however, since I see it *is* present in the repo here, I suspect the issue might be that it's not being bundled into the pipenv wheel / uploaded to PyPI...

nateprewitt:
aha! Thanks @msegado, it is in our repo, but it isn't being bundled as you can see by inspecting the [5.1.0 archive](https://pypi.python.org/packages/37/af/73e27a4711128f7bc1e0c46722c0dcc0a7313223efcc076a0212fb6d9806/pipenv-5.1.1.tar.gz#md5=20778b03c20269228fb2705a7a3511c6). This is an issue with our release process. I believe we'll need to add an entry to MANIFEST.in to include `cacert.pem`.

nateprewitt:
Alright, I've pushed 35518da to address this.

@msegado and @aeleos, feel free to try installing the master branch and seeing if you're still experiencing the issue. If not, this will be included in our next release.

iamthad:
[...]

PR #456, opened 2017-08-31
Fix for SSLError: cacert.pem missing

nils-werner:
The file `cacert.pem` was never included in wheels because the setuptools flag `include_package_data=True` wasn't set, causing issue #396.

This PR fixes that.

I don't know where you're trying to go with future versions, as your `pipenv/vendors` directory is almost empty now. So I have based this branch on the `v5.1.2` tag.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -73,4 +73,5 @@ setup(
         'Programming Language :: Python :: Implementation :: CPython',
         'Programming Language :: Python :: Implementation :: PyPy'
     ],
+    include_package_data=True,
 )


nateprewitt:
Hey @nils-werner, thanks for taking the time to put this together. Unfortunately we'd already added this in 06acad1 but it seems a release was never cut with that fix.

We're currently in a state where we're moving away from vendoring, but have had some CI issues I haven't had time to address yet. For that reason we really can't cut a release with a fix for this but I'll try to get something out asap.

Thanks again for your contribution!

nils-werner:
Maybe this could go in a hotfix `5.1.3` release, to fix the issue until the new release without vendoring is released?

nateprewitt:
We may be able to go back to 06acad1 and cut a release from that point but the vendoring removal commits have already been merged into master.

I'm currently traveling without a laptop so I'm unable to verify that tests are working across platforms at that commit. If @kennethreitz has the time to look into this, that may be an option. 

I'm however slightly uncomfortable with doing this. Either placing the cacerts.pem file on the path or installing requests 2.14 or earlier should fix this temporarily. I think we can get this fixed in the next couple days on master which would be preferred.

I'll leave it to Kenneth to make a final call here.

PR #466, opened 2017-09-02
Re-enable flake8 testing, add build stages to Travis testing

erinxocon:
I fixed some of the style errors flake8 was showing from the changes we made yesterday.  I also excluded the setup.py file, which I suppose isn't necessary since we are only testing flake8 on 3.6 and the reason for exclusion was a 2.7 issue.  I've also added build stages so the style test only runs after all of the pytest suites run.  Seemed complicated but it works.


kennethreitz:
this looks ready to merge!

PR #472, opened 2017-09-05
Windows tests

erinxocon:
This is essentially just a port of the existing tests to run on windows.  All of the work was done converting the `tests/test_pipenv.py` containing the original Unix tests.  I've just replaced Unix commands with cmd/ps commands where appropriate.  A rewrite of the tests using the `os` and `shutil` modules could be done to eliminate the two sets of tests, but this works for now.  Plus we know that `delegator` really works well with windows cmd/ps commands!  I didn't have to port `tests/test_project.py` or `tests/test_utils.py` as they ran fine on windows already with no platform specific code.  The only problem is that app veyor fails for python 2.7, despite all the tests running fine on my computer.  I couldn't figure it out as I don't know as much about app veyor.  @kennethreitz or @nateprewitt any ideas?


kennethreitz:
Merged. We'll fix windows 2.7 together later :)

erinxocon:
Fixed with PR #481

PR #474, opened 2017-09-05
hyphen missing in pip index argument

jgelens:
The index argument was not being parsed because of the missing hyphen.


kennethreitz:
thanks!

PR #481, opened 2017-09-06
Vendorize background

erinxocon:
Since futures is a lot different from 2.7 to 3.4 I vendorized both and renamed some namespaces. 
 Did a version check to decide which one to use...it's seems messy.


pipenv/vendor/background.py
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env python
+# -*- coding:utf-8 -*-
+
+import sys
+import multiprocessing
+
+if sys.version_info.major < 3:
```
kennethreitz: concurrent.futures is built in on 3.4+ i believe.

erinxocon: 3.4 actually!  removed the vendorized version for 3

erinxocon:
@kennethreitz python3 now uses the built in futures, while python2 uses the vendorized one.  All green across the board!  Ready to merge?

kennethreitz:
✨🍰✨

PR #496, opened 2017-09-08
Windows patch

erinxocon:
This is a patch for windows.

There were two main issues:

1) `shellquotes()` from the `utils.py` module was not returning a valid windows command.  I changed it to something that both Unix and windows can understand.  This fixed a lot but not all of the failures.  

2) There was an issue in the test from the windows `test_pipenv.py` - `test_pipenv_run`.  This test was failing because there wasn't a virtualenv for it to install stuff too since `pipenv install --python` was never being run.  When this test failed, it got stuck in that directory, having never cleaned itself up properly.  This caused a bunch of issues with Mock and once `test_pipenv_run` was fixed the rest of the tests ran as they should.


pipenv/cli.py
```diff
@@ -1251,7 +1251,7 @@ def update(dev=False, three=None, python=None, dry_run=False, bare=False, dont_u
     ensure_project(three=three, python=python, validate=False)
     # --dry-run
     if dry_run:
-        dont_upgrade = True
+        # dont_upgrade = True
```
erinxocon: @kennethreitz this variable is assigned but never used.  I commented it out for now, but do you intend to use it going forward or can it be removed?

Issue #499, opened 2017-09-08
Version resolution mismatch between PyPI and pip for non-semver packages causes keyerrors when installing dependencies


techalchemy:
The pip version resolver strips leading zeroes for semver compliance when parsing version information for packages, but the pypi json api retains them.  

```bash
 ~/g/pipenv   master $+…  vf tmp --python=python3.6          
Running virtualenv with interpreter /usr/bin/python3.6
Using base prefix '/usr'
New python executable in /home/hawk/.virtualenvs/tempenv-18a097382f8c/bin/python3.6
Also creating executable in /home/hawk/.virtualenvs/tempenv-18a097382f8c/bin/python
Installing setuptools, pip, wheel...done.
◰³ tempenv-18a097382f8c  ~/g/pipenv   master $+…  pip3.6 install pipenv 
◰³ tempenv-18a097382f8c  ~/g/pipenv   master $+…  cd /tmp/
◰³ tempenv-18a097382f8c  /tmp  mkdir newdir
◰³ tempenv-18a097382f8c  /tmp  cd newdir
◰³ tempenv-18a097382f8c  /t/newdir  pipenv install dateparser
Creating a Pipfile for this project...
Installing dateparser...
Collecting dateparser
  Using cached dateparser-0.6.0-py2.py3-none-any.whl
Collecting python-dateutil (from dateparser)
  Using cached python_dateutil-2.6.1-py2.py3-none-any.whl
Collecting pytz (from dateparser)
  Using cached pytz-2017.2-py2.py3-none-any.whl
Collecting ruamel.yaml (from dateparser)
  Downloading ruamel.yaml-0.15.33-cp36-cp36m-manylinux1_x86_64.whl (557kB)
Collecting tzlocal (from dateparser)
Collecting regex (from dateparser)
Requirement already satisfied: six>=1.5 in /home/hawk/.virtualenvs/tempenv-18a097382f8c/lib/python3.6/site-packages (from python-dateutil->dateparser)
Installing collected packages: python-dateutil, pytz, ruamel.yaml, tzlocal, regex, dateparser
Successfully installed dateparser-0.6.0 python-dateutil-2.6.1 pytz-2017.2 regex-2017.7.28 ruamel.yaml-0.15.33 tzlocal-1.4

Adding dateparser to Pipfile's [packages]...
Locking [dev-packages] dependencies...
Locking [packages] dependencies...
Traceback (most recent call last):
[...]

PR #501, opened 2017-09-08
Pypi pip version bugfix

techalchemy:
* Uses pip version parser to parse package versions for semver compliance enforcement
* Uses this for json api version numbers in PyPI which are not represented in a way that is consistent with pip
* Opted to use the pip parsing standard for our codebase because it is implemented in the language and readily accessible

See #499


techalchemy:
I have a fix coming for the build failure shortly

PR #506, opened 2017-09-09
Automatically set system flag if running command with abs path

dschep:
paths starting with the os separator are absolute paths and shouldn't be resolved with the `which` helper and should have the virtualenv activated automatically for them.


kennethreitz:
✨🍰✨

uranusjr:
There is `os.path.isabs` that works on Windows. This patch doesn’t.

PR #517, opened 2017-09-10
Use os.path.isabs to detect absolute path commands

uranusjr:
See https://github.com/kennethreitz/pipenv/pull/506#issuecomment-328366839


kennethreitz:
✨🍰✨

PR #519, opened 2017-09-10
Leverage pip's requests to allow for error catching

JBKahn:
I logged the issue here: https://github.com/pypa/pip/issues/4717

Basically, the only thing standing in our way of using a third party private server (gemfury in this case) is that the pip code only catching request errors if you use their session when calling it.

What are your thoughts on this?

### DIFF
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -58,7 +58,7 @@ def resolve_deps(deps, sources=None, verbose=False):
 
     pip_options, _ = pip_command.parse_args(pip_args)
 
-    pypi = PyPIRepository(pip_options=pip_options, session=requests)
+    pypi = PyPIRepository(pip_options=pip_options, session=pip._vendor.requests)
 
     if verbose:
         logging.log.verbose = True


kennethreitz:
looks good!

JBKahn:
Awesome, when are you planning on cutting your next release?

JBKahn:
Also, the tests never finished running and there might be things to fix for broken assertions, would you like me to help fix them?

kennethreitz:
Sure!

JBKahn:
I took a quick look and it doesn't seem to have broken any tests, thankfully.

That being said, this PR into pip will break some of the paths you have in the code base and pip doesn't seem to be locked to a particular version in some way. If I have time before they merge 
 the code they discussed in https://github.com/pypa/pip/issues/4717 then I'll update the paths in the code to fall back to the old ones if the new ones aren't valid yet.

kennethreitz:
My plan is to depend hard on latest pip

JBKahn:
K, then I'll prepare a PR if I have time.

PR #543, opened 2017-09-14
Fixes for the failing tests

erinxocon:
All the unix tests are working.  

Windows 2.7 has unicode errors
Windows 3.x can't find python3.6 for all versions of 3.  I suspect it's because of the pep508 identifiers.  The unix tests throw and error for this but it doesn't stop it from passing.  

Will try and fix windows in the morning.


kennethreitz:
You're amazing!

PR #553, opened 2017-09-15
fix a bug on windows - can not get the python version

neoni123:
I find a bug on windows.
The program can not get the python version when writing the pipfile.


erinxocon:
Thanks for this!  Still need to do some work on windows tests, so I'm merging it into this branch for more work!

PR #555, opened 2017-09-15
Windows tests

erinxocon:
This is a fix for the windows tests.  Manual testing confirmed it was working, but the tests were not updated to reflect some changes in the cli api as well as changes in the pipfile spec.  

The tests are still failing for one reason though:  tablib is not letting `shutil` do its work in one of the tests.  Seems like git is hanging on a hash keeping the file open, which then can't get removed by `shutil`

```>                   os.unlink(fullname)
E PermissionError: [WinError 5] Access is denied: 
'test_requirements_to_pip\\.venv\\src\\tablib\\.git\\objects\\pack\\pack-f62460c8a29b488d06169e6eb93733b7a8d3aea7.idx'
c:\python34-x64\lib\shutil.py:375: PermissionError

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -10,6 +10,21 @@ environment:
 
   matrix:
 
+    - PYTHON: "C:\\Python27-x64"
+      PYTHON_VERSION: "2.7.x"
+      PYTHON_ARCH: "64"
+      TOXENV: "py27"
+
+    - PYTHON: "C:\\Python34-x64"
+      PYTHON_VERSION: "3.4.x"
+      PYTHON_ARCH: "64"
+      TOXENV: "py34"
+
+    - PYTHON: "C:\\Python35-x64"
+      PYTHON_VERSION: "3.5.x"
+      PYTHON_ARCH: "64"
+      TOXENV: "py35"
+
     - PYTHON: "C:\\Python36-x64"
       PYTHON_VERSION: "3.6.x"
       PYTHON_ARCH: "64"
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -28,7 +28,7 @@ def python_version(path_to_python):
 
     try:
         TEMPLATE = 'Python {}.{}.{}'
-        c = delegator.run('{0} --version'.format(path_to_python), block=False)
+        c = delegator.run([path_to_python,'--version'], block=False)
         c.return_code == 0
     except Exception:
         return None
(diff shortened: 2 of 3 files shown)


erinxocon:
@kennethreitz taking your advice, I just didn't remove the folder, and made sure to get out of it before the test ends so the rest of the tests function properly.  Seems hacky, but it does work.

erinxocon:
all ✅!

jabagawee:
Tests still fail for me in this branch if given a clean Windows environment and a fresh checkout. Have you tried it without local state?

erinxocon:
@jabagawee No I have not.  I need to set up a clean windows environment for testing.  It seems to be working for me locally.  I'll have to investigate this morning.

erinxocon:
@jabagawee can you post the stack trace of your error from pytest?

jabagawee:
It's alright, they've been broken since basically the start of this test_windows directory anyways. One insight that may help is that mocking out `delegator.run` in `which_pip()` breaks `Project.virtualenv_location`, which needs a string value in

````
@property
def virtualenv_location(self):
...
        c = delegator.run('pew dir "{0}"'.format(self.virtualenv_name))  # returns a Mock
        loc = c.out.strip()
...
self._virtualenv_location = loc
return loc  # is a Mock
````

Since `c` is just a `Mock` object, it doesn't have an `out` property. Then in `which`, we call:

````
p = os.sep.join([project.virtualenv_location] + ['Scripts\{0}.exe'.format(command)])
````

but `project.virtualenv_location` is a `Mock`, so `os.sep.join` fails here

jabagawee:
full gist at https://gist.github.com/jabagawee/763fae8baf52d5066169849f8ec740a1

erinxocon:
Huh!  Would you mind cloning this branch and then submitting a PR with these changes?  Otherwise I will check it out in the morning.

jabagawee:
Not sure what you need, since I have no changes in my checkout. Fresh environment + fresh checkout = not working for me 😢. I'd be glad to help you debug more tomorrow though.

jabagawee:
Well don't I feel dumb right about now. It turned out that the difference between my setup and AppVeyor's was that I wasn't entering a virtualenv before running my tests. Doing that cleaned up all the test failures for myself, mainly because the tests went down the intended path of execution. Is it impossible to test the other branch since we are in the venv during tests? Should we write more tests that mock out venv-state to see how it would go through the other path?

PR #556, opened 2017-09-15
Update crude installation instructions to properly find get-pipenv.py

jabagawee:
Changed the raw-file link for get-pipenv.py so that calling curl would lead to
the actual file instead of a 302 Redirect to the file.


kennethreitz:
✨🍰✨

PR #558, opened 2017-09-15
Output friendly and concise message when updating pipenv failed.

isudox:
Updating pipenv might fail occasionally, and the default output is not quite clear.
So I use crayons to beautify the error message and make it more friendly.


kennethreitz:
✨🍰✨

PR #560, opened 2017-09-15
Fix for TypeError: argument of type 'NoneType' is not iterable

jacebrowning:
This should handle the case where `python_version()` can return `None`:

screen shot 2017-09-15 at 10 14 57 am

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -398,14 +398,16 @@ def ensure_project(three=None, python=None, validate=True, system=False, warn=Tr
 
                 path_to_python = which('python')
 
-                if project.required_python_version not in python_version(path_to_python):
+                full_python_version = python_version(path_to_python)
+                if full_python_version is None or \
+                    project.required_python_version not in full_python_version:
                     puts(
                         '{0}: Your Pipfile requires {1} {2}, '
                         'but you are using {3} ({4}).'.format(
                             crayons.red('Warning', bold=True),
                             crayons.white('python_version', bold=True),
                             crayons.blue(project.required_python_version),
-                            crayons.blue(python_version(path_to_python)),
+                            crayons.blue(full_python_version or "<unknown>"),
                             crayons.green(shorten_path(path_to_python))
                         ), err=True
                     )
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -2,6 +2,7 @@
 import os
 import hashlib
 import tempfile
+import logging
 
 import delegator
 import pip
@@ -13,7 +14,7 @@ import six
 from piptools.resolver import Resolver
 from piptools.repositories.pypi import PyPIRepository
 from piptools.scripts.compile import get_pip_command
-from piptools import logging
+from piptools import logging as piptools_logging
 
 # List of version control systems we support.
 VCS_LIST = ('git', 'svn', 'hg', 'bzr')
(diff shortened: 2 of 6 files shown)


pipenv/utils.py
```diff
@@ -30,7 +31,8 @@ def python_version(path_to_python):
         TEMPLATE = 'Python {}.{}.{}'
         c = delegator.run([path_to_python,'--version'], block=False)
         c.return_code == 0
```
jacebrowning: Was this supposed to be an assert?

pipenv/cli.py
```diff
@@ -398,14 +398,16 @@ def ensure_project(three=None, python=None, validate=True, system=False, warn=Tr
 
                 path_to_python = which('python')
 
-                if project.required_python_version not in python_version(path_to_python):
+                full_python_version = python_version(path_to_python)
+                if full_python_version is None or \
+                    project.required_python_version not in full_python_version:
```
jacebrowning: `python_version()` can return `None` in some cases so we need to handle that.

pipenv/cli.py
```diff
@@ -398,14 +398,16 @@ def ensure_project(three=None, python=None, validate=True, system=False, warn=Tr
...
                     puts(
                         '{0}: Your Pipfile requires {1} {2}, '
                         'but you are using {3} ({4}).'.format(
                             crayons.red('Warning', bold=True),
                             crayons.white('python_version', bold=True),
                             crayons.blue(project.required_python_version),
-                            crayons.blue(python_version(path_to_python)),
+                            crayons.blue(full_python_version or "<unknown>"),
```
kennethreitz: this code path should never be hit, `which('python')` should *always* return — if it's not, there's a bug we need to fix.

pipenv/utils.py
kennethreitz: looks like it

kennethreitz: fixed in master

pipenv/cli.py
jacebrowning: It was getting hit here: https://ci.appveyor.com/project/kennethreitz/pipenv/build/1.0.549#L114

pipenv/utils.py
kennethreitz: this might be the bug

pipenv/cli.py
kennethreitz: yeah, there's a bug on windows with it. trying to figure it out...

kennethreitz: i think that missing assert might be the key

kennethreitz:
rebase with master

pipenv/cli.py
jacebrowning: As the code is implemented `python_version()` _can_ return `None`, so I think that either `python_version()` should raise an exception rather than return `None` or it's callers should treat `None` as a valid response and handle it.

kennethreitz:
https://github.com/kennethreitz/pipenv/pull/561

pipenv/cli.py
kennethreitz: `python_version` will only return `None` of the executable it was passed isn't a Python, though..

Which I suppose, is possible, but is unlikely.

kennethreitz:
got this merged manually — and passing.  :)  Thanks for the help!

kennethreitz:
@jacebrowning ps. NO_SPIN is automatic on Travis CI now :)

jacebrowning:
@kennethreitz Thanks! I'll update my templates.

PR #564, opened 2017-09-15
Fix popen_spawn when specifying a path on Windows which contains "\" separators

eli-b:
I've opened an identical PR for pexpect: https://github.com/pexpect/pexpect/pull/446

### DIFF
--- a/pipenv/vendor/pexpect/popen_spawn.py
+++ b/pipenv/vendor/pexpect/popen_spawn.py
@@ -40,7 +40,7 @@ class PopenSpawn(SpawnBase):
             kwargs['creationflags'] = subprocess.CREATE_NEW_PROCESS_GROUP
 
         if not isinstance(cmd, (list, tuple)):
-            cmd = shlex.split(cmd)
+            cmd = shlex.split(cmd, posix=sys.platform != 'win32')
 
         self.proc = subprocess.Popen(cmd, **kwargs)
         self.closed = False


kennethreitz:
what exactly does this do?

kennethreitz:
everything is behaving as expected right now.

eli-b:
This lets me run `pipenv install --three --python=C:\Miniconda2\envs\python3\python.exe -d` on Windows without getting `WindowsError: [Error 2] The system cannot find the file specified`. `shlex.split` without `posix=False` treats backslashes as escapes instead of preserving them.

There's an [open StackOverflow question](https://stackoverflow.com/questions/41185376/windowserror-error-2-the-system-cannot-find-the-file-specified-in-pexpect-pop/46248216#46248216) about this issue in popen_expect.

I should have explained the problem this solves for me.

uranusjr:
Is `posix=(os.sep == '/')` maybe a more pragmatic solution?

techalchemy:
given that we are specifically interested in whether we are dealing with a posix compliant platform I would recommend `posix=(os.name == 'posix')`

eli-b:
I just copied the test from earlier in the function where `sys.platform == 'win32` is used. Testing for a POSIX system directly might be clearer.

techalchemy:
I have a slight change to this over in #570 for reference

PR #569, opened 2017-09-16
Quick fix for dealing with relative paths!

techalchemy:
We can do this ez-mode instead of being super complicated about it, maybe

### DIFF
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -166,14 +166,15 @@ def convert_deps_from_pip(dep):
 
     req = [r for r in requirements.parse(dep)][0]
     # File installs.
-    if req.uri and not req.vcs:
+    if (req.uri or (os.path.exists(req.path) if req.path else False)) and not req.vcs:
 
         # Assign a package name to the file, last 7 of it's sha256 hex digest.
-        req.name = hashlib.sha256(req.uri.encode('utf-8')).hexdigest()
+        hashable_path = req.uri if req.uri else req.path
+        req.name = hashlib.sha256(hashable_path.encode('utf-8')).hexdigest()
         req.name = req.name[len(req.name) - 7:]
 
         # {file: uri} TOML (spec 3 I guess...)
-        dependency[req.name] = {'file': req.uri}
+        dependency[req.name] = {'file': hashable_path}
 
     # VCS Installs.
     if req.vcs:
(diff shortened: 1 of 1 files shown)


pipenv/utils.py
```diff
@@ -260,7 +261,8 @@ def convert_deps_to_pip(deps, r=True):
 
         # Support for files.
         if 'file' in deps[dep]:
-            dep = deps[dep]['file']
+            dep = '-e {0}'.format(deps[dep]['file'])
+            print('Adding dep: {}'.format(dep))
```
kennethreitz: we don't use print, anywhere.

pipenv/utils.py
```diff
@@ -260,7 +261,8 @@ def convert_deps_to_pip(deps, r=True):
 
         # Support for files.
         if 'file' in deps[dep]:
-            dep = deps[dep]['file']
+            dep = '-e {0}'.format(deps[dep]['file'])
```
kennethreitz: i think the lockfile format should be file: too?

kennethreitz: i think this breaks support for files

kennethreitz: we don't want to magically make non-editable things editable

pipenv/utils.py
techalchemy: oh yes, i was using that for debugging...

pipenv/utils.py
techalchemy: 2 things:

1. If we want to use *file:* uris in the lockfile should we be resolving the relative paths or can we leave them as relative?  
    * If we need to resolve them to use the URIs I would say we should probably just exclude them from the pipfile or change the spec to accommodate them somehow? I don't have a suggestion for this 
2. I agree this will probably break support for files by making non-editable things editable but I think I have a solution.  Let me push a change and see if you think it will work

techalchemy: Ok the latest commit still doesn't use `file://` URIs but will handle editable files correctly and in a non-breaking manner (i.e. similarly to how we do it with VCS currently). The first pass was rushed.  I had to go out to dinner!

I'll need some input / would love some feedback about what to put in the lockfile.  If you're feeling brave we can revisit this tomorrow.  We just need to make a ruling on what to use as a name if we are moving between relative and absolute paths and what, if anything, we can hash.

kennethreitz:
this looks solid to me!

PR #570, opened 2017-09-16
Test for posix compliance by checking posix compliance

techalchemy:
* Instead of testing for platform, which is not always accurate, it's better to test posix compliance by just checking whether we are posix compliant


kennethreitz:
✨🍰✨

PR #573, opened 2017-09-16
fix: bold used in puts instead of crayon

Woile:
bold was being used in puts instead of in crayon, breaking the code.

reproduce doing: `pipenv uninstall -d requests`


kennethreitz:
nice catch!

PR #574, opened 2017-09-16
fix: uninstall dev expecting tuple, not dict

Woile:
Trying to do:
`pipenv uninstall -d nose`
raises a `KeyError` exception.

This fix the issue.


kennethreitz:
✨🍰✨

PR #576, opened 2017-09-16
fix: uninstall dev now removes package from Pipfile

### LINKED ISSUE #575: Bug: uninstall dev is not removing package from Pipfile
Doing `pipenv uninstall -d nose` does not remove the package from `Pipfile`.

The cause is because [in here](https://github.com/kennethreitz/pipenv/blob/master/pipenv/cli.py#L1475) happens this `pipfile_remove = False`.
Is there a reason? Removing that line fixes the issue.
I would expect the package to be removed from the Pipfile, whether its a dev or non dev package.
I can submit a PR if required.

Woile:
Doing `pipenv uninstall -d <package>` does not remove the file from the
Pipfile. There was a flag preventing from doing so in `cli.uninstall`

Includes test

Closes #575

Note:
`test_cli_usage` is not passing for me because running `pipenv install regex` does not return `code == 0`

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -1472,7 +1472,6 @@ def uninstall(
             )
             package_names = project.parsed_pipfile['dev-packages']
             package_names = package_names.keys()
-            pipfile_remove = False
         else:
             puts(
                 crayons.white('No {0} to uninstall.'.format(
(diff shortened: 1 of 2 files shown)


kennethreitz:
✨🍰✨

kennethreitz:
will be included in the next release

kennethreitz:
released!

PR #584, opened 2017-09-17
Improve pyenv detection

rfleschenberg:
On my system, pipenv fails to detect the presence of pyenv, so it doesn't offer to download and install missing Python versions.

 I think `PYENV_ROOT` will only be set if the user explicitly sets it. It seems better to use `PYENV_SHELL` here. That one is set by `pyenv init`.

Also, I think it's a bit cleaner to use a `bool`  for the `PYENV_INSTALLED` variable.

### DIFF
--- a/pipenv/environments.py
+++ b/pipenv/environments.py
@@ -51,4 +51,4 @@ PIPENV_TIMEOUT = int(os.environ.get('PIPENV_TIMEOUT', 120))
 
 PIPENV_INSTALL_TIMEOUT = 60 * 15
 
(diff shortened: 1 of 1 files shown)


kennethreitz:
Interesting, I thought Pyenv *required* `PYENV_ROOT` to be set in order to operate.

kennethreitz:
let's have to use either.

kennethreitz:
https://github.com/kennethreitz/pipenv/commit/9cd15b70404c41c5a92785706f0ce85a82dd3a75

PR #616, opened 2017-09-20
Fix reST syntax of lists

cocoatomo:
A list item containing indented the second (and following) lines become an item of definition list.

For example, following reST code

```reST
- ``--all`` — This parameter will purge all files from the virtual environment,
              but leave the Pipfile untouched.
```

converted to

```HTML
   <li>
      <dl class="first docutils">
         <dt><code class="docutils literal"><span class="pre">--all</span></code> — This parameter will purge all files from the virtual environment,</dt>
         <dd>but leave the Pipfile untouched.</dd>
      </dl>
   </li>
```

In the example, the second line should be positioned at the same level of the first line.

In addition, sphinx splits that list item into two msgstr: "``--all`` — This parameter will purge all files from the virtual environment," and "but leave the Pipfile untouched."
This makes maintaining translations difficult when the original document is updated.

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -384,7 +384,7 @@ $ pipenv uninstall
 as well as one additonal, ``--all``.
 
     - ``--all`` — This parameter will purge all files from the virtual environment,
-                  but leave the Pipfile untouched.
+      but leave the Pipfile untouched.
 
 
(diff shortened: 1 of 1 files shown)


kennethreitz:
thanks!

PR #618, opened 2017-09-21
remove stray apostrophe 🤓

gregplaysguitar:
See https://english.stackexchange.com/questions/22603/why-doesnt-its-have-an-apostrophe


kennethreitz:
oops! i know the rules :)

PR #619, opened 2017-09-21
Windows concurrency

techalchemy:
* I added some helper functions but essential the issue on windows boiled down to a problem with paths
* I don't think it's safe to explicitly use forward or backward slashes rather than `os.path.join()`
* There are places where we are currently using `os.sep` and concatenating -- is there a reason? Should we switch to `os.path.join`?
* `Project` objects currently resolve virtualenv locations using forward slashes on windows, but `os.path.join` just appends the remaining items with the normal path separator, so we can use `os.path.normpath` to normalize
* However this pre-escapes the path for us which means we can't drop the command into a double-quoted string and pass it to delegator, so instead I am shellquoting everything that isn't handled this way now

Lets see if it passes tests.


techalchemy:
well shit, broke appveyor

techalchemy:
JK appveyor was broken already

kennethreitz:
let's get windows tests working first

kennethreitz:
got merge conflict handled

kennethreitz:
running windows tests now

kennethreitz:
looks like they're passing! This will require some additional work.

kennethreitz:
e.g. they're still running non-concurrently.

PR #627, opened 2017-09-21
Fix bug with misprint of "project" variable in update method of CLI

vovanbo:
Add test for this case


nateprewitt:
Thanks for his @vovanbo!

PR #631, opened 2017-09-21
Remove bad packages json graph

techalchemy:
* Had to change the JSON flag to 'is_json' but couldn't find how to let the user still pass it in the same way

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -32,7 +32,7 @@ from .project import Project
 from .utils import (
     convert_deps_from_pip, convert_deps_to_pip, is_required_version,
     proper_case, pep423_name, split_vcs, resolve_deps, shellquote, is_vcs,
-    python_version, suggest_package
+    python_version, suggest_package, remove_json_packages
 )
 from .__version__ import __version__
 from . import pep508checker, progress
@@ -1937,8 +1937,8 @@ def check(three=None, python=False):
 
 @click.command(help=u"Displays currently–installed dependency graph information.")
 @click.option('--bare', is_flag=True, default=False, help="Minimal output.")
-@click.option('--json', is_flag=True, default=False, help="Output JSON.")
-def graph(bare=False, json=False):
+@click.option('--is_json', is_flag=True, default=False, help="Output JSON.")
+def graph(bare=False, is_json=False):
     try:
         python_path = which('python')
     except AttributeError:
@@ -1951,10 +1951,10 @@ def graph(bare=False, json=False):
         )
         sys.exit(1)
 
-    if json:
+    if is_json:
         bare = True
 
-    j = '--json' if json else ''
+    j = '--json' if is_json else ''
 
     cmd = '"{0}" {1} {2}'.format(
         python_path,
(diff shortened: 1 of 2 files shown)


kennethreitz:
nope, need to change the import instead.

kennethreitz:
`import json as simplejson`

pipenv/cli.py
```diff
@@ -1988,7 +1988,12 @@ def graph(bare=False, json=False):
             else:
                 click.echo(crayons.white(line, bold=False))
     else:
-        click.echo(c.out)
+        dep_graph = json.loads(c.out)
+        # Clean top level of dependency graph
+        dep_graph = [p for p in dep_graph if p['package']['package_name'] not in BAD_PACKAGES]
```
nateprewitt: Isn't this identical to just calling remove_json_packages? We could do that and reduce the likelihood of these two code chunks getting out of sync.

pipenv/cli.py
```diff
@@ -1958,10 +1958,10 @@ def graph(bare=False, json=False):
         )
         sys.exit(1)
 
-    if json:
+    if is_json:
```
nateprewitt: What's the rationale for adding the extra letters to the flag? It seems like just json should be sufficiently expressive.

pipenv/utils.py
```diff
@@ -770,3 +770,15 @@ def find_requirements(max_depth=3):
                 if os.path.isfile(r):
                     return r
     raise RuntimeError('No requirements.txt found!')
+
+
+def remove_json_packages(pkg, bad_pkgs):
+    """Crawl pkg for bad packages in bad_pkgs to remove"""
+    # Check for dependencies to remove
+    pkgs = pkg.get('dependencies', False)
```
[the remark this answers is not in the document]
kennethreitz: Agreed, and None is default, so...

kennethreitz:
i think i'm going to rewrite this inline.

techalchemy:
I wasn't sure if pipdeptree would always make the dependency list flat. If it does I'm not sure putting it inline is going to work. 

As for conditionals it's an old habit when I truth test things to explicitly make them true or false. 

As for the extra letters, assigning something to be called 'json' when we have an import I need to use which is called 'json' was causing a collision. I believe Kenneth opted to re-namespace the import everywhere instead of changing the command line flag.

kennethreitz:
i redid the thingy

kennethreitz:
we only need to scrub top-level packages, we don't need to remove setuptools if it's a dependency of something else.

kennethreitz:
my solution was much simpler :)

techalchemy:
hm yes that is a fair point, you wouldn't need my solution in that case

pipenv/utils.py
```diff
@@ -770,3 +770,15 @@ def find_requirements(max_depth=3):
...
+    if pkgs:
+        pkgs = [p for p in pkgs if p['package_name'] not in bad_pkgs]
+        # Recursively crawl for bad package entries and remove them
+        pkgs = [remove_json_packages(p, bad_pkgs) for p in pkgs]
```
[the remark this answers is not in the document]
techalchemy: FYI I considered the performance impact of iterating over a list twice.  I decided that it is probably not very significant, and opted to make this more legible.

I wasn't sure if we wanted to remove nested packages, so I designed for that and figured we could refactor if not.  But yes, if you wanted to prune nested dependency trees it would return what you wanted.  If there were no nested trees then each dependency would have no items in its 'dependencies' list and the `if pkgs` test would fail.

Academic point, but this code does what it set out to do.

pipenv/cli.py
techalchemy: No it isn't.  `remove_json_packages`, perhaps misnamed, only crawls the dependency tree.  It didn't remove top level packages.

PR #649, opened 2017-09-22
POC for plugin support (I don't expect it to be accepted, sending only as RFC)

rochacbruno:
## Just a `POC` for plugin extensions

> I don't expect it to be accepted, sending only as a Request For Comments

### How it works?

1. Create a new plugin for example `pipenv_hello.py`:

```py
import click
@click.command()
def main():
    """The Hello Plugin"""
    click.echo("Hello from pipenv plugin")
```

2. Distribute it under `pipenv.extension` namespace

```py
# -*- coding: utf-8 -*-


setup(
    name='pipenv_hello',
    version='1.0.0',
    description='Example plugin to pipenv',
    packages=find_packages(),
    install_requires=['pipenv'],
    entry_points={
        'pipenv.extension': [
            'hello = pipenv_hello:main',
        ],
    }
)
```

3. Once in PyPI plugin can be installed

```bash
$ pipenv install pipenv_hello
```
or `pip install pipenv_hello`

4.  If the plugin provides new commands It will now be vailable

```bash
$ pipenv --help
...
Commands:
  ...
  hello      The Hello Plugin
  ...
```

5. Use it

```bash
$ pipenv hello
Hello from pipenv plugin
```
6. If the plugin only register `hooks` you will see it when running `pipenv listhooks` (not implemented yet)

## What is a plugin?
- Can register new `click.command` or `click.group` as subcommands to `pipenv` main command line group
- Can (or not) register `hooks`/`signals`
- has access to current project artifacts `Pipfile, virtualenv etc..`

## Ideas not implemented yet?
- Register hooks (or signals) that would be useful for example to intercept 'pre' and 'post' install operations and also 'pre' and 'post' for any command like 'lock', 'install'.
*not sure yet if hook system as in `pluggy` or signals as in `blinker` that would be useful to intercept `pre` and `post` operations.

## Stable?
- NO! just a POC/**RFC** needs tests, not intended to be merged.

## Use cases?

One can create a new plugin called `pipenv_publish` or `pipenv_setup` when installed with `pip install pipenv_<plugin_name>` it will add more subcommands to `pipenv` i.e: `setup` and `publish`

- `pipenv setup` would use pipenv helper functions to parse `Pipfile` requirements and a section called `package` to generate a `setup.py` file.

- `pipenv publish` would check the existence of that file and then use `twine` to upload to `PyPI`

- Other plugin ideas would be `pipenv bumpversion` (to use with publish) `pipenv build` `pipenv test` etc....

As you can see my motivation is primarily to use the same `pipenv` environment to distribute my packages (as I am developing frameworks that makes sense)

However, this open opportunity for other plugins to be developed.

For example, all the `poet` and `hatch` functionalities can be migrated to `pipenv` via plugins.

https://github.com/sdispater/poet
https://github.com/ofek/hatch

### DIFF
--- a/examples/plugin/pipenv_hello.py
+++ b/examples/plugin/pipenv_hello.py
@@ -1,0 +1,7 @@
+import click
+
+
+@click.command()
+def main():
+    """The Hello Plugin"""
+    click.echo("Hello from pipenv plugin")
--- a/examples/plugin/setup.py
+++ b/examples/plugin/setup.py
@@ -1,0 +1,32 @@
+# -*- coding: utf-8 -*-
+from setuptools import setup, find_packages
+
+
+setup(
+    name='pipenv_hello',
+    license='MIT',
+    version='1.0.0',
+    description='Example plugin to pipenv',
+    author='Bruno Rocha',
+    author_email='rochacbruno@gmail.com',
+    url='https://github.com/rochacbruno/pipenv_hello',
+    packages=find_packages(),
+    install_requires=['pipenv'],
+    entry_points={
+        'pipenv.extension': [
+            'hello = pipenv_hello:main',
+        ],
+    },
+    classifiers=[
+        'License :: OSI Approved :: MIT License',
+        'Programming Language :: Python',
+        'Programming Language :: Python :: 2.7',
+        'Programming Language :: Python :: 3',
+        'Programming Language :: Python :: 3.3',
+        'Programming Language :: Python :: 3.4',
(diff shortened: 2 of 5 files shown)


kennethreitz:
Thanks for putting this together — however, I don't think this project wants to go in this direction at this time. 

I've put a lot of thought into this.

kennethreitz:
I'll think about it some more.

kennethreitz:
I just don't think this is a good idea.

PR #666, opened 2017-09-23
Add --pip-options to support pip options

Brightcells:
Such as
```shell
pipenv install requests --pip-options="-U"
```


kennethreitz:
use environment variables.

kennethreitz:
thanks for the suggestion, though! Pip supports environment variables for all it's options, and i encourage any users that need that functionality to use them instead.

PR #670, opened 2017-09-23
add: --index in import_requirements

### LINKED ISSUE #669: Missing --index in parse_requirements
Hi guys, I think here:

https://github.com/kennethreitz/pipenv/blob/master/pipenv/cli.py#L236

the `--index` should be tested as well, not only `-i`. I'm not sure though. I can make a PR if I'm correct.

Regards!

Woile:
Closes #669


kennethreitz:
thanks!

PR #672, opened 2017-09-23
Remove extra touch of the Pipfile during run

### LINKED ISSUE #546: Pipfile touched even on pipenv run
I have a watch command tracking my `Pipfile` but on every call of `pipenv run` the command is executed, even if the file wasn't changed. I tracked this down to the call of `Project.touch_pipfile`. I'm not sure if it required, for `Pipfile` to be touched every time, probably it is just a side effect of calling  `cli.py@ensure_project`.

Would it be reasonable to only touch the `Pipfile` if it either doesn't exist or if there is actually a change to its contents?

jacebrowning:
Fixes #546

`touch_pipfile` is already called from `ensure_project` if needed.

### DIFF
--- a/pipenv/project.py
+++ b/pipenv/project.py
@@ -311,11 +311,14 @@ class Project(object):
 
     @property
     def pipfile_is_empty(self):
-        self.touch_pipfile()
+        if not self.pipfile_exists:
+            return True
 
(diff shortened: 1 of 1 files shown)


kennethreitz:
thanks!

PR #679, opened 2017-09-24
Fixed stdlib file not found error

numshub:
This should fix https://github.com/kennethreitz/pipenv/issues/677


kennethreitz:
thanks for fixing this!

PR #685, opened 2017-09-24
simplify code in find_a_system_python

tomviner:
Thanks for recommending [`PIPENV_DEFAULT_PYTHON_VERSION`](https://twitter.com/kennethreitz/status/912068103588245507). I was taking a look at the source, to see how it worked and this function looked ripe for simplifying. 

(Not sure if micro contributions like this are always helpful, so feel free to feedback on that :smile: )

Also I see it ignores anything more detailed that x.y.


kennethreitz:
✨🍰✨

PR #687, opened 2017-09-25
Windows tests

erinxocon:
This fixes issue #686.  The windows tests were failing mainly due to the python_requires in the pipfile saying that 3.6 was required.  This made the tests only pass on windows when using 3.6.  I don't know if this is intended and we need to redirect it, or if removing it from the pipfile is a fine solution.  I also added in tests for pypy, python3.7-dev for travis and they seem to work fine (we can disable them if we need speed, the pypy test is slow at 9-10 minutes).  I also added 32-bit versions of pythons to the test matrix in appveyor.yml.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,8 @@ python:
   - "3.5"
   - "3.6"
   - "pypy"
-  - "pypy3"
-  # TODO: pkg_config issues
+  - "3.7-dev"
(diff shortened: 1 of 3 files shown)


erinxocon:
Oh I also got xdist working on app veyor, tests suite runs about 5 minutes faster now!

erinxocon:
#691 contains this PR

PR #691, opened 2017-09-25
More tests

erinxocon:
This takes the remaining tests that were in legacy that didn't use Mock and converts them to use the new test suite.


.travis.yml
```diff
@@ -4,9 +4,10 @@ python:
   - "3.4"
   - "3.5"
   - "3.6"
-  # - "3.7-dev"
-  # - "pypy"
+  - "pypy"
+  - "3.7-dev"
   # - "pypy3" # TODO: pkg_config issues
+  # - "3.7-dev"
```
nateprewitt: We can probably nix this since it's been added above, yeah?

kennethreitz:
I don't think we need to be testing against this many pythons on appveyor. It's slow already.

erinxocon:
@nateprewitt Now that we have a better test suite, we could set up a coverage map.  We stopped testing against the old windows test suite a while ago when Ken and I created the new version.  I've now ported any test that was both in test_pipenv.py for the windows and unix versions (mostly the same but a couple odd ones out here and there).  Any remaining tests were used on both platforms are are in test_legacy.py.  So we're not actually loosing anything??

erinxocon:
@kennethreitz That's what I thought.  However the default installer for python is 32-bit I believe.  Should we target those versions over the 64 bit ones?  Or just go back to testing the 64 bit ones?

kennethreitz:
just pick one, i'd say. they'll have the same behavior

kennethreitz:
I'm not convinced coverage will help us as we're subprocessing everything out.

kennethreitz:
That's also the last of my concerns right now.

nateprewitt:
I'm not referring to coverage as a number, or generated from coverage.py, but rather all of those tests were added for specific bugs. I just wanted to make sure we're not losing any of that knowledge by deleting all the legacy code.

kennethreitz:
those tests were unmaintainable, so i'm happy to see them go.

kennethreitz:
and when they blew up, it was very unclear what was going wrong. these new tests are superior in all ways. :)

erinxocon:
No, we're not losing anything.  Any important test from the legacy tests has been migrated now.

PR #719, opened 2017-09-26
Display warning message if PYENV_ROOT is not set and abort if installed python is not the desired one

Perados:


kennethreitz:
don't want to abort if it's the wrong python, just warn.

kennethreitz:
PYENV_ROOT is almost always set, from what i can tell. i think you're just a special case :)

pipenv/cli.py
```diff
@@ -390,13 +389,23 @@ def abort():
...
+                    '{0}{1}versions{1}*{1}bin'.format(
+                        PYENV_ROOT,
+                        os.sep
+                    )
+                ):
+                    add_to_path(found)
+            else:
+                click.echo(
+                    u'{0}: PYENV_ROOT is not set. New python paths will '
+                    u'probably not be exported properly after installation.'
```
kennethreitz: no u here

Perados:
But then the wrong python would be used for the venv. That's not the desired behaviour, is it?

Haha well, I was only using pyenv to install different versions of python, then using the paths with virtualenvwrapper, that's why I never needed to set PYENV_ROOT.

kennethreitz:
take out the sys.exit and i'll merge

pipenv/cli.py
```diff
@@ -390,13 +389,23 @@ def abort():
...
+                        PYENV_ROOT,
+                        os.sep
+                    )
+                ):
+                    add_to_path(found)
+            else:
+                click.echo(
+                    '{0}: PYENV_ROOT is not set. New python paths will '
+                    'probably not be exported properly after installation.'
+                    .format(
```
kennethreitz: `''.format`

Perados:
@kennethreitz Alright, I removed the sys.exit. :)

PR #722, opened 2017-09-27
Tests

erinxocon:
added some tests and added a skipif for one of them to only run it on travis


nateprewitt:
This all looks good to me 😁!

PR #741, opened 2017-09-27
Add shell integration section to documentation

numshub:
The default activate.fish behaviour will set an ugly prompt, especially when using a virtualenv aware prompt like bobthefish.
This behaviour can be changed by manually setting an environmental variable.
A clean, elegant prompt is a must for such a refined piece of software as pipenv.

I propose a shell integration section to provide this info. In the future it could grow with additional tips (relevant to other shells, even if I do not see why to use them instead of beautiful fish).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -103,6 +103,13 @@ Magic shell completions are now enabled! There is also a `fish plugin <https://g
 
 Fish is the best shell. You should use it.
 
+Shell Integration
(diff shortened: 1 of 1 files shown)


kennethreitz:
this belongs in the docs, not in the readme

kennethreitz:
and you should try shell --fancy :)

PR #749, opened 2017-09-28
Issue #748: cover the --site-packages option

### LINKED ISSUE #748: Mention `PIP_IGNORE_INSTALLED` in combination with `--site-packages`
When using the `--site-packages` option, `PIP_IGNORE_INSTALLED` can be useful to make sure that any pip-installable dependencies actually are installed into the virtual environment, with the system site-packages then being used only to resolve imports that aren't covered by the Python level dependency graph at all (with my current use case for that being Fedora's Python bindings for DNF and libsolv).

I was mainly writing an issue instead of a PR in order to ask where the appropriate place in the docs to add this would be. I'm currently thinking a new "Relying on platform-provided Python components" section at the end of https://github.com/kennethreitz/pipenv/blob/master/docs/advanced.rst might work.

ncoghlan:
Also mentions using `PIP_IGNORE_INSTALLED` to still
install everything that's pip-installable into the virtual environment,
even if a suitable version is already installed globally.

Closes #748

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -377,3 +377,21 @@ To enable completion in fish, add this to your config::
 Magic shell completions are now enabled!
 
 ✨🍰✨
+
+☤ Working with platform-provided Python components
+--------------------------------------------------
+
+It's reasonably common for platform specific Python bindings for
+operating system interfaces to only be available through the system
+package manager, and hence unavailable for installation into virtual
+environments with `pip`. In these cases, the virtual environment can
+be created with access to the system `site-packages` directory::
+
+    `$ pipenv --three --site-packages
+
+To ensure that all `pip`-installable components actually are installed
+into the virtual environment and system pacakges are only used for
+interfaces that don't participate in Python-level dependency resolution
+at all, use the `PIP_IGNORE_INSTALLED` setting::
+
+    $ PIP_IGNORE_INSTALLED=1 pipenv install --dev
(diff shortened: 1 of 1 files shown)


docs/advanced.rst
```diff
@@ -377,3 +377,21 @@ To enable completion in fish, add this to your config::
...
+☤ Working with platform-provided Python components
+--------------------------------------------------
+
+It's reasonably common for platform specific Python bindings for
+operating system interfaces to only be available through the system
+package manager, and hence unavailable for installation into virtual
+environments with `pip`. In these cases, the virtual environment can
+be created with access to the system `site-packages` directory::
+
+    `$ pipenv --three --site-packages
```
nateprewitt: Just wanna nix the extra ` and I think we're set.

docs/advanced.rst
```diff
@@ -377,3 +377,21 @@ To enable completion in fish, add this to your config::
...
+
+To ensure that all `pip`-installable components actually are installed
+into the virtual environment and system pacakges are only used for
```
nateprewitt: s/pacakges/packages/g

nateprewitt:
Thanks @ncoghlan! :sparkles:

kennethreitz:
fixed the capitalization :)

campanelli-resonon:
@nateprewitt The "one liner" `$ PIP_IGNORE_INSTALLED=1 pipenv install --dev` doesn't work for me on Windows.

PR #771, opened 2017-09-29
Updated README.rst

avallbona:
Fixed the documentation because the option -j does not work anymore as a shortcut of --jumbotron


nateprewitt:
This looks right, thanks for your first contribution @avallbona!

PR #779, opened 2017-09-30
let's get green again

erinxocon:


kennethreitz:
whoopsie daisy — at least I added a test :)

Issue #788, opened 2017-10-01
git remote address modified changing underscore to a hyphen...


minskmaz:
I noticed that after pipenv announces successful install of a private repo from gitlab:

pipenv install -e git+git@gitlab.com:_orangenero/db.git#egg=db;

... I get an error from pip

pip.exceptions.InstallationError: Command "git clone -q git@gitlab.com:-orangenero/db.git /tmp/pip-8ySH2I-export" failed with error code 128 in None

Notice pip has been passed a modified gitlab groupname that transforms the underscore to a hyphen. In this case:

"_orangenero" to "-orangenero"

or, 

git clone -q git@gitlab.com:-orangenero/db.git /tmp/pip-8ySH2I-export

nateprewitt:
@minskmaz, I believe we addressed this late in the 7.X releases or early 8.X. Could you please provide the version of pipenv you're currently using?

minskmaz:
pipenv, version 8.2.5

minskmaz:
@nateprewitt pipenv, version 8.2.5

erinxocon:
Can reproduce.  Working on a fix now.

kennethreitz:
fix merged

minskmaz:
Thank you @nateprewitt / @kennethreitz but I'm not certain its fixed. I'm now on version 8.2.6 as you see here:

Removing intermediate container e57341c44ad5
Step 10/15 : RUN pipenv --version
 ---> Running in f93d1756481f
pipenv, version 8.2.6

But I am still getting the pip error below after pipenv installs the egg:

Installing -e git+git@gitlab.com:_orangenero/allspark.git#egg=allspark…
Obtaining allspark from git+git@gitlab.com:_orangenero/allspark.git#egg=allspark
  Updating /root/.local/share/virtualenvs/iota-9BD0xIDx/src/allspark clone
Installing collected packages: allspark
  Found existing installation: allspark 0.0.post0.dev7+ng5af9e93
    Uninstalling allspark-0.0.post0.dev7+ng5af9e93:
      Successfully uninstalled allspark-0.0.post0.dev7+ng5af9e93
  Running setup.py develop for allspark
Successfully installed allspark

Adding -e git+git@gitlab.com:_orangenero/allspark.git#egg=allspark to Pipfile's [packages]…
Locking [dev-packages] dependencies…
Locking [packages] dependencies…
fatal: strange pathname '-orangenero/allspark.git' blocked
[...]

PR #793, opened 2017-10-01
788 fix

erinxocon:
This is a fix for the bug in issue #788

### DIFF
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -739,7 +739,7 @@ def pep440_version(version):
 def pep423_name(name):
     """Normalize package name to PEP 423 style standard."""
     name = name.lower()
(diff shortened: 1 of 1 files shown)


pipenv/utils.py
```diff
@@ -738,8 +738,11 @@ def pep440_version(version):
 
 def pep423_name(name):
     """Normalize package name to PEP 423 style standard."""
-
-    return name.lower().replace('_', '-')
+    name = name.lower()
+    if not(name in VCS_LIST or name in FILE_LIST):
```
kennethreitz: shouldn't we be doing `s.starswith`?

nateprewitt:
We should probably change the name and docstring of this function because it's no longer doing what it was designed to do. Maybe just normalize_package(_name?).

Issue #806, opened 2017-10-01
pip-tools stack trace if VCS doesn't include -e


c17r:
```sql
$ pipenv install 'git+http://github.com/alsoicode/django-medusa.git@v0.3.2#egg=django-medusa'
Courtesy Notice: Pipenv found itself running within a virtual environment, so it will automatically use that environment, instead of creating its own for any project.
Installing git+http://github.com/alsoicode/django-medusa.git@v0.3.2#egg=django-medusa…
Collecting django-medusa from git+http://github.com/alsoicode/django-medusa.git@v0.3.2#egg=django-medusa
  Cloning http://github.com/alsoicode/django-medusa.git (to v0.3.2) to /private/var/folders/b4/4mq3zd1x5ds1wbzd00z0xpmm0000gn/T/pip-build-mtemz9re/django-medusa
Requirement already satisfied: django in /Users/christian/.virtualenvs/hnhiring-HX-OrkMQ/lib/python3.6/site-packages (from django-medusa)
Requirement already satisfied: pytz in /Users/christian/.virtualenvs/hnhiring-HX-OrkMQ/lib/python3.6/site-packages (from django->django-medusa)
Installing collected packages: django-medusa
  Running setup.py install for django-medusa: started
    Running setup.py install for django-medusa: finished with status 'done'
Successfully installed django-medusa-0.3.2

Adding git+http://github.com/alsoicode/django-medusa.git@v0.3.2#egg=django-medusa to Pipfile's [packages]…
Locking [dev-packages] dependencies…
Locking [packages] dependencies…
Traceback (most recent call last):
  File "/Users/christian/.pyenv/versions/3.6.2/bin/pipenv", line 11, in <module>
    load_entry_point('pipenv==8.2.6', 'console_scripts', 'pipenv')()
  File "/Users/christian/.pyenv/versions/3.6.2/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/Users/christian/.pyenv/versions/3.6.2/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/Users/christian/.pyenv/versions/3.6.2/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
[...]

PR #807, opened 2017-10-02
Non editable vcs fix

techalchemy:
See #806 - issue is easily reproducible


erinxocon:
@techalchemy Thanks for the fix!  Can you write or update the test suite for this bug?

techalchemy:
but, what about being lazy :(

erinxocon:
@techalchemy as awesome as being lazy is, having a well tested master python tool is pretty cool too right :)

techalchemy:
for sure, glad to see you policing it :) I'll write some tests for it in the morning after I confirm that it also fixes the alternate variant of the issue I just read in full

PR #821, opened 2017-10-03
Fix check --unused

krismolendyke:
- Add missing `mapping` file from vendored `pipreqs` to dist
- Resolve non-stdlib package names using `mapping` file before checking PyPI
- Do not `lower` needed package names after resolving them with `pipreqs`


erinxocon:
Thanks for the contribution @krismolendyke!  I'm gonna let @kennethreitz push the all powerful merge button on this! :)

krismolendyke:
Absolutely.  When I went looking for existing tests before I missed [this](https://github.com/kennethreitz/pipenv/blob/0acd7a2acffb7f14fe4f56ca3959d1dbfad37674/tests/test_pipenv.py#L500).  I'll update to prove this functionality.

PR #824, opened 2017-10-04
Fix when PIPENV_SHELL is empty

alercunha:
After [this commit](https://github.com/kennethreitz/pipenv/commit/04ce86abe081246a27cd94d3e20db2a6d4000452) the exception KeyError is no longer raised in case the env variable `SHELL` is missing.

This patch fixes the issue without making any changes in the current flow. 

One might want to make this check earlier when `PIPENV_SHELL` is set at [environments.py#L90](https://github.com/kennethreitz/pipenv/blob/master/pipenv/environments.py#L90) and then remove all validations of `PIPENV_SHELL` afterwards.

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -1155,11 +1155,11 @@ def activate_virtualenv(source=True):
     suffix = ''
 
     # Support for fish shell.
-    if 'fish' in PIPENV_SHELL:
+    if PIPENV_SHELL and 'fish' in PIPENV_SHELL:
         suffix = '.fish'
 
     # Support for csh shell.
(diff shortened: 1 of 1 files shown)


erinxocon:
Thanks for your first contribution to pipenv @alercunha!  This looks good but I'm going to let @kennethreitz or @nateprewitt merge 👍

alercunha:
Thanks @erinxocon. My pleasure. Nice little project of yours, very useful. I'm here to help.

nateprewitt:
This seems fine to me since we've already been importing PIPENV_SHELL. Thanks for your contribution, @alercunha!

PR #830, opened 2017-10-04
Add check unused test for mapped package name

krismolendyke:
As promised in https://github.com/kennethreitz/pipenv/pull/821


erinxocon:
Thanks @krismolendyke!

PR #835, opened 2017-10-05
Bugfix/git ssh vcs url

### LINKED ISSUE #788: git remote address modified changing underscore to a hyphen...

techalchemy:
* Fix Requirement objects that use git+git@host:user/repo.git syntax
* Requirements are not natively being recognized as VCS requirements
* This converts them to VCS by moving req.path to req.uri and adding
* Adds req.vcs flag
* Remove some redundancies from project package filtering
* Fixes #788 (Maybe) -- see discussion in the issue

### DIFF
--- a/pipenv/project.py
+++ b/pipenv/project.py
@@ -60,15 +60,16 @@ class Project(object):
             if hasattr(v, 'keys'):
                 # When a vcs url is gven without editable it only appears as a key
                 if is_vcs(v) or is_vcs(k):
+                    # Non-editable VCS entries can't be resolved by piptools
                     if 'editable' not in v:
                         continue
                     else:
                         ps.update({k: v})
                 else:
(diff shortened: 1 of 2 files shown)


techalchemy:
* I thought I mentioned this in the commit but I'm not actually sure if this commit belongs in pipenv
* This is actually a problem with Requirements failing to recognize git+git@location addresses as VCS

PR #842, opened 2017-10-06
Reset the InstallRequirement.prepared attribute when yeilding dependencies

### LINKED ISSUE #839: pipenv requirement resolution failure for the ipython package
I think that you are cutting off the resolution algorithm in `pipenv/patched/piptools/resolver.py` too soon.  The three observations that lead me to this belief are:

(1)  In a clean empty python27 pipenv, typing `pipenv install ipython` succeeds in installing ipython but fails to create `Pipfile.lock` with the following error message:
```
RuntimeError: No stable configuration of concrete packages could be found for the given constraints after 16 rounds of resolving.
This is likely a bug.
```

(2) In a new empty python27 pipenv, typing `pipenv install appnope backports.shutil-get-terminal-size decorator enum34 ipython-genutils pathlib2 pexpect pickleshare prompt-toolkit ptyprocess pygments scandir simplegeneric six traitlets wcwidth ipython` succeeds, in both the installation and in the creation of `Pipfile.lock`.  The magic list of packages I used here was just the packages that step (1) told me were necessary to install for `ipython`.  This suggests that the problem is with the resolution method, not the interdependencies of this particular set of packages.

(3) Indeed, rerunning experiment (1), but after manually changing `pipenv/patched/piptools/resolver.py` so as to allow it an extra 100 rounds of resolution, succeeds in both the installation of `ipython` and in the creation of `Pipfile.lock`.

vphilippon:
Fixes #839
This well ensure the dependencies are well-fetched on the next iteration too.

I think this patched section could be removed, but I can't give a 100% guarantee right away, so I'll submit this fix for now and will leave the "full" resolving of markers later.

### DIFF
--- a/pipenv/patched/piptools/resolver.py
+++ b/pipenv/patched/piptools/resolver.py
@@ -276,9 +276,12 @@ class Resolver(object):
             return
         elif ireq.markers:
             for dependency in self.repository.get_dependencies(ireq):
+                dependency.prepared = False
                 yield dependency
+            return
         elif ireq.extras:
             for dependency in self.repository.get_dependencies(ireq):
+                dependency.prepared = False
                 yield dependency
             return
         elif not is_pinned_requirement(ireq):


kennethreitz:
✨🍰✨

PR #845, opened 2017-10-06
Typo fix + rebuild man

therefromhere:
`make man` to cover recent docs changes (including #843 )

I assume removal of the "Automatic Python Installation" section from man that this causes is intentional? (https://github.com/kennethreitz/pipenv/compare/master...therefromhere:rebuild-man?expand=1#diff-b5e9ceb23f5465ad53594720c58e859aL684 )


kennethreitz:
thanks!

PR #849, opened 2017-10-07
cli: Use max() to determine latest version

auscompgeek:
This uses the `max()` builtin to determine the latest pipenv version
instead of sorting the available versions and taking the last version.

This is clearer, idiomatic, and avoids the need for intermediate lists.


kennethreitz:
✨🍰✨

PR #852, opened 2017-10-07
Ensure python version is installed before deleting existing virtual env

joaorafaelm:
When specifying a python (`--three, --two or --python`) version, `pipenv` deletes the existing virtual env **before** checking if the specified version is installed.

With this change, pipenv can prevent mistyped versions, or even undesirable actions; suppose someone cannot download a new python version due to a limited network etc.


erinxocon:
Thanks for this @joaorafaelm

PR #855, opened 2017-10-08
set timeout on fetching warehouse hashes

JackDanger:


pipenv/utils.py
```diff
@@ -495,6 +495,8 @@ class PipCommand(pip.basecommand.Command):
                         collected_hashes = list(list(resolver.resolve_hashes([result]).items())[0][1])
 
                 except (ValueError, KeyError):
+                    if verbose:
+                        print('Error fetching {}'.format(name))
                     pass
```
joaorafaelm: pass statement can be removed 😉

JackDanger: Thanks for the review! I could have looked at this for another hour and not noticed that. :)

nateprewitt:
This seems sensible, thanks @JackDanger!

PR #864, opened 2017-10-09
Use consistant os check with os.name (import nt does not guaranty to be on windows)

boeddeker:
A file, folder or package can produce a name conflict with nt on non windows systems.
Also https://github.com/python/cpython/blob/master/Lib/pathlib.py uses `os.name`.

### DIFF
--- a/pipenv/vendor/pathlib.py
+++ b/pipenv/vendor/pathlib.py
@@ -28,16 +28,15 @@ except NameError:
     basestring = str
 
 supports_symlinks = True
-try:
+if os.name == 'nt':
     import nt
-except ImportError:
-    nt = None
-else:
     if sys.getwindowsversion()[:2] >= (6, 0) and sys.version_info >= (3, 2):
         from nt import _getfinalpathname
     else:
         supports_symlinks = False
         _getfinalpathname = None
+else:
+    nt = None
 
 
 __all__ = [


nateprewitt:
Hey @boeddeker, thanks for taking the time to put this PR together! Unfortunately this is a package that pipenv vendors and we'd rather not patch it ourselves if possible. If you'd like to open a PR with the upstream repo, we'd be happy to upgrade to a newer version when it's merged.

boeddeker:
I had to do it because I have the problem that I have a package called nt. 

Can you tell me what the upstream repo is? 
I can not find what is meant with "vendor".

I found
`$ cat ./pipenv/patched/pip/_vendor/vendor.txt`
but there is no pathlib.

`$ grep -r "pathlib" *` says https://bitbucket.org/pitrou/pathlib/issue/12/ and that site says https://pypi.python.org/pypi/pathlib2/ and that package is for python 2.7.

So my conclusion is python 3.6 should use the standard lib instead of the "vendor" code?

nateprewitt:
Sorry for the lack of clarity in my last response. So the upstream repository would be where pathlib is housed, which appears to be bitbucket from your research.

We copy the code and house it with pipenv so you won't have to download separate dependencies. It also provides us the benefit of knowing people are using the same version of dependencies that we've verified to work.

So I agree that Python 3 doesn't need to import this, but I think your underlying issue is bigger than patching this one library. The "nt" check is actually fairly common and if you continue to use 3rd party Python libraries, I believe it's likely you'll encounter this elsewhere. If possible, I'd highly suggest changing the name of your file or appending an underscore to differentiate your copy.

boeddeker:
Ok, then I will live with my fork.

The upstream repo is dead and for the successor (https://pypi.python.org/pypi/pathlib2/), I am not sure, that it supports py3. Further pathlib2 has already such a fix like my provided one.

Offtopic: I am using a python package called "nt" since 2.5 years and I only knew since some month that this package does not work under windows. So normally this name not a problem under python as long it is not under windows.
I tried many repos and this is the first package that I try that has this check. I know that the best is to change the name, but I am not the only one that uses this software, so I have to convince the others.

nateprewitt:
@boeddeker, we actually aren't using pathlib anymore. I've merged #867 to remove it and this should be fixed in the next release.

boeddeker:
@nateprewitt thanks for the fix.

PR #867, opened 2017-10-09
remove unused vendored pathlib

nateprewitt:
We've had pathlib vendored since the 3.x series but we don't actually use it anywhere. Lets remove it since it's just increasing download size and causing issues in #864.


nateprewitt:
Green, so I'm merging.

PR #818, opened 2017-10-02
Fix install for local sdist packages

techalchemy:
* Fixes #817
* Packages werent being identified as local files
* Therefore they were not being given a path
* We were trying to resolve them with pypi

Issue #817, opened 2017-10-02
Pipenv does not support installing from local source packages


chrish42:
I created a virtualenv with pipenv, and installed some required packages from PyPI with it, which works well. However, I also have in my project a Python source distribution tarball (created by running `python setup.py sdist` in another project), that I'd like to install in the virtualenv. When running `pipenv install mytarball.tar.gz`, the source tarball is successfully installed in the virtualenv, but then I get an error at the end of the process:

```
Adding tarball-1.5.2.tar.gz to Pipfile's [packages]…
Locking [dev-packages] dependencies…
Locking [packages] dependencies…
Traceback (most recent call last):
  File "/Users/christianhudon/miniconda3/bin/pipenv", line 11, in <module>
    sys.exit(cli())
  File "/Users/christianhudon/miniconda3/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/Users/christianhudon/miniconda3/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/Users/christianhudon/miniconda3/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/christianhudon/miniconda3/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/christianhudon/miniconda3/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/Users/christianhudon/miniconda3/lib/python3.6/site-packages/pipenv/cli.py", line 1852, in install
    do_lock(system=system, pre=pre)
  File "/Users/christianhudon/miniconda3/lib/python3.6/site-packages/pipenv/cli.py", line 1083, in do_lock
    pre=pre
  File "/Users/christianhudon/miniconda3/lib/python3.6/site-packages/pipenv/utils.py", line 421, in resolve_deps
    resolved_tree.update(resolver.resolve(max_rounds=PIPENV_MAX_ROUNDS))
  File "/Users/christianhudon/miniconda3/lib/python3.6/site-packages/pipenv/patched/piptools/resolver.py", line 88, in resolve
    self.their_constraints))
[...]

PR #868, opened 2017-10-10
Introduces a test for PR #818 and issue #817 local sdists

techalchemy:


tests/test_pipenv.py
```diff
@@ -651,3 +652,26 @@ def test_urls_work(self):
             dep = p.lockfile['default'][key]
 
             assert 'file' in dep
+
+    @pytest.mark.install
+    @pytest.mark.files
+    def test_local_zipfiles(self):
+
+        with PipenvInstance() as p:
+            zip_contents = requests.get('https://github.com/divio/django-cms/archive/release/3.4.x.zip')
```
nateprewitt: I'd prefer to either find a way to stub this out or bundle a condensed example.zip package if we can. Network calls tend to be where a lot of test time and failures come from.

techalchemy: Sure I wasn't totally sure about the best practice for this so I just copied the test above it (though to be fair that is testing a URL as well).  I didn't want to just dump this in the repo without talking ot someone though.  Might want to find something smaller to be honest.  Where should I put it?

nateprewitt: Sorry I missed this last night @techalchemy. I'd create a directory under tests with "test_artifacts" are something like that and drop it in there.

techalchemy:
OK @nateprewitt sorry for being slow but I got around to making those changes.  I hope this doesn't break travis/appveyor, but we shall see

nateprewitt:
No problem, thanks for getting things put together :) I won't have a chance to look at this till I get home but if we get green builds, I think we'll be set to merge.

techalchemy:
@nateprewitt the zip I used was grabbed from tag 0.12.1 on github a few hours ago with this command exactly:

```console
$ wget -O tablib.zip https://github.com/kennethreitz/tablib/archive/v0.12.1.zip
```

I went ahead and grabbed the tar.gz from pypi however, and since this test is more about ensuring that installing an archive in the current directory works (i.e. with the command `pipenv install tablib.zip`), the same test is valid using a tar.gz instead so I just grabbed the one you linked and dropped it in as a replacement with no modification.

nateprewitt:
@techalchemy, ah ok, there must be something on master that varies from the last release which is likely why things weren't lining up. Thanks for taking care of that so quick. Once the Appveyor wraps, I'll merge.

nateprewitt:
:sparkles:

PR #874, opened 2017-10-10
 Fix pep508 logic assuming single-digit minor versions

joshfriend:
If the python version were to reach `3.10.0`, the logic from pep508 would result in a `python_version` of `3.1` instead of `3.10`

### DIFF
--- a/pipenv/patched/pip/_vendor/packaging/markers.py
+++ b/pipenv/patched/pip/_vendor/packaging/markers.py
@@ -266,7 +266,7 @@ def default_environment():
         "platform_version": platform.version(),
         "python_full_version": platform.python_version(),
         "platform_python_implementation": platform.python_implementation(),
-        "python_version": platform.python_version()[:3],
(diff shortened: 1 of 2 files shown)


nateprewitt:
@joshfriend, thanks for taking the time to put this together! I agree that this will eventually be a problem, however I have a quick note here. The initial patch is to a currently unmodified file from pip itself. Rather than keeping track of that on our end, it would be much better to get this merged upstream.

If you wouldn't mind opening a PR there, and seeing it through to merge, that would be really helpful. At that time, I think we can look at merging this. I'm going to close this out for now though since we're at least 3.5 years out from a 3.x release greater than 9. If it's not fixed before then, we'll know it when pip does. Please feel free to ping here when things are set up in the pip repo.

Thanks again!

joshfriend:
> The initial patch is to a currently unmodified file from pip itself

I somehow did not look at the file path and realize that, sorry 😝 

The PEP508 document also has it wrong.

I'll send a fix to pip first though, even if its a long ways away from being a problem :)

Issue #887, opened 2017-10-12
Unexpected VCS dependencies added to the Pipfile.lock


k4nar:
When a VCS dependency is present in the Pipfile, any VCS dependency in the virtualenv will be added to the Pipfile.lock. This can include packages installed with `pip install -e <path>`, as it's represented in the `pip freeze` as `-e git+ssh://...` if the path corresponds to a git repo.

In my case this wasn't an issue before #835 as git dependencies where not parsed correctly, but since this PR I'm seeing unexpected git repositories showing up in my Pipfile.lock.

I think that the culprit is in this loop, which adds any VCS dependency present in the `pip freeze` to the Pipfile.lock without checking if they are in the Pipfile:
https://github.com/kennethreitz/pipenv/blob/8685ba80d75c6489bbb4dce19861c9e91c2636ed/pipenv/cli.py#L1050-L1061

I'm working on a fix.

techalchemy:
I’m pretty sure that is the intended behavior and anything that alters it will be breaking the functionality instead of fixing it.  Pipenv by default is built to treat whichever virtualenv it manages as a representation of a dependency state. Pipfiles are a basis for installation (so you dont have to feed in the output of pip freeze anymore), while lockfiles are a more detailed representation of the present state of the environment (I.e. if it appears in pip freeze it should appear in the lockfile).

#835 actually FIXED an issue preventing vcs dependencies from showing up correctly in the lockfile. 

Out of curiosity, what problem is this causing for you?

k4nar:
In my project if I do a `pipenv install -e .`, and I have another VCS dependency in my Pipfile, every time I run `pipenv lock` the Pipfile.lock will be edited with the following:
```json
        "e1839a8": {
            "git": "ssh://git@github.com/<orga>/<repo>.git",
            "ref": "<commit>",
            "editable": "."
        },
```

Instead of:
```json
        "e1839a8": {
            "path": ".",
            "editable": "."
        },
```

Also if I happen to install locally a VCS dependency which is not present in my Pipfile, it will be added to the Pipfile.lock. This is not true for non-VCS dependencies, and I'm pretty sure this is not an expected behavior.

techalchemy:
Hmmmmmmmmmmmm I think you are right, the example helped. Good catch! I did double check to confirm that we don't include local packages in the lockfile that are not installed directly via pipenv.

```fish
 ◰³ tempenv-4507121013db0  ~/g/tmp2  pipenv install requests
[...]

PR #889, opened 2017-10-12
Don't lock VCS dependencies not listed in the Pipfile

k4nar:
Refs #887

The regressions test fails without the second commit. `python_dateutil` is added to the Pipfile.lock even though it's not listed in the Pipfile.


techalchemy:
Thanks for putting this well-resarched pull request together,  As I mentioned in #887 I think this is indeed a bug and this appears to address it in a way that is both thorough and effective. I've been through a good chunk of the VCS dependency code at some point or other so I feel comfortable giving this a thumbs up, pending the final review of @nateprewitt @kennethreitz or @erinxocon

PR #895, opened 2017-10-13
Shells with PIPENV_IGNORE_VIRTUALENVS on win

### LINKED ISSUE #894: PIPENV_IGNORE_VIRTUALENVS and pipenv shell
On my Windows 10 machine, I have PIPENV_IGNORE_VIRTUALENVS set to 1 and most pipenv commands seem to work using the .venv correctly... with the exception of `pipenv shell`

    Adam@ADAM-DESKTOP C:\Users\Adam\projects\project1
    > pipenv install
    Using base prefix 'c:\\python36'
    New python executable in C:\Users\Adam\projects\project1\.venv\Scripts\python.exe
    
    Virtualenv location: C:\Users\Adam\projects\project1\.venv
    Updated Pipfile.lock!
    Installing dependencies from Pipfile.lock…
      ================================ 0/0 — 00:00:00
    To activate this project's virtualenv, run the following:
     $ pipenv shell
    
    Adam@ADAM-DESKTOP C:\Users\Adam\projects\project1
    > pipenv --venv
    C:\Users\Adam\projects\project1\.venv
    
    Adam@ADAM-DESKTOP C:\Users\Adam\projects\project1
    > pipenv shell
    ERROR: Environment 'project1-z9NXaua1' does not exist. Create it with 'pew new project1-z9NXaua1'.

techalchemy:
* Allow PIPENV_IGNORE_VIRTUALENVS to work with pipenv shell
* Fixes #894

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -2038,7 +2038,8 @@ def do_shell(three=None, python=False, fancy=False, shell_args=None):
         import subprocess
         # Tell pew to use the project directory as its workon_home
         with temp_environ():
-            os.environ['WORKON_HOME'] = project.project_directory
+            if PIPENV_VENV_IN_PROJECT:
+                os.environ['WORKON_HOME'] = project.project_directory
             p = subprocess.Popen([cmd] + list(args), shell=True, universal_newlines=True)
             p.communicate()
             sys.exit(p.returncode)
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -25,19 +25,17 @@ import pipdeptree
 import requirements
 import semver
 import flake8.main.cli
-import pew
 from pipreqs import pipreqs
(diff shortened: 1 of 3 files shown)


techalchemy:
Hm I don’t known if it’s a good idea to try to vendor all of psutil. It breaks on windows unless installed at the system level with access to certain dlls to build the requisite pyd file

techalchemy:
Ok I think this might be ok, but I think we need to revisit vendoring psutil because it will never work correctly if we put it in a virtualenv on windows (since it can't build or link)

techalchemy:
ok I worked around the psutil issues, wrote some not that fun tests but hopefully this is good to go

erinxocon:
This should be good.  We started relying on psutil when using pew so that should help resolve some things.

Issue #881, opened 2017-10-12
Feature request: pipenv graph --why


mattjegan:
# Motivation

I tried to run `pipenv install` and recieved the warning:

```
Warning: Your dependencies could not be resolved. You likely have a mismatch in your sub-dependencies.
  You can use $ pipenv install --skip-lock to bypass this mechanism, then run $ pipenv graph to inspect the situation.
```

Upon running `pipenv install --skip-lock` I ran `pipenv graph` too try inspect the dependancy tree for the specific package that couldn't be resolved. For a large project, it takes a considerable amount of time to find the package since it is a sub-dependancy in a large tree. It would be useful to be able to get just the subtrees I'm looking for. 

# Proposal

I propose that a `--why` flag could be added to `pipenv graph` such that we can get the subtrees associated with some package. 

## Example

The currrent behaviour would look like this:
```
$ pipenv graph

.
.
.
blessed==1.14.2
  - six [required: >=1.9.0, installed: 1.11.0]
  - wcwidth [required: >=0.1.4, installed: 0.1.7]
.
.
.
cement==2.10.2
ConfigArgParse==0.12.0
crayons==0.1.2
  - colorama [required: Any, installed: 0.3.9]
.
.
.
django-cacheops==4.0.1
  - django [required: >=1.8, installed: 1.11.6]
    - pytz [required: Any, installed: 2017.2]
  - funcy [required: <2.0,>=1.8, installed: 1.9.1]
  - redis [required: >=2.9.1, installed: 2.10.6]
  - six [required: >=1.4.0, installed: 1.11.0]
.
.
.
```

But upon trying to search for `six` usages we would get something like this:

```
$ pipenv graph --why six

blessed==1.14.2
  - six [required: >=1.9.0, installed: 1.11.0]
django-cacheops==4.0.1
  - six [required: >=1.4.0, installed: 1.11.0]
```

I believe this would make manual dependency resolving much faster for users. 

**Also, thanks for the awesome work** 😄

nateprewitt:
Hey @mattjegan, thanks for putting this proposal together! I think that this could be useful but is likely a tedious command if you're looking for multiple packages. I believe we'll be going with the `--reverse` option (#904) as an alternative since it gives single command for a complete overview of "why?" questions.

Hopefully this solves the issue that you were looking to address. Please let us know if you have any more ideas on things you'd like to see in Pipenv!

PR #904, opened 2017-10-13
Add pipenv graph --reverse

vphilippon:
Outputs the reversed dependencies graph.

Ex:
```
requests==2.18.4
  - lib_with_good_taste==1.0.0 [requires requests>=2.18]
    - package_using_lib_with_good_taste==1.2.3 [requires lib_with_good_taste]
  - another_package [requires requests<3.0,>=2.0]
```

Inspired by #881. Even if it's not exactly the same feature, this is a quick-win that covers essentially the same need: have a compact view of "who" required a given package. Plus, you get the full chain with this.

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -2306,7 +2306,8 @@ def check(three=None, python=False, unused=False, style=False, args=None):
 @click.command(help=u"Displays currently–installed dependency graph information.")
 @click.option('--bare', is_flag=True, default=False, help="Minimal output.")
 @click.option('--json', is_flag=True, default=False, help="Output JSON.")
-def graph(bare=False, json=False):
+@click.option('--reverse', is_flag=True, default=False, help="Reversed dependency graph.")
+def graph(bare=False, json=False, reverse=False):
     try:
         python_path = which('python')
     except AttributeError:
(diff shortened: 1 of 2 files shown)


techalchemy:
Useful feature to be sure, we should make a note to add it to the documentation as well

vphilippon:
@techalchemy It's already in it, automatically thanks to `sphinx-click`.
It will appear under that section: https://docs.pipenv.org/#pipenv-graph

techalchemy:
well damn, shows how much i know about our CLI library.  I couldn't even figure out if there is a way to take a command line argument and assign it to a variable named something different :) I'm pretty sure this must be doable since you can do it with invoke and other similar less-featured tools

nateprewitt:
Thanks for putting this together @vphilippon. I think the straight `--reverse` output looks good, but running both `--reverse` and `--json` gives unexpected results. I'm assuming this is a bug in pipdeptree but the json output entries aren't reversed.

We may want to disable this combo or at least warn that the results won't be what users expect.

Also could we get a slightly more expanded test? This one only shows that pipenv doesn't completely explode when passing that parameter. It could still be exiting out with a non-0 exit code, or exiting 0 with non-reversed data.

Thanks!

vphilippon:
@nateprewitt Thanks for the review, I'll improve the tests and block the `--json --reverse` combination.
I'll get on that soon:tm: when I have some free time.

vphilippon:
I've pushed my branch to keep my progress, but you should wait before merging.

In its current form, my test could fail in the future if a dependency of `requests` or `pipenv` was to change and gain a dependency (with a name with a lower lexicographic order than `requests`) and that dependency have a dependency on `certifi`, `chardet`, or `idna` (ex: `request==2.18.4` depends on `X`, which depends on `foo`, and tomorrow, `foo` starts depending on  `certifi`. `foo` would appear directly under `certifi` in the reversed graph, before `requests`, making this test suddenly fail).

I'll have to improve that test, that was just a first simple idea what was likely too simple.

vphilippon:
Looks like done, except for one test that failed on a connection error, but I'm pretty sure its unrelated to the change.

PR #906, opened 2017-10-14
fix typo Pipenv.lock -> Pipfile.lock

Luit:


nateprewitt:
Good catch @Luit, thanks! :sparkles:

PR #907, opened 2017-10-14
Move COPY . /app after pipenv install

uda:
When working with concrete projects and/or auto-builds, you won't want every push to require a full dependency pull, besides the network issue there is the timing issue.

You should rebuild the dependencies only if Pipfile / Pipfile.lock is changed.

P.S the new empty line was unintentional, added auto-magically by the Github online editor.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,10 +15,12 @@ WORKDIR /app
 
 # COPY Pipfile Pipfile
 # COPY Pipfile.lock Pipfile.lock
-# COPY . /app
 
 # -- Install dependencies:
 # RUN pipenv install --deploy --system
+
+# -- After installing requirements copy the rest of the code
+# COPY . /app
 
 ENTRYPOINT []
 CMD [ "/bin/bash" ]


nateprewitt:
Hey @uda, sorry if I'm being slow here. What is it about copying the `app` directory over that is causing the Pipfile.lock to rebuild every time? Is `app` changing your Pipfile for some reason?

mattjmcnaughton:
@nateprewitt My understanding is that a Docker container is composed of layers. Anytime a layer changes, all of the layers above must be rebuilt. So before this change, if any of the code in the app directory changes, we will need to rebuild all of the above layers, including the one which install the deps through Pipfile. Typically with containers, the goal is to put more expensive layers which are less likely to change (i.e. installing dependencies) early in the Dockerfile, and more likely to change operations (i.e copying the application code) later in the Dockerfile. Hope that's helpful :)

uda:
Thanks @mattjmcnaughton, exactly.

erinxocon:
Closing this PR as we're using onbuild for the Pipfile and Pipfile.lock now as per PR #918

uda:
Thanks, That definitely looks better

PR #910, opened 2017-10-15
Add cacert.pem from requests to releases. Fixes #871

### LINKED ISSUE #871: "Could not find a version that matches coloredlogs" when creating lockfile
Pipfile contents:

```toml
[[source]]
verify_ssl = true
name = "pypi"

[packages]

flask = "*"
peewee = "*"
"psycopg2" = "*"
coloredlogs = "*"
"openvpn-status-parser" = {git = "https://github.com/ojarva/openvpn-status-parser.git"}
"python-dateutil" = "*"
"dbus-python" = "*"
gunicorn = "*"
telepot = "*"
"pymorphy2" = "*"
"pymorphy2-dicts-ru" = "*"
telethon = "*"
```

Running `pipenv install` results in 

```
Could not find a version that matches coloredlogs
Tried: (no version found at all)
```

`pipenv install --skip-lock` works fine and installs all required packages.

Edit: using the latest versions of `pipenv` and all dependencies from PyPI on Python 3.6.2 on Arch Linux.

NighTurs:
In #871 it was found that problem is with cacert.pem missing from vendored requests. Although cacert.pem exists in repository it is ignored during build. By digging through commits I found that previously it was included through MANIFEST.in, but was changed in 6ac67d9 commit. Looking at commit message and change it doesn't seem like it was intended. Likely it was broken even before that, because at the moment of that change path pipenv/vendor/requests/cacert.pem didn't exist (I guess there was some restructuring going on before that).

There is a note about cacert.pem in pip/_vendor/README.rst on how you can optionally remove it, which seems strange if cacert.pem should be ignored by build like it is right now.

Still I might be wrong, but it fixes #871 for me.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,4 +1,5 @@
 include README.rst LICENSE NOTICES HISTORY.txt pipenv/patched/safety.zip
+include pipenv/patched/pip/_vendor/requests/cacert.pem
 include pipenv/vendor/pipreqs/stdlib
 include pipenv/vendor/pipreqs/mapping
 include pipenv/pipenv.1


nateprewitt:
So I added this specifically for this reason in 35518da and I'm not sure why it was removed. @kennethreitz was this a slip up or did you have specific concerns about bundling the cacerts.pem?

We'll need to bundle a newer version of Requests if we have a reason not to include the certs file.

erinxocon:
For now we're going to merge this unless someone has a reason not to!  Thanks @NighTurs

PR #918, opened 2017-10-16
Uses ONBUILD in Dockerfile to install deps while building

xsteadfastx:
Copies `Pipfile` and `Pipefile.lock`, installs deps on building a docker
image using the official pipenv image.

This is a approach i saw on the official python onbuild images. Maybe this makes things a little more convenient...

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,19 @@
-FROM python:3.6.2
+FROM python:3.6.3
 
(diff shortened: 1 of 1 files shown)


erinxocon:
Looks good to me!

erinxocon:
Thanks for the first time contribution @xsteadfastx!

PR #926, opened 2017-10-17
Fix pip file deps conversion

krismolendyke:
Fix for determining if a package name is a locally installed package.  I ran into this because a project I work on depends on the `ansible` Python package _and_ has a top-level `ansible/` directory which is _not_ a Python package.  

The current logic detects the `ansible/` filesystem directory as a Python package file installation and sets the package name to the hash of the directory path, which is nonsense.  Then, `pip` tries to uninstall something like `deadbeef` and fails, not uninstalling anything after that non-existent "package".

I kept this change simple and stupid as I believe that the file (un)installation support is really just about _files_, and I didn't want to drag `importlib` into this.

re: 

- https://github.com/kennethreitz/pipenv/pull/818
- https://github.com/kennethreitz/pipenv/pull/868

### DIFF
--- a/tests/test_pipenv.py
+++ b/tests/test_pipenv.py
@@ -239,19 +239,7 @@ class TestPipenv:
         with PipenvInstance() as p:
             shutil.copy(source_path, os.path.join(p.path, file_name))
             os.mkdir(os.path.join(p.path, "tablib"))
-
             c = p.pipenv('install {}'.format(file_name))
-            key = [k for k in p.pipfile['packages'].keys()][0]
-            dep = p.pipfile['packages'][key]
-
-            assert 'file' in dep or 'path' in dep
-            assert c.return_code == 0
-
-            key = [k for k in p.lockfile['default'].keys()][0]
-            dep = p.lockfile['default'][key]
-
-            assert 'file' in dep or 'path' in dep
-
             c = p.pipenv('uninstall --all --verbose')
             assert c.return_code == 0
             assert 'tablib' in c.out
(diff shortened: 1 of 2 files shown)


techalchemy:
good catch, I was knee deep in some of this stuff when I wrote it so I didn't have the wherewithal to step back from this. This is a good fix.

tests/test_pipenv.py
```diff
@@ -227,6 +227,36 @@ def test_uninstall(self):
...
+        # Not sure where travis/appveyor run tests from
+        test_dir = os.path.dirname(os.path.abspath(__file__))
+        source_path = os.path.abspath(os.path.join(test_dir, 'test_artifacts', file_name))
+
+        with PipenvInstance() as p:
+            shutil.copy(source_path, os.path.join(p.path, file_name))
+            os.mkdir(os.path.join(p.path, "tablib"))
+
+            c = p.pipenv('install {}'.format(file_name))
+            key = [k for k in p.pipfile['packages'].keys()][0]
```
nateprewitt: I should have caught this when the original test was added. Why are we creating a new list for keys when keys() already returns a list?

Let's not continue this trend into new tests.

krismolendyke: Probably because in Python 2 `keys()` returns a `list` but in Python 3 `keys()` returns a _view_ that would need to be coerced to a `list` to be indexed.

techalchemy: we do this in a number of places in the codebase for this reason and in the other tests that check on this as well

krismolendyke: OK rather than deal with this silly 2/3 issue I just removed the copy/pasta of the offending test.  The original test remains but this new one does not propagate that behavior.

nateprewitt:
Thanks @krismolendyke!

krismolendyke:
🙇

PR #935, opened 2017-10-18
Force vcs installs to block to avoid race condition

### LINKED ISSUE #919: Editable dependencies have a race condition when installing
When installing editable dependencies, they must update `site-packages/easy-install.pth` with the path to the src of the repo checkout, for it to be able to be imported. By default pipenv installed all dependencies in blocks that run concurrently, but since multiple editable installs edit the same file this creates a race condition where multiple files may be trying to edit at the same time, resulting in some not being included.

In about 10 runs of `pipenv install` with the Pipfile below, 4 only included the second (variant) path, 2 only included the first (affect) path, and 4 included both paths. This can be worked around using `--sequential` at the cost of installs being much slower.

Ideally, pipenv could gather all editable dependencies and run them all sequentially at the end of the install process so as to not cause the race condition.

Python 2.7
Pipenv 8.2.7

Pipfile:
```
[[source]]
verify_ssl = true
name = "pypi"

[dev-packages]

[packages]
django-affect = {ref = "master", git = "https://github.com/jsatt/django-affect.git", editable = true}
django-variant = {ref = "master", git = "https://github.com/jsatt/django-variant.git", editable = true}
```

jsatt:
Fixes #919

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -783,11 +783,11 @@ def do_install_dependencies(
                 del v['hash']
 
     # Convert the deps to pip-compatible arguments.
-    deps_list = [(d, ignore_hashes) for d in convert_deps_to_pip(deps, project, r=False, include_index=True)]
+    deps_list = [(d, ignore_hashes, blocking) for d in convert_deps_to_pip(deps, project, r=False, include_index=True)]
     failed_deps_list = []
 
     if len(vcs_deps):
-        deps_list.extend((d, True) for d in convert_deps_to_pip(vcs_deps, project, r=False))
+        deps_list.extend((d, True, True) for d in convert_deps_to_pip(vcs_deps, project, r=False))
 
     # --requirements was passed.
     if requirements:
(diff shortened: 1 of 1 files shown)


pipenv/cli.py
```diff
@@ -783,11 +783,11 @@ def cleanup_procs(procs, concurrent):
                 del v['hash']
 
     # Convert the deps to pip-compatible arguments.
-    deps_list = [(d, ignore_hashes) for d in convert_deps_to_pip(deps, project, r=False, include_index=True)]
+    deps_list = [(d, ignore_hashes, blocking) for d in convert_deps_to_pip(deps, project, r=False, include_index=True)]
```
techalchemy: Since you changed the variable below to `block` shouldn't this also be  `block` instead of `blocking`

jsatt: No. This `blocking` is coming from above (line 743). It's providing the regular dependencies a value of whether the install should block or not based on the `concurrent` arg passed into the function. On line 790, I'm always setting that block value to be true for vcs dependencies.

The `block` below (line 801) is unpacking the value that is added to the tuples here for use when passing into the `pip_install` call. I made it a point to name them differently so it doesn't overload the `blocking` variable and cause potential issues and confusion down the road.

nateprewitt:
From an 👀-only check, this looks ok to me. I would like a maintainer or one of our regular contributors to verify this is a properly diagnosed bug before we merge though. I'll try to set aside some time in the next few days if no one else gets to it.

pipenv/cli.py
techalchemy: Thanks for the explanation, that makes good sense to me.

erinxocon:
@nateprewitt I can verify that this is a bug.  This looks good to me, I'm going to merge!

erinxocon:
need to write a test for this

Issue #540, opened 2017-09-14
issues with relative packages


AlJohri:
@kennethreitz I'm having trouble installing local packages in relative directories via `-e`.

<img width="769" alt="screen shot 2017-09-13 at 7 54 55 pm" src="https://user-images.githubusercontent.com/2790092/30406072-7008675e-98bd-11e7-8a60-bfb9f57cfe10.png">

This is the initial requirements.txt file:

```
-e ../../lib/python
arrow==0.10.0
blessings==1.6
boto3==1.4.7
```

This is what the converted Pipfile looks like:
```
[[source]]
url = "https://pypi.python.org/simple"
verify_ssl = true

[packages]
"-e file:///users/johria/development/heliograf/lib/python" = "*"
arrow = "==0.10.0"
blessings = "==1.6"
```

`heliograf` in this case is the name in setup.py of the relative package.

I believe the issue may be that the Pipfile.lock doesn't have the directory?

```
        "heliograf": {
            "hashes": [],
            "version": "==0.1.0"
        },
```

--------------------

While the above is starting from the requirements.txt, this _kind of_ works: `pipenv install -e ../../lib/python`

It adds another entry into the Pipefile like this:

```
"-e ../../lib/python" = "*"
```

I can see that the package is installed properly by running `pipenv run python` and importing the package. In addition, doing a `pipenv shell` followed by a `pip freeze` shows:

```
-e git+git@github.com:WPMedia/heliograf.git@4dbd528e2f2daeb4b4462b29d2dc8c54341428b5#egg=heliograf&subdirectory=lib/python
```

potentially related issues:
- https://github.com/kennethreitz/pipenv/issues/66
- https://github.com/kennethreitz/pipenv/issues/67

versions:
- python version: 3.6.2
- pipenv version: 7.0.1
- python installed via: pyenv

techalchemy:
The actual problem you're experiencing seems to be that you have a setup.py file somewhere with an explicit pinned dependency on a version of heliograf that doesn't exist in a repository pipenv can query.  This is fine, because you have that package installed and you claim it is the right version, but note that even pip doesn't know which version it is because it is installed from VCS.  I'm not sure this is a pipenv issue.

kennethreitz:
try using an absolute path instead, see if that changes anything

AlJohri:
[...]

PR #936, opened 2017-10-18
Fix is_file

krismolendyke:
Related to #926, `utils.is_file` seems like it should check not only that a file exists but that a file is a file and not a directory.

I added some tests for `utils.is_file` but I don't have any pytest experience so please let me know if there's a better way to accomplish that logic.

### DIFF
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -771,7 +771,7 @@ def is_file(package):
     if hasattr(package, 'keys'):
         return any(key for key in package.keys() if key in ['file', 'path'])
 
-    if os.path.exists(str(package)):
+    if os.path.isfile(str(package)):
         return True
 
     for start in FILE_LIST:
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -534,8 +534,8 @@ def convert_deps_from_pip(dep):
     extras = {'extras': req.extras}
 
     # File installs.
-    if (req.uri or (os.path.isfile(req.path) if req.path else False) or
-            os.path.isfile(req.name)) and not req.vcs:
+    if (req.uri or (is_file(req.path) if req.path else False) or
+        is_file(req.name)) and not req.vcs:
         # Assign a package name to the file, last 7 of it's sha256 hex digest.
         if not req.uri and not req.path:
             req.path = os.path.abspath(req.name)
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -2,6 +2,8 @@
 
 import os
 import pytest
+import shutil
+import tempfile
 from mock import patch, Mock
 
 import pipenv.utils
@@ -106,6 +108,25 @@ class TestUtils:
         with pytest.raises(ValueError) as e:
             dep = pipenv.utils.convert_deps_from_pip(dep)
             assert 'pipenv requires an #egg fragment for vcs' in str(e)
+
+    @pytest.mark.parametrize('entry, expected', [
+        ('http://example.com/package', True),
+        ('https://example.com/package', True),
+        ('ftp://example.com/package', True),
+        ('file:///tmp/package', True)
+    ])
+    def test_is_file(self, entry, expected):
+        pipenv.utils.is_file(entry) is expected
+
+    def test_is_file_files(self):
+        with tempfile.NamedTemporaryFile() as temp_file:
+            assert pipenv.utils.is_file(temp_file.name)
+
+        dir_name = tempfile.mkdtemp()
+        try:
+            assert pipenv.utils.is_file(dir_name) is False
+        finally:
+            shutil.rmtree(dir_name)
 
     @pytest.mark.parametrize('version, specified_ver, expected', [
         ('*', '*', True),


nateprewitt:
@techalchemy, I think you've got a good grasp on most of the places is_file is used. This seems sensical to me, but I want to double check this won't have unintended consequences. Any thoughts?

techalchemy:
WE can't actually do this unless we add an additional check everywhere we check for `is_file` to see if something is a directory.  The reason is that often someone attempts to install something such as `-e ../package/` which would install a package in a relative path using its `setup.py`. See for instance #540 -- but if you want to make the naming more accurate we can make the function name `is_path` or something

techalchemy:
The fact that all the tests pass with this change is concerning, I guess I need to add one to cover this condition

krismolendyke:
That was my concern, too, and why in #926 i didn't push this change into `utils.is_file`.  Happy to take a different approach here but I still need figure out a way to discern between filesystem directories and Python packages.  Maybe we are back in an `importlib` or `__init__.py` sniffing scenario?

techalchemy:
What is driving the need to make this distinction?  I'm not understanding the issue you are attempting to fix here

krismolendyke:
My particular issue stems from this logic https://github.com/kennethreitz/pipenv/blob/9c4b68605ce60c4650b30289cabcf941fa7e5fa7/pipenv/project.py#L72.

As mentioned in #926 I have a project that requires the `ansible` package _and_ has a project root directory named `ansible/`.  Since that logic detects the `ansible/` directory as a file, it does not add the legitimate `ansible` package from my `Pipfile` to the package list.

krismolendyke:
I should note that the `ansible/` directory is nothing directly related to any Python packages.  It is simply a place where playbooks and other Ansible stuff is kept.  It's nothing like an editable Python package installation.

techalchemy:
Do you have an issue open for this? I'm not clear on how pipenv is detecting your project root directory name is ansible? Can you provide instructions for how to replicate this?

techalchemy:
For example I created a new project directory called ansible and ran `pipenv install ansible` without issue.

```pycon
    >>> project.project_directory
    '/tmp/ansible'

    >>> project.packages
    {'ansible': '*'}
```

```console
 ◰³ tempenv-522680565567  /t/ansible  cat Pipfile
[[source]]
verify_ssl = true
name = "pypi"

[packages]
ansible = "*"

[dev-packages]

 ◰³ tempenv-522680565567  /t/ansible  pipenv install
Courtesy Notice: Pipenv found itself running within a virtual environment, so it will automatically use that environment, instead of creating its own for any project.
Updated Pipfile.lock (114e81)!
Installing dependencies from Pipfile.lock (114e81)…
  🐍   ▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉ 14/14 — 
```

krismolendyke:
It's a work project so I can't share 😉 but the layout is similar to:

```sh
$ ll -R project-root
total 0
-rw-r--r--  1 kris  staff     0B Oct 18 16:49 Pipfile
-rw-r--r--  1 kris  staff     0B Oct 18 16:49 Pipfile.lock
drwxr-xr-x  2 kris  staff    68B Oct 18 16:49 ansible/
```

More generally, this is a problem whenever a Python package name parsed from a section of the `Pipfile` matches the name of a directory immediately next to `Pipfile`.  I can work up a failing test case later.

If in [your example](https://github.com/kennethreitz/pipenv/pull/936#issuecomment-337723560) you simply made `ansible` an immediate sub-directory of the project root, I think the issue will present itself.

krismolendyke:
I'm going to close this because it seems like mucking with `utils.is_file` is not the best approach.  I'll open another pull request w/ a more clear failing test case that illustrates the issue.

Thanks for the discussion.

techalchemy:
Gotcha.  I mainly have two concerns and I am not sure which approach is best

* I want to be sure that people who intend to install a local package can do that (e.g. pipenv install xyz.zip or pipenv install xyz.whl), as well as local folders which are packages (maybe that requires an additional check for directory paths or enforces ./ and ../ on those paths which are directories?)
* Alternatively we could take a different approach and alter the way projects spit back their package lists?
* is_file is used in a lot of places, but perhaps should be renamed 'is_local_path' or something similar... it was never really meant to test remote URIs
* Thanks to you as well for your patience, this is a situation I hadn't considered but seems like a valid use case, I'm just not sure about the optimal approach

krismolendyke:
No sweat.  I have some thoughts on those topics and others that I'll let bake a bit longer.  In the meantime I have a local fix for my immediate issue.

techalchemy:
Excellent! I’d be glad to hear them and continue the discussion on how to handle this.  If you don’t have an issue opened on this I’ll go ahead and create one to track this

krismolendyke:
🆒  go ahead and create an issue, thanks.  I'll collect my thoughts there.  If I had to boil this down I'd say that the basic issue is distinguishing local/editable "installations" of Python packages from identically named "plain" filesystem files/directories.

Issue #911, opened 2017-10-15
`pipenv install .` in empty dir gives unexpected error


iandanforth:
##### Describe you environment

        "host-environment-markers": {
            "implementation_name": "cpython",
            "implementation_version": "0",
            "os_name": "posix",
            "platform_machine": "x86_64",
            "platform_python_implementation": "CPython",
            "platform_release": "16.4.0",
            "platform_system": "Darwin",
            "platform_version": "Darwin Kernel Version 16.4.0: Thu Dec 22 22:53:21 PST 2016; root:xnu-3789.41.3~3/RELEASE_X86_64",
            "python_full_version": "2.7.10",
            "python_version": "2.7",
            "sys_platform": "darwin"
        }

1. Pipenv version: 

version 8.2.7

##### Expected result

Either

1. Behaviour equivalent to `pipenv install` with no errors
or
2. An error explaining why specifying the local dir is bad, what is missing, etc.

##### Actual result

Traceback (most recent call last):
  File "/usr/local/bin/pipenv", line 11, in <module>
    load_entry_point('pipenv==8.2.7', 'console_scripts', 'pipenv')()
  File "/Library/Python/2.7/site-packages/pipenv/vendor/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/Library/Python/2.7/site-packages/pipenv/vendor/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/Library/Python/2.7/site-packages/pipenv/vendor/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Library/Python/2.7/site-packages/pipenv/vendor/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Library/Python/2.7/site-packages/pipenv/vendor/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/Library/Python/2.7/site-packages/pipenv/cli.py", line 1793, in install
    c = pip_install(package_name, ignore_hashes=True, allow_global=system, no_deps=False, verbose=verbose, pre=pre)
  File "/Library/Python/2.7/site-packages/pipenv/cli.py", line 1316, in pip_install
    if [x for x in requirements.parse(package_name.split('--hash')[0].split('--trusted-host')[0])][0].vcs:
  File "/Library/Python/2.7/site-packages/pipenv/vendor/requirements/parser.py", line 50, in parse
[...]

Issue #939, opened 2017-10-19
Installing packages with relative paths fails -- discussion


techalchemy:
See #936 -- basically boils down to `is_file` checks for local files.  This  issue is to open a broader discussion for the proper way to handle checks for local files during installation.  Previously we have had several issues (#817, #540 for instance) where users needed to install a local sdist in the project directory (in that case a file, properly speaking, as a `tar.gz` or a `.zip`), and a relative path to a local directory containing a valid `setup.py` (e.g. `pipenv install -e ../somedirectory/` or the equivalent of `pip install -e ../...`).

As a consequence we currently are mainly checking only that a path exists locally before attempting to install it as a local path, no matter whether the user supplied a preceding `./` if it is in the current / project directory (we could require this), or whether the path is a file which can be installed with pip, or, if it is a directory, if that directory contains a python package at all and if so, if it can be installed with pip.

Essentially I would propose that we start the discussion with this set of options and whether they are suitable/viable:

* Our initial position for filesystem locations was to require `file://` URIs, but we budged on this because pip allows local paths and because users often use this syntax. We could enforce a local path 'uri' by enforcing `./` and `../` syntax and asking if the user meant `./path` if we find a matching one before proceeding
* We could ignore any directories in the project root that are not explicitly installed in editable mode - what are the potential consequences of this?
* We could attempt to pass through existing directories to a `pip install` as we are doing now, but instead of failing and pretending to succeed, we could figure out a way to fail silently and proceed to pypi or whatever repository the user has chosen.  This seems like the most idiomatic approach potentially, although I'm not sure how complex the implementation is.
* We could be dumb and just check for the presence of `setup.py` in any directory that matches a requested path and, if it is missing, dont bother trying to install that directory and just query the repository

I personally am only opposed to the first option, I don't really have strong feelings on the others and could see one or any combination of those working well.  There could easily be other options I'm not considering.  I'd be curious to hear what others have to say

/cc @krismolendyke @kennethreitz @erinxocon @nateprewitt (when you guys have time)

techalchemy:
See #914 also

kstohr:
Apologies, I just submitted a query on this issue... and didn't make the connection. 

The user should determine when/whether local packages should be added to the pipfile explicitly rather than automatically detecting and installing which could cause issues/complications for projects in development. The above mentioned `pipenv install -e ./path/to/setup.py` seems intuitive to me...

techalchemy:
Closing this -- resolved via #958

PR #941, opened 2017-10-19
catch dot and act force package name to false

erinxocon:


erinxocon:
Fixes #911

pipenv/cli.py
```diff
@@ -1776,7 +1776,11 @@ def install(
     # Capture -e argument and assign it to following package_name.
     more_packages = list(more_packages)
     if package_name == '-e':
-        package_name = ' '.join([package_name, more_packages.pop(0)])
+        package_name = ' '.join([package_name, more_packages.po(0)])
```
nateprewitt: We're missing a 'p' on 'pop' here.

pipenv/cli.py
```diff
@@ -1776,7 +1776,11 @@ def install(
...
+
+    # Capture . argument and asign it to nothing
+    if package_name == '.':
+        package_name = False
```
nateprewitt: I'm pretty sure we're going to blow up farther down the line if the user passes more than one package name because we now how a list of booleans and strings.

We may want to take another approach on this or make a more stringent check that has its own logic and exit strategy.

erinxocon: @nateprewitt Agree.  It was brought to my attention that the "." is commonly used on unix to specify current directory so I thought I'd add in a check for it.  It is a bit hairy now.

pipenv/cli.py
erinxocon: Should be fixed now in master

pipenv/cli.py
techalchemy: installing . should be totally valid, it's specified in Pipenvs pipfile -- how will this interact with `-e`?  Also if you are in a directory that is a valid python package you technically should be allowed to `install .`.  Pip allows you to `pip install .` or `pip install -e .` I believe

nateprewitt: @techalchemy you're right, but I think the issue here is the user is trying to install a package with a Pipfile but no setup.py. It's the equivalent of 'pipenv install' but we currently fail on it because the users package doesn't have the machinery pip-tools expects.

I think the solution here is to try to install because the users directory could very well be a valid package but when we hit he exception warn that they aren't in an installable directory and possibly redirect them to 'pipenv install' because that's what they likely want.

That or we alias this to pipenv install when a setup.py isn't present but that's likely going to be another parsing rats nest.

techalchemy: So this discussion is highly relevant because of #936 and #939 + ongoing bugs with local installations, and my proposed approach is in line with yours (slightly different context but generally the same) -- if a user passes a path that points to a _directory_ on the filesystem, it would be most idiomatic to attempt to install it (since it might be a valid package) and then fail -- in this case with a warning about not being installable, but in the case I'm thinking of we would want to check pypi to see if there is a matching package there as well (not applicable to installing '.' obviously)

nateprewitt: @techalchemy yeah, I think we're on the same page. We just need to get the cascade right if we're going to try different sources (local, url, pypi; not necessarily in that order) and implement it in one place.

PR #947, opened 2017-10-22
Fix a typo in do_lock() error handler

richardcooper:


nateprewitt:
Thanks @richardcooper!

Issue #882, opened 2017-10-12
IndexError on running pipenv install


mccartykim:
With a fresh install on Windows 10, I get the following failure and traceback. This is with Python 3.6 installed for all users, with Pipenv installed from `pip install --user pipenv`. I'm curious if this perhaps has something to do with the ["find_windows_executable" function](https://github.com/kennethreitz/pipenv/blob/8685ba80d75c6489bbb4dce19861c9e91c2636ed/pipenv/utils.py#L857).

```
PS C:\Users\mccar\Documents\Software Development\bittorrent> pipenv install
Creating a virtualenv for this project…

Virtualenv location:
Creating a virtualenv for this project…

Virtualenv location:
Pipfile.lock not found, creating…
Locking [dev-packages] dependencies…
Traceback (most recent call last):
  File "c:\program files (x86)\python36-32\lib\runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "c:\program files (x86)\python36-32\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "C:\Users\mccar\AppData\Roaming\Python\Python36\Scripts\pipenv.exe\__main__.py", line 9, in <module>
  File "C:\Users\mccar\AppData\Roaming\Python\Python36\site-packages\pipenv\vendor\click\core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "C:\Users\mccar\AppData\Roaming\Python\Python36\site-packages\pipenv\vendor\click\core.py", line 697, in main
    rv = self.invoke(ctx)
  File "C:\Users\mccar\AppData\Roaming\Python\Python36\site-packages\pipenv\vendor\click\core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "C:\Users\mccar\AppData\Roaming\Python\Python36\site-packages\pipenv\vendor\click\core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "C:\Users\mccar\AppData\Roaming\Python\Python36\site-packages\pipenv\vendor\click\core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "C:\Users\mccar\AppData\Roaming\Python\Python36\site-packages\pipenv\cli.py", line 1785, in install
[...]

PR #950, opened 2017-10-22
test for installing from remote file

Dreamersoul:
i am opening this pull request even though i didn't have a chance to run the test today (windows tests are broken i think #882) will take a look using my mac tomorrow
in the mean time please let me know if i used the correct decorators for the test function and if i am missing something from the test case

thank you for the amazing work

this is  a test for: `pipenv install -r http://something.com/req.txt`

### DIFF
--- a/tests/test_pipenv.py
+++ b/tests/test_pipenv.py
@@ -804,3 +804,23 @@ requests = "==2.14.0"
             dep = p.lockfile['default'][key]
 
             assert 'file' in dep or 'path' in dep
+
+    @pytest.mark.install
+    @pytest.mark.files
+    @pytest.mark.urls
+    def test_install_remote_requirments(self):
+        with PipenvInstance() as p:
+            # using a github hosted requirements.txt file
+            c = p.pipenv('install -r https://raw.githubusercontent.com/kennethreitz/pipenv/3688148ac7cfecefb085c474b092c31d791952c1/tests/test_artifacts/requirements.txt')
+            assert c.return_code == 0
+
+            print(p.lockfile['default'])
+            # check Pipfile with versions
+            assert 'requests' in p.pipfile['packages']
+            assert p.pipfile['packages']['requests'] == u'==2.18.4'
+            assert 'records' in p.pipfile['packages']
+            assert p.pipfile['packages']['records'] == u'==0.5.2'
+
+            # check Pipfile.lock
+            assert 'requests' in p.lockfile['default']
+            assert 'records' in p.lockfile['default']
--- a/tests/test_pipenv.py
+++ b/tests/test_pipenv.py
@@ -812,9 +812,8 @@ requests = "==2.14.0"
         with PipenvInstance() as p:
             # using a github hosted requirements.txt file
             c = p.pipenv('install -r https://raw.githubusercontent.com/kennethreitz/pipenv/3688148ac7cfecefb085c474b092c31d791952c1/tests/test_artifacts/requirements.txt')
-            assert c.return_code == 0
 
-            print(p.lockfile['default'])
+            assert c.return_code == 0
             # check Pipfile with versions
             assert 'requests' in p.pipfile['packages']
             assert p.pipfile['packages']['requests'] == u'==2.18.4'


tests/test_pipenv.py
```diff
@@ -796,3 +796,21 @@ def test_local_zipfiles(self):
...
+    @pytest.mark.install
+    @pytest.mark.files
+    @pytest.mark.urls
+    def test_install_remote_requirments(self):
+        with PipenvInstance() as p:
+            # using a github hosted requirements.txt file
+            c = p.pipenv('install -r https://github.com/kennethreitz/pipenv/blob/3688148ac7cfecefb085c474b092c31d791952c1/tests/test_artifacts/requirements.txt')
+            assert c.return_code == 0
+
+            # should i check versions here? or there is another test for that? hmm
```
nateprewitt: It wouldn't hurt to check versions.

tests/test_pipenv.py
```diff
@@ -796,3 +796,21 @@ def test_local_zipfiles(self):
...
             assert 'file' in dep or 'path' in dep
+
+
+    @pytest.mark.install
+    @pytest.mark.files
+    @pytest.mark.urls
+    def test_install_remote_requirments(self):
+        with PipenvInstance() as p:
+            # using a github hosted requirements.txt file
+            c = p.pipenv('install -r https://github.com/kennethreitz/pipenv/blob/3688148ac7cfecefb085c474b092c31d791952c1/tests/test_artifacts/requirements.txt')
```
nateprewitt: This is going to need to be `https://raw.githubusercontent.com/kennethreitz/pipenv/3688148ac7cfecefb085c474b092c31d791952c1/tests/test_artifacts/requirements.txt` actually. The link I originally provided is a full HTML page from github, not the raw file.

Dreamersoul:
Thanks for the amazing input will look into it today

Dreamersoul:
commits are squashed and test was passing locally. we should be all set 👍

tests/test_pipenv.py
```diff
@@ -796,3 +796,23 @@ def test_local_zipfiles(self):
             dep = p.lockfile['default'][key]
 
             assert 'file' in dep or 'path' in dep
+
```
erinxocon: This test fails on windows.  We need it to work before merging.

erinxocon: This test fails on windows. We need it to work before merging.

erinxocon: This might actually be a bug on windows in pipenv due to how windows opens and places files locks on things.

Dreamersoul: could this be the problem?
https://bugs.python.org/msg27445

it says the file is probably locked by the process
we could probably unlock the file or something?

Dreamersoul:
so after investigation turns out `tempfile.mkstemp` actually opens the file and returns the file descriptor so we have two solutions:
1. close the file immediately so basically using mkstemp just to create the file
2. we could hold on to the descriptor and close it before deleting

which one do you recommend? also if i got it wrong can someone suggest another approach

i have already implemented the second approach do you want me to push it?

techalchemy:
Oh this test... it’s probably more likely that it’s failing due to issues relating to loca file installation. I’m in the middle of untangling how that wound up broken

techalchemy:
Ah this is for the other test. I guess my question would be whether using a descriptor is the best option at all. We already create temp files for requirements as we resolve and add to Pipfiles. Is there any reason not to just iterate over requirements from the remotesfile contents? I haven’t looked at complexity for that but you’re not streaming the file content into the descriptor right? So it’s going to be all loaded in memory anyway

Dreamersoul:
@techalchemy I was thinking of streaming it but went with using temp files!
I wasn't using file descriptors but the tempfile library automatically opens the file

I'm just downloading the file you can check at the end of the utils.py file to see how I am doing it :)

nateprewitt:
I've pushed b1be8f7 which will properly capture and then close the file descriptor for Windows before deleting the temp file.

If you rebase onto master @Dreamersoul, I think we're about ready.

Dreamersoul:
Awesome! I checked your commit and turns out I did the same locally! Once I have access to my laptop I'll rebase and push :+1:

Dreamersoul:
Can someone trigger a rebuild? The tests failed because of pip and not my test :sweat_smile:

nateprewitt:
Looks like we're set. Thanks again @DreamerSoul!

tests/test_pipenv.py
```diff
@@ -804,3 +804,23 @@ def test_local_zipfiles(self):
...
+    @pytest.mark.install
+    @pytest.mark.files
+    @pytest.mark.urls
+    def test_install_remote_requirments(self):
+        with PipenvInstance() as p:
+            # using a github hosted requirements.txt file
+            c = p.pipenv('install -r https://raw.githubusercontent.com/kennethreitz/pipenv/3688148ac7cfecefb085c474b092c31d791952c1/tests/test_artifacts/requirements.txt')
+            assert c.return_code == 0
+
+            print(p.lockfile['default'])
```
nateprewitt: Ah, actually can you remove this print?

Dreamersoul: I thought I did *face palm*

nateprewitt:
✨

PR #957, opened 2017-10-23
Run tests against Python 3.7

jacebrowning:
This might have helped surface https://github.com/kennethreitz/pipenv/issues/956 sooner.


jacebrowning:
Oddly, this passes on 3.7.

nateprewitt:
Thanks for checking in on this @jacebrowning. I don't think we're going to enable this until 3.7 is at least out of alpha which isn't for a few months. We're also 8 months out from release so there's no rush here.

Issue #870, opened 2017-10-10
extras_require and length of name > 5 fails.


dfee:
I have a setup.py that looks like this:

```python3
from setuptools import setup, find_packages

setup(
    name='graphene_utils',
    version='1.0',
    packages=find_packages(),
    include_package_data=True,
    install_requires=[
        'graphene>=2.0.dev',
    ],
    extras_require={
        'sqlalchemy': [
            'sqlalchemy',
        ],
        'test': [
            'pylint',
        ],
    },
    entry_points='''
    [console_scripts]
    run=server:run
    ''',
)
```

When attempting `pipenv install "-e .[sqlalchemy]"` it fails with:
```bash
~/code/graphene_utils (virtualenvs/graphene_utils-Dw3J0BOS)
$ pipenv install "-e .[sqlalchemy]"
Courtesy Notice: Pipenv found itself running within a virtual environment, so it will automatically use that environment, instead of creating its own for any project.
Discovering imports from local codebase…
Installing -e .[sqlal…
⠸
Error:  An error occurred while installing -e .[sqlal!
.[sqlal should either be a path to a local project or a VCS url beginning with svn+, git+, hg+, or bzr+
```

However, renaming the extra_requires to `sqla` works fine.

nateprewitt:
Yeah, this is definitely a bug. I haven't had a chance to look into it yet, but I'll try to set aside some time this weekend. If anyone else would like to pick this up before then, feel free.

spott:
Are there any workarounds for this?

nateprewitt:
We still haven't root caused the issue, so the only suggestion at this time is you'll have to use pip or a non-wheel (I wouldn't recommend with something like tensorflow).

spott:
What are the consequences of using pip inside `pipenv shell`?

Does this just screw with the lock file?

nateprewitt:
@spott yes, it will install but nothing will be added to the lockfile or Pipfile. This means you'd have to run this step on every install.

spott:
@nateprewitt: I'll have to run which step?  pipenv lock?

nateprewitt:
Tensorflow won't be added to the Pipfile when you install it via pip. This means it will also not be included in the lockfile when locking.

You'll need to run pipenv shell and install Tensorflow via pip anytime you run pipenv install.

spott:
@nateprewitt:  Ah.  Thanks.  hopefully this bug will be installed when I next need to do that.

[...]

Issue #949, opened 2017-10-22
Package name can collide with top-level directory


shvenkat:
## Description

A package specified in `Pipfile` is not installed if there is a top-level directory with the same name. If the directory is removed/renamed, the next `pipenv install` installs the package.

## Environment

1. OS Type: Mac OS 10.11.6
1. Python version: Python 3.6.0
1. Pipenv version: pipenv, version 8.2.7

## Steps to reproduce

```
mkdir /tmp/pipenv_test && cd /tmp/pipenv_test
echo -e "[requires]\npython_version = \"3.6\"\n\n[packages]\nrequests = \"*\"" > Pipfile
mkdir requests
PIPENV_VENV_IN_PROJECT=1 pipenv install
```
 
## Expected result

Package `requests` is installed into the virtual environment in `.venv`.

## Actual result

Package `requests` is *not* installed into the virtual environment in `.venv`. Deleting or renaming the directory `requests`, and running `PIPENV_VENV_IN_PROJECT=1 pipenv install` does result in the package `requests` being installed.

nateprewitt:
Thanks for opening this @shvenkat! This is yet another data point on multiple issues we have open around local file installs.

@techalchemy, this suggests to me that we probably should require a `./` minimum when people are installing local packages (e.g. `./requests`). Otherwise, it's going to be difficult to disambiguate what they're referencing with the entry in the Pipfile.

techalchemy:
@nateprewitt this would align with pip's behavior around directories (files can be implicitly resolved I think).  This is kind of an odd case and we had talked about attempting (and if we fail, doing the normal method) package installation if we find a directory match.  Maybe that is overcomplicating though, especially given that it does not replicate the current behavior of pip.

```console
 ◰³ tempenv-4d12275072fe  /t/new  unzip dls/tablib.zip
Archive:  dls/tablib.zip
4c300e65a50eef72b91fa1909c9f68679723955e
 ◰³ tempenv-4d12275072fe  /t/new  ls
dls  Pipfile  tablib
 ◰³ tempenv-4d12275072fe  /t/new  pip install tablib 
Collecting tablib
  Using cached tablib-0.12.1.tar.gz
Collecting odfpy (from tablib)
Collecting openpyxl (from tablib)
...
```

[...]

Issue #914, opened 2017-10-16
Cannot install from a local wheel


pfmoore:
I have a project where I want to install a package (`cassiopeia`) that depends on `pycurl`. I'm on Windows, and `pycurl` currently does not provide a binary wheel for Python 3.6, nor does it build without a prebuilt libcurl being available. I can however get a wheel from Christoph Gohlke's website.

I'd like to hold the pycurl wheel in my project directory, under `wheels/pycurl*.whl`. I could then supply `--find-links` to pip and install pycurl from there.

Specifying an environment variable `PIP_FIND_LINKS` sort of works, but I can't find any way to record the fact that I want to include the `wheels` directory in the search path permanently, so that I don't have to (remember to) set an environment variable every time I rebuild the environmnent.

I found https://github.com/pypa/pipfile which gives examples of more complex ways of specifying things in a pipfile, but I can't make this work.

##### Describe you environment

1. Windows
1. Python version: `python 3.6.2`
1. Pipenv version: `pipenv, version 8.2.6`

##### Expected result

To be able to configure "find-links = ./wheels" somewhere, probably in the Pipenv file, and have it pick up wheels from that location. (I'd expect to be able to use a relative path, to mean "relative to the project directory", or an absolute path for wheels I store globally).

##### Actual result

I need to manually set `PIP_FIND_LINKS` before installing via pipenv, and then unset it afterwards, so that it doesn't affect other invocations of pip.

##### Steps to replicate

`pipenv install`. Pipfile is

```
[[source]]
url = "https://pypi.python.org/simple"
verify_ssl = true
name = "pypi"

[packages]
cassiopeia = "*"
pycurl = {file = "./wheels/pycurl-7.43.0-cp36-cp36m-win_amd64.whl"}

[dev-packages]
```

Output is:

```
>pipenv install
Creating a virtualenv for this project…
Using real prefix 'C:\\Users\\pfmoore\\AppData\\Local\\Programs\\Python\\Python36'
New python executable in C:\Users\pfmoore\.virtualenvs\riot_data-4b7hkvZN\Scripts\python.exe
Installing setuptools, pip, wheel...done.

Virtualenv location: C:\Users\pfmoore\.virtualenvs\riot_data-4b7hkvZN
Pipfile.lock not found, creating…
Locking [dev-packages] dependencies…
Locking [packages] dependencies…
Updated Pipfile.lock (17f190)!
Installing dependencies from Pipfile.lock (17f190)…
[...]

PR #958, opened 2017-10-23
Relative file and path installation

techalchemy:
* Issues - #949, #939, #936 and to a lesser extent #817 #540 #914 #960 #870 and more
* Fixed: Local file path installation (resolves in pipfile as relative
path)
* Pass file:// URI to Requirements library for resolving

TODO:
* Ignore non-explicit directory paths lacking os.sep or ./
* Fix extras_require (per #870)
* Add tests

### DIFF
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -885,6 +885,23 @@ def find_windows_executable(bin_path, exe_name):
     return find_executable(exe_name)
 
 
+def convert_path_to_uri(path):
+    """Given a path, return a file:// URI pointer"""
+    # Adapted and borrowed from pip
+    drive, path = os.path.splitdrive(os.path.abspath(path))
+    filepath = path.split(os.path.sep)
+    url = '/'.join(filepath)
+    if drive:
+        return 'file:///' + drive + url
+    return 'file://' + url
+
+
+def convert_file_uri_to_path(uri):
+    """Given a URI, return an OS path"""
+    path = urlparse(uri)
+    return os.path.abspath(os.path.join(path.netloc, path.path))
+
+
 def walk_up(bottom):
     """Mimic os.walk, but walk 'up' instead of down the directory tree.
     From: https://gist.github.com/zdavkeos/1098474
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -23,6 +23,10 @@ try:
     from urllib.parse import urlparse
 except ImportError:
     from urlparse import urlparse
+try:
+    from pathlib import Path
+except ImportError:
+    from pathlib2 import Path
 
 from distutils.spawn import find_executable
 from contextlib import contextmanager
@@ -41,7 +45,7 @@ specifiers = [k for k in lookup.keys()]
 
 # List of version control systems we support.
 VCS_LIST = ('git', 'svn', 'hg', 'bzr')
-SCHEME_LIST = ('http://', 'https://', 'ftp://', 'file:///')
+SCHEME_LIST = ('http://', 'https://', 'ftp://', 'file://')
 
 requests = requests.Session()
 
@@ -264,6 +268,35 @@ packages = [
     'structlog', 'mandrill', 'googlemaps', 'easy-thumbnails', 'automaton',
     'webcolors'
 ]
+
+
+def get_requirement(dep):
+    """Pre-clean requirement strings passed to the requirements parser.
+    
+    Ensures that we can accept both local and relative paths, file and VCS URIs,
+    remote URIs, and package names, and that we pass only valid requirement strings
+    to the requirements parser. Performs necessary modifications to requirements
+    object if the user input was a local relative path.
+    """
+    path = None
+    # Only operate on local, existing, non-URI formatted paths
+    if (is_file(dep) and isinstance(dep, six.string_types) and
+            not any(dep.startswith(uri_prefix) for uri_prefix in SCHEME_LIST)):
+        dep_path = Path(dep)
+        # Only parse if it is a file or an installable dir
+        if dep_path.is_file() or (dep_path.is_dir() and pip.utils.is_installable_dir(dep)):
+            if dep_path.is_absolute():
+                path = dep
+            else:
+                path = get_converted_relative_path(dep)
+            dep = dep_path.resolve().as_uri()
+    req = [r for r in requirements.parse(dep)][0]
+    # If the result is a local file with a URI and we have a local path, unset the URI
+    # and set the path instead
+    if req.local_file and req.uri and not req.path and path:
+        req.path = path
+        req.uri = None
+    return req
 
 
 def cleanup_toml(tml):
@@ -539,12 +572,11 @@ def convert_deps_from_pip(dep):
 
     dependency = {}
 
-    req = [r for r in requirements.parse(dep)][0]
+    req = get_requirement(dep)
     extras = {'extras': req.extras}
 
     # File installs.
-    if (req.uri or (os.path.isfile(req.path) if req.path else False) or
-            os.path.isfile(req.name)) and not req.vcs:
+    if (req.uri or req.path or (os.path.isfile(req.name) if req.name else False)) and not req.vcs:
         # Assign a package name to the file, last 7 of it's sha256 hex digest.
         if not req.uri and not req.path:
             req.path = os.path.abspath(req.name)
@@ -778,6 +810,17 @@ def is_vcs(pipfile_entry):
     return False
 
 
+def is_installable_file(path):
+    """Determine if a path can potentially be installed"""
+    if hasattr(path, 'keys') and any(key for key in path.keys() if key in ['file', 'path']):
+        path = urlparse(path['file']).path if 'file' in path else path['path']
+    if not isinstance(path, six.string_types) or path == '*':
+        return False
+    lookup_path = Path(path)
+    return lookup_path.is_file() or (lookup_path.is_dir() and
+            pip.utils.is_installable_dir(lookup_path.resolve().as_posix()))
+
+
 def is_file(package):
     """Determine if a package name is for a File dependency."""
     if hasattr(package, 'keys'):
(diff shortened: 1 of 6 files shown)


erinxocon:
We're gonna want @kennethreitz to look at this before merging, he wanted to take a look at the dependency resolution stuff before it was commited.

techalchemy:
Yah it’s not anywhere near ready yet. I have about 5 more trains of thought I want to explore on this before we consider merging. Implementation details on resolving requirements working around the file URI, etc. 

This used to work without workarounds like this so I fear we merged a regression at some point. It might be simplest to start by figuring out where.

erinxocon:
@techalchemy let me know how I can help out here.  You know more about this than I about this.

techalchemy:
@erinxocon I think it might actually be a lot less complicated than I was thinking.  I need to rethink this

techalchemy:
Here was how we used to handle this: https://github.com/techalchemy/pipenv/blob/c7056ea4067c213e0702167f5207cc0eb45dea42/pipenv/cli.py#L836 -- this got reverted somewhere

techalchemy:
ffffffffffffff i broke my own tests with this one, but it is much closer, more to come

nateprewitt:
@techalchemy could we get a quick breakdown of what pathlib is buying us here? That's a pretty significant import.

techalchemy:
@nateprewitt I imported it for cross-platform backward- and forward- path-to-uri conversions, consistent path resolution and normalization, etc.  I was only importing `Path` specifically, but I vendored `pathlib2` for backward compatibility and it may be lighter weight to import `Path` from there.

The longer-term plan was to refactor a lot of our path-handling functions and replace them with `pathlib.Path` since it does a better, more consistent job normalizing, resolving, and formatting across platforms

techalchemy:
Appveyor test are off by a backslash, but as-is this code provides the same behavior as we have for file URIs

techalchemy:
@nateprewitt @erinxocon @kennethreitz All tests are passing on this, and it fixes (currently broken) functionality for installing local files via relative paths.  This was a regression and has caused a lot of issues (I've seen a number of threads on reddit and there have been mentions of reservations about pipenv related to inability to install local wheels, for example).  This does not fix #870 and does not provide the thorough refactor of `is_file` which is needed, I will do that in a separate branch.

nateprewitt:
Thanks @techalchemy, this on my list of reviews for tonight.

tests/test_utils.py
```diff
@@ -180,6 +180,14 @@ def test_is_valid_url(self):
         assert pipenv.utils.is_valid_url(url)
         assert pipenv.utils.is_valid_url(not_url) is False
 
+    @pytest.mark.parametrize('input_path, expected', [
+        ('artifacts/file.zip', os.path.join('.', 'artifacts', 'file.zip')),
+        ('./artifacts/file.zip', os.path.join('.', 'artifacts', 'file.zip')),
+        ('../otherproject/file.zip', os.path.join('.', '..', 'otherproject', 'file.zip'))
+    ])
+    def test_get_converted_relative_path(self, input_path, expected):
```
nateprewitt: So I'm a little unsure about this test. It's generally good to have concrete expected values, otherwise we're only ensuring that `os.path.join` is just a broken in `get_converted_relative_path` as it is on it's own.

pipenv/utils.py
```diff
@@ -873,6 +892,17 @@ def find_windows_executable(bin_path, exe_name):
     return exec_files[0]
 
 
+def convert_file_uri_to_path(uri):
```
nateprewitt: What are we envisioning this function being used for in the future? I'm inclined not to add it until it has a use.

pipenv/utils.py
```diff
@@ -264,6 +267,22 @@
...
 
 
+def get_requirement(dep):
+    """Resolve a requirement no matter where it points"""
+    path = None
+    if (is_file(dep) and isinstance(dep, six.string_types) and
+            not any(dep.startswith(uri_prefix) for uri_prefix in FILE_LIST)):
+        dep_path = Path(dep)
+        if dep_path.exists() and not dep_path.is_absolute():
+            path = get_converted_relative_path('{}'.format(dep))
```
nateprewitt: Why can't we pass `dep` here? Haven't we already asserted it's `str`-like?

pipenv/utils.py
```diff
@@ -532,12 +551,12 @@ def convert_deps_from_pip(dep):
 
     dependency = {}
 
-    req = [r for r in requirements.parse(dep)][0]
+    req = get_requirement(dep)
```
nateprewitt: We just modified the path here in the `get_requirement` function but we modify it again below. Is the path adding needed in the function? It looks like we don't use it in either of the two calls this is used.

techalchemy:
I do have a refactor of is_file that is now working, i will PR it separately when I have some tests.  Slightly messier than I thought but not too bad

pipenv/utils.py
nateprewitt: On second pass, doesn't this exacerbate the problem with local packages vs pypi packages?

If I have a directory named `ansible` in my package, and specify that I was to install `ansible`, I would say the correct behaviour is to install from pypi. If I want to install from my local directory, it's up to the me to dictate that (`./ansible`).

With this approach, it looks like a pypi install will never be possible.

techalchemy: I think we can pass `dep`, I wrote this section in kind of a hurry which will explain the next piece of information.

I was looking at this briefly yesterday before I stopped working on this, but I think you're right about the behavior of `./ansible`.  That is to say, with this current code, `pipenv install ansible` will   accurately install ansible from pypi actually, but it won't write it to the lockfile.  I actually think I have a different approach to handle this in general though and I'm curious about whether it will work.  I will update the PR shortly.

tests/test_utils.py
techalchemy: Yeah I agree,  I was fishing for a solution that would accommodate windows which had a special caveat for python 2.7.  I'm just going to write a separate one for windows.

tests/test_utils.py
```diff
@@ -180,6 +180,14 @@ def test_is_valid_url(self):
         assert pipenv.utils.is_valid_url(url)
         assert pipenv.utils.is_valid_url(not_url) is False
 
+    @pytest.mark.parametrize('input_path, path', [
+        ('artifacts/file.zip', './artifacts/file.zip'),
+        ('./artifacts/file.zip', './artifacts/file.zip'),
+        ('../otherproject/file.zip', './../otherproject/file.zip')
```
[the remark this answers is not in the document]
techalchemy: the goal here was to make everything explicitly relative to the project path.  I'm not 100% sure that is best, but it was the path of least resistance when I was writing the code so I figured it was a good starting point

pipenv/utils.py
techalchemy: uhhhh this is an artifact that is no longer needed.  Will dig up the axe.

pipenv/utils.py
techalchemy: This is the only one I need to push back on -- yes it is needed, and yes it is used.  `get_requirement` takes a string and uses `requirements.parse()` to parse it.  The root of the issue we had with handling relative paths is that `requirements.parse()` can only take either a requirement name (without any path separators, it must be a valid package name according to the pep standards) or a URI.

In the case of a relative path, we need to instantiate a `Requirement` object by converting any existing relative paths to their corresponding `file://` URI.  However, we want to store the relative path in the `Pipfile`, which only happens if `Requirement.path` is set and `Requirement.uri` is not set.  According to the spec, the `file` property is for URIs, but we are writing `path` for the relative path.  So we store the relative path to put in the `req.path` property, but we resolve the `file://` uri for parsing Requirements:

```python
        dep_path = Path(dep)
        if dep_path.exists() and not dep_path.is_absolute():
            path = get_converted_relative_path(dep)
            dep = dep_path.resolve().as_uri()
        ...
    if req.local_file and req.uri and not req.path and path:
        req.path = path
        req.uri = None
    return req
```

The modifications we are doing in `convert_deps_from_pip` don't have any impact to `req.path` in the cases where this code does.  This code can surely be refactored, but the code I have runs only when the user supplies a local relative path (on reflection it should work on absolute paths also) whereas the other code modifies `path` only when there is a file supplied as the `name`, but there is no `path` or `uri`.

I basically left this code alone because there is only one conditional that possibly duplicates anything I've done and I didn't want to refactor the entire function which does quite a bit of heavy lifting, at least not in this PR

techalchemy: Hm I reevaluated some of this and in order to fix the `ansible` case above some of this needed to be refactored.  Update incoming.

techalchemy:
So there is some additional work to do on the file checks, but this fixes the `ansible` example (I didn't write a test for that yet) by using `pip.utils.is_installable_dir` for checking directories.

techalchemy:
@nateprewitt This is currently functioning but I'm ok with including a refactor of some of the file related handling as well.  I'm happy to have a more real-time discussion about the approach if you like

pipenv/utils.py
```diff
@@ -266,6 +269,25 @@
 ]
 
 
+def get_requirement(dep):
+    """Resolve a requirement no matter where it points"""
```
kennethreitz: This is a terrible docstring :)

kennethreitz:
I approve.

nateprewitt:
@techalcyemy Alright, if we get the docstrig cleaned up, we can merge.

nateprewitt:
@techalchemy can we squash this down to a commit or two?

techalchemy:
@nateprewitt yes but completionism compels me to wait for CI

techalchemy:
That squash was not fun but it is now done so feel free to review

nateprewitt:
The failure on Windows 2.7 looks like it may be legitimate.

rzeszotko-so1:
I happened to have this issue and I just tested the fix - `pipenv install foo.tar.gz` will use the new relative path mechanism, but if you run `pipenv install` while having `foo.tar.gz` listed in requirements.txt, the absolute path to `foo.tar.gz` will still get hardcoded in Pipfile.lock.

Issue #942, opened 2017-10-19
Allow creating requirements.txt files without dev-packages


tuukkamustonen:
To install a set of pinned packages in remote systems, there is:

    pipenv install --system  # won't install dev dependencies

However, this requires both `Pipfile` and `Pipfile.lock` to be present, and also `pipenv` to be installed in the target system. It's way more inconvenient than just being able to (maybe first copy package wheel into remote system and) say `pip install <mypackage>` (and letting pip to pull in the pinned dependencies specified in `setup.py`'s `install_requires`).

(Or is there some benefit in using `pipenv install --system` instead?)

To be able to generate pinned `requiremets.txt` that excludes development packages, add `--no-dev` option to `pipenv lock`, that seems to by default include also dev packages. Or maybe even change to default to match that of `pipenv install`, and exclude dev-packages by default?

As per https://github.com/kennethreitz/pipenv/issues/209#issuecomment-278185133 this is already supported in code:

    from pipenv.project import Project
    from pipenv.utils import convert_deps_to_pip

    pfile = Project(chdir=False).parsed_pipfile
    requirements = convert_deps_to_pip(pfile['packages'], r=False)

Although there is workaround for this need already, supporting it right in the cli would/could be nice.

erinxocon:
I think we could add a method that doesn't add dev packages tot he requirements file.  It's tough because normally you'd just have everything in your requirements.txt or maybe had two for production and dev dependencies.  The reason `pipenv install --system` instead of just pip is that with the PIpfile.lock, you get a deterministic build every time you install.

tuukkamustonen:
> The reason pipenv install --system instead of just pip is that with the PIpfile.lock, you get a deterministic build every time you install.

Yeah, but you get the same result if you depend on `requirements.txt` that was created by `pip-compile` from `requirements.in` (from pip-tools package). Right? I don't see any difference.

> maybe had two for production and dev dependencies.

Yeah, it's pretty common practice to have separate `requirements.txt` and `dev-requirements.txt` (or rather `requirements.in` and `dev-requirements.in` if using pip-tools).

techalchemy:
@tuukkamustonen I think if you prefer using piptools with pip-compile directly you are certainly able to do that, pipenv uses pip-tools to build lockfiles so technically you can do that yourself, if you like. If you don't see the difference between pipenv and using pip-compile from pip-tools then you likely wont get anything out of using pipenv.

piotr-dobrogost:
[...]

PR #972, opened 2017-10-25
Lock requirements enhancement

erinxocon:
Addresses #942

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -2004,9 +2004,10 @@ def uninstall(
 @click.option('--python', default=False, nargs=1, help="Specify which version of Python virtualenv should use.")
 @click.option('--verbose', is_flag=True, default=False, help="Verbose mode.")
 @click.option('--requirements', '-r', is_flag=True, default=False, help="Generate output compatible with requirements.txt.")
+@click.option('--dev', '-d', is_flag=True, default=False, help="Generate output compatible with requirements.txt for the development dependencies.")
 @click.option('--clear', is_flag=True, default=False, help="Clear the dependency cache.")
 @click.option('--pre', is_flag=True, default=False, help=u"Allow pre–releases.")
-def lock(three=None, python=False, verbose=False, requirements=False, clear=False, pre=False):
+def lock(three=None, python=False, verbose=False, requirements=False, dev=False, clear=False, pre=False):
 
     # Ensure that virtualenv is available.
     ensure_project(three=three, python=python)
@@ -2016,7 +2017,7 @@ def lock(three=None, python=False, verbose=False, requirements=False, clear=Fals
         pre = project.settings.get('pre')
 
     if requirements:
-        do_init(dev=True, requirements=requirements)
+        do_init(dev=dev, requirements=requirements)
 
     do_lock(verbose=verbose, clear=clear, pre=pre)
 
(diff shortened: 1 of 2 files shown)


pipenv/cli.py
```diff
@@ -2004,9 +2026,10 @@ def uninstall(
 @click.option('--python', default=False, nargs=1, help="Specify which version of Python virtualenv should use.")
 @click.option('--verbose', is_flag=True, default=False, help="Verbose mode.")
 @click.option('--requirements', '-r', is_flag=True, default=False, help="Generate output compatible with requirements.txt.")
+@click.option('--dev', '-d', is_flag=True, default=False, help="Generate output compatible with requirements.txt for the development dependencies.")
```
tuukkamustonen: @erinxocon I think with `--dev` it includes _also_ normal packages? Should change this text to "Include dev packages into requirements.txt output"? Now it makes one think that only dev packages will be outputted.

tuukkamustonen: Hmm, I think my assumption is wrong, as you wrote:

> pipenv lock -r -d will give you just the development deps and development vcs deps

Ok, then the help text is according to this specification.

Apparently it's impossible to create such requirements.txt files now, that would include _both_ standard and dev packages. I don't have use case for that, so I think it's ok. Someone else might want that, though.

erinxocon: The current functionality separates them.  `--dev` does only development dependencies, and `-r` only does anything listed under the default section!  Is this not what was needed?

erinxocon: If the all packages is needed I can add it at a later time!

tuukkamustonen: > The current functionality separates them. --dev does only development dependencies, and -r only does anything listed under the default section! Is this not what was needed?

Indeed, this change should fulfill my needs. Thanks :)

tuukkamustonen: @erinxocon Note that there is some inconsistence:

    pipenv install  # by default only standard packages, with --dev ALSO dev packages
    pipenv update  # by default only standard packages, with --dev ALSO dev packages
    pipenv lock -r  # by default only standard packages, with --dev ONLY dev packages

erinxocon: In this case is that the wanted use case though? I'll let my fellow contributors put in their 2 cents!

tuukkamustonen: So to clarify, I needed `requirements.txt` file that excludes dev packages. After the change you've made here, I can do that with `pipenv lock -r` as you've instructed. So yeah, my use is now supported.

I have no opinion if `pipenv lock -r -d` should include _only_ dev packages or _also_ dev packages. I personally have no use case for neither. I was just pointing out that there's now slight inconsistency in API, compared to `install` and `update`. So you might want to consider changing `--dev` behavior here accordingly.

erinxocon: I appreciate the input @tuukkamustonen!   A very keen eye on the api discrepancies!

piotr-dobrogost: It seems `--dev` in `pipenv lock -r --dev` should be consistent with `pipenv install` and `pipenv update` indeed. In the future, if there's a need, additonal flag like `--only-dev` could be added.

ajenkins-cargometrics: I just came here to post a bug stating that `lock -r -d` only outputs dev dependencies.  In isolation, I don't think there's anything inherently wrong with the current behavior, but the fact that `lock -d` is inconsistent with `install -d` and `update -d` makes it seem like a bug.  Not only that, but if I run `pipenv lock -h`, the help text for `--dev` says "Install both develop and default packages.", further adding to the confusion.

PR #976, opened 2017-10-25
Fix #973 Fails if two conflicting "requires" sections are present.

### LINKED ISSUE #973: Fails if two conflicting "requires" sections are present.
I stumbled upon this since, at some point, an extra `requires` section was added to my `Pipifile`. I'm not able to reproduce the circumstances that lead to such a malformed configuration file, but still the reaction of `pipenv` is quite baffling. 

1. OS Type: OS X 10.12.6 (16G29)
1. Python2 version: `$ python -V` 2.7.14
1. Python3 version: `$ python3 -V` 3.6.3
1. Pipenv version: `$ pipenv --version` 8.2.7

##### Expected result

Running `pipenv check` on

```
[[source]]
verify_ssl = true
name = "pypi"

[requires]
python_version = "3.6"

[requires]
python_version = "2.7"

[packages]

[dev-packages]

```

should return a more informative error about the presence of two conflicting `requires` sections than a stacktrace.

##### Actual result

The command crashes emitting a stacktrace (related to the `requires` section already defined, but not so readable).

```
$ pipenv check
  File "/usr/local/bin/pipenv", line 11, in <module>
    sys.exit(cli())
  File "/usr/local/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 722, in __call__
  File "/usr/local/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 1066, in invoke
  File "/usr/local/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 895, in invoke
  File "/usr/local/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 535, in invoke
  File "/usr/local/lib/python3.6/site-packages/pipenv/cli.py", line 2208, in check
    p = pipfile.Pipfile.load(project.pipfile_location)
  File "/usr/local/lib/python3.6/site-packages/pipenv/vendor/pipfile/api.py", line 126, in load
    pipfile.data = p.parse()
  File "/usr/local/lib/python3.6/site-packages/pipenv/vendor/pipfile/api.py", line 82, in parse
    config.update(toml.loads(content))
  File "/usr/local/lib/python3.6/site-packages/pipenv/vendor/toml.py", line 344, in loads
    str(currentlevel))
toml.TomlDecodeError: What? requires already exists?{'source': [{'url': 'https://pypi.python.org/simple', 'verify_ssl': True, 'name': 'pypi'}], 'requires': {'python_version': '3.6'}}

```

##### Steps to replicate

Just run `pipenv check` in an empty directory containing just the above-mentioned `Pipfile`.

mirosval:
Add a more explicit error message if a section of the Pipfile is duplicated

### DIFF
--- a/pipenv/project.py
+++ b/pipenv/project.py
@@ -227,6 +227,14 @@ class Project(object):
         with open(self.pipfile_location) as f:
             contents = f.read()
 
+        sections = re.findall("\[[a-zA-Z-]+\]", contents)
+        for section in set(sections):
+            section_count = sections.count(section)
+            if section_count > 1:
+                raise RuntimeError(
+                    "Your Pipfile contains section '{}' multiple times ({}x)".format(section, section_count)
+                )
+
         # If any outline tables are present...
         if ('[packages.' in contents) or ('[dev-packages.' in contents):
 
--- a/tests/test_project.py
+++ b/tests/test_project.py
@@ -1,4 +1,5 @@
 import pipenv.project
+import pytest
 from pipenv.vendor import delegator
 
 
@@ -153,3 +154,22 @@ class TestProject():
         assert lockfile['_meta']['hash'] == {'sha256': 'ff0b0584610a7091156f32ca7d5adab8f29cb17263c6d63bcab42de2137c4787'}
 
         delegator.run('rm -fr test_internal_lockfile')
+
+    def test_duplicate_sections(self):
+        proj = pipenv.project.Project()
+
+        # Create test space.
+        delegator.run('mkdir test_duplicate_sections')
+
+        with open('test_duplicate_sections/Pipfile', 'w') as f:
+            f.write('[requires]\n'
+                    'python_version = "3.6"\n\n'
+                    '[requires]\n'
+                    'python_version = "2.7"\n')
+
+        proj._pipfile_location = 'test_duplicate_sections/Pipfile'
+
+        with pytest.raises(RuntimeError):
+            _ = proj.parsed_pipfile
+
+        delegator.run('rm -fr test_duplicate_sections')


erinxocon:
Thanks @mirosval for this!  Always good to present nice errors to the user.  And with a test to go along! ✨🍰✨

nateprewitt:
Ermm, I don't think we wanted to merge this in this state @erinxocon. This breaks when you're using extras and I'm no convinced we want to universally enforce no duplicate sections.

I'm going to revert this and @mirosval if you wouldn't mind opening another PR that focuses on packages, dev-packages and requires that'd be great.

erinxocon:
@nateprewitt Sorry about that, got a little overzealous!  I overlooked that the regex excludes vcs and extras.  What sections do you think could be duplicated?  I can't think of a use case here.

nateprewitt:
@erinxocon I don't have a concrete example but regexes while powerful get hard to manage when you need exclusions. I don't believe there's anything in the toml spec restricting duplicate keys and it seems better to be explicit about what we want rather than having to switch to that later.

erinxocon:
@nateprewitt Should we just pass on this error and use the first or last parsed duplicate section?

nateprewitt:
Got a chance to confirm this flag breaks on multiple sources and I think we're fine with the basic idea of this patch, I just want to tighten the scope. We will never have enough sections to really need a regex and instead of taking on the cognitive burden of a regex with specific exclusions, let's just declare a constant tuple with the three sections.

erinxocon:
@nateprewitt Dig it!

mirosval:
@nateprewitt No worries! I'm happy to improve this. The reason I resorted to regex is that the original issue resulted in a pretty scary toml parser error. So I think that it would break for any duplicate section with the same error.

mirosval:
So according to the TOML spec (https://github.com/toml-lang/toml#table) a `Table` (earlier referred to here as Section) can only appear once:

```toml
[a]
b=1

[a]
c=1
```

is invalid. However an `Array of Tables` is valid:
```toml
[[products]]
name = "Hammer"
sku = 738594937

[[products]]

[[products]]
name = "Nail"
sku = 284758393
color = "gray"
```

The Regex as it currently stands will not allow this. And now I'm thinking that perhaps the correct place to fix this would be the Toml package itself, not in Pipenv. If we want to fix it in Pipenv then perhaps a regex like this:

`^\s*\[[^\]\[]+\]\s*$` (https://regex101.com/r/y1A23D/1) but I agree this is very ugly.

techalchemy:
This already works in pipenv

nateprewitt:
@mirosval, I don't know if Uiri who maintains the Python toml library will see any need to make a change since the traceback is already clear. While the phrasing is a little unconventional it does say that "requires" was a duplicate entry. In that way @techalchemy is right that this already "works".

I don't have an issue catching this and presenting a nicer looking error via click. That said this is purely cosmetic at that point and in a pretty weird case at that. The user should never have this happen unless they specifically copy paste bad excess content into their Pipfile.

PR #979, opened 2017-10-26
Switched to `short_help` kwarg from `help` for command help text.

timworx:
The truncated help text for the "Commands:" section (on `pipenv`) threw me off some while first looking at and learning Pipenv.

I made a quick PR to see how it would be with displaying the full help text. 

Eliminating the auto-truncation by the `click` library was achieved by switching from the `help` kwarg to `short_help` in the `click.command` decorator.

screen shot 2017-10-26 at 5 16 11 pm

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -1729,7 +1729,7 @@ def do_py(system=False):
         click.echo(crayons.red('No project found!'))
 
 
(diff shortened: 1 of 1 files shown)


nateprewitt:
This seems sensible to me. Thanks @timworx, great first contribution!

PR #984, opened 2017-10-27
Honor PIPENV_VENV_IN_PROJECT setting after virtualenv creation

jonathn:
After creating the virtualenv by passing PIPENV_VENV_IN_PROJECT=1 to pipenv,
if _any_ future pipenv commands do not also pass PIPENV_VENV_IN_PROJECT=1
a second virtualenv gets created in ~/.local/share/virtualenvs/ instead of just using the existing one.
This goes against the principle of least surprise.
This commit will reuse <project_dir>/.venv if it exists.

### DIFF
--- a/pipenv/project.py
+++ b/pipenv/project.py
@@ -147,13 +147,13 @@ class Project(object):
         if self._virtualenv_location:
             return self._virtualenv_location
 
-        # The user wants the virtualenv in the project.
-        if not PIPENV_VENV_IN_PROJECT:
+        # Locate the virtualenv in the project directory.
+        loc = os.sep.join(self.pipfile_location.split(os.sep)[:-1] + ['.venv'])
+
+        # If this is neither found nor requested, use pew.
+        if not PIPENV_VENV_IN_PROJECT and not os.path.isdir(loc):
             c = delegator.run('pew dir "{0}"'.format(self.virtualenv_name))
             loc = c.out.strip()
-        # Default mode.
-        else:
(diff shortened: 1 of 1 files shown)


nateprewitt:
Hey @jonathn, thanks for the initiative with this. So PIPENV_VENV_IN_PROJECT is intended to be a binary opt in. Either you want this as the default or you want pew to handle your venvs.

This is currently respected without an issue but you need to actually set the environment variable rather than just passing it to one command.

If you need this in a per project basis, please use a .env file.

jonathn:
It sounds like you are implying that if I create a <project_dir>/.env file containing:
```
PIPENV_VENV_IN_PROJECT=1
```
then pipenv will pick up and use that setting all the time?
That would be great, but afaict the code only loads the .env file for `shell` and `run` commands and only after the virtualenv has already been initialized which is too late.

Is there no way to set these variables in the Pipfile, and if not can we add that functionality?
(the only variable you couldn't set that way would be PIPENV_PIPFILE due to the chicken-and-egg problem)

jonathn:
^ above question was for @nateprewitt

nateprewitt:
Sorry @jonathn, this got lost in the queue. It looks like `.env` isn't being used in this case which is probably not correct. We'll need to explore the implications of running the `.env` check before our `ensure_virtualenv` function.

As for adding this to the Pipfile, I'm afraid that's unlikely. This option is unrelated to dependency management and is instead an implementation variable that will vary between users.

Just to get an idea of your current use case. Why do you need the local `.venv` directory in some projects but not others?

Issue #837, opened 2017-10-05
pipenv uninstall -d somepackage removes all development packages


MrLeeh:
At first I want to say many thanks to @kennethreitz  for this great package. It makes working with virtual environment much easier and offers a straight workflow that fits perfectly in my toolchain.

So here is what I just experienced. I' m not sure if it is a bug or intentional. I just tried to uninstall a package which I before installed as a development package. So what I did was 

    $ pipenv uninstall -d somepackage

However the result was very unexpected for me. pipenv uninstalled all my dev packages. I then had a look at the help and it was completely clear and said that the *-d* or *--dev* flag uninstalls **all development packages**. So it was completely my fault but I could imagine this happening to quite a few folks.

I would suggest that the script should check for inappropriate parameters (like package names) and interrupt execution if they occur to prevent unintended uninstalls.

erinxocon:
Hi @MrLeeh!  Thanks for using pipenv! I think this is by design.  If you just do `pipenv uninstall dev-package` it will look through packages and dev packages to uninstall just that package.  --dev for install makes sense as we need to ensure the package gets put in the correct section, but for uninstall having it remove all also makes sense.  I'll let @kennethreitz weigh in here on this feature request!

brettdh:
@erinxocon i got bit by this today as well. I'd say it *doesn't* make sense for `--dev` to have such different meanings for `install` and `uninstall`, particularly since the meaning with `uninstall` is so surprising and destructive. My feature request would be to rename the option on `uninstall` to more clearly communicate that it removes all dev packages.

techalchemy:
@brettdh would it be sufficient to just have a confirmation? I.e. _Are you sure you want to uninstall all dev packages?_

erinxocon:
So I think I found a use case where we need to do some more checking for this.  If you for some reason have a package in both dev and production, and issue an uninstall, it removed it from both the dev and packages sections.  Specifiying --dev would only remove the package from the dev section, not both.

erinxocon:
So after looking at the code, this wouldn't be possible without a lot of rework.  If you have the same package installed for dev and regular packages, it's the same reference, so removing it from dev automatically removes it from the regular and vice versa.  For now the --dev functionality will stay as it is.  I did however rename it from `--dev` to `--all-dev` to be more consistent with the `--all` removal functionality.

So to summerize

`--all` removes everything and purges the environment
`--all-dev` removes all dev packages
[...]

PR #990, opened 2017-10-29
updated --dev flag to --all-dev

erinxocon:
Changed name of the --dev flag to --all-dev to be more specific about what it does.  --dev now does nothing

### DIFF
--- a/tests/test_pipenv.py
+++ b/tests/test_pipenv.py
@@ -271,6 +271,28 @@ class TestPipenv:
             assert 'tablib' in c.out
             assert 'tablib' not in p.pipfile['packages']
 
+    @pytest.mark.run
+    @pytest.mark.uninstall
+    def test_uninstall_all_dev(self):
+        with PipenvInstance() as p:
+            c = p.pipenv('install --dev requests pytest')
+            assert c.return_code == 0
+            assert 'requests' in p.pipfile['dev-packages']
+            assert 'pytest' in p.pipfile['dev-packages']
+            assert 'reqests' in p.lockfile['develop']
+            assert 'pytest' in p.lockfile['develop']
+
+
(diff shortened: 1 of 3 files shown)


erinxocon:
This address #837

tests/test_pipenv.py
```diff
@@ -271,6 +271,28 @@ def test_uninstall_all_local_files(self):
             assert 'tablib' in c.out
             assert 'tablib' not in p.pipfile['packages']
 
+    @pytest.mark.run
+    @pytest.mark.uninstall
+    def test_uninstall_all_dev(self):
+        with PipenvInstance() as p:
+            c = p.pipenv('install --dev requests pytest')
```
techalchemy: Not sure if you're still working on this and if I'm just prematurely reviewing but you may want to install some non-dev stuff and test that it isn't getting removed/deleted

erinxocon: Good thought

erinxocon:
@techalchemy This can be reverted easily.

My thinking was that since the perceived default behavior of --dev is not what it actually does.  The error message is nice, but then I feel like most people will select No from that prompt and be frustrated as to why it didn't work.  I'm not sure though, UX isn't my fotré.  I struggled too with the behavior of --all and --all-dev being different in terms of scrubbing the Pipfile, but I don't know what to do there...

nateprewitt:
Errm, this is a breaking change and we'd need to do a major release for our next release now.

nateprewitt:
I see we've already released it :/ Well... hmm. I guess we wait and see if there's outcry.

erinxocon:
Sorry about that.   I take responsibility.  I need to incorporate these ideas before pressing merge.

PR #999, opened 2017-10-30
Add instructions to install via homebrew in README

OmeGak:
Related to https://github.com/kennethreitz/pipenv/issues/996.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -44,6 +44,12 @@ Installation
 
     $ pip install pipenv
 
+Or if you are on macOS, you can install Pipenv via `Homebrew <https://brew.sh/>`_ with:
+
+::
+
+    $ brew install pipenv
+
 ✨🍰✨
 
 ☤ User Testimonials


nateprewitt:
@OmeGak thanks for putting this together! We have a few things that need to be reviewed before we merge this though.

@erinxocon @techalchemy, before this is merged we need to verify a.) that this works as expected b.) what the release cycle is (will it be in sync with GitHub? If not what's the delay?) and c.) are all dependencies in the formula required.

I'm going to try to get to these today but I'd prefer their answered before we advertise this as an option.

erinxocon:
@nateprewitt I'll test with brew on my virtual test mac tonight.

erinxocon:
So this does work on my mac, but it's much slower than a native installation by about 100%.

erinxocon:
On a brand new mac, this works with the system python and has all the functionality pipenv should.  I wanted to test on a new mac that has never had its PATH modified or new shell installed, etc.

nateprewitt:
@erinxocon, I guess I should have clarified, I'm not super concerned it will install. The first point should really be a parent of the following two.

Could you check what's installed in `pip freeze` on the new machine? It looks like this installs copies of:

 * backports.shutil_get_terminal_size
 * certifi
 * chardet
 * configparser
 * enum34 
 * flake8 
 * idna
 * mccabe
 * pathlib
 * pycodestyle
 * pyflakes
 * requests
 * shutilwhich
 * urllib3
 * virtualenv
 * virtualenv-clone

I don't know why these are being installed separately (or why testing modules would be required for pipenv to run).

erinxocon:
@nateprewitt It doesn't seem to install them in a place where pip can read them.

nateprewitt:
Thanks @OmeGak! ✨

OmeGak:
No probs. I was really excited to see Pipenv in Homebrew. :D

PR #1000, opened 2017-10-31
Ignore prerelease python installs from pyenv

### LINKED ISSUE #998: Pipenv wrongly chooses pre-release pythons if available
### Setup:
Using pyenv, install a python and a release candidate or beta of the same version. pyenv removes rc/beta pythons from pyenv when the final build is available, so you might have to create the build definition yourself. Put the following in `$PYENV_ROOT/plugins/python-build/share/python-build/2.7.14rc1`:

```bash
install_package "openssl-1.0.2k" "https://www.openssl.org/source/openssl-1.0.2k.tar.gz#6b3977c61f2aedf0f96367dcfb5c6e578cf37e7b8d913b4ecb6643c3cb88d8c0" mac_openssl --if has_broken_mac_openssl
install_package "readline-6.3" "https://ftpmirror.gnu.org/readline/readline-6.3.tar.gz#56ba6071b9462f980c5a72ab0023893b65ba6debb4eeb475d7a563dc65cafd43" standard --if has_broken_mac_readline
install_package "Python-2.7.14rc1" "https://www.python.org/ftp/python/2.7.14/Python-2.7.14rc1.tgz#4d1ca11c5451e1b7cafcabc2003cf80fb3745bac45ba51c39b855b0011a27408" ldflags_dirs standard verify_py27 ensurepip
```

(I had managed to have both the RC and final build installed because I had installed 2.7.14rc1 before the final version was out, then forgot to remove it).

Next, use `pyenv shell 2.7.14` to set the version of python for the current shell

### The Problem

When I run:

```
$ pipenv install --python 2.7.14
Using /Users/josh/.pyenv/versions/2.7.14rc1/bin/python2.7 to create virtualenv…
```

Pipenv is prefering the prerelease python over the one I specifically asked for

### Explanation

1. Pipenv detects pyenv is installed
2. Pipenv scans `$PYENV_ROOT/versions` for installed pythons
3. Pipenv prepends each python to `$PATH` as it finds them in the PATH
4. Pipenv runs `which python2.7` and finds `['/Users/josh/.pyenv/versions/2.7.14rc1/bin/python2.7', '/Users/josh/.pyenv/versions/2.7.14/bin/python2.7', '/usr/bin/python2.7']` (shortened a bit, i have a lot of pythons installed)
5. Pipenv runs `python --version` on each and parses the output **discarding the rc/beta status**
6. Since 2.7.14rc1 came first in the path, Pyenv picks it since without the `rc1` suffix, it is the version I had asked for.

I dug through `cli.py` for a while trying to fix it and found some issues:

* RC/Beta pythons sort last in the glob results when pipenv lists the contents of `$PYENV_ROOT/versions`, but end up first in `$PATH`
* Even if the list of pythons is reversed when added to `$PATH` to restore the sort order, if `2.7.14` is the active python, it is already in the path, so Pipenv skips prepending that one, and only prepends `2.7.14rc1`, causing it to take priority.

I have not yet been able to determine a way to fix this that doesn't break `pipenv --python 2.7` installing the latest 2.7.* version, for example.

techalchemy:
- Fixes #998

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -392,13 +392,20 @@ def ensure_python(three=None, python=None):
         """Adds all pyenv installations to the PATH."""
         if PYENV_INSTALLED:
             if PYENV_ROOT:
+                pyenv_paths = {}
                 for found in glob(
-                    '{0}{1}versions{1}*{1}bin'.format(
+                    '{0}{1}versions{1}*'.format(
                         PYENV_ROOT,
                         os.sep
                     )
                 ):
-                    add_to_path(found)
+                    pyenv_paths[os.path.split(found)[1]] = '{0}{1}bin'.format(found, os.sep)
+
+                for version_str, pyenv_path in pyenv_paths.items():
+                    version = pip._vendor.packaging.version.parse(version_str)
+                    if version.is_prerelease and pyenv_paths.get(version.base_version):
+                        continue
+                    add_to_path(pyenv_path)
             else:
                 click.echo(
                     '{0}: PYENV_ROOT is not set. New python paths will '


erinxocon:
This looks good to me from a code perspective.  I'll let @nateprewitt chime in since he's more knowledgeable about pyenv.

nateprewitt:
Hrmm, so this solves the issue but causes another. We now refuse to add pre-releases to the path and I believe that means we'll now ignore a user passing a full path to a valid pre-release python binary.

techalchemy:
I don’t know that it actually causes this additional issue since I don’t think we are currently parsing version info from the path itself other than in this code that is implemented here (but possibly we should consider doing so?) Currently we parade version info from `python —version` due to Windows path related issues not conveying python versions if I remember correctly.  I have no idea if there is anything we can pull from sys.* to indicate pre-release status, but that would be something to look at. 

As it stands if you have both a pre-release and a release on your path the current process prefers the pre-release. A possibly more elegant and idiomatic approach would be to create a list of stuff to slap on top of `$PATH` and always put prereleases at the end instead of ignoring them entirely.  Thoughts?

techalchemy:
Always happy to do the easy ones :)

joshfriend:
@techalchemy Thanks for this! IDK why, but I just couldn't come up with any solutions for it myself.

`sys.version_info` will tell you the pre-release status, here's a few examples:

```python
>>> sys.version_info
sys.version_info(major=3, minor=7, micro=0, releaselevel='alpha', serial=1)
>>> sys.version_info
sys.version_info(major=3, minor=6, micro=0, releaselevel='beta', serial=1)
>>> sys.version_info
sys.version_info(major=2, minor=7, micro=14, releaselevel='candidate', serial=1)
>>> sys.version_info
sys.version_info(major=3, minor=6, micro=2, releaselevel='final', serial=0)
```

Right now, pipenv is still going to add the `/bin` from each python install it finds in pyenv to the `$PATH`. This is **a LOT** of python installs for me 😄 . Perhaps theres a way to not pollute the `PATH` so much? I could see a situation where the contents of one python's `bin` is different from another, but because _all_ are in the path, they just get mixed together.

techalchemy:
@joshfriend that is pretty hilarious. I did notice that myself and was considering alternatives but wasn’t sure about which was best. I’m open to suggestions and will kick this around a bit today and see if something clever pops up

PR #1001, opened 2017-10-31
Add a function to standardize console output.

isudox:
i.e.
INFO  Creating pipenv environment.
WARN  Network connect interrupted.
ERROR  Stack over flow!


nateprewitt:
Thanks for the initiative @isudox! I don't know if this will fall in line with how Kenneth has wanted to handle messaging from pipenv though. I think we're going to pass on this but please let us know if you have any other thoughts.

Issue #1002, opened 2017-10-31
Pipenv doesn't resolve --system python correctly.


galuszkak:
##### Describe you environment

1. OS Type `Docker Alpine Python` -> https://github.com/docker-library/python/blob/1d59eb2dd813c64891bf554a8ea01754aba25816/3.6/alpine3.6/Dockerfile
1. Python version: `Python 3.6.3`
1. Pipenv version: `pipenv, version 8.3.1`

##### Expected result

If I set PIP_PYTHON_PATH it's ignored and overrided by pipenv (my question is why, what's reasoning behind that?) because https://github.com/kennethreitz/pipenv/blob/master/pipenv/utils.py#L367 It's overriding this ENV variable when I use flag `--system`. I've catched that issue because when building my docker image it was failing on `pipenv install --system`

Why I know this? Because there is other function that works incorrectly.
```(Pdb) python_path
'/bin/python'
(Pdb) which('python')
'/bin/python'
(Pdb) print(sys.executable)
'/usr/local/bin/python'

/ # /bin/python
sh: /bin/python: not found
```

I think issue is somewhere here: https://github.com/kennethreitz/pipenv/blob/1bf6f5101a3bf1dca78be814e742411cc5bdb2dd/pipenv/cli.py#L1440
```
            p = os.sep.join([location] + ['bin/{0}'.format(command)])
```
this should return `/usr/local/bin/python` but for some strange reason it's returning `/bin/python`

##### Actual result

`pipenv install --system` should install not failed that /bin/python is not found.

##### Steps to replicate

`pipenv install --system` on that docker image will fail

techalchemy:
@galuszkak this is interesting. So `sys.executable` is set according to pdb, but we aren’t passing the fact that you used `—system` when we actually call `which()`, so it’s trying to use the project virtualenv location which presumably doesn’t exist. This is the relevant invocation which should be easy enough to fix when I get into the office: https://github.com/kennethreitz/pipenv/blob/master/pipenv/utils.py#L406

As for why we are doing all of this, it’s mainly so that we can trick the pip-fools package resolver to resolve python packages for versions of python that are different from the active interpreter

erinxocon:
Using this Dockerifle, which is based on the one commited to this repo I cannot reproduce this issue, not saying it doesn't exist, looks like something is up, buy this dockerfile seems to work:

```Dockerfile
FROM python:alpine3.6

# -- Install Pipenv:
[...]

PR #1005, opened 2017-10-31
fixed problem with not respecting --system (closes: #1002)

galuszkak:
This should fix problem in #1002 .

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 import contextlib
 import codecs
+import functools
 import os
 import sys
 import shutil
@@ -788,7 +789,6 @@ def do_install_dependencies(
             dev_deps.update(lockfile['develop'])
             dev_vcs_deps.update(lockfile.get('develop-vcs', {}))
 
-
     # Install default dependencies, always.
     deps.update(lockfile['default'] if not only else {})
     vcs_deps.update(lockfile.get('default-vcs', {}))
@@ -821,7 +821,6 @@ def do_install_dependencies(
 
             click.echo('\n'.join(d[0] for d in dev_deps_list))
             sys.exit(0)
-
 
     procs = []
 
@@ -1051,7 +1050,7 @@ def do_lock(verbose=False, system=False, clear=False, pre=False):
         verbose=verbose,
         python=python_version(which('python', allow_global=system)),
         clear=clear,
-        which=which,
+        which=functools.partial(which, allow_global=system),
         which_pip=which_pip,
         project=project,
         pre=pre
@@ -1112,7 +1111,7 @@ def do_lock(verbose=False, system=False, clear=False, pre=False):
         sources=project.sources,
         verbose=verbose,
         python=python_version(which('python', allow_global=system)),
-        which=which,
+        which=functools.partial(which, allow_global=system),
         which_pip=which_pip,
         project=project,
         pre=pre


galuszkak:
Is that fix fine @techalchemy and it is what was on Your mind? Sounds, tests are passing so it should be OK.

techalchemy:
@galuszkak awesome, thanks for putting this PR together.  This seems like a good approach, I just need to sit with it for a minute and think about whether this can have any impact on package resolution with pip-tools.

/cc @nateprewitt @vphilippon if you guys have any reservations at a glance

galuszkak:
@nateprewitt my initial idea behind fix was that, --system flag isn't passed down, and I didn't want to change interface of resolve_deps for now.

Happy to find real cause. I supply more info now in #1002 to fix that issue. Hope this helps.

nateprewitt:
Thanks @galuszkak!

techalchemy:
@nateprewitt I think it comes down to our approach, and as with many other decisions we are going to have to make a call about whether to accept a band-aid and think about a long-term fix in a refactor or whether we want to start thinking about refactoring (but I'm not clear when these things will be merged in that case, for instance local files are still broken afaik)

Here is the relevant code, note that `allow_global` is functionally a lie since it doesn't search the system path but instead returns the executable from the current interpreter.  This is only relevant currently when passing `which('python')` specifically.

```python
def which(command, location=None, allow_global=False):
    if location is None:
        location = project.virtualenv_location

    if not allow_global:
        if os.name == 'nt':
            p = find_windows_executable(os.path.join(location, 'Scripts'), command)
        else:
    else:
        if command == 'python':
            p = sys.executable

return p
```

galuszkak:
@techalchemy I don't think so. If some is invoking pipenv outside of virtualenv `sys.executable` should return system interpreter which used right now.

I think this function should be like this then which I think it is much more consistent:
```
def which(command, location=None, allow_global=False):
    if location is None and not allow_global:
        if not project.virtualenv_location:
            raise RuntimeError('There is no correct virtualenv here')
        
        location = project.virtualenv_location
        
    if not allow_global:
        if os.name == 'nt':
            p = find_windows_executable(os.path.join(location, 'Scripts'), command)
        elif location:
        elif command == 'python':
            p = sys.executable
    else:
        if command == 'python':
            p = sys.executable

return p
```

techalchemy:
But here is the issue with that approach:

- We don’t require a virtualenv to run code
- `Project.virtualenv_location` can return a value whether a virtualenv exists or not
- We don’t really want to fall back to`sys.executable` when `allow_global=False`

> @techalchemy I don’t think so.

I’m not sure what this was a response to but pipenv should function the way I described. The `which` function specifically however currently only respects `allow_global` in the context of returning a path to `python`. Since the function is misleadingly naked ‘which’ rather than ‘which_python’ for example I think this is a problematic behavior. As a developer I would expect, for example, to be able to ask for the path to a global `flake8`, or `anaconda` or whatever using this function.

galuszkak:
> We don’t require a virtualenv to run code

Yes, I didn't stated differently in that matter. Proposed code actually check if You are looking in virtualenv or global. If neither then You have exception raised, because, we don't know what is actually searched.

> Project.virtualenv_location can return a value whether a virtualenv exists or not

If virtualenv doesn't exists this should either give None or ''. Why this should have different behavior, I'm not really getting that point :) .

> We don’t really want to fall back tosys.executable when allow_global=False
Far point, this is just because I was looking mostly on my use cases and saw that someone started with `command == 'python'` check.

If we want real `which` function then we should use `shutils.which` and backport `which` for python 2.7 . But I'm not sure if this is what You are trying to say. We can always pass `$PATH` env to `shutils.which` to look in virtualenv or global.

techalchemy:
> We don’t require a virtualenv to run code
>> Yes, I didn't stated differently in that matter

The code in the current codebase will return the path to an executable inside the virtualenv whether the virtualenv exists or not, but your change doesn't allow us to do that.  Currently we don't require any actual virtualenv to exist.  I didn't write this code but I also don't feel safe assuming you can break this functionality.  This appears to be an intended behavior of `Project.virtualenv_location`.

> If virtualenv doesn't exists this should either give None or ''. Why this should have different behavior? I'm not really getting that point :) .

There is a key factor here -- this is your opinion about that the code _should_ do.  I am not making any statements about what it should or shouldn't do, I'm just explaining what the code _actually does_:

```python
    @property
    def virtualenv_location(self):

        # if VIRTUAL_ENV is set, use that.
        if PIPENV_VIRTUALENV:
            return PIPENV_VIRTUALENV

        # Use cached version, if available.
        if self._virtualenv_location:
            return self._virtualenv_location

        # The user wants the virtualenv in the project.
        if not PIPENV_VENV_IN_PROJECT:
            c = delegator.run('pew dir "{0}"'.format(self.virtualenv_name))
            loc = c.out.strip()
        # Default mode.
        else:
            loc = os.sep.join(self.pipfile_location.split(os.sep)[:-1] + ['.venv'])

        self._virtualenv_location = loc
return loc
```

I didn't mean to clutter this PR up with the sidebar comment to @nateprewitt which was intended just to be a quick dialogue about whether this PR is the right approach to take for the current problem with the current code as is.  This is not the place to discuss how to refactor pipenv.

galuszkak:
@techalchemy thanks, I think You are right, my naive approach is probably taken from fact that I don't have time to dive too much to code base here.

@nateprewitt can You elaborate what are our options? I still believed that first approach fixes the main issue which is not passing allow_global flag to existing system so it can be executed with same params in different places.

galuszkak:
Friendly ping, what are we doing with this PR? Is there any objection to move it forward, or we should dive somewhere else to fix the issue? cc @techalchemy @nateprewitt

techalchemy:
@nateprewitt are you OK to move forward on this / is the `import partial` a dealbreaker?

nateprewitt:
@techalchemy Like I said, I'm really not a fan of this, but I don't have the time right now to come up with a better solution.

If we merge this, the only thing I ask is we do this uniformly. Right now this patch only affects `dev-packages` and not `packages`. We need to add the same logic on [line 1117](https://github.com/Solution4Future/pipenv/blob/e3df9f4451af1bc1cf3bcc7e7807549d3d6476b8/pipenv/cli.py#L1117).

galuszkak:
@nateprewitt I will update PR then.

galuszkak:
PR updated.

techalchemy:
@nateprewitt I am mainly interested in the nature of your objection and I am happy to work around it if it is a dealbreaker.  I just feel we are currently in a holding pattern on fixing some bugs and want to understand if that is because you are generally opposed to importing things, or if you have specific concerns about the particular imports

I wouldn't think at a glance that `from functools import partial` would add significant overhead but if you do I don't mind doing some kind of diagnostic or thinking of a new approach

kennethreitz:
This is a good solution, but closing for the solution proposed in #1002

Issue #1003, opened 2017-10-31
pipenv should require a minimum flake8 version in setup.py


nsoranzo:
Be sure to check the existing issues, both open and closed.

Describe the issue briefly here.

##### Describe you environment

1. OS Type: Linux Ubuntu 14.04.5 LTS
1. Python version: ` 2.7.6
1. Pipenv version: 8.3.1

##### Expected result

After "successfully" installing pipenv with:
```
$ pip install --user pipenv
```
I tried to create a new virtualenv with:
```
pipenv --python=python2.7
```

##### Actual result

```
$ pipenv install flake8
Traceback (most recent call last):
  File "/usr/users/ga002/soranzon/.local/bin/pipenv", line 9, in <module>
    load_entry_point('pipenv==8.3.1', 'console_scripts', 'pipenv')()
  File "/usr/users/ga002/soranzon/.local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 570, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/usr/users/ga002/soranzon/.local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2755, in load_entry_point
    return ep.load()
  File "/usr/users/ga002/soranzon/.local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2409, in load
    return self.resolve()
  File "/usr/users/ga002/soranzon/.local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 2415, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
  File "/usr/users/ga002/soranzon/.local/lib/python2.7/site-packages/pipenv/__init__.py", line 17, in <module>
    from .cli import cli
  File "/usr/users/ga002/soranzon/.local/lib/python2.7/site-packages/pipenv/cli.py", line 27, in <module>
    import flake8.main.cli
ImportError: No module named cli
```

##### Steps to replicate

The issue is due to the ancient version of flake8 installed from Ubuntu:
```
$ /usr/bin/flake8 --version
2.1.0 (pep8: 1.4.6, pyflakes: 0.8.1, mccabe: 0.2.1) CPython 2.7.6 on Linux
```

After running:
```
$ pip install --user --upgrade flake8
```
everything works now.

erinxocon:
This should be fixed in master soon!  Thanks for the report @nsoranzo!

erinxocon:
[...]

PR #1007, opened 2017-11-01
Added a minimum version of flake8

erinxocon:
should address #1003


nateprewitt:
I don't think we need to be this strict. We should probably figure out where the problem arose from (somewhere between 2.1 and 3.5) and choose an option that's a little more permissive.

erinxocon:
@nateprewitt Looks like there was an api change in version 3.0 that cases this.  The minimum version can therefore be 3.0.0.

Issue #856, opened 2017-10-08
pipenv doesn't respect pip.conf


hrbonz:
I'm using [devpi](https://devpi.net/docs/devpi/devpi/latest/+d/index.html) as a private pypi repository where I can proxy cache pypi packages and add my own in-house packages.

My `pip.conf` looks like the following:
```
[global]
index_url = https://pypi.priv.xxx/prod/+simple/
[search]
index = https://pypi.priv.xxx/prod/
```

Couldn't find another issue mentioning this problem.

kennethreitz:
These entries need to go into your Pipfile.

hrbonz:
@kennethreitz I don't think I've made my case really clear so let me try to convince you with some use cases I've run into while getting started with pipenv.
First of, I'm just starting to dive into pipenv, how it works and its code. I know that you can specify a source in `Pipfile`, I also saw you can use named indexes, that looks similar to what can be done in `~/.pypirc`.

## Creating a new pipenv project/environment

```
$ mkdir foobar
$ cd foobar
$ pipenv install --verbose requests
⠋New python executable in /home/xxx/.local/share/virtualenvs/foobar-JdBU33Mf/bin/python         
Installing setuptools, pip, wheel...done.                                                      
                                                                                               
Virtualenv location: /home/xxx/.local/share/virtualenvs/foobar-JdBU33Mf                         
Installing requests… 
⠙Installing u'requests'
$ "/home/xxx/.local/share/virtualenvs/foobar-JdBU33Mf/bin/pip" install   --verbose   "requests"
-i https://pypi.python.org/simple --exists-action w
Collecting requests
  1 location(s) to search for versions of requests:
  * https://pypi.python.org/simple/requests/
  Getting page https://pypi.python.org/simple/requests/
[...]
Successfully installed certifi-2017.7.27.1 chardet-3.0.4 idna-2.6 requests-2.18.4 urllib3-1.22
Cleaning up...

Adding requests to Pipfile's [packages]…
  PS: You have excellent taste! ✨ 🍰 ✨
Locking [dev-packages] dependencies…
Locking [packages] dependencies…
Updated Pipfile.lock (76e6d4)!
```

## Install a package from an additional index

When installing a package that pypi doesn't know, it won't work right away (it does work with pip or pip-tools as they use `pip.conf` settings).
```
$ pipenv install --verbose palantir
Installing palantir…
⠋Installing u'palantir'
[...]

Issue #988, opened 2017-10-28
Lock file gets incorrect hashes when a custom index is specified


pfmoore:
I am specifying a custom index in `Pipfile` to host a wheel for `pycurl` (which is also on PyPI, but does not have a wheel for my environment there). My index just hosts the wheel, with a SHA256 hash specified.

When I do `pipenv install`, I get an error saying that the hash is not found, and showing a list of expected hashes. I don't know where it got that list from, but it's not from my index... The install fails, and doesn't install `pycurl`.

##### Describe you environment

1. OS is Windows 10
1. Python version: `$ python -V`. Python 3.6.2
1. Pipenv version: `$ pipenv --version`. pipenv, version 8.2.7

Pipenv file:

```toml
[[source]]
url = "https://pypi.python.org/simple"
verify_ssl = true
name = "pypi"

[[source]]
url = "https://pfmoore.github.io/simple"
verify_ssl = true
name = "pycurl"

[packages]
pycurl = {version="*", index="pycurl"}

[dev-packages]
```

##### Expected result

I expected pipenv to download the wheel (which is valid for my environment, confirmed via `pip install`) and install it.

##### Actual result

Output from `pipfile install --verbose`:

```
Creating a virtualenv for this project…
Using real prefix 'C:\\Users\\Gustav\\AppData\\Local\\Programs\\Python\\Python36'
New python executable in C:\Users\Gustav\.virtualenvs\riot_data-g6PtQnzc\Scripts\python.exe
Installing setuptools, pip, wheel...done.

Virtualenv location: C:\Users\Gustav\.virtualenvs\riot_data-g6PtQnzc
Pipfile.lock not found, creating…
Locking [dev-packages] dependencies…
Locking [packages] dependencies…
Updated Pipfile.lock (3d54a4)!
Installing dependencies from Pipfile.lock (3d54a4)…
[...]

PR #1008, opened 2017-11-01
respect pip configuration at Pipfile creation

hrbonz:
Respect pip index configuration (pip.conf or environment variable) when
creating a new Pipfile.

I'm really not sure about this implementation but I wanted to start the conversation, I've made some choices that I think have a minimal impact.

I couldn't find a convenient method in the `pip` project that would just hand me out the best option and respect the configuration precedence so I implemented it by hand.

Change mentioned in #856

### DIFF
--- a/pipenv/project.py
+++ b/pipenv/project.py
@@ -18,6 +18,8 @@ from .utils import (
 from .environments import PIPENV_MAX_DEPTH, PIPENV_VENV_IN_PROJECT
 from .environments import PIPENV_VIRTUALENV, PIPENV_PIPFILE
 
+from pip.baseparser import ConfigOptionParser
+
 if PIPENV_PIPFILE:
     if not os.path.isfile(PIPENV_PIPFILE):
         raise RuntimeError('Given PIPENV_PIPFILE is not found!')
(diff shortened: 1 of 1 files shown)


hrbonz:
I feel like I should add some tests here

techalchemy:
Hey @hrbonz thanks for putting this together. Unfortunately altering which source is added to the Pipfile doesn’t quite respect `pip.conf` because of the way pip already generates its repository index list (which already should respect `pip.conf`) and because of the way the piptools resolver works (which isn’t affected by the `Pipfile`).

hrbonz:
@techalchemy unfotunately, `pip.conf` is not respected because of the configuration precedence and the use of [sources](https://github.com/kennethreitz/pipenv/blob/master/pipenv/cli.py#L1400) [as indexes on the command line](https://github.com/kennethreitz/pipenv/blob/master/pipenv/utils.py#L374) with `-i`/`--index-url` or `--extra-index-url`.

That is the base of the issue I described in #856

techalchemy:
@hrbonz I see what you're saying, but I'm not actually sure this is the right approach here.  It's not clear that we actually want to always fall back to the default indexes in the host pip config, since I'm not sure that provides sufficient portability.  I do think we would want to respect environment settings, however, but I think there is some discussion to have here (e.g. in #988 this came up)

We have broader design questions to answer but for now, I would be much more prepared to accept something that prioritizes environment level settings (e.g. `PIP_INDEX_URL` followed by an iterative list over `PIP_EXTRA_INDEX_URL`) -- this approach will probably need to take into account `verify_ssl` from the `PIP_TRUSTED_HOST` environment variable

techalchemy:
And before you rework, /cc @nateprewitt would you be willing to approve a PR to use `PIP_INDEX_URL` from the environment if it exists as the primary pip index (i.e. first entry in `[[sources]]` in the Pipfile?  This seems like something that might impact existing users so I would want a second set of eyes on it

kennethreitz:
We should honor the environment variable instead.

kennethreitz:
not today, but thank you

hrbonz:
There must be something I don't get, why force users to set an environment variable when there is already a configuration for a standard tool that gives you this information? This is, moreover, the tool you use under the hood.
Environment variables are useful in some situations but really not ideal most of the time.

PR #1011, opened 2017-11-01
Use the default PYENV_ROOT if it is not set

jamieconnolly:
When running `pipenv install`, it doesn't see that I have already installed the Python version through pyenv. I get the following warning:
```
Warning: PYENV_ROOT is not set. New python paths will probably not be exported
properly after installation.
```

If I run `PYENV_ROOT=~/.pyenv pipenv install` then it successfully sees the installed version.

I don't export a `PYENV_ROOT`, as pyenv defaults to using `~/.pyenv` (seen [here](https://github.com/pyenv/pyenv/blob/2f396596d768bf7aa300f7e89e89ae4cc0f748aa/libexec/pyenv#L54)).

This PR tells Pipenv to use this default if there's no `PYENV_ROOT` set, so that it doesn't break when used with the default installation of pyenv.

### DIFF
--- a/pipenv/environments.py
+++ b/pipenv/environments.py
@@ -81,9 +81,9 @@ PIPENV_INSTALL_TIMEOUT = 60 * 15
 
 PIPENV_DONT_USE_PYENV = os.environ.get('PIPENV_DONT_USE_PYENV')
 
-PYENV_ROOT = os.environ.get('PYENV_ROOT')
+PYENV_ROOT = os.environ.get('PYENV_ROOT', os.path.expanduser('~/.pyenv'))
 
-PYENV_INSTALLED = (bool(os.environ.get('PYENV_SHELL')) or bool(PYENV_ROOT))
+PYENV_INSTALLED = (bool(os.environ.get('PYENV_SHELL')) or bool(os.environ.get('PYENV_ROOT')))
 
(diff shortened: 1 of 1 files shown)


jamieconnolly:
I'm not sure why Python 2.7 is failing on Travis although it seems to be unrelated to this PR. A previous PR also fails because of the same error (see [here](https://travis-ci.org/kennethreitz/pipenv/jobs/295641251)).

This is the error that it outputs so you (hopefully) don't have to go digging around.

```
pip.exceptions.PreviousBuildDirError: pip can't proceed with requirements 'configparser==3.5.0 from https://pypi.python.org/packages/7c/69/c2ce7e91c89dc073eb1aa74c0621c3eefbffe8216b3f9af9d3885265c01c/configparser-3.5.0.tar.gz#md5=cfdd915a5b7a6c09917a64a573140538' due to a pre-existing build directory (/tmp/tmpuRAPWNbuild/configparser). This is likely due to a previous installation that failed. pip is being responsible and not assuming it can delete this. Please delete it and try again.
```

nateprewitt:
@jamieconnolly, that was a transient failure. Things worked fine on a rerun.

This looks good to me, so let's do it. Thanks for the contribution!

Issue #1010, opened 2017-11-01
Intelligently Warn when version is provided incorrectly


kennethreitz:
Message from a fan:

> I tried using pipenv and after pipenv install requests noticed it used version "*" in the pipfile. I edited that to "2.18.4" to fix it to an exact version but got an error Your dependencies could not be resolved. You likely have a mismatch in your sub-dependencies. After googling found that the exact format is supposed to be "==2.18.4".
> 
> I found pipenv because I was looking for something like npm or composer for Python, so I thought the version constraints would be similar. Maybe you can show an error when the version constrain format is wrong, or better yet, add it as an alias to the correct one?

I love this idea, and think we should implement it.

nateprewitt:
@kennethreitz, could you clarify which approach you were leaning towards 😁? I think we can definitely warn, aliasing may be a bit murkier.

kennethreitz:
Warning.

techalchemy:
@nateprewitt @kennethreitz if we only find a valid version (we can parse as I am doing for pyenv right now) with no specifiers we can assume it should be a strict pin and also issue a warning, wouldn't that be the most elegant / user friendly?

nateprewitt:
@techalchemy, the mantra we've adopted for a lot of the areas in Requests is "don't make assumptions for the user". Whenever we have, there's usually a vocal group who doesn't want the tool to behave in X way because it's not the correct assumption for use case Y.

By having a single standardized format there isn't ambiguity or additional code to guess what the user may have meant. The warning is a minimal change and guides the user in the right direction pretty quickly.

techalchemy:
@nateprewitt That's totally valid.  I do agree with the user in question that including the specifier if you want to pin a specific version when we dont include a specifier for the wildcard case is kind of confusing though.  The change would be pretty substantial and touches a lot of moving parts so I agree a warning is easiest.

erinxocon:
I added a warning to pip-tools if it can't try any version, to check the version number/specifier and look at PEP440 for any more information.  The warning message can probably be simplified for the end user.  That's just my initial take.

erinxocon:
![image](https://user-images.githubusercontent.com/208567/32309947-a1451cfa-bf54-11e7-81e1-eb2245017555.png)

How it looks in the terminal.

erinxocon:
This should be addressed in master currently.

PR #1019, opened 2017-11-02
Added nice warning message for version specifiers

erinxocon:
This modifies the pip-tools exceptions class to prompt the user the check their version specifier/number if none be resolved.

### DIFF
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -290,7 +290,7 @@ def cleanup_toml(tml):
         # Insert a newline after the heading.
         if after:
             new_toml.append('')
-    
+
     # adding new line at the end of the TOML file
     new_toml.append('')
     toml = '\n'.join(new_toml)
@@ -467,7 +467,12 @@ def resolve_deps(deps, which, which_pip, project, sources=None, verbose=False, p
                     crayons.red('$ pipenv graph')
                 ),
                 err=True)
+
             click.echo(crayons.blue(e))
+
+            if 'no version found at all' in str(e):
+                click.echo(crayons.blue('Please check your version specifier and version number. See PEP440 for more information.'))
+
             sys.exit(1)
 
     for result in resolved_tree:


erinxocon:
Addresses #1010

techalchemy:
@erinxocon this is a nice addition to the warnings for locking.  Do we need to add some kind of handling for the install command? To stay aligned with the approach this takes (rather than parsing ourselves for specifiers), we could handle `requirements` related exceptions in calls to `convert_deps_to_pip` or something like that.

One question I have is whether we have any guidelines or process for deciding when to vendor/patch/update things as compared with extending.  I don't want to make this minor change into the hill-to-die-on for patched libraries, but it would be helpful for me when making decisions as well as when reviewing prs to understand what our approach is here

erinxocon:
@techalchemy I'm not sure about the process for updating patched/vendored libs.  That's a good question for @nateprewitt!  

Currently we have click handling a lot of those issues, hence me adding the verbage to the pathced pip-tools rather than pipenv.

techalchemy:
Ok. The install piece doesn’t even seem to fail properly so this I’m not sure you would be able to handle that

nateprewitt:
I'm of the overly opinionated opinion you shouldn't modify other source code that you're not going to contribute back. I brought this up with Kenneth when he added the pip and pip-tools modifications. It means we're now stuck maintaining our own fork (that isn't *really* in a proper repo) that will continue to fall out of sync because we can't merge back in easily.

So, I'd personally really prefer we don't make any more patches to vendored code if possible. If we can catch this exception and raise something cleaner through click/exceptions we own, that'd be my preference. That said, if you two think this is the right approach, that's two votes against one.

techalchemy:
@nateprewitt I tend to favor extending/catching/monkeypatching as compared with source modifications for the same reasons as well but I haven't spent any time looking at how we are handling this currently or what it would take to tack onto the piptools exception vs modifying it directly,

That said you'll notice that I am new to being an open source maintainer so I'm not being overly assertive with opinions like this or merging PRs or whatnot, while I figure out the dynamics of all of that.  And above all I favor pragmatism-- I'd rather have a small improvement in the near-term if it's a practice we've already mostly adopted, as compared with making no progress at all because the prospect of reimplementing a ton of functionality is daunting

erinxocon:
@techalchemy @nateprewitt This is now handled in utils.py from pipenv and not in our patched pip-tools.

techalchemy:
That was fast!

erinxocon:
@nateprewitt Going to merge this for inclusion in the next update.

PR #1022, opened 2017-11-02
Add homebrew to the install instructions

richardcooper:
#999 added instructions for installing via homebrew to the README. This PR adds an expanded version of the same instructions to the docs

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -43,12 +43,18 @@ The problems that Pipenv seeks to solve are multi-faceted:
 Install Pipenv Today!
 ---------------------
 
+Pipenv is a python package and so can be installed using ``pip`` as you would expect.
 ::
 
     $ pip install pipenv
     ✨🍰✨
 
-If you have excellent taste, there's also a  `fancy installation method <https://docs.pipenv.org/install.html#fancy-installation-of-pipenv>`_.
+If you have excellent taste, there are various other installation methods which
+prevent pipenv and its dependencies from interfering with the rest of your
+Python installation. These include
+`Pipsi <https://docs.pipenv.org/install.html#fancy-installation-of-pipenv>`_,
+`Nix <https://docs.pipenv.org/install.html#referentially-transparent-installation-of-pipenv>`_
+and `Homebrew <https://docs.pipenv.org/install.html#homebrew-installation-of-pipenv>`_.
 
 .. toctree::
    :maxdepth: 2
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -59,6 +59,9 @@ If you installed Python from source, with an installer from `python.org`_, or
 via `Homebrew`_ you should already have pip. If you're on Linux and installed
 using your OS package manager, you may have to `install pip <https://pip.pypa.io/en/stable/installing/>`_ separately.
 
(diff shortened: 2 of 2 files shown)


docs/index.rst
```diff
@@ -43,12 +43,18 @@ The problems that Pipenv seeks to solve are multi-faceted:
...
 ---------------------
 
+Pipenv is a python package and so can be installed using ``pip`` as you would expect.
 ::
 
     $ pip install pipenv
     ✨🍰✨
 
-If you have excellent taste, there's also a  `fancy installation method <https://docs.pipenv.org/install.html#fancy-installation-of-pipenv>`_.
+However, installing pipenv using pip should be considered a last resort. To
```
nateprewitt: I'm not sure we want this wording. Pip is a fine option for installing pipenv and is the common case. I don't think we want to push users away from that because it provides an additional barrier for entry.

docs/install.rst
```diff
@@ -233,6 +236,23 @@ Nix provides atomic upgrades and rollbacks, it's reliable and reproducible thank
 
 .. _pragmatic_installation:
 
+☤ Mac Installation of Pipenv
```
nateprewitt: Let's change this to "Homebrew" instead of "Mac".

docs/install.rst
```diff
@@ -233,6 +236,23 @@ Nix provides atomic upgrades and rollbacks, it's reliable and reproducible thank
...
+===============================
+
+Homebrew is a popular open-source package management system for macOS.
+
+Installing pipenv via homebrew will keep pipenv and all of its dependencies in
```
nateprewitt: Homebrew is capitalized in the official documentation and we should be making sure we use that consistently. There are multiple places in this PR this needs to be tweaked.

docs/install.rst
```diff
@@ -59,6 +59,9 @@ If you installed Python from source, with an installer from `python.org`_, or
 via `Homebrew`_ you should already have pip. If you're on Linux and installed
 using your OS package manager, you may have to `install pip <https://pip.pypa.io/en/stable/installing/>`_ separately.
 
+If you plan to install pipenv using Homebrew you can skip this step. The
```
nateprewitt: I don't know if this is necessary.

docs/install.rst
nateprewitt: Let's also move this section underneath the pragmatic installation section.

docs/install.rst
richardcooper: I would like this to stay. System python on macOS doesn't include pip. Homebrew python doesn't either, it gives you either `pip2` or `pip3` but no `pip`.

So its very likely that `pip --version` will fail for anyone interested in the Homebrew install method. There's no point making them go and work out how to install pip for themselves when it's not needed.

I'm not actually sure if it's needed for the Pipsi or Nix method either.

richardcooper:
@nateprewitt I've made most of the changes you've requested.

I'm not sure how this is causing Travis CI to fail. It seems unrelated.

techalchemy:
@richardcooper the CI failure was unrelated but I did restart it so the ugly red x will go away

nateprewitt:
Thanks @richardcooper!

PR #1026, opened 2017-11-03
Only check for updates once per 24 hours

### LINKED ISSUE #1023: Pipenv is really slow to exit if the network is not working
When the network is very slow, Pipenv takes an extremely long time to exit after doing almost anything (e.g. `pipenv` will print the usage message, and then block for a while – but `pipenv --help` will behave as normal).

1. OS Type: Ubuntu 16.04
1. Python version: 3.6.3
1. Pipenv version: 8.3.1, installed via pipsi

Here is the result of `pip list` in the virtualenv that Pipenv is installed in:

```
Package          Version    
---------------- -----------
certifi          2017.7.27.1
chardet          3.0.4      
flake8           3.5.0      
idna             2.6        
mccabe           0.6.1      
pew              1.1.0      
pip              9.0.1      
pipenv           8.3.1      
pycodestyle      2.3.1      
pyflakes         1.6.0      
requests         2.18.4     
setuptools       36.6.0     
urllib3          1.22       
virtualenv       15.1.0     
virtualenv-clone 0.2.6      
wheel            0.30.0
```

##### Expected result

Pipenv shouldn't suddenly take longer to exit because the network is behaving strangely.

##### Actual result

If the network is slow, it takes almost exactly 21 seconds to run many Pipenv commands, including:

- `pipenv`
- `pipenv --completion`

but *not* `pipenv --help` (finishes in about 1 second, regardless of network state AFAICT).

If I press ctrl-C after running `pipenv`, when it has printed the usage message but not yet exited, I get a traceback like this:

```
Error in atexit._run_exitfuncs:
  File "/usr/lib/python3.6/concurrent/futures/thread.py", line 40, in _python_exit
    t.join()
  File "/usr/lib/python3.6/threading.py", line 1056, in join
    self._wait_for_tstate_lock()
  File "/usr/lib/python3.6/threading.py", line 1072, in _wait_for_tstate_lock
    elif lock.acquire(block, timeout):
KeyboardInterrupt
Exception ignored in: <function ThreadPoolExecutor._adjust_thread_count.<locals>.weakref_cb at 0x7fce19f11a60>
  File "/usr/lib/python3.6/concurrent/futures/thread.py", line 131, in weakref_cb
  File "/usr/lib/python3.6/queue.py", line 145, in put
  File "/usr/lib/python3.6/threading.py", line 347, in notify
TypeError: 'NoneType' object is not callable
```

##### Steps to replicate

~~Have a WiFi card that randomly decides to stop working~~ Run `sudo tc qdisc add dev INTERFACE root netem delay 10s` (where `INTERFACE` might be something like `wlp2s0` – obtain it from either `ip link` or `ifconfig`) to add a 10s delay to all your traffic, then run `pipenv`.

The delay can then be removed with `sudo tc qdisc del dev INTERFACE root netem`.

[Source for these commands](http://bencane.com/2012/07/16/tc-adding-simulated-network-latency-to-your-linux-server/)

techalchemy:
- Store stamp file in app cache dir (os specific)
- Vendor appdirs (is this a good idea?)
- Fixes #1023 and some other issues I dont know about
- This works but is just a suggestion to get my thoughts on paper

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -1622,9 +1622,10 @@ def cli(
         # Awesome sauce.
         click.echo(crayons.normal(xyzzy, bold=True))
 
-    if not update and need_update_check():
-        # Spun off in background thread, not unlike magic.
-        check_for_updates()
+    if not update:
+        if need_update_check():
+            # Spun off in background thread, not unlike magic.
+            check_for_updates()
     else:
         # Update pip to latest version.
         ensure_latest_pip()
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -138,13 +138,11 @@ def add_to_path(p):
 @background.task
 def check_for_updates():
     """Background thread -- beautiful, isn't it?"""
-    if not need_update_check():
-        return
     try:
+        checked_for_updates()
         r = requests.get('https://pypi.python.org/pypi/pipenv/json', timeout=0.5)
         latest = max(map(semver.parse_version_info, r.json()['releases'].keys()))
         current = semver.parse_version_info(__version__)
-        checked_for_updates()
 
         if latest > current:
             click.echo('{0}: {1} is now available. You get bonus points for upgrading ($ {})!'.format(
@@ -158,6 +156,7 @@ def check_for_updates():
 
 def ensure_latest_self(user=False):
     """Updates Pipenv to latest version, cleverly."""
+    checked_for_updates()
     try:
         r = requests.get('https://pypi.python.org/pypi/pipenv/json', timeout=2)
     except requests.RequestException as e:
@@ -165,7 +164,6 @@ def ensure_latest_self(user=False):
         sys.exit(1)
     latest = max(map(semver.parse_version_info, r.json()['releases'].keys()))
     current = semver.parse_version_info(__version__)
-    checked_for_updates()
 
     if current < latest:
 
@@ -1624,7 +1622,7 @@ def cli(
         # Awesome sauce.
         click.echo(crayons.normal(xyzzy, bold=True))
 
-    if not update:
+    if not update and need_update_check():
         # Spun off in background thread, not unlike magic.
         check_for_updates()
     else:
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -35,7 +35,8 @@ from .utils import (
     convert_deps_from_pip, convert_deps_to_pip, is_required_version,
     proper_case, pep423_name, split_vcs, resolve_deps, shellquote, is_vcs,
     python_version, suggest_package, find_windows_executable, is_file,
-    prepare_pip_source_args, temp_environ, is_valid_url, download_file
+    prepare_pip_source_args, temp_environ, is_valid_url, download_file,
+    need_update_check, checked_for_updates
 )
 from .__version__ import __version__
 from . import pep508checker, progress
@@ -137,10 +138,13 @@ def add_to_path(p):
 @background.task
 def check_for_updates():
     """Background thread -- beautiful, isn't it?"""
+    if not need_update_check():
+        return
     try:
         r = requests.get('https://pypi.python.org/pypi/pipenv/json', timeout=0.5)
         latest = max(map(semver.parse_version_info, r.json()['releases'].keys()))
         current = semver.parse_version_info(__version__)
+        checked_for_updates()
 
         if latest > current:
             click.echo('{0}: {1} is now available. You get bonus points for upgrading ($ {})!'.format(
@@ -161,6 +165,7 @@ def ensure_latest_self(user=False):
         sys.exit(1)
     latest = max(map(semver.parse_version_info, r.json()['releases'].keys()))
     current = semver.parse_version_info(__version__)
+    checked_for_updates()
 
     if current < latest:
 
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -36,7 +36,7 @@ from .utils import (
     proper_case, pep423_name, split_vcs, resolve_deps, shellquote, is_vcs,
     python_version, suggest_package, find_windows_executable, is_file,
     prepare_pip_source_args, temp_environ, is_valid_url, download_file,
-    need_update_check, checked_for_updates
+    need_update_check, touch_update_stamp
 )
 from .__version__ import __version__
 from . import pep508checker, progress
@@ -139,7 +139,7 @@ def add_to_path(p):
 def check_for_updates():
     """Background thread -- beautiful, isn't it?"""
     try:
-        checked_for_updates()
+        touch_update_stamp()
         r = requests.get('https://pypi.python.org/pypi/pipenv/json', timeout=0.5)
         latest = max(map(semver.parse_version_info, r.json()['releases'].keys()))
         current = semver.parse_version_info(__version__)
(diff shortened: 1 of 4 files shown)


techalchemy:
@erinxocon @nateprewitt do whatever you want to do with this, cherrypick ideas or whatever. Gonna go to bed so I just wanted to have this out there for those of you who work all night

erinxocon:
@nateprewitt @techalchemy I actually worked out an extremely similar solution in my branch https://github.com/kennethreitz/pipenv/blob/fix-1023/pipenv/cli.py

Mine does not vendorize Appdirs.

pipenv/cli.py
```diff
@@ -137,10 +138,13 @@ def add_to_path(p):
 @background.task
 def check_for_updates():
     """Background thread -- beautiful, isn't it?"""
+    if not need_update_check():
+        return
     try:
         r = requests.get('https://pypi.python.org/pypi/pipenv/json', timeout=0.5)
         latest = max(map(semver.parse_version_info, r.json()['releases'].keys()))
         current = semver.parse_version_info(__version__)
+        checked_for_updates()
```
anowlcalledjosh: Can I suggest putting this before the call to `requests.get()`? That way, if the user gets bored of waiting for the update check to time out and hits ctrl-C, the update check won't run for the next 24 hours (by which time hopefully the network will be working again) rather than the update check running (and blocking) every time Pipenv runs.

techalchemy:
@erinxocon @nateprewitt im ok with not vendoring appdirs, I just wasn’t sure when to make the call to do so. The package isn’t really maintained anymore which was the only reason I vendored it 

We wound up with almost the same exact solution— yours uses the config dir, mine the cache dir, and mine keeps the file IO in the background as well (but I’m not actually sure how that’s handled). I guess we should just try timing a few operations with both approaches and see which has better performance

erinxocon:
The backgorund decorator just spawns the process in a threadpool that is running in the background.  I chose to only instantiate a background process if needed by keeping the IO in the main thread.  In my fix, I also have it set so that if the user pushes ctrl-c before the update is over, it will not check again for 24 hours.

techalchemy:
Ahhhh ok. Is it undesirable to spawn that process for I/O?

erinxocon:
@techalchemy Unsure, I just thought there was no use in spinning up a whole BG thread to see if we needed to do the update seemed inefficient, but really with today's machines it would be fine either way.  I'm gonna push one or two commits to your branch and then I think we can merge and close the issue!

erinxocon:
@techalchemy I also agree with you about vendoring app dirs after looking at it's maintenance history.

techalchemy:
loving your decisiveness. Hack away, I'm gonna do some tests real quick and see whether things are faster if we put the io in the background or in the foreground

pipenv/cli.py
```diff
@@ -137,10 +138,13 @@ def add_to_path(p):
 @background.task
 def check_for_updates():
     """Background thread -- beautiful, isn't it?"""
+    if not need_update_check():
```
erinxocon: It's gonna be more work for me to fork and submit a PR to your branch @techalchemy.  All I was going to do, pending the outcome of your tests, was to remove statement to be outside of the thread.  This would move some IO to the main cli but I don't think it will be an issue.

pipenv/environments.py
```diff
@@ -88,3 +88,5 @@
 SESSION_IS_INTERACTIVE = bool(os.isatty(sys.stdout.fileno()))
 
 PIPENV_SHELL = os.environ.get('SHELL') or os.environ.get('PYENV_SHELL')
+
+PIPENV_CACHE_DIR = os.environ.get('PIPENV_CACHE_DIR', user_cache_dir('pipenv'))
```
erinxocon: Does the cache dir get cleared on reboot?  Or is it persistent?

pipenv/utils.py
```diff
@@ -946,3 +947,23 @@ def download_file(url, filename):
...
         f.write(r.content)
+
+
+def need_update_check():
+    mkdir_p(PIPENV_CACHE_DIR)
+    p = os.sep.join((PIPENV_CACHE_DIR, '.pipenv_update_check'))
+    if not os.path.exists(p):
+        return True
+    out_of_date_time = time() - (24 * 60 * 60)
+    if os.path.isfile(p) and os.stat(p).st_mtime <= out_of_date_time:
```
erinxocon: I think os.getmtime() would be better for cross platform compatibility?

pipenv/cli.py
erinxocon: I would agree that this should happen before requests.get()

pipenv/utils.py
techalchemy: ^^ it might be, I actually don't know.  Let me check

pipenv/cli.py
techalchemy: oh yeah give me one sec and i will let you know, I was going to do the actual math on it but i'll eyeball and see if there is even a difference at a glance

pipenv/environments.py
techalchemy: It persists across reboots

pipenv/utils.py
techalchemy: I believe `st_mtime` is platform-independent but `getmtime` is slightly more readable and I prefer it personally

pipenv/cli.py
nateprewitt: I'm pretty sure network IO will always be larger than thread spawn time. I'd suggest keeping network calls out of the main thread unless there's significant data showing otherwise.

techalchemy: OK your approach is faster than mine by like ~0,03s on average so I'd say just move that into the cli as you were going to do:

```python
    if not update and need_update_check():
        # Spun off in background thread, not unlike magic.
        check_for_updates()
```

techalchemy:
@erinxocon I incorporated a bunch of your stuff as well, using `os,utime` instead of opening a file pointer all the time now

pipenv/cli.py
techalchemy: @nateprewitt we were mainly talking about the disk IO associated with touching/opening a file pointer, i.e. just whether we should determine if we need to check for an update inside the background thread that will do the network call, or determine whether we need to check for an update before we even spawn the background thread (thereby preventing it from spawning)

pipenv/utils.py
```diff
@@ -946,3 +947,26 @@ def download_file(url, filename):
...
+    if not os.path.exists(p):
+        return True
+    out_of_date_time = time() - (24 * 60 * 60)
+    if os.path.isfile(p) and os.path.getmtime(p) <= out_of_date_time:
+        return True
+    else:
+        return False
+
+
+def checked_for_updates():
```
nateprewitt: Only note I have is maybe renaming this. The name implies it's Boolean in nature, but it will always be true. We probably don't have a reason to return True either at the bottom.

techalchemy: @nateprewitt renamed to `touch_update_stamp()`

techalchemy:
Hmm I think this breaks appveyor as is?

erinxocon:
Looks like your last commit fixed app veyor @techalchemy

techalchemy:
@erinxocon this one should be good now

erinxocon:
@techalchemy as soon as the tests finish I think we can merge!

nateprewitt:
🛬

PR #1031, opened 2017-11-05
Stop matching packages beginning with vcs names

### LINKED ISSUE #1030: pipenv does not install gitlint
`gitlint` is not installed when using `--dev --deploy` flags.

Here's my `Pipfile`:

```
[[source]]

verify_ssl = true


[requires]

python_version = '3.6'


[dev-packages]

django-debug-toolbar = "*"
"pep8-naming" = "*"
"flake8-builtins" = "*"
"flake8-commas" = "*"
"flake8-quotes" = "*"
"flake8-blind-except" = "*"
"flake8-bugbear" = "*"
"flake8-comprehensions" = "*"
"flake8-pep3101" = "*"
pytest = "*"
"pytest-flake8" = "*"
pytest-isort = "*"
pytest-django = "*"
pytest-cov = "*"
gitlint = "*"
pre-commit = "*"
xenon = "*"
radon = "*"
pylint = "*"
pytest-xdist = "*"


[packages]

django = "<1.12"
django-split-settings = "*"
django-axes = "*"
"psycopg2" = "*"
gunicorn = "*"
unipath = "*"
python-decouple = "*"
djangorestframework = "*"
djangorestframework-jwt = "*"
```

Here's the command I am using: `pipenv install --dev --deploy`. The same happens with just `--dev`.

It was working some versions ago.

1. OS Type: macos
1. Python version: `Python 3.6.2` 

##### Expected result

I expect every dependency from `Pipfile` to be installed.

##### Actual result

Here's my `pip freeze`:
```
» pip freeze
apipkg==1.4
aspy.yaml==1.0.0
astroid==1.5.3
attrs==17.2.0
cached-property==1.3.1
certifi==2017.7.27.1
chardet==3.0.4
colorama==0.3.9
coverage==4.4.1
Django==1.11.7
django-axes==2.3.3
django-debug-toolbar==1.8
django-split-settings==0.2.5
djangorestframework==3.7.1
djangorestframework-jwt==1.11.0
execnet==1.5.0
flake8==3.5.0
flake8-blind-except==0.1.1
flake8-bugbear==17.4.0
flake8-builtins==1.0
flake8-commas==0.4.3
flake8-comprehensions==1.4.1
flake8-pep3101==1.1
flake8-polyfill==1.0.1
flake8-quotes==0.12.0
gunicorn==19.7.1
identify==1.0.6
idna==2.6
isort==4.2.15
lazy-object-proxy==1.3.1
mando==0.6.4
mccabe==0.6.1
nodeenv==1.2.0
pep8-naming==0.4.1
pre-commit==1.3.0
psycopg2==2.7.3.2
py==1.4.34
pycodestyle==2.3.1
pyflakes==1.6.0
PyJWT==1.5.3
pylint==1.7.4
pytest==3.2.3
pytest-cache==1.0
pytest-cov==2.5.1
pytest-django==3.1.2
pytest-flake8==0.9.1
pytest-forked==0.2
pytest-isort==0.1.0
pytest-xdist==1.20.1
python-decouple==3.1
pytz==2017.3
PyYAML==3.12
radon==2.1.1
requests==2.18.4
six==1.11.0
sqlparse==0.2.4
Unipath==1.1
urllib3==1.22
virtualenv==15.1.0
wrapt==1.10.11
xenon==0.5.4
```

##### Steps to replicate

1. Copy `Pipfile`
2. `pipenv install --dev`
3. Check for `gitlint` in `pip freeze` and `Pipfile.lock`

techalchemy:
- Fixes #1030

### DIFF
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -771,6 +771,9 @@ def is_vcs(pipfile_entry):
     if hasattr(pipfile_entry, 'keys'):
         return any(key for key in pipfile_entry.keys() if key in VCS_LIST)
     elif isinstance(pipfile_entry, six.string_types):
+        # This syntax is only valid with git
+        if pipfile_entry.startswith('git+git@'):
+            pipfile_entry = pipfile_entry.replace('git+git@', 'git+git://')
         return bool(requirements.requirement.VCS_REGEX.match(pipfile_entry))
     return False
 
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -771,7 +771,7 @@ def is_vcs(pipfile_entry):
     if hasattr(pipfile_entry, 'keys'):
         return any(key for key in pipfile_entry.keys() if key in VCS_LIST)
     elif isinstance(pipfile_entry, six.string_types):
-        return pipfile_entry.startswith(VCS_LIST)
+        return bool(requirements.requirement.VCS_REGEX.match(pipfile_entry))
     return False
 
 
--- a/tests/test_pipenv.py
+++ b/tests/test_pipenv.py
@@ -326,9 +326,13 @@ class TestPipenv:
         with PipenvInstance() as p:
             c = p.pipenv('install git+https://github.com/requests/requests.git#egg=requests')
             assert c.return_code == 0
-            assert 'requests' in p.pipfile['packages']
+            # edge case where normal package starts with VCS name shouldn't be flagged as vcs
+            c = p.pipenv('install gitdb2')
+            assert c.return_code == 0
+            assert all(package in p.pipfile['packages'] for package in ['requests', 'gitdb2'])
             assert 'git' in p.pipfile['packages']['requests']
             assert p.lockfile['default']['requests'] == {"git": "https://github.com/requests/requests.git"}
+            assert 'gitdb2' in p.lockfile['default']
 
     @pytest.mark.e
     @pytest.mark.vcs
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -124,8 +124,10 @@ class TestUtils:
         ({'some_value': 5, 'other_value': object()}, False),
         ('package', False),
         ('git+https://github.com/requests/requests.git#egg=requests', True),
-        ('git+git@github.com:requests/requests.git#egg=requests', True)
+        ('git+git@github.com:requests/requests.git#egg=requests', True),
+        ('gitdb2', False)
     ])
+    @pytest.mark.vcs
     def test_is_vcs(self, entry, expected):
         assert pipenv.utils.is_vcs(entry) is expected
 


techalchemy:
Hm it's weird that the requirements regex parser doesn't pick up `git+git@github.com:requests/requests.git#egg=requests`

erinxocon:
This looks good to me, going to commit to master.  I think this warrants a v8.3.3 release as it's a small but significant fix.  I think @nateprewitt is the only one who can cut a release right now,

erinxocon:
Also @techalchemy thanks for the tests!  It's important we keep adding tests!

nateprewitt:
Erghh, I was in the middle of typing up a review on this. We can't flatten the auth'd URL to a plain git+git:// url because I'm pretty sure that will break SSH auth.

techalchemy:
@nateprewitt I'm not flattening it though

techalchemy:
@nateprewitt I only flattened it for the purpose of the conditional test (since the requirements regex doesn't match on git+git@, only on git+git:// or git+ssh:// and I wasn't sure which was more appropriate)

I'm happy to change the way it gets flattened for comparison, but it doesn't make any changes in the pipfile itself

erinxocon:
That was my belief as well.  Apologies for merging.  It had been laying around for a couple hours and I didn't see anything major in the way the comparison was being made in relation to the original issue, but more experienced eyes...  :)

techalchemy:
yeah I was going to double check, I was like 80% sure it was more appropriate to make it `git+ssh://git@` but I decided that since it wasn't getting written anywhere it didn't matter, pip accepts it without any transformation so it only needs to pass the `VCS_REGEX` check

techalchemy:
@nateprewitt no problem I still think we should do it the right way, I saw your branch and I remembered encountering the approach `packaging` is currently taking to this, using `pyparsing`.

Ultimately we should probably just push a commit back up to `Requirements` because they need to handle pip-accepted syntax IMO.  But we also need to handle it, and it would be better if we did so without making the code confusing to read or making it convey confusing ideas.  Readability is also important.  I'd be willing to sacrifice a microsecond or two for it even 👍 

This was my brief re-hack to stop being dumb about this

https://github.com/techalchemy/pipenv/commit/f22bc12e70a4928f655930b9461bef26ddbe6203

techalchemy:
Note I didn’t PR that since I’m not sure about the import overhead or whether it stays the same in pip 10

techalchemy:
@nateprewitt this is actually how pip handles it, so I'm going to PR this

https://github.com/pypa/pip/blob/28061a71c515a43484d1da9868c33e2984b78fe2/src/pip/_internal/vcs/git.py#L271

```python
    def get_url_rev(self):
        """
        Prefixes stub URLs like 'user@hostname:user/repo.git' with 'ssh://'.
        That's required because although they use SSH they sometimes doesn't
        work with a ssh:// scheme (e.g. Github). But we need a scheme for
        parsing. Hence we remove it again afterwards and return it as a stub.
        """
        if '://' not in self.url:
            assert 'file:' not in self.url
            self.url = self.url.replace('git+', 'git+ssh://')
            url, rev = super(Git, self).get_url_rev()
            url = url.replace('ssh://', '')
        else:
            url, rev = super(Git, self).get_url_rev()

    return url, rev
```

PR #1034, opened 2017-11-05
Fix git url scheme parsing consistent with pip

techalchemy:
- Follow pip methodology for boolean testing
- Be less confusing about what we're doing and why

### DIFF
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -771,9 +771,9 @@ def is_vcs(pipfile_entry):
     if hasattr(pipfile_entry, 'keys'):
         return any(key for key in pipfile_entry.keys() if key in VCS_LIST)
     elif isinstance(pipfile_entry, six.string_types):
-        # This syntax is only valid with git
-        if pipfile_entry.startswith('git+git@'):
-            pipfile_entry = pipfile_entry.replace('git+git@', 'git+git://')
+        # Add scheme for parsing purposes, this is also what pip does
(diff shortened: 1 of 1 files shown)


techalchemy:
@nateprewitt this is actually how pip handles it, so I'm going to PR this

https://github.com/pypa/pip/blob/28061a71c515a43484d1da9868c33e2984b78fe2/src/pip/_internal/vcs/git.py#L271

```python
    def get_url_rev(self):
        """
        Prefixes stub URLs like 'user@hostname:user/repo.git' with 'ssh://'.
        That's required because although they use SSH they sometimes doesn't
        work with a ssh:// scheme (e.g. Github). But we need a scheme for
        parsing. Hence we remove it again afterwards and return it as a stub.
        """
        if '://' not in self.url:
            assert 'file:' not in self.url
            self.url = self.url.replace('git+', 'git+ssh://')
            url, rev = super(Git, self).get_url_rev()
            url = url.replace('ssh://', '')
        else:
            url, rev = super(Git, self).get_url_rev()

    return url, rev
```

techalchemy:
wtf is that bizarre error... also NP, I feel like the code is getting confusing enough as is.  @nateprewitt do you have any recommendations about what you use for profiling if we want to start to take a stab at refactoring but make sure we don't lose too much performance?

nateprewitt:
@techalchemy vmprof is what Kenneth and I have used historically. It does a pretty nice job putting stuff in a flame graph.

Issue #993, opened 2017-10-30
Dependency could not be resolved when I used asterisk.


Kjwon15:
I am using `requests` and `pyrebase` as dependency to my project.
My Pipfile looks like this: (non-relevant are removed)
```
[[source]]

url = "https://pypi.python.org/simple"
verify_ssl = true
name = "pypi"


[packages]

requests = "*"
pyrebase = "*"


[dev-packages]


```

##### Describe you environment

1. OS Type: Ubuntu 17.10 artful
1. Python version: Python 3.6.2
1. Pipenv version: pipenv, version 8.3.1

##### Expected result

Install `requests 2.11.1` and goes well.

##### Actual result

Pipenv installs `requests 2.18.4`(which is latest version) and conflict with `pyrebase`(which requires `2.11.1`)

##### Steps to replicate

`pipenv install requests pyrebase`

erinxocon:
This is how our locking mechanism works.  The pipfile specifies the most recent version of requests, where as pyrebase specifically requests an older version.  Since pyrebase specifically requests an older version of requests, you get a conflict when you are trying to lock while also installing the most current version as it can't match them up.  

If you just install pyrebase, you will have access to requests as it will be in your site packages.  So you don't have to install requests again.  Everything actually installs ok, the way you are doing it, it just can't create a lock file for a deterministic build.  You can use the `--skip-lock` command to ignore this error message if you really need the most recent version of requests installed.

Kjwon15:
So `*` means *most recent* instead of wildcard?

erinxocon:
@Kjwon15 Indeed!

nateprewitt:
This seems like a bug to me. Shouldn't pip-tools be resolving the correct version of `requests` to install alongside `pyrebase`?

Is this a dupe of one of the many "our patched version of pip-tools is broken" tickets?

techalchemy:
@nateprewitt i think it’s an order of operations thing. I think if you specifically ask for requests = “*” it is technically pinning the most recent version of requests and then moving on. This is confusing I agree, considering I believe we advertise the syntax as ‘unpinned’ when converting from requirements.txt. If the intention is to strictly pin the most current version of everything I would imagine this could cause problems for people who think these are not pinned at all and assume we will resolve dependencies and land on versions with no conflicts (which seems preferable)

If you want strict pins that goes in the lockfile, I would think?

erinxocon:
[...]

Issue #875, opened 2017-10-10
Dependencies could not be resolved. Regression?


helderco:
Hey Kenneth, this used to work so I don't know what's happening.

```
$ pipenv --version
pipenv, version 8.2.6

$ cat Pipfile
[[source]]
url = "https://pypi.python.org/simple"
verify_ssl = true

[packages]
django-cms = "*"
django = "*"

$ pipenv lock             
Locking [dev-packages] dependencies…
Locking [packages] dependencies…
Warning: Your dependencies could not be resolved. You likely have a mismatch in your sub-dependencies.
  You can use $ pipenv install --skip-lock to bypass this mechanism, then run $ pipenv graph to inspect the situation.
Could not find a version that matches django<1.11,==1.11.6,>=1.8
[...]

PR #1038, opened 2017-11-06
Dependency resolution - Do not pin candidates on first selection

vphilippon:
Trying to address #875, #993, #1060

What this PR still needs:
- [x] A test to reproduce the issue where pre-pinning candidate causes the resolution to fail.
- [x] A test that covers what the initial patch from Kenneth intended to fix.

Details can be found on this comment: https://github.com/kennethreitz/pipenv/issues/875#issuecomment-337717817

### DIFF
--- a/tests/test_pipenv.py
+++ b/tests/test_pipenv.py
@@ -843,6 +843,29 @@ requests = "*"
             c = p.pipenv('install')
             assert c.return_code == 0
 
+    @pytest.mark.extras
+    @pytest.mark.lock
+    @pytest.mark.requirements
+    @pytest.mark.complex
+    def test_complex_lock_deep_extras(self):
+        # records[pandas] requires tablib[pandas] which requires pandas.
+
+        with PipenvInstance() as p:
+            with open(p.pipfile_path, 'w') as f:
+                contents = """
+[packages]
+records = {extras = ["pandas"], version = "==0.5.2"}
+                """.strip()
+                f.write(contents)
+
+            c = p.pipenv('lock')
+            assert c.return_code == 0
+            assert 'tablib' in p.lockfile['default']
+            assert 'pandas' in p.lockfile['default']
+
+            c = p.pipenv('install')
+            assert c.return_code == 0
+
     @pytest.mark.lock
     @pytest.mark.deploy
     def test_deploy_works(self):
--- a/tests/test_pipenv.py
+++ b/tests/test_pipenv.py
@@ -1,4 +1,5 @@
 import os
+from pkg_resources import parse_version
 import re
 import tempfile
 import shutil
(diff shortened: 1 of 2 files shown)


vphilippon:
Note to self: need to test with `records[pandas]`, it must result in a lock with pandas included, which was the original intended patch.

tests/test_pipenv.py
```diff
@@ -820,6 +820,27 @@ def test_complex_deps_lock_and_install_properly(self):
             c = p.pipenv('install')
             assert c.return_code == 0
 
+    @pytest.mark.lock
+    @pytest.mark.requirements
+    @pytest.mark.complex
+    def test_complex_lock_changing_candidate(self):
+        # The requests candidate will change from latest to <2.12.
```
erinxocon: Do we want to check the version numbers in the lock file to see if it reports <2.12

erinxocon:
@vphilippon Thanks so much for this!  This will be a great fix for deep dependency resolution.  I think the tests are great with one comment.  I'll let @kennethreitz or @nateprewitt take a look as well.  I'm not as familiar with the pip internals, but this does appear to work for me when I checkout this PR and test.

tests/test_pipenv.py
vphilippon: That would be good idea to make sure that we're resolving *and* giving a valid result, indeed!

vphilippon: Updated the test

erinxocon:
@vphilippon Weird the app veyor test for python34 took like 20 minutes...App Veyor can be odd

vphilippon:
@erinxocon Yeah, the flaky CI does worry me a bit, as I rely on it for many tests (I have a bunch of issues locally with pipenv, which means I can't run all the tests locally before pushing)

So the Appveyor tests passed, that's nice. That one Travis job that is failing (Py3.4) bugs me, as I relaunched it, to no avail. I'll give it a 3rd try.

erinxocon:
@vphilippon Huh the 3.4 test on travis still fails.  It looks like all of the tests run ok, but it just stalls.  Do we have any tests that would leave a hanging resource or pointer open?

erinxocon:
@vphilippon It looks like all of the tests pass for python34, but it's hanging, despite all of the tests passing I don't think we can merge until we resolve this.  Any ideas? I'm not seeing something obvious.

vphilippon:
@erinxocon No idea from my side either. I'm not sure what could be hanging the tests. That does need investigation, although I'm not sure when I'll be able to look into this.

nateprewitt:
I bounced it to get a fresh set of tests run. Let's see if this is tied to host state or not.

vphilippon:
IT WORKED!

nateprewitt:
Looking good, thanks for getting this put together @vphilippon!

PR #1041, opened 2017-11-08
Check if virtualenv location really exists

### LINKED ISSUE #969: Pipenv cannot create virtualenv
When I try to create a new env using `pipenv --two` or `pipenv --three`,  pipenv enters an infinite loop repeatedly printing 

```
Virtualenv already exists!
Removing existing virtualenv…
```

If I try `pipenv install`, pipenv tries to install the packages to `/usr/lib/python3.6` and the output of `pipenv --venv` is empty.

1. OS Type: Archlinux
1. Python version: Python 3.6.2
1. Pipenv version:  pipenv, version 8.2.7

techalchemy:
- Check if virtualenv_location exists as an os path as part of the virtualenv_exists check
- Fixes #969

### DIFF
--- a/pipenv/project.py
+++ b/pipenv/project.py
@@ -103,7 +103,7 @@ class Project(object):
     @property
     def virtualenv_exists(self):
         # TODO: Decouple project from existence of Pipfile.
-        if self.pipfile_exists:
+        if self.pipfile_exists and os.path.exists(self.virtualenv_location):
             if os.name == 'nt':
                 extra = ['Scripts', 'activate.bat']
             else:


pipenv/project.py
```diff
@@ -103,7 +103,7 @@ def requirements_exists(self):
     @property
     def virtualenv_exists(self):
         # TODO: Decouple project from existence of Pipfile.
-        if self.pipfile_exists:
+        if self.pipfile_exists and os.path.exists(self.virtualenv_location):
```
nateprewitt: I might be missing something here but what does this catch that the exists call in the return statement doesn't?

jck: This: https://github.com/kennethreitz/pipenv/issues/969#issuecomment-342704601

nateprewitt: Thanks @jck, I see the issue now. This looks sensible to me. I think that we really need to rethink how a lot of the project functions work because they've been pretty mangled together. We need clear cut cases of when we return  `''` vs `None`, and what they mean.

techalchemy: @nateprewitt agreed but also better docstribgs could help for example.  It’s not totally clear to me why we return an empty string here which is why I didn’t understand this problem at all. But this check was safer and easier

PR #1047, opened 2017-11-08
Ensure 'run <command>' arguments are not stolen by 'run' arguments

jonathn:
Without this patch:
`pipenv run myscript -- --foo` is treated as
`pipenv run myscript --foo`

and
`pipenv run myscript --foo --python --debug` is treated as
`pipenv run --python=--debug myscript --foo`

..etc

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -2217,6 +2217,7 @@ def inline_activate_virtualenv():
     short_help="Spawns a command installed into the virtualenv.",
     context_settings=dict(
         ignore_unknown_options=True,
+        allow_interspersed_args=False,
         allow_extra_args=True
     )
 )


nateprewitt:
So this seems like a good idea, and something we discussed adding some time ago. The only concern I have is this will likely break commands that currently function with the "bad" behavior. This may have to sit in the queue till we're ready for a 9.0 release.

techalchemy:
@nateprewitt @jonathn isn't it convention to use `--` to delineate whether arguments get passed to the originating command or a subcommand?

nateprewitt:
@techalchemy the -- forces the command line to treat everything after it as arguments. I've only ever seen it used with unruley cli apps or weird nested command combinations.

I wouldn't consider it normal to have to pass that. If you're referring to the first example, I haven't checked to but I'm assuming he's expecting -- --foo to pass foo to pipenv rather than myscript.

techalchemy:
@nateprewitt I would expect the opposite behavior -- allowing interspersed arguments is likely going to just add more overhead to the CLI and we are already struggling with that (completion takes me a full second to load on my vm).  I'm not really an expert with click but if we subcommand it out properly it should be possible for `run` to not parse any arguments (besides --help, which would be fine if we allow interspersed args then) at all, no?

nateprewitt:
> allowing interspersed arguments is likely going to just add more overhead

> it should be possible for run to not parse any arguments

Yes, that's what this patch is trying to accomplish. This *disables* interspersed arguments.

techalchemy:
@nateprewitt oh ok my bad, I didn't read carefully/assumed this patch was _enabling_ because of a different discussion I saw somewhere else on this topic recently.  Ignore everything I said, then, this is a good change (for 9.x)

nateprewitt:
Thanks for the proactiveness here, @jonathn! This is on the roadmap for 9.0.0, but it may be a little bit before we actually get this functionality merged in. We need to sit down and figure out what breaking changes we have in the pipeline and get them all done at once.

kennethreitz:
Use quotes. `pipenv run 'myscript -- --foo'` should work.

kennethreitz:
I might be wrong...

kennethreitz:
I'm fine with us merging this, if needed.

kennethreitz:
Major release though.

Issue #953, opened 2017-10-23
Simple commands (such as autocompletion and --version) take a relatively long time


PolarPayne:
Evaluating the autocomplete takes a very long time (closer to a second).
`eval "$(env _PIPENV_COMPLETE=source-bash pipenv)"`

Autocompletions also take closer to a second, as does running pretty much any command (no matter how simple).

I had to disable the autocompletion as it adds too much time to my terminals startup (actually I just took the output of the command to my .bashrc for now).

##### Describe you environment

1. Linux (Ubuntu 17.04)
1. Python version: Python 3.5.3
1. Pipenv version: pipenv, version 8.2.7

##### Expected result

These simple commands shouldn't take more than a few milliseconds.

##### Steps to replicate

```bash
$ time env _PIPENV_COMPLETE=source-bash pipenv --verbose

_pipenv_completion() {
    local IFS=$'\t'
    COMPREPLY=( $( env COMP_WORDS="${COMP_WORDS[*]}" \
                   COMP_CWORD=$COMP_CWORD \
                   _PIPENV_COMPLETE=complete-bash $1 ) )
    return 0
}

complete -F _pipenv_completion -o default pipenv

real    0m0,763s
user    0m0,708s
sys     0m0,056s
```

```bash
$ time pipenv --version
pipenv, version 8.2.7

real    0m0,646s
user    0m0,600s
sys     0m0,040s
```

techalchemy:
I don't know much about completion but this seems slow. /cc  @erinxocon @nateprewitt

nateprewitt:
@techalchemy it is slower than it needs to be. I fixed this a few months back and it looks like the big wave of changes in the last month reintroduced it.

It's just that the cli method has become so overloaded it does a lot of checks with every call that aren't necessary in cases like this.

We can fix it again but at best I think we'll be around 400ms just because of click and imports.

techalchemy:
@nateprewitt Is there some way we can pare down imports or pre-compile autocompletion or something like that? Maybe that's just a massive refactor though?

erinxocon:
@nateprewitt @techalchemy I think the cli might need a bit of a refactor, or at least the options each command can accept needs to be looked at.  

@PolarPayne Can I ask what kind of hardware you are running on?

PolarPayne:
I'm running on a new Lenovo X1 Carbon, it has an SSD and is pretty fast (it's not a potato).

erinxocon:
[...]

PR #1069, opened 2017-11-15
Improvements to startup speed

PolarPayne:
See issue #953.

The main issue with slow startup times seems to be that `from .cli import cli` takes quite long, this commit moves these imports to conditional blocks so they're run as late as possible, and not at all if not needed.

I can understand if this "fix" feels too hacky, but I don't know what exactly might be the root cause of the slow startup problems. Once the root cause is found and fixed, it would be great to add some kind of a regression test for it so it doesn't happen again.


PolarPayne:
Realized that this won't actually even work, and shouldn't even be done in this way.

PR #1074, opened 2017-11-16
Fixed the remark about the best shell.

ninjaaron:
Was a little misinformation in the doc. Couldn't let it slide.

_ok, so fish really is the best shell in terms of sanity, except for the fact that it breaks anything that relies on POSIX compat. Also, doesn't doesn't do escape sequences correctly with rsync or sftp, which ZSH does. I'd actually be using it if it did that one thing. ZSH also has a lot of the same benefits for scripting; i.e. arrays have normal-looking syntax for index operations, you don't need to do the special `"${myvar[@]}"` incantation to iterate on them, don't have to quote every effing variable to keep it from splitting on whitespace, recursive globbing (and every other kind of globbing), etc. It's like Fish, but more mature and will run most POSIX shell scripts. Everything the heart wants._

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -110,7 +110,7 @@ Alternatively, with bash, put this in your ``.bashrc`` or ``.bash_profile``::
 
 Magic shell completions are now enabled! There is also a `fish plugin <https://github.com/fisherman/pipenv>`_, which will automatically activate your subshells for you!
 
-Fish is the best shell. You should use it.
+ZSH is the best shell. You should use it.
 
 ☤ Usage
 -------


techalchemy:
Lol. Thanks for the PR 🎆 As a side note you can use bass with fish to run posix-mode things. I switched off of zsh mainly due to low setup overhead and speed... zsh runs significantly slower for me

ninjaaron:
Yeah, I think I have noticed that ZSH is a bit slower, though not enough to be a deal-breaker (which I can't say of xonsh, unfortunately). Just gots to have me that correct escaping for remote filenames with rsync. Maybe it would be more productive to file a bug with fish instead of making pull requests on pipenv... I also remember some screwy behavior about command substitution the last time I used it, but I can't remember exactly what the problem was at the moment. It was in the bugtracker. It's probably fixed now.

techalchemy:
While amusing I did go ahead and close this one, Kenneth is pretty opinionated (and I am a convert too)

PR #1078, opened 2017-11-16
Fix #1054, 'Using <python> to create virtualenv' now outputs to stderr

### LINKED ISSUE #1054: `Using /usr/local/bin/python3.6m to create virtualenv…` is printed to stdout when executing `pipenv lock -r > requirements.txt` if there's no existing virtualenv.
This can be observed when `pipenv lock -r > requirements.txt` is used to generate `requirements.txt` in an environment without an existing virtualenv.

##### My environment
1. OS: CI(Docker Image) `python:3.6`(based on `Debian 9.0 stretch`)
1. Pipenv version: `pipenv, version 8.3.2`

##### Expected result
Only dependencies from `Pipfile` should be printed into `requirements.txt`.

##### Actual result
When the former command is executed under an environment without an existing virtualenv, it not only prints the dependencies but also prints `Using /usr/local/bin/python3.6m to create virtualenv…` into the file.

##### Steps to replicate
As described above.

##### Workarounds or Solutions
The current workaround I found is to execute any `PipEnv` commands prior to the generation of `requirements.txt` as this would let `PipEnv` create the Virtual Environment.

The solution would be to suppress the output.

Moreover, IMO, `PipEnv` should only create virtualenvs when it's needed. `pipenv lock -r` shouldn't create virtualenvs since there's no actual use of the virtualenv under this situation and it consumes a long time and a lot of resources to create a virtualenv and install dependencies.

MarcDufresne:
See #1054 for details.

Basically, using `pipenv lock -r > requirements.txt` in a project that didn't already have a virtualenv would output `Using /path/to/python to create virtualenv...` in stdout. This would cause the `requirements.txt` file to contain the said line.

Example without fix:
```
$ cat requirements.txt
Using /usr/local/bin/python3.6m to create virtualenv…
pandocfilters==1.4.2
snowballstemmer==1.2.1
ipython-genutils==0.2.0
pygments==2.2.0
pytest-asyncio==0.8.0
prompt-toolkit==1.0.15
```

Example with this diff:
```
$ cat requirements.txt
pandocfilters==1.4.2
snowballstemmer==1.2.1
ipython-genutils==0.2.0
pygments==2.2.0
pytest-asyncio==0.8.0
prompt-toolkit==1.0.15
```

(Both examples have been shortened for brevity)

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -937,7 +937,7 @@ def do_create_virtualenv(python=None, site_packages=False):
             crayons.normal('Using', bold=True),
             crayons.red(python, bold=True),
             crayons.normal(u'to create virtualenv…', bold=True)
-        ))
+        ), err=True)
 
     # Use virtualenv's -p python.
     if python:


nateprewitt:
I haven't tested this but do we actually need the virtualenv for the requirements locking?

If not, it would probably be better to simply skip that step than mislabeling the output here.

MarcDufresne:
Pretty sure you don't need the virtualenv to generate the requirements, as they should be sourced from the Pipfile.lock. 
I agree that the better solution would be to *not* generate a virtualenv, but I think this is a deeper issue in `pipenv`. AFAIK *all* commands will first try to generate a virtualenv.

kimond:
According to the documentation, https://docs.pipenv.org/advanced.html#generating-a-requirements-txt nothing about to be or not into a virtualenv is specified. But like @MarcDufresne said I think that all commands will try to generate a virtualenv. 
Should there be one more option to skip the creation of a virtualenv or just avoid to handle it when using `pipenv lock -r`?

nateprewitt:
For some reason we moved back to using ensure_project on every command call. These were originally set up to only call it on the commands that *must* have a virtualenv.

Like I said I haven't tested this but we should only need to move the requirement check in `lock` to above the creation call. Then extend the check in do_init to not run when requirements is true.

Also as a side note, why are we creating the virtualenv in do_init? That seems redundant.

MarcDufresne:
I think we could still merge this PR, moving this specific message to stderr shouldn't be an issue. Pipenv already uses stderr in a lot of places.

I have found many messages that use stderr even without being actual errors, e.g.:
- Locking [dev-packages] dependencies...
- Locking [packages] dependencies...
- Updated Pipfile.lock
- Pipfile.lock not found, creating...

And many others, including package installation and virtualenv creation.
(And I also found actual errors that do not use stderr ¯\\\_(ツ)\_/¯)

MarcDufresne:
Indeed, makes sense. I feel like this is a "good enough" fix for now, but a rework of the virtualenv creation process would be better in the short term.

nateprewitt:
Thanks @MarcDufresne! ✨

MarcDufresne:
Thanks to you!

techalchemy:
Just to chime in I agree with @nateprewitt that we definitely create virtualenvs in far more places than is required, and the code overall requires some cleanup.  On the other hand I have something of a 'does this make something better' philosophy to try and prevent the perfect from being the enemy of the good and I think this does

PR #1082, opened 2017-11-17
Modify fake sys.python_version for special callers

techalchemy:
- Return the real python version for specific function callers that use
it to decide whether to decode/re-encode os.path results into unicode
(which only happens on python2)

Without this patch, the recent hackery causes the following issue when using the latest version of pipenv installed via python3.6 and a virtualenv python of 2.7:

`pipenv install --python=2.7.14 apscheduler`

```console
Adding apscheduler to Pipfile's [packages]…
  File "/home/hawk/.virtualenvs/tempenv-67cf248132d9f/bin/pipenv", line 11, in <module>
  File "/home/hawk/git/pipenv/pipenv/vendor/click/core.py", line 722, in __call__
  File "/home/hawk/git/pipenv/pipenv/vendor/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/home/hawk/git/pipenv/pipenv/vendor/click/core.py", line 1066, in invoke
  File "/home/hawk/git/pipenv/pipenv/vendor/click/core.py", line 895, in invoke
  File "/home/hawk/git/pipenv/pipenv/vendor/click/core.py", line 535, in invoke
  File "/home/hawk/git/pipenv/pipenv/cli.py", line 1950, in install
  File "/home/hawk/git/pipenv/pipenv/cli.py", line 1121, in do_lock
    pre=pre
  File "/home/hawk/git/pipenv/pipenv/utils.py", line 529, in resolve_deps
  File "/home/hawk/git/pipenv/pipenv/utils.py", line 491, in actually_resolve_reps
    resolved_tree.update(resolver.resolve(max_rounds=PIPENV_MAX_ROUNDS))
  File "/home/hawk/git/pipenv/pipenv/patched/piptools/resolver.py", line 102, in resolve
    has_changed, best_matches = self._resolve_one_round()
  File "/home/hawk/git/pipenv/pipenv/patched/piptools/resolver.py", line 200, in _resolve_one_round
    for dep in self._iter_dependencies(best_match):
  File "/home/hawk/git/pipenv/pipenv/patched/piptools/resolver.py", line 296, in _iter_dependencies
    dependencies = self.repository.get_dependencies(ireq)
  File "/home/hawk/git/pipenv/pipenv/patched/piptools/repositories/pypi.py", line 153, in get_dependencies
    result = reqset._prepare_file(self.finder, ireq)
  File "/home/hawk/git/pipenv/pipenv/patched/pip/req/req_set.py", line 639, in _prepare_file
    abstract_dist.prep_for_dist()
  File "/home/hawk/git/pipenv/pipenv/patched/pip/req/req_set.py", line 135, in prep_for_dist
    self.req_to_install.assert_source_matches_version()
  File "/home/hawk/git/pipenv/pipenv/patched/pip/req/req_install.py", line 559, in assert_source_matches_version
    display_path(self.source_dir),
  File "/home/hawk/git/pipenv/pipenv/patched/pip/utils/__init__.py", line 125, in display_path
    path = path.decode(sys.getfilesystemencoding(), 'replace')
AttributeError: 'str' object has no attribute 'decode'
```

The caller function is in `pip.utils.__init__.py`:

```python
def display_path(path):
    """Gives the display value for a given path, making it relative to cwd
    if possible."""
    path = os.path.normcase(os.path.abspath(path))
    if sys.version_info[0] == 2:
        path = path.decode(sys.getfilesystemencoding(), 'replace')
        path = path.encode(sys.getdefaultencoding(), 'replace')
    if path.startswith(os.getcwd() + os.path.sep):
        path = '.' + path[len(os.getcwd()):]
    return path
```

This may not be a best practice but it allows us to add an exception for this particular function call

### DIFF
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -392,10 +392,24 @@ def clean_pkg_version(version):
     return six.u(pep440_version(str(version).replace('==', '')))
 
 
+PatchedPyVersion = namedtuple('PatchedSysVersion', ['major', 'minor', 'micro'])
+
+class PatchedSysVersion(PatchedPyVersion):
+    def __getitem__(self, *args, **kwargs):
+        """Don't fake python versions for pip.utils.display_path, because it uses
+        sys.version_info to determine whether it needs to decode the result of
+        calls to os.path.
+        """
+        caller = sys._getframe().f_back.f_code.co_name
+        if caller == 'display_path':
+            return HackedPythonVersion.BACKUP_VERSION_INFO.__getitem__(*args, **kwargs)
+        return super(PatchedSysVersion, self).__getitem__(*args, **kwargs)
+
+
 class HackedPythonVersion(object):
     """A Beautiful hack, which allows us to tell pip which version of Python we're using."""
 
-    PatchedSysVersion = namedtuple('PatchedSysVersion', ['major', 'minor', 'micro'])
+    BACKUP_VERSION_INFO = sys.version_info
 
     def __init__(self, python_version, python_path):
         self.python_version = python_version
@@ -405,7 +419,9 @@ class HackedPythonVersion(object):
         os.environ['PIP_PYTHON_VERSION'] = str(self.python_version)
         os.environ['PIP_PYTHON_PATH'] = str(self.python_path)
         self.backup_version_info = sys.version_info
-        sys.version_info = self.PatchedSysVersion(*map(int, self.python_version.split('.')))
+        sys.version_info = PatchedSysVersion(
+            *map(int, self.python_version.split('.'))
+        )
 
     def __exit__(self, *args):
         # Restore original Python version information.


techalchemy:
Adding you all since I really dont know how we should address this and I doubt this will be the last issue we have to handle around it

vphilippon:
It looks like the `sys.version_info` hack doesn't even fix the issue in the end. I would suggest to simply revert it for now. I see no reason in keeping it right now.

nateprewitt:
#1088 is up to revert. If the original patch isn't working as intended, I'd prefer we go with that approach rather than complicate the patching further.

If we have consensus on that, any of the other maintainers should go ahead and merge the reversion.

nateprewitt:
I believe this is superseded by #1090 now. If that's correct, I propose we close this out.

PR #1088, opened 2017-11-17
Revert "In HackedPythonVersion, patch sys.version_info"

nateprewitt:
Reverts kennethreitz/pipenv#1076

### DIFF
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -1,5 +1,4 @@
 # -*- coding: utf-8 -*-
-from collections import namedtuple
 import os
 import hashlib
 import tempfile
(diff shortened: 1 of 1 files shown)


pipenv/utils.py
```diff
@@ -529,18 +522,17 @@ def resolve_deps(deps, which, which_pip, project, sources=None, verbose=False, p
             resolved_tree = actually_resolve_reps(deps, index_lookup, markers_lookup, project, sources, verbose, clear, pre)
         except RuntimeError:
             # Don't exit here, like usual.
-            resolved_tree = None
+            pass
 
     # Second (last-resort) attempt:
-    if resolved_tree is None:
-        with HackedPythonVersion(python_version='.'.join([str(s) for s in sys.version_info[:3]]), python_path=backup_python_path):
```
vphilippon: We should keep the '.' and indentation fixes though

nateprewitt: That's fine,  let's get a PR up that pulls out what we need to remove.

PR #1091, opened 2017-11-18
Warn users who try to use pipenv run before installing packages

### LINKED ISSUE #1080: a way to tell that the virtualenv and Pipfile.lock are in sync?
Ran the following

```bash
pipenv run ...
Using /usr/bin/python2.7 to create virtualenv…
⠋Running virtualenv with interpreter /usr/bin/python2.7
New python executable in /home/user/.virtualenvs/...-RVyaMy2x/bin/python2.7
Also creating executable in /home/user/.virtualenvs/...-RVyaMy2x/bin/python
...
ImportError: No module named yaml
```

1. Ubuntu 16.04
1. Python 2.7.12
1. pipenv, version 8.3.2

##### Expected result

Expected one of the following:

1. Automatically install dependencies when `pipenv run ...` is called and all dependencies aren't in the virtualenv
1. Add optional flag to do 1
1. Add command to test if dependencies are installed so that I can run `pipenv install` before running `pipenv run ...` (as it stands now, `pipenv install` is fairly time consuming so I don't want to run it unconditionally) 
1. Modify `pipenv check` to check if all dependencies are installed and correct

I can try to make this change, just wanted to make sure it's something that you'd be okay merging.

##### Actual result

Virtualenv was installed but packages were not.

##### Steps to replicate

Run `pipenv run ...` on a python file with `Pipfile.lock` present, without running `pipenv install` first.

techalchemy:
- Fixes #1080
- Currently, pipenv run will create a virtualenv if one doesnt exist
- However, it will not install packages if it needs to
- This allows us to install packages only if we have a new venv

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -2241,7 +2241,17 @@ def inline_activate_virtualenv():
 @click.option('--three/--two', is_flag=True, default=None, help="Use Python 3/2 when creating virtualenv.")
 @click.option('--python', default=False, nargs=1, help="Specify which version of Python virtualenv should use.")
 def run(command, args, three=None, python=False):
-    # Ensure that virtualenv is available.
+    # Make sure we have created the virtualenv if applicable
+    # This isn't required if we are in CI or are using system python
+    if not project.virtualenv_exists and not PIPENV_USE_SYSTEM and not 'CI' in os.environ:
+        click.echo(
+            u'This project has no virtualenv yet! Use $ {0}, for example, to create one.'.format(
+                crayons.red(u'pipenv install', bold=True)
+            ),
+            err=True
+        )
+        sys.exit(1)
+
     ensure_project(three=three, python=python, validate=False)
 
     load_dot_env()
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -843,7 +843,10 @@ def is_installable_file(path):
         path = urlparse(path['file']).path if 'file' in path else path['path']
     if not isinstance(path, six.string_types) or path == '*':
         return False
-    lookup_path = Path(path)
+    try:
+        lookup_path = Path(path)
+    except OSError:
+        return False
     return lookup_path.is_file() or (lookup_path.is_dir() and
             pip.utils.is_installable_dir(lookup_path.resolve().as_posix()))
 
--- a/tests/test_pipenv.py
+++ b/tests/test_pipenv.py
@@ -598,14 +598,32 @@ requests = {version = "*"}
     @pytest.mark.run
     @pytest.mark.dotenv
     def test_env(self):
-
         with PipenvInstance(pipfile=False) as p:
+            p.pipenv('install')
             with open('.env', 'w') as f:
                 f.write('HELLO=WORLD')
 
             c = p.pipenv('run python -c "import os; print(os.environ[\'HELLO\'])"')
             assert c.return_code == 0
             assert 'WORLD' in c.out
+
+    @pytest.mark.run
+    @pytest.mark.cli
+    def test_run_before_install(self):
+        with temp_environ():
+            # This is designed to work with CI
+            if 'CI' in os.environ:
+                del os.environ['CI']
+            with PipenvInstance() as p:
+                with open(p.pipfile_path, 'w') as f:
+                    contents = """
+[packages]
+requests = "*"
+                    """.strip()
+                    f.write(contents)
+                c = p.pipenv('run python -c "print(\'hello\')"')
+                assert c.return_code == 1
+                assert 'has no virtualenv' in c.err
 
     @pytest.mark.e
     @pytest.mark.install


cs01:
Super fast fix, thanks! How long until this makes its way into PyPI?

nateprewitt:
@cs01 We're unfortunately in a bit of a code freeze until we have everything ready for 9.0.0.

I wanted to get it landed this week but we'll likely look at releasing something after Thanksgiving.

techalchemy:
Wow appveyor is 100% broken... erm...

techalchemy:
This fix definitely works but i dont know if it's a good idea

techalchemy:
These CI builds ran super slow... it’s possible this check is the cause. Don’t merge yet, I’ll just add some logic to bypass in most cases

erinxocon:
@techalchemy It could be that we just have a lot of builds going on right now, with my new branch and PR

techalchemy:
@kennethreitz this is partly why I needed --system available

pipenv/cli.py
```diff
@@ -2242,6 +2242,16 @@ def inline_activate_virtualenv():
...
     # Ensure that virtualenv is available.
+    if not project.virtualenv_exists and not PIPENV_USE_SYSTEM:
+        click.echo(
+            crayons.red(
+                u'This project has no virtualenv yet! Use $ pipenv install, for example, to create one.',
+                bold=True
+            ),
+            err=True
+        )
+        sys.exit(1)
```
cs01: any chance an exception such as `PipenvMissingVirtualenv` could be thrown so it could be caught and handled?

techalchemy: Probably not since this is a CLI application, I'm guessing we will want to stick to exit codes

kennethreitz: `$ pipenv install` should be the same color as literally every other shell example we spit out.

pipenv/utils.py
```diff
@@ -843,6 +843,15 @@ def is_installable_file(path):
         path = urlparse(path['file']).path if 'file' in path else path['path']
     if not isinstance(path, six.string_types) or path == '*':
         return False
+    # If the string starts with a valid specifier operator, test if it is a valid
```
kennethreitz: this seems like this should be in a seperate pull request, i have no idea what this is doing.

pipenv/cli.py
kennethreitz: @cs01 absolutely not.

pipenv/utils.py
techalchemy: it was required to pass appveyor tests in this one.

kennethreitz:
explain what is happening

techalchemy:
@kennethreitz explain what is happening to what

pipenv/cli.py
cs01: Ah yeah that makes sense. I only asked because I want to be able to programatically tell if I can skip running `pipenv install` or not. [This](https://github.com/kennethreitz/pipenv/issues/1080#issuecomment-345867519) would take care of that too.

techalchemy:
If you're asking what was happening with regard to the specifier check, the appveyor builds are [here](https://ci.appveyor.com/project/kennethreitz/pipenv/build/1.0.1796/job/okpqonwpk82gb1si) and I excerpted the error below.  This cropped up whenever the dependency resolution issue got extra visibility recently but I'm not sure what caused it or why. Let me try to remove it since it's not affecting any other prs, and i'll rebase onto master and squash

```console
OSError: [WinError 123] The filename, directory name, or volume label syntax is incorrect: '>=3.3.0,<4'
Command exited with code 1
```

techalchemy:
ok seems like whenever we reverted that stuff it fixed appveyor's issues so we're all good on that end

techalchemy:
Py27 failure is transient: https://ci.appveyor.com/project/techalchemy/pipenv/build/job/0o0fv1wi20e725tv

[Build status](https://ci.appveyor.com/project/techalchemy/pipenv/branch/bugfix/1080-install-packages-on-run-after-create)

kennethreitz:
rebase required

pipenv/cli.py
```diff
@@ -2241,7 +2241,17 @@ def inline_activate_virtualenv():
 @click.option('--three/--two', is_flag=True, default=None, help="Use Python 3/2 when creating virtualenv.")
 @click.option('--python', default=False, nargs=1, help="Specify which version of Python virtualenv should use.")
 def run(command, args, three=None, python=False):
-    # Ensure that virtualenv is available.
+    # Make sure we have created the virtualenv if applicable
+    # This isn't required if we are in CI or are using system python
+    if not project.virtualenv_exists and not PIPENV_USE_SYSTEM and not 'CI' in os.environ:
+        click.echo(
+            u'This project has no virtualenv yet! Use $ {0}, for example, to create one.'.format(
```
kennethreitz: remove "for example"

Issue #1083, opened 2017-11-17
How to install setup.py optional dependencies for a package with extras?


tedmiston:
Possibly related issues: https://github.com/kennethreitz/pipenv/issues/858, https://github.com/kennethreitz/pipenv/issues/870, https://github.com/kennethreitz/pipenv/issues/1014.

I'm attempting to install a package which uses the `extras_require` kwarg in its setup function.

The package is in a repo I created [here](https://github.com/astronomerio/boa).  You can see that the setup.py provides two extras: `dev` and `test`.  I'm attempting to use pipenv to create a virtual env for this package.

In a virtual environment I would run `pip install .[dev]` but running the same in pipenv either doesn't seem to work or I'm doing something wrong.

##### Describe your environment

1. macOS 10.13.1
1. Python version: Python 3.6.3
1. Pipenv version: pipenv, version 8.3.2

##### Expected result

I expected a virtual environment to be created by pipenv.

##### Actual result

```
Installing .[dev]…
Traceback (most recent call last):
  File "/Users/taylor/.pyenv/versions/3.6.3/lib/python3.6/site-packages/pkg_resources/_vendor/packaging/requirements.py", line 90, in __init__
    req = REQUIREMENT.parseString(requirement_string)
  File "/Users/taylor/.pyenv/versions/3.6.3/lib/python3.6/site-packages/pkg_resources/_vendor/pyparsing.py", line 1617, in parseString
    raise exc
  File "/Users/taylor/.pyenv/versions/3.6.3/lib/python3.6/site-packages/pkg_resources/_vendor/pyparsing.py", line 1607, in parseString
    loc, tokens = self._parse( instring, 0 )
  File "/Users/taylor/.pyenv/versions/3.6.3/lib/python3.6/site-packages/pkg_resources/_vendor/pyparsing.py", line 1379, in _parseNoCache
    loc,tokens = self.parseImpl( instring, preloc, doActions )
  File "/Users/taylor/.pyenv/versions/3.6.3/lib/python3.6/site-packages/pkg_resources/_vendor/pyparsing.py", line 3376, in parseImpl
    loc, exprtokens = e._parse( instring, loc, doActions )
  File "/Users/taylor/.pyenv/versions/3.6.3/lib/python3.6/site-packages/pkg_resources/_vendor/pyparsing.py", line 1379, in _parseNoCache
    loc,tokens = self.parseImpl( instring, preloc, doActions )
[...]

Issue #858, opened 2017-10-08
In pipfile `editable install` dependencies was strange.


aodag:
I use setup.py and editable install with extras requires using pipenv.

```
pipenv install -e .[dev,testing]
```

that caused entries to dependency in pipfile.

```
[packages]
"-e .[dev,testing]" = "*"
```

Is that entries correct?

kennethreitz:
this is valid, but not optimal

erinxocon:
@aodag I'm curious if you have tried this in the most recent version of pipenv?

Issue #848, opened 2017-10-07
pienv can't detect environment markers from requirements.txt


aodag:
i have requirements.txt such as below:

```
pyramid ; python_version < '3.0'
```

using pipenv 
```
pipenv install -r requirements.txt --three
```

i got py3 environment and pyramid was installed.
however requirements.txt has environment marker, so pipenv shouldn't install pyramid.

kennethreitz:
this is correct

kennethreitz:
we'd have to look into using the pip parser to do this.

techalchemy:
@kennethreitz @aodag this actually will be fixed in #1095 I believe.

PR #1095, opened 2017-11-19
Properly split and parse requirements lines

### LINKED ISSUE #870: extras_require and length of name > 5 fails.

techalchemy:
- Split out markers and local paths
- Fixes #870 and #1083 and #858 and #848 
- Also splits markers properly using pip's approach
- The requirements parser we use really needs a lot of help

### DIFF
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -277,15 +277,33 @@ def get_requirement(dep):
     remote URIs, and package names, and that we pass only valid requirement strings
     to the requirements parser. Performs necessary modifications to requirements
     object if the user input was a local relative path.
+    
+    :param str dep: A requirement line
+    :returns: :class:`requirements.Requirement` object
     """
     path = None
+    # Split out markers if they are present - similar to how pip does it
+    # See pip.req.req_install.InstallRequirement.from_line
+    if not any(dep.startswith(uri_prefix) for uri_prefix in SCHEME_LIST):
+        marker_sep = ';'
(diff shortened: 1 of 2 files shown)


techalchemy:
This may need the fix I just pushed to #1091 in order for appveyor to pass

kennethreitz:
✨🍰✨

Issue #1087, opened 2017-11-17
.env file with values containing "\n" are wrongly loaded


francbartoli:
If a value of the .env file contains a new line string `\n` then the resulting variable is wrong with the value `\\n`

##### Describe your environment

1. Mac OS X 10.9
1. Python version: 2.7.13
1. Pipenv version: 8.3.2

##### Expected result

.env file
-------
PRIVATE_KEY=-----BEGIN CERTIFICATE-----\nMIIDXTCCAkWgAwIBAgIJAJC1HiIAZAiIMA0GCSqGSIb3Df\nBAYTAkFVMRMwEQYDVQQIDApTb21lLVN0YXRlMSEwHwYDVx\naWRnaXRzIFB0eSBMdGQwHhcNMTExMjMxMDg1OTQ0WhcNMT\nA ....\nMANY LINES LIKE THAT ....\nJjyzfN746vaInA1KxYEeI1Rx5KXY8zIdj6a7hhphpj2E04\nC3Fayua4DRHyZOLmlvQ6tIChY0ClXXuefbmVSDeUHwc8Yu\nB7xxt8BVc69rLeHV15A0qyx77CLSj3tCx2IUXVqRs5mlSb\nvA==\n-----END CERTIFICATE-----

print os.environ['PRIVATE_KEY']
`-----BEGIN CERTIFICATE-----\nMIIDXTCCAkWgAwIBAgIJAJC1HiIAZAiIMA0GCSqGSIb3Df\nBAYTAkFVMRMwEQYDVQQIDApTb21lLVN0YXRlMSEwHwYDVx\naWRnaXRzIFB0eSBMdGQwHhcNMTExMjMxMDg1OTQ0WhcNMT\nA ....\nMANY LINES LIKE THAT ....\nJjyzfN746vaInA1KxYEeI1Rx5KXY8zIdj6a7hhphpj2E04\nC3Fayua4DRHyZOLmlvQ6tIChY0ClXXuefbmVSDeUHwc8Yu\nB7xxt8BVc69rLeHV15A0qyx77CLSj3tCx2IUXVqRs5mlSb\nvA==\n-----END CERTIFICATE-----`

##### Actual result

[...]

PR #1096, opened 2017-11-19
Fixes encoding issue with environment variables from .env files

erinxocon:
Addresses #1087


pipenv/vendor/dotenv/main.py
```diff
@@ -102,13 +102,11 @@ def parse_dotenv(dotenv_path):
...
             k, v = line.split('=', 1)
-
             # Remove any leading and trailing spaces in key, value
-            k, v = k.strip(), v.strip().encode('unicode-escape').decode('ascii')
+            k, v = k.strip(), v.strip()
 
             if len(v) > 0:
                 quoted = v[0] == v[len(v) - 1] in ['"', "'"]
-
                 if quoted:
```
techalchemy: It might be better if we put this logic before the decision to encode, since I think it does make sense to escape things by default. Currently this is written such that it does:

1. Encode using unicode-escape, decode back to ascii
2. If the string is quoted, decode using unicode-escape

It might make more sense to just skip the encode/decode step for quoted variables instead

nateprewitt:
Yeah, this is definitively an issue for dotenv, not pipenv. I would say we encourage the initiator of #1087 to do open a ticket there. We shouldn't be maintaining another package in this repo.

erinxocon:
@nateprewitt @techalchemy I can submit a PR back to dotenv but then we still have to pull the change into our code since it's vendeored.

erinxocon:
@nateprewitt @techalchemy This package doesn't look too maintained.  I am submitting a PR to them, however I have made the change in our copy and moved it to the patched folder per out conversation with @kennethreitz

erinxocon:
Turns out there is an open PR on dotenv that addresses this issue.  I pulled that in and tested it and it seems to work fine.  Waiting on tests..

erinxocon:
https://github.com/theskumar/python-dotenv/pull/32/commits/b6e63fe3400c3a2019e0a582f45ef373bcd6dd2b#diff-c0dd4b817d885690b2e59dffae5fdc98

techalchemy:
hopefully they accept that PR sometime and we can put it back in vendored 📦

PR #1097, opened 2017-11-19
Pass 'allow_global' to resolve_deps

### LINKED ISSUE #1002: Pipenv doesn't resolve --system python correctly.

techalchemy:
- Fixes #1002


techalchemy:
Once I remember how to cause the problem

galuszkak:
@techalchemy I think there is bug here.
Second invocation of resolve_deps doesn't use allow_global this line (L1114):
https://github.com/kennethreitz/pipenv/pull/1097/files#diff-e1e68886e05fef92bb74ed1699e5179eR1114

I think You should add it.

nateprewitt:
Yep, this is the same thing that wasn't addressed in the last attempted patch for this. Good spot @galuszkak, thanks!

techalchemy:
OK, and now we pray to the cli gods.  That was a good catch, and I realized we aren't passing the `--system` flag through or picking it up in a majority of our CLI functions so I added it.  That should be helpful

techalchemy:
@kennethreitz and this is the rest of why i needed `--system`

pipenv/cli.py
```diff
@@ -1764,7 +1766,7 @@ def do_py(system=False):
 @click.option('--dev', '-d', is_flag=True, default=False, help="Install package(s) in [dev-packages].")
 @click.option('--three/--two', is_flag=True, default=None, help="Use Python 3/2 when creating virtualenv.")
 @click.option('--python', default=False, nargs=1, help="Specify which version of Python virtualenv should use.")
-@click.option('--system', is_flag=True, default=False, help="System pip management.")
+@click.option('--system', is_flag=True, default=False, help="System pip management.", envvar='PIPENV_USE_SYSTEM')
```
kennethreitz: we don't use envvar anywhere else -- I don't think it's inherently bad, but i think we should stick to the conventions established.

kennethreitz:
I don't think the title of this pull request encapsulates what's going on here, which is a *lot*. 

What are you trying to accomplish, specifically?

pipenv/cli.py
techalchemy: kk, I wasn't sure so I'll nix it

pipenv/cli.py
```diff
@@ -1970,7 +1972,7 @@ def uninstall(
         system = True
 
     # Ensure that virtualenv is available.
-    ensure_project(three=three, python=python)
+    ensure_project(three=three, python=python, system=system)
```
techalchemy: @kennethreitz are you ok with this change in particular?  this is for uninstall

techalchemy: this essentially just doesn't create a virtualenv

techalchemy:
This doesn't work, because now we still `pipenv install --system` for our CI pipeline, but then we `pipenv run pytest` in order to run the tests.  Since `pipenv install --system` doesnt make a virtualenv we can't use `pipenv run`.  I feel like the easiest solution and the best one for consistency is to keep the api call the same no matter what the user is doing, whether it just acts as a wrapper to a direct call or invokes something inside a virtualenv.

I don't know.  What is the best option here?

nateprewitt:
From a reviewing standpoint, can we get a few changes made

It's difficult to tell what this issue is about now, could we get a brief synopsis of the problem and the proposed solution in the original post?

There are now multiple sytlistic changes in the PR which make telling which changes are relevant and which are cosmetic difficult. Let's spin cleanup off to a new PR if we need to make those changes. We don't have a hard cap at 80 though, so I'm not sure they're needed.

The test also isn't very clear what it's trying to prove. Can we get a docstring on that?

techalchemy:
Ok.  Let me describe the problem.

I am a user.  I am using pipenv with `--system`, inside of a docker container, but my pipenv install is not inside a virtualenv.  I **make a folder directly off my filesystem root**.  For some reason that is required to cause the bug.

When we resolve currently, we create a fake python version using `which('python')`.  Under the circumstances I described, `which('python')` returns `/bin/python` because there is no project yet, there is no virtualenv, and `--system` doesn't get dropped through to the resolver.  That causes an error.

I would like to note that this error doesn't occur in master due to the use of a backup resolver:

```python
 backup_python_path = shellquote(sys.executable)
```

But I don't really feel like that is a justification for leaving what is definitively a bug in the code.

I have now spent a significant amount of time explaining why we should pass a necessary argument to a function that expects it, trying to prove that the issue did exist, create a test for a thing that can't currently be reproduced, and this feels pretty silly at this point so I'll undo all of my changes and if that still isn't sufficient I'll just close the pull request and we can leave the bug in the code.

galuszkak:
>  create a test for a thing that can't currently be reproduced

@techalchemy in original issue I attached how to reproduce it. If You are unable to do so I can create instruction or code that reproduce that issue from ground, and put it in repository. 

Thanks for hard work on that @techalchemy

techalchemy:
@galuszkak no worries I grabbed your instructions and included them in the docstring for the test. It just took me awhile to figure out what exactly caused it to error out, turns out the projects first parent has to be `/` as well. We’ll see if this one passes muster.

kennethreitz:
my inclination here is that when using install --system, lock and other
commands are not useful or needed, and that virtualenv creation is
neccessary for those features (e.g. the current implementation is just
fine).

galuszkak:
@kennethreitz for me that contradicts purpose. So proposed solution with Docker workflow is using --system (per documentation), but when I want to pindown version through docker workflow You are saying it's not necessary. For me it doesn't matter if it's with --system flag or not, I expect that for example `pipenv lock` will work correctly with or without that flag, because people have different workflows, and if they are using docker they don't need virtualenv because they isolate on container level, and they still need Pipfile.lock for their peer workers when they will work on that codebase.

kennethreitz:
I envision a dock or user generating a locked file before sending things to Docker, always

kennethreitz:
In other words this functionality is offered for deployments not for development workflow

galuszkak:
@techalchemy I think this PR needs to be rebased. 

Is this PR waiting still for review of @kennethreitz ?

PR #1098, opened 2017-11-19
Stop excluding files from package list...

### LINKED ISSUE #1079: Editables from pipfile are installed shallow
Dependencies of editables are not included in the lock file, so they do not get installed during a fresh `pipenv install`, this is due to the `no_deps` switch in https://github.com/kennethreitz/pipenv/blob/023696e46993bf55cd2807eae1801f8884a6357a/pipenv/cli.py#L775

A minimal example is demonstrated in the following gist
https://gist.github.com/PaulVisscher/f20d2d56fff66414b85ed77622a3aa66

1. Debian GNU/Linux 8
1. Python version: Python 3.6.3
1. Pipenv version: 3.8.2

##### Expected result
Dependencies of editable packages are added to the lock file and installed

##### Actual result
Editable packages are installed shallow: dependencies of editable packages are not added to the lock file, and thus not installed.
```
pipenv shell
python
>>> import foo
  File "/dummy/foo/foo/__init__.py", line 1, in <module>
    import jsonschema
ModuleNotFoundError: No module named 'jsonschema'
```

##### Steps to replicate
 - Replicate the structure from the linked [gist](https://github.com/kennethreitz/pipenv/blob/023696e46993bf55cd2807eae1801f8884a6357a/pipenv/cli.py#L775) or download the included test folder from the [test-case.zip](https://github.com/kennethreitz/pipenv/files/1480495/test-case.zip).
- Run
```
pipenv install
pipenv shell
````
- Run
```
python
>>> import foo
````

### PR DESCRIPTION
**Pull request by techalchemy**
- Fixes #1079 and #909
- I don't know why I implemented this logic in the first place
- I might be an idiot

### DIFF
--- a/pipenv/project.py
+++ b/pipenv/project.py
@@ -56,23 +56,35 @@ class Project(object):
     def _build_package_list(self, package_section):
         """Returns a list of packages for pip-tools to consume."""
         ps = {}
+        # TODO: Separate the logic for showing packages from the filters for supplying pip-tools
         for k, v in self.parsed_pipfile.get(package_section, {}).items():
             # Skip editable VCS deps.
             if hasattr(v, 'keys'):
-                # When a vcs url is given without editable it only appears as a key
-                if is_vcs(v) or is_vcs(k):
+                # When a vcs url is gven without editable it only appears as a key
+                # Eliminate any vcs, path, or url entries which are not editable
+                # Since pip-tools can't do deep resolution on them
+                # Exempt setuptools-installable directories
+                if (is_vcs(v) or is_vcs(k) or (is_installable_file(k) and os.path.isfile(k)) or
+                        any((prefix in v and
+                             (os.path.isfile(v[prefix]) or is_valid_url(v[prefix])))
+                            for prefix in ['path', 'file'])):
                     # Non-editable VCS entries can't be resolved by piptools
                     if 'editable' not in v:
                         continue
                     else:
                         ps.update({k: v})
                 else:
-                    if not (is_installable_file(k) or is_installable_file(v) or
-                            any(file_prefix in v for file_prefix in ['path', 'file'])):
-                        ps.update({k: v})
+                    ps.update({k: v})
             else:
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: pipenv/project.py [hunk 1]
```diff
@@ -67,13 +67,11 @@ def _build_package_list(self, package_section):
                     else:
                         ps.update({k: v})
                 else:
-                    if not (is_installable_file(k) or is_installable_file(v) or
-                            any(file_prefix in v for file_prefix in ['path', 'file'])):
-                        ps.update({k: v})
+                    ps.update({k: v})
             else:
                 if not (any(is_vcs(i) for i in [k, v]) or
```
nateprewitt: We really need a comment expressing what this is doing and why.

pipenv/project.py
```diff
@@ -67,13 +67,11 @@ def _build_package_list(self, package_section):
...
-                        any(is_installable_file(i) for i in [k, v]) or
-                        any(is_valid_url(i) for i in [k, v])):
```
nateprewitt: Any reasons we're deviating from the use of `i` in the other checks? They should probably all be `entry` or `i`.

techalchemy: Yeah I was contemplating swapping everything to `entry` but I had to re-think all my logic anyway so I just left it as `i` across the board (which is not very descriptive but uh, sometimes it keeps our lines under 80 characters?)

techalchemy:
CI is lying, transient failures.  Things are fine

techalchemy:
Btw appveyor passes on my own account:

[Build status](https://ci.appveyor.com/project/techalchemy/pipenv/branch/bugfix/1079-shallow-editable-installs)

You can fix a lot of transient appveyor failures if you add the following to https://ci.appveyor.com/project/kennethreitz/pipenv/settings/build  => before build script => cmd:

```cmd
rmdir /s /q "c:\users\appveyor\appdata\local\temp\1\tmp*\configparser"
rmdir /s /q "c:\users\appveyor\appdata\local\pip-tools\cache"
```

PR #1109, opened 2017-11-22
Prompt for confirmation before replacing existing virtualenv

darkfeline:
Fixes https://github.com/kennethreitz/pipenv/issues/997

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -544,6 +544,9 @@ def ensure_python(three=None, python=None):
 def ensure_virtualenv(three=None, python=None, site_packages=False):
     """Creates a virtualenv, if one doesn't exist."""
 
+    def abort():
+        sys.exit(1)
+
     global USING_DEFAULT_PYTHON
 
     if not project.virtualenv_exists:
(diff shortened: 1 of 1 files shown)


erinxocon:
@nateprewitt Erg, I understand the need for this, but I don't like it.  I don't want to be prompt everytime I need to recreate a virtualenv, which I do a lot.  However this is probably not the end user experience and this enhancement is good.  I see that we are using `PIPENV_YES` env variable to override this behavior.  I don't want to add another env variable...but...should we separate out saying yes to everything from just saying yes to virtualenv destruction and creation?  I believe the PIPENV_YES flag was implemented mainly for CI so it wouldn't hang on some of the interactive functionality.

pipenv/cli.py
```diff
@@ -580,6 +583,8 @@ def ensure_virtualenv(three=None, python=None, site_packages=False):
         ensure_python(three=three, python=python)
 
         click.echo(crayons.red('Virtualenv already exists!'), err=True)
+        if not (PIPENV_YES or click.confirm('Remove existing virtualenv?', default=False)):
```
nateprewitt: Alright I'm going to propose an addendum here. Rather than prompting every user every time we create a new environment, let's only do this if VIRTUAL_ENV environment variable is set. That will tell us if virtualenv is already active and keep people from overwriting the potentially unrelated environment.

@erinxocon thoughts?

erinxocon: I like this approach, no extra commands or env variables.  

Question: Is there any _common_ scenario where you can somehow manually activate a virtualenv without this variable being set as well?

kennethreitz: @nateprewitt agreed

kennethreitz: @erinxocon no

erinxocon:
After making the changes @nateprewitt requested I'm good with merging this!

kennethreitz:
This (original function) never should have shipped — sorry for the oversight.

kennethreitz:
Looking forward to this fix being shipped. Let's try to get this out ASAP.

PR #1112, opened 2017-11-23
Use the pipenv cache path in the patched pip-tools

vphilippon:
This will avoid having `pipenv` and the "real" `pip-tools` stepping on each other.

I encountered some issues while working on both projects due to that overlap, so lets fix this.


erinxocon:
I think this can be merged!  Thanks @vphilippon!

PR #1114, opened 2017-11-23
Do sequential installs on editable dependencies

### LINKED ISSUE #1111: easy-install.pth contains incomplete number of paths for editable = true packages.

1. OS Type: Debian Stable 9.2
1. Python version: Python 2.7.13
1. Pipenv version: pipenv, version 8.3.2

##### Expected result

When running pipenv install with a Pipfile containing 15 local editable packages:

```
pyramid-mustache = {path = "vendor/pyramid-mustache/", editable = true}
Dalton = {path = "vendor/Dalton/", editable = true}
pyelasticsearch = {path = "vendor/pyelasticsearch/", editable = true}
deus_lib = {path = "deus_lib/", editable = true}
lexdb = {path = "lexdb/", editable = true}
bias = {path = "bias/", editable = true}
winslow = {path = "winslow/", editable = true}
usertool = {path = "usertool/", editable = true}
analytics_services = {path = "analytics_services/", editable = true}
citadel = {path = "citadel/", editable = true}
lex-nlp-scripts = {path = "lexnlp-product/core/scripts/", editable = true}
dashboarding = {path = "dashboarding/", editable = true}
east = {path = "east/", editable = true}
webhook_handler = {path = "webhook_handler/", editable = true}
qatool = {path = "qatool/", editable = true}
```

I expect 15 entries in `easy-install.pth`. 

##### Actual result

I get 13, 11, etc, and each time you run it you get a different set of local pth entries.

No changes to Pipfile or to Pipfile.lock, just rerunning pipenv install

Examples:
```
/home/gavin/LexMachina/src/deus_lex/vendor/Dalton
/home/gavin/LexMachina/src/deus_lex/east
/home/gavin/LexMachina/src/deus_lex/vendor/pyramid-mustache
/home/gavin/LexMachina/src/deus_lex/citadel
/home/gavin/LexMachina/src/deus_lex/vendor/pyelasticsearch
/home/gavin/LexMachina/src/deus_lex/webhook_handler
/home/gavin/LexMachina/src/deus_lex/usertool
/home/gavin/LexMachina/src/deus_lex/analytics_services
/home/gavin/LexMachina/src/deus_lex/lexdb
/home/gavin/LexMachina/src/deus_lex/deus_lib
/home/gavin/LexMachina/src/deus_lex/winslow
/home/gavin/LexMachina/src/deus_lex/lexnlp-product/core/scripts
/home/gavin/LexMachina/src/deus_lex/qatool
```

```
/home/gavin/LexMachina/src/deus_lex/webhook_handler
/home/gavin/LexMachina/src/deus_lex/vendor/Dalton
/home/gavin/LexMachina/src/deus_lex/east
/home/gavin/LexMachina/src/deus_lex/dashboarding
/home/gavin/LexMachina/src/deus_lex/vendor/pyramid-mustache
/home/gavin/LexMachina/src/deus_lex/citadel
/home/gavin/LexMachina/src/deus_lex/bias
/home/gavin/LexMachina/src/deus_lex/usertool
/home/gavin/LexMachina/src/deus_lex/lexdb
/home/gavin/LexMachina/src/deus_lex/analytics_services
/home/gavin/LexMachina/src/deus_lex/winslow
/home/gavin/LexMachina/src/deus_lex/lexnlp-product/core/scripts
/home/gavin/LexMachina/src/deus_lex/qatool
/home/gavin/LexMachina/src/deus_lex/vendor/pyelasticsearch
```
[OUTPUT.txt](https://github.com/kennethreitz/pipenv/files/1497769/OUTPUT.txt)

##### Steps to replicate

Sadly I have no idea how to make a smaller test case as I have 0 idea what is going on. Happy to try and create one if someone can give me some hint as to how the heck this is even possible.

techalchemy:
- Fixes #1111 
- Uses existing infrastructure to add editable dependencies to the list
of sequentially installed packages
- Only alters functions used for this purpose
- Essentially there is a race condition because multiple editors are
trying to write to lib/site-packages/easy-install.pth for local editable
packages, so they overwrite one another

### DIFF
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -919,24 +919,100 @@ def proper_case(package_name):
     return good_name
 
 
-def split_vcs_editable(split_file):
-    """Split VCS and editable dependencies out from file."""
+def split_section(input_file, section_suffix, test_function):
+    """
+    Split a pipfile or a lockfile section out by section name and test function
+    
+        :param dict input_file: A dictionary containing either a pipfile or lockfile
+        :param str section_suffix: A string of the name of the section
+        :param func test_function: A test function to test against the value in the key/value pair
+    
+    >>> split_section(my_lockfile, 'vcs', is_vcs)
+    {
+        'default': {
+            "six": {
+                "hashes": [
+                    "sha256:832dc0e10feb1aa2c68dcc57dbb658f1c7e65b9b61af69048abc87a2db00a0eb",
+                    "sha256:70e8a77beed4562e7f14fe23a786b54f6296e34344c23bc42f07b15018ff98e9"
+                ],
+                "version": "==1.11.0"
+            }
+        },
+        'default-vcs': {
+            "e1839a8": {
+                "editable": true,
+                "path": "."
+            }
+        }
+    }
+    """
+    pipfile_sections = ('packages', 'dev-packages')
+    lockfile_sections = ('default', 'develop')
+    if any(section in input_file for section in pipfile_sections):
+        sections = pipfile_sections
+    elif any(section in input_file for section in lockfile_sections):
+        sections = lockfile_sections
+    # return the original file if we can't find any pipfile or lockfile sections
+    else:
+        return input_file
 
-    if 'packages' in split_file or 'dev-packages' in split_file:
-        sections = ('packages', 'dev-packages')
-    elif 'default' in split_file or 'develop' in split_file:
-        sections = ('default', 'develop')
-
-    # For each vcs or editable entry in a given section, move it to section-editable.
     for section in sections:
-        entries = split_file.get(section, {})
-        editable_dict = {}
+        split_dict = {}
+        entries = input_file.get(section, {})
         for k in list(entries.keys()):
-            if is_vcs(entries[k]) or (hasattr(entries[k], 'keys') and entries[k].get('editable') is True):
-                editable_dict[k] = entries.pop(k)
-        split_file[section + '-editable'] = editable_dict
+            if test_function(entries.get(k)):
+                split_dict[k] = entries.pop(k)
+        input_file['-'.join([section, section_suffix])] = split_dict
+    return input_file
 
-    return split_file
+
+def split_file(file_dict):
+    """Split VCS and editable dependencies out from file."""
+    sections = {
+        'vcs': is_vcs,
+        'editable': lambda x: hasattr(x, 'keys') and x.get('editable')
+    }
+    for k, func in sections.items():
+        file_dict = split_section(file_dict, k, func)
+    return file_dict
+
+
+def merge_deps(file_dict, project, dev=False, requirements=False, ignore_hashes=False, blocking=False, only=False):
+    """
+    Given a file_dict, merges dependencies and converts them to pip dependency lists.
+        :param dict file_dict: The result of calling :func:`pipenv.utils.split_file`
+        :param :class:`pipenv.project.Project` project: Pipenv project
+        :param bool dev=False: Flag indicating whether dev dependencies are to be installed 
+        :param bool requirements=False: Flag indicating whether to use a requirements file
+        :param bool ignore_hashes=False:
+        :param bool blocking=False:
+        :param bool only=False:
+        :return: Pip-converted 3-tuples of [deps, requirements_deps]
+    """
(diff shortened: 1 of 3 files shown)


erinxocon:
@techalchemy Looks like we've got failing tests from this change!

techalchemy:
Yah forgot to change the function name in the test

techalchemy:
This appears permanently queued in appveyor.  It passes on my appveyor however:

[Build status](https://ci.appveyor.com/project/techalchemy/pipenv/branch/bugfix/1111-sequential-editable-installs)

gcarothers:
I can confirm this fixes #1111 on the original Pipfile and using a simple `pipenv install` as expected `easy-install.pth` contains all 15 entries.

nateprewitt:
Thanks for checking in on that @gcarothers!

I wrote up a response for this the other day and apparently didn't post it. To briefly summarize my thoughts, I think this solves the issue but from an architectural perspective, I'd really like to avoid this overloading what this function does. Our recent trend of repurposing functions to do multiple things, not because they're conceptually related, but because they happen to occur at the same place in the code is creating a lot of unnecessary coupling.

Something slightly cleaner might be to create something that does a general division/split operation and then implement precedence for how things are divided with ordered function calls. That'll allow us to scale a little easier in the future if we need more of these.

gcarothers:
I mean architecturally the fact that all locally referenced packages still have to go in the same easy-install.pth seems like an artifact of a bad old time ;) In the meantime without this fix it's not really possible for any project with N local packages to use pipenv?

nateprewitt:
@gcarthors it's definitely possible, you just need to pass --sequential with your install. We're just fixing bad defaults here causing a race condition with local package heavy installs.

techalchemy:
@nateprewitt I refactored just this one thing for the moment, for readability, and split functionality.  There are other places where we can use this type of refactoring but one thing at a time

pipenv/utils.py
```diff
@@ -917,21 +917,100 @@ def proper_case(package_name):
...
-    # For each vcs entry in a given section, move it to section-vcs.
     for section in sections:
-        entries = split_file.get(section, {})
-        vcs_dict = dict((k, entries.pop(k)) for k in list(entries.keys()) if is_vcs(entries[k]))
-        split_file[section + '-vcs'] = vcs_dict
+        split_dict = {}
+        entries = input_file.get(section, {})
+        for k in list(entries.keys()):
+            if test_function(entries.get(k)):
+                split_dict[k] = entries.pop(k)
```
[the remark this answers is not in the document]
techalchemy: It doesn't, although my initial pass at it used `for k, v in entries.items()` which did cause a runtime error.  This mimics the prior pattern of iterating over the list of keys rather than the dictionary itself, which (should?) allow us to manipulate it as we iterate.  That's how we handled it in the past and it _was_ working ok, tests were passing for me, but that doesn't prove it won't break things.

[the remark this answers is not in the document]
nateprewitt: You're right, copying it into the new list does resolve the issue.

pipenv/utils.py
```diff
@@ -950,11 +950,12 @@ def split_section(input_file, section_suffix, test_function):
     lockfile_sections = ('default', 'develop')
     if any(section in input_file for section in pipfile_sections):
         sections = pipfile_sections
-    elif any(section in input_file for section in lockfile_sections):
-        sections = lockfile_sections
-    # return the original file if we can't find any pipfile or lockfile sections
     else:
-        return input_file
+        if any(section in input_file for section in lockfile_sections):
```
nateprewitt: Why did this need to be moved?

techalchemy: Hm I think I misunderstood what you were asking for in your review where you asked me to move this into the else statement for readability. To clarify -- did you mean that you prefer to test something like `if not any(section in input_file for section in pipfile_sections+lockfile_sections):` and do the alternative checks afterward?

nateprewitt: Sorry, I was simply asking that you move the comment into the else statement rather than having it out of alignment. There should need be any code changes.

techalchemy: kk sorry, I thought the request was odd but it seemed simple enough so i just did (what i thought was) it

PR #1127, opened 2017-11-27
Fix #1124

### LINKED ISSUE #1124: Too few arguments for format string?
https://github.com/kennethreitz/pipenv/blob/33d739322865bbc2305068009ebd9fb2cfef9229/pipenv/cli.py#L223

`which_pip()` seems to only return a single string but `str.format` seems to be expecting two separate positional arguments.

```python
>>> '{0} {1}'.format('oops!')
IndexError: tuple index out of range
```

I am guessing that `windows` is supposed to be that second argument and if that is the case, the closing bracket should simply be moved to include it.

```python
# old
c = delegator.run('"{0}" install {1} pip --upgrade'.format(which_pip()), windows, block=False) 

# new?
c = delegator.run('"{0}" install {1} pip --upgrade'.format(which_pip(), windows), block=False) 
```

I wasn't 100% sure that this was the correct solution though but if it is, I could also send a PR to fix it.

Thanks!

delirious-lettuce:
This PR fixes #1124 .

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -220,7 +220,7 @@ def ensure_latest_pip():
 
             windows = '-m' if os.name == 'nt' else ''
 
-            c = delegator.run('"{0}" install {1} pip --upgrade'.format(which_pip()), windows, block=False)
+            c = delegator.run('"{0}" install {1} pip --upgrade'.format(which_pip(), windows), block=False)
             click.echo(crayons.blue(c.out))
     except AttributeError:
         pass


nateprewitt:
Thanks again for catching this, @delirious-lettuce! Once the tests finish rerunning we'll go ahead and merge this.

delirious-lettuce:
@nateprewitt ,

Sounds good to me! Glad I could help (even if it was just moving a parens)!

PR #1128, opened 2017-11-27
Only strip VCS prefix for URIs with vcs+ patterns

### LINKED ISSUE #1126: Pipenv fails to install from github by ref and tag
Trying to install from github using ref and tag fails, because pipenv "destroys" the format of the url removing `git:` from the reference.

1. OS Type OSX 10.12.6
1. Python version: `$ python -V` 3.6.3
1. Pipenv version: `$ pipenv --version` 8.3.2

##### Expected result

The reference in `Pipfile` looks like this (which is correct)
`pytest-tags = {git = "git://github.com/Projectplace/pytest-tags.git", editable = true, ref = "1.5"}`

But in the `Pipfile.lock` file, it's
`"git": "//github.com/Projectplace/pytest-tags.git"`

Workaround is to re-add `git:` to the pipfile.lock manually.

##### Actual result

```Pipfile.lock not found, creating…
Updated Pipfile.lock (24ab2e)!
Installing dependencies from Pipfile.lock (24ab2e)…
  🐍   ▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉ 19/19 — 00:00:05
An error occurred while installing -e git+//github.com/Projectplace/pytest-tags.git@c8cf057c03a1fa30769c0328efa41d4826e03bcf#egg=pytest-tags! Will try again.
Obtaining pytest-tags from git+//github.com/Projectplace/pytest-tags.git@c8cf057c03a1fa30769c0328efa41d4826e03bcf#egg=pytest-tags
  Cloning github.com/Projectplace/pytest-tags.git (to c8cf057c03a1fa30769c0328efa41d4826e03bcf) to /Users/jimbrannlund/.local/share/virtualenvs/dpo-automation-NuB7OARD/src/pytest-tags
fatal: repository 'github.com/Projectplace/pytest-tags.git' does not exist

Command "git clone -q github.com/Projectplace/pytest-tags.git /Users/jimbrannlund/.local/share/virtualenvs/dpo-automation-NuB7OARD/src/pytest-tags" failed with error code 128 in None```

### PR DESCRIPTION
**Pull request by techalchemy**
- Fixes #1126
- Does a simple check before stripping the VCS prefix on
a `requirement.uri` in `pipenv.utils.convert_deps_from_pip`

### DIFF
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -45,7 +45,7 @@ specifiers = [k for k in lookup.keys()]
 
 # List of version control systems we support.
 VCS_LIST = ('git', 'svn', 'hg', 'bzr')
-SCHEME_LIST = ('http://', 'https://', 'ftp://', 'file://', 'git://')
+SCHEME_LIST = ('http://', 'https://', 'ftp://', 'file://')
 
 requests = requests.Session()
 
@@ -661,10 +661,10 @@ def convert_deps_from_pip(dep):
             req.path = None
 
         # Crop off the git+, etc part.
-        if '+' in req.uri:
+        if req.uri.startswith('{0}+'.format(req.vcs)):
             req.uri = req.uri[len(req.vcs) + 1:]
         dependency.setdefault(req.name, {}).update({req.vcs: req.uri})
-        print('dependency: {}'.format(dependency))
+
         # Add --editable, if it's there.
         if req.editable:
             dependency[req.name].update({'editable': True})
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -45,7 +45,7 @@ specifiers = [k for k in lookup.keys()]
 
 # List of version control systems we support.
 VCS_LIST = ('git', 'svn', 'hg', 'bzr')
-SCHEME_LIST = ('http://', 'https://', 'ftp://', 'file://')
+SCHEME_LIST = ('http://', 'https://', 'ftp://', 'file://', 'git://')
 
 requests = requests.Session()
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: pipenv/utils.py [hunk 1]
```diff
@@ -45,7 +45,7 @@
 
 # List of version control systems we support.
 VCS_LIST = ('git', 'svn', 'hg', 'bzr')
-SCHEME_LIST = ('http://', 'https://', 'ftp://', 'file://')
```
nateprewitt: Why is this getting added to the scheme list?

pipenv/utils.py
```diff
@@ -661,8 +661,10 @@ def convert_deps_from_pip(dep):
             req.path = None
 
         # Crop off the git+, etc part.
-        dependency.setdefault(req.name, {}).update({req.vcs: req.uri[len(req.vcs) + 1:]})
-
+        if '+' in req.uri:
```
nateprewitt: So I don't know if the `+` check by itself will be sufficient. That's a legal character in the path as well. We need to determine if there's a + explicitly in the scheme.

pipenv/utils.py
```diff
@@ -661,8 +661,10 @@ def convert_deps_from_pip(dep):
...
+            req.uri = req.uri[len(req.vcs) + 1:]
+        dependency.setdefault(req.name, {}).update({req.vcs: req.uri})
+        print('dependency: {}'.format(dependency))
```
nateprewitt: Let's remove this.

techalchemy: ooh my bad

pipenv/utils.py
techalchemy: yeah good call

pipenv/utils.py
techalchemy: erm, it shouldn't be necessary, sorry for the sloppy PR that was my early hope that just adding it to the scheme list would pick it up in the logic

PR #1141, opened 2017-12-01
Added mimesis to the list of packages. 

lk-geimfari:
Added `mimesis`  to suggestions in interactive mode. 

[Mimesis](https://github.com/lk-geimfari/mimesis) is a fast and easy tool for generating mock data.


nateprewitt:
Thanks for the pull request @lk-geimfari! This list is generated from a query returning the top 1000 most used packages. That will likely remain the way the values here are generated, otherwise we'll be spending a lot of time merging commits for new packages that would like to be added.

For that reason, I think we'll avoid doing manual updates but thanks again for the initiative!

PR #1165, opened 2017-12-05
Update README.rst for a dead URL

triplepoint:
Replace dead link to "python packaging" documentation in the first line of the documentation.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -21,7 +21,7 @@ Pipenv: Python Development Workflow for Humans
 
 ---------------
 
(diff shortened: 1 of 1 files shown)


triplepoint:
Though maybe this was the preferred link? https://packaging.python.org/tutorials/managing-dependencies/#managing-dependencies

nateprewitt:
@triplepoint let's use the second link because I believe that's what we merged elsewhere. Thanks for catching the miss here.

nateprewitt:
LGTM, thanks again!

PR #1167, opened 2017-12-05
fix windows os_name marker in doc advanced.rst

jwaschkau:
See https://docs.python.org/dev/library/platform.html#module-platform and https://docs.python.org/2/library/os.html#os.name


jwaschkau:
Hi@kennethreitz,
why did this merge request get closed. The current documentation is incorrect:
grafik

PR #1178, opened 2017-12-07
Improve test & code clarity around requirements generation

idoru:
- Test now checks for dev dependency "pytest==3.1.1" rather than the
characters of the string
- Test verifies that dev dependencies are excluded by default
- Corrected intent of returned variables from merge_deps call in cli.py


techalchemy:
@idoru thanks for the first time contribution.  This actually looks really good.  I thought I had concerns but upon more careful consideration I think this is good to go.

@nateprewitt this doesn't alter any behavior but does provide some cleanup

kennethreitz:
✨🍰✨

PR #1183, opened 2017-12-09
WIP: Stop Treating Specifiers as Paths on Windows

techalchemy:
- For some reason pipenv treats pipfile entries with leading specifiers as paths on windows
- Might want to add a fix for 9.0.1

- [x] Add failing test
- [x] Fix path checking for entries with leading specifiers

[Build status](https://ci.appveyor.com/project/techalchemy/pipenv/branch/bugfix/windows-specifier-check)

### DIFF
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -869,6 +869,16 @@ def is_installable_file(path):
         path = urlparse(path['file']).path if 'file' in path else path['path']
     if not isinstance(path, six.string_types) or path == '*':
         return False
+    # If the string starts with a valid specifier operator, test if it is a valid
+    # specifier set before making a path object (to avoid breakng windows)
+    if any(path.startswith(spec) for spec in '!=<>'):
+        try:
+            pip.utils.packaging.specifiers.SpecifierSet(path)
+        # If this is not a valid specifier, just move on and try it as a path
+        except pip.utils.packaging.specifiers.InvalidSpecifier:
+            pass
+        else:
+            return False
     lookup_path = Path(path)
     return lookup_path.is_file() or (lookup_path.is_dir() and
             pip.utils.is_installable_dir(lookup_path.resolve().as_posix()))
--- a/tests/test_pipenv.py
+++ b/tests/test_pipenv.py
@@ -434,7 +434,7 @@ setup(
 tablib = "<0.12"
                 """.strip()
                 f.write(contents)
-            c = p.pipenv(install)
+            c = p.pipenv('install')
             assert c.return_code == 0
             assert 'tablib' in p.pipfile['packages']
             assert 'tablib' in p.lockfile['default']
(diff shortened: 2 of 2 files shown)


pipenv/utils.py
```diff
@@ -869,6 +869,16 @@ def is_installable_file(path):
         path = urlparse(path['file']).path if 'file' in path else path['path']
     if not isinstance(path, six.string_types) or path == '*':
         return False
+    # If the string starts with a valid specifier operator, test if it is a valid
+    # specifier set before making a path object (to avoid breakng windows)
+    if any(path.startswith(spec) for spec in '!=<>'):
```
therefromhere: I've not tested this, but shouldn't "~" also be in this check?  as per https://github.com/pypa/pipenv/blob/master/pipenv/utils.py#L689 ,  https://github.com/pypa/pipenv/blob/master/pipenv/vendor/pipreqs/pipreqs.py#L248

nateprewitt: Good spot @therefromhere. Yes, it should.

therefromhere: added #1186 for this

techalchemy: @nateprewitt @therefromhere I strongly disagree.  `~` is a valid path value and there is no reason to try to do a specifier check on it, especially since entries in the pipfile like this on windows are not broken in Pipenv as is.  All this will do is slow down the installation by doing extra things to valid, currently fully functional entries.

techalchemy: See https://github.com/pypa/pipenv/pull/1186#issuecomment-350517997 for more info

nateprewitt: @techalchemy I don't think I'm following your reasoning. We want this to be an incomplete definition of what we're trying to check because it happens to work now? What's keeping that from changing if we move to another path library? We're talking about an additional 20µs per call that has an upper bound of maybe 20? 50?

For all of the things pipenv has going against it for speed, this is would not be one of them.

techalchemy: @nateprewitt when I excluded it I was specifically concerned about having actual paths with leading '~' being succesfully parsed as specifier sets, which would therefore return false.  I can't currently think of an example case where this would occur, however, after testing a few examples, so perhaps we should include the change from #1186 after all and see if it breaks anything

nateprewitt: In the context I believe you're referencing, `~` as a home specifier would always be followed by a separator if the path was longer than `~`. If it's not longer than `~`, it will never evaluate to a specifier because it the specifier is `~=`.

The only case we *could* miss here is a file in the top level directory of the project who's name is `~=something`. That's not impossible but pretty far fetched and within the realm of something I'm comfortable not supporting.

techalchemy: No I agree with you I definitely am convinced, it's correct to add it to the list.  I was wrong to exclude it!

PR #624, opened 2017-09-21
run ensure_project in cli do_shell method

AlJohri:
fixes #622

Issue #324, opened 2017-04-20
Cannot load zsh auto-completion


Zhomart:
If I tried what docs suggest it doesn't work at all:

```
$ cat ~/.zshrc
 eval (env _PIPENV_COMPLETE=source-zsh pipenv)
```

Then I tried following:

```
$ cat ~/.zshrc
 eval "$(env _PIPENV_COMPLETE=source-zsh pipenv)"
```

Now I get following error:

```
zsh compinit: insecure directories, run compaudit for list.
Ignore insecure directories and continue [y] or abort compinit [n]?
```

nateprewitt:
Hey @Zhomart, this was a feature implemented by @kennethreitz and I'm not entirely sure how it was originally supposed to function. He may have some insight on this.

kennethreitz:
the only documented (and tested) way of doing this is for fish, currently.

kennethreitz:
I'd look here for guidance: https://github.com/click-contrib/click-completion

nateprewitt:
I'm going to close this out since there doesn't seem to be anything further actionable here. Please feel free to let us know if you have any more information, questions, or reproducible examples, @Zhomart.

PR #1185, opened 2017-12-09
use current directory if project directory doesn't exist

nateprewitt:
#624 was partially reverted in ecbb386 without a solution to the closed issue. Ideally, we would have `project.project_directory` return the current directory if a `Pipfile` doesn't exist. I don't know what the implications of that will be though, and we're likely to introduce follow on issues mucking with it right now.

This should get us out the door for 9.0.1 without losing what #624 was trying to accomplish. I'd like a second pair of eyes to green light this before we merge.

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -123,7 +123,10 @@ project = Project()
 
 def load_dot_env():
     if not PIPENV_DONT_LOAD_ENV:
-        denv = dotenv.find_dotenv(PIPENV_DOTENV_LOCATION or os.sep.join([project.project_directory, '.env']))
+        # If the project doesn't exist yet, check current directory for a .env file
+        project_directory = project.project_directory or '.'
+
+        denv = dotenv.find_dotenv(PIPENV_DOTENV_LOCATION or os.sep.join([project_directory, '.env']))
         if os.path.isfile(denv):
             click.echo(crayons.normal('Loading .env environment variables…', bold=True), err=True)
         dotenv.load_dotenv(denv, override=True)
@@ -2056,6 +2059,9 @@ def lock(three=None, python=False, verbose=False, requirements=False, dev=False,
 
 def do_shell(three=None, python=False, fancy=False, shell_args=None):
 
+    # Ensure that virtualenv is available.
+    ensure_project(three=three, python=python, validate=False)
+
     # Set an environment variable, so we know we're in the environment.
     os.environ['PIPENV_ACTIVE'] = '1'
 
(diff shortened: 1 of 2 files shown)


pipenv/cli.py
```diff
@@ -2056,6 +2059,9 @@ def lock(three=None, python=False, verbose=False, requirements=False, dev=False,
 
 def do_shell(three=None, python=False, fancy=False, shell_args=None):
 
+    # Ensure that virtualenv is available.
+    ensure_project(three=three, python=python, validate=False)
+
```
techalchemy: Doesn't this still always create a virtualenv though?

pipenv/cli.py
```diff
@@ -123,7 +123,10 @@ def spinner():
 
 def load_dot_env():
     if not PIPENV_DONT_LOAD_ENV:
-        denv = dotenv.find_dotenv(PIPENV_DOTENV_LOCATION or os.sep.join([project.project_directory, '.env']))
+        # If the project doesn't exist yet, check current directory for a .env file
+        project_directory = project.project_directory or '.'
+
+        denv = dotenv.find_dotenv(PIPENV_DOTENV_LOCATION or os.sep.join([project_directory, '.env']))
         if os.path.isfile(denv):
```
techalchemy: I guess my question is how does this impact the shell command?

nateprewitt: We don't raise an exception because we can't join on `None`.

pipenv/cli.py
nateprewitt: This is the state the codebase was in prior to #324. Kenneth added this 3 months ago, so I don't think we should completely remove it unless we know why we're doing it.

techalchemy: I agree and this looks sensible to me but I think it will still cause the issue which caused him to comment it out in the first place, or if it doesn't I'm not seeing why not

nateprewitt: I'm not entirely sure it doesn't. However, that line has been in either `shell` or `do_shell` since 7.0.0 and I'm hesitant to remove it without understanding why it was there and we're now suddenly having an issue with it.

We should probably file a follow up issue with the specific case and address that separately though.

techalchemy:
@nateprewitt I think we can merge this if @kennethreitz will accept a PR on his fish plugin.  The issue he is running into is that pipenv now prints to stderr when we fail to find a virtualenv but the fish plugin only redirects stdout.  I'll submit a PR to fix it.

nateprewitt:
With fisherman/pipenv#8 merged, I'm merging this.

PR #1186, opened 2017-12-10
missing support for ~= in check? - see comment in #1183

therefromhere:
see https://github.com/pypa/pipenv/pull/1183/files#r155932921

also fixed comment typo

### DIFF
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -870,8 +870,8 @@ def is_installable_file(path):
     if not isinstance(path, six.string_types) or path == '*':
         return False
     # If the string starts with a valid specifier operator, test if it is a valid
-    # specifier set before making a path object (to avoid breakng windows)
-    if any(path.startswith(spec) for spec in '!=<>'):
+    # specifier set before making a path object (to avoid breaking windows)
+    if any(path.startswith(spec) for spec in '!=<>~'):
         try:
             pip.utils.packaging.specifiers.SpecifierSet(path)
         # If this is not a valid specifier, just move on and try it as a path


techalchemy:
@therefromhere thanks for the PR. It is missing but I don't think it's necessary to run a specifier check on every entry that starts with a `~` because it's a valid path value.

If you want to PR the typo correction I'd be happy to accept that but pipfiles like the following currently work completely fine on windows, I see no reason to expose them to extra steps of logic when they don't need it:

```ini
[[source]]

verify_ssl = true
name = "pypi"


[dev-packages]


[packages]

django = "~=1.11"
```

nateprewitt:
@techalchemy why does ~=1.11 work and <= not? This seems like #1183 was fixed at the wrong level if that's the case.

techalchemy:
@nateprewitt I believe it has to do with the way pathlib handles testing paths on different platforms. Posix paths can be tested against the filesystem with arbitrary? values or at least with values included in the specifier set, while the windows check fails with: `WindowsError: [Error 123] The filename, directory name, or volume label syntax is incorrect: '<2.0'`. This happens as we are consuming pipfile entries and determining how to convert them into pip requirements for resolving into the lockfile, so we first have to determine if they are filesystem paths (specifically if they are files).

Essentially I think it is the correct behavior to fix the check where it is breaking, this is the right place to do it, and I don't think we should fix the parts that are working

therefromhere:
fair enough @techalchemy , I'll cancel this PR then.

techalchemy:
@therefromhere thanks for opening this PR and spurring a discussion, in the end I am now convinced that you are in fact correct and that I was wrong to exclude `~` from the list of specifiers! Thanks for submitting the PR and causing the discussion, I think we should merge the original one

nateprewitt:
I'd like to get appveyor running before we merge any more PRs, but after that I think we're happy with this. Thanks @therefromhere!

PR #1193, opened 2017-12-11
Pass clear flag to lock when updating

uda:
When updating all with the `--clear` flag, it isn't passed to the `do_lock` call, so the cache isn't actually cleared.
This PR fix this behavior.


kennethreitz:
✨🍰✨

PR #1221, opened 2017-12-19
Force Windows local drive names to uppercase

### LINKED ISSUE #1218: Existing virtualenvs ignored on Windows when using cmd.exe
This is an issue caused by some odd quirk in how cmd.exe handles path names on Windows.

As a result, it is possible to end up in a situation, where running a pipenv command (such as pipenv shell) in a directory, for which a valid virtualenv already exists, causes pipenv to ignore it and create a brand new one, with the same name (except for the hash).

I have recently run into this issue at work. I have a workstation running Windows 10 and Python 2.7, which usually gets called from cmd.exe. What I have just discovered, is that depending on what command I use when changing directories, a slightly different path gets reported as pipfile_location in project.py:

> cd c:\tmp\test

results in "c:\tmp\test" getting set as the current directory in cmd.exe and "c:\tmp\test\Pipfile" getting returned from pipfile_location

vs.

> cd C:\tmp\test

which leads to "C:\tmp\test" getting set as the current directory in cmd.exe and "C:\tmp\test\Pipfile" being the reported location of the Pipfile.

Note the lower- vs. upper-case drive letter, which results in a completely different hash value for the pipfile location, and in consequence, a completely different name being returned from virtualenv_name in project.py, which then triggers the creation of a brand new virtualenv. 
This seems to affect only drive letters -- directory names are fine ("cd C:\TMP" produces "C:\tmp" on my machine). PowerShell does not seem to suffer from this.

1. OS Type : Windows 10 x64 (build 1709)
1. Python version: 2.7.14
1. Pipenv version: 9.0.1

##### Expected result

'pipenv shell' activates the existing environment

##### Actual result

'pipenv shell' creates a new environment alongside the existing one

##### Steps to replicate

Provide the steps to replicate (which usually at least includes the commands and the Pipfile).

With Python 2.7 (and pipenv) installed and added to the PATH env variable:

1. Start the command prompt (Win+R, cmd.exe)
2. From within the command prompt, create a directory somewhere on a disk (e.g. C:\temp\test)
3. Go to that directory, typing the drive letter in upper case (cd C:\temp\test)
4. Call pipenv to create a new virtualenv for that directory (e.g. pipenv install pylint)
5. Change the current working directory to some other directory (e.g. cd C:\)
6. Go back to the test directory, typing the drive letter in lower case (cd c:\temp\test)
7. Call pipenv shell (or pipenv --venv) inside the test directory -- pipenv will
create a new virtualenv, ignoring the one created in step 4

uranusjr:
This fixes #1218.

Windows generally use upper cased drive names, but allow (without normalizing) lower cased names in cmd.exe, which results in inconsistencies when hashing the full path (to get the name of the project's virtualenv). Python does not provide a good solution[*], so we need to roll our own.

[*]: Python does have os.path.normcase(), but it always converts the whole paths to lowercase. This would break virtually *all* existing virtualenvs for Windows users.

UNC host names (which Python also treats as drives), on the other hand, can actually be either cased. I am not sure if they are case-sensitive, or should be coerced to what case, so this patch keeps with the existing behaviour, and does not try to coerce them.

### DIFF
--- a/pipenv/project.py
+++ b/pipenv/project.py
@@ -14,7 +14,7 @@ import toml
 from .utils import (
     mkdir_p, convert_deps_from_pip, pep423_name, recase_file,
     find_requirements, is_file, is_vcs, python_version, cleanup_toml,
-    is_installable_file, is_valid_url
+    is_installable_file, is_valid_url, normalize_drive,
 )
 from .environments import PIPENV_MAX_DEPTH, PIPENV_VENV_IN_PROJECT
 from .environments import PIPENV_VIRTUALENV, PIPENV_PIPFILE
@@ -23,7 +23,7 @@ if PIPENV_PIPFILE:
     if not os.path.isfile(PIPENV_PIPFILE):
         raise RuntimeError('Given PIPENV_PIPFILE is not found!')
     else:
-        PIPENV_PIPFILE = os.path.abspath(PIPENV_PIPFILE)
+        PIPENV_PIPFILE = normalize_drive(os.path.abspath(PIPENV_PIPFILE))
 
 
 class Project(object):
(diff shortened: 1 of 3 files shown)


kennethreitz:
✨🍰✨

PR #1233, opened 2017-12-21
Clean up temporary files created during installation

OddBloke:
At the moment, pipenv will create a temporary file per non-leaf package during dependency resolution, and one per resolved package during installation.

This PR adds cleanup to those two cases, and a test to ensure that no temporary artifacts remain after installation.


kennethreitz:
i like this idea, but it appears to break our tests?

PR #1241, opened 2017-12-22
Don't shellquote backup Python executable

### LINKED ISSUE #1240: Traceback on dependency resolution error with editable git
When I have an editable git repository and another requirement that causes a dependency resolution failure in my Pipfile, I get a `FileNotFoundError: [Errno 2] No such file or directory: '"/usr/bin/python3"': '"/usr/bin/python3"'` error.

My reproducer uses https://github.com/OddBloke/sphinx-git and python-novaclient==2.26.0 (which always causes a dependency resolution error, as mentioned in #1232).

(As far as I can tell, the git repository and the other requirement don't need to be at all related to end up with this traceback.)

1. OS Type: Ubuntu 17.10
1. Python version: Python 3.6.3
1. Pipenv version: pipenv, version 9.0.1

##### Expected result

Dependency resolution error output, with no traceback.

##### Actual result

A traceback:

```pytb
  File "/home/daniel/.local/bin/pipenv", line 11, in <module>
    sys.exit(cli())
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 722, in __call__
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 1066, in invoke
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 895, in invoke
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 535, in invoke
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/cli.py", line 1865, in install
    do_init(dev=dev, allow_global=system, ignore_pipfile=ignore_pipfile, system=system, skip_lock=skip_lock, verbose=verbose, concurrent=concurrent, deploy=deploy, pre=pre)
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/cli.py", line 1311, in do_init
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/cli.py", line 1102, in do_lock
    pre=pre
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/utils.py", line 557, in resolve_deps
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/utils.py", line 507, in actually_resolve_reps
    resolved_tree.update(resolver.resolve(max_rounds=PIPENV_MAX_ROUNDS))
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/patched/piptools/resolver.py", line 102, in resolve
    has_changed, best_matches = self._resolve_one_round()
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/patched/piptools/resolver.py", line 200, in _resolve_one_round
    for dep in self._iter_dependencies(best_match):
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/patched/piptools/resolver.py", line 274, in _iter_dependencies
    for dependency in self.repository.get_dependencies(ireq):
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/patched/piptools/repositories/pypi.py", line 153, in get_dependencies
    result = reqset._prepare_file(self.finder, ireq)
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/patched/pip/req/req_set.py", line 523, in _prepare_file
    abstract_dist.prep_for_dist()
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/patched/pip/req/req_set.py", line 134, in prep_for_dist
    self.req_to_install.run_egg_info()
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/patched/pip/req/req_install.py", line 438, in run_egg_info
    command_desc='python setup.py egg_info')
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/patched/pip/utils/__init__.py", line 667, in call_subprocess
    cwd=cwd, env=env)
  File "/usr/lib/python3.6/subprocess.py", line 709, in __init__
    restore_signals, start_new_session)
  File "/usr/lib/python3.6/subprocess.py", line 1344, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: '"/usr/bin/python3"': '"/usr/bin/python3"'
```

##### Steps to replicate

1. Put this in a Pipfile:
```
[[source]]

verify_ssl = true
name = "pypi"


[packages]

sphinx-git = {git = "https://github.com/OddBloke/sphinx-git.git", editable = true}
python-novaclient = "~=2.26.0"

[dev-packages]
```
2. Run `pipenv install`.

```
$ pipenv install --verbose
Could not find a version that matches pbr!=2.1.0,<2.0,>=0.11,>=2.0.0
Tried: 0.5.2.5.g5b3e942, 0.5.0, 0.5.1, 0.5.2, 0.5.4, 0.5.5, 0.5.6, 0.5.7, 0.5.8, 0.5.10, 0.5.11, 0.5.12, 0.5.13, 0.5.14, 0.5.15, 0.5.16, 0.5.17, 0.5.18, 0.5.19, 0.5.20, 0.5.21, 0.5.22, 0.5.23, 0.6, 0.7.0, 0.8.0, 0.8.1, 0.8.2, 0.9.0, 0.9.0, 0.10.0, 0.10.0, 0.10.1, 0.10.1, 0.10.2, 0.10.2, 0.10.3, 0.10.3, 0.10.4, 0.10.4, 0.10.5, 0.10.5, 0.10.6, 0.10.6, 0.10.7, 0.10.7, 0.10.8, 0.10.8, 0.11.0, 0.11.0, 0.11.1, 0.11.1, 1.0.0, 1.0.0, 1.0.1, 1.0.1, 1.1.0, 1.1.0, 1.1.1, 1.1.1, 1.2.0, 1.2.0, 1.3.0, 1.3.0, 1.4.0, 1.4.0, 1.5.0, 1.5.0, 1.6.0, 1.6.0, 1.7.0, 1.7.0, 1.8.0, 1.8.0, 1.8.1, 1.8.1, 1.9.0, 1.9.0, 1.9.1, 1.9.1, 1.10.0, 1.10.0, 2.0.0, 2.0.0, 2.1.0, 2.1.0, 3.0.0, 3.0.0, 3.0.1, 3.0.1, 3.1.0, 3.1.0, 3.1.1, 3.1.1
CRITICAL:pip.utils:Error [Errno 2] No such file or directory: '"/usr/bin/python3"': '"/usr/bin/python3"' while executing command python setup.py egg_info
  File "/home/daniel/.local/bin/pipenv", line 11, in <module>
    sys.exit(cli())
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 722, in __call__
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 1066, in invoke
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 895, in invoke
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 535, in invoke
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/cli.py", line 1865, in install
    do_init(dev=dev, allow_global=system, ignore_pipfile=ignore_pipfile, system=system, skip_lock=skip_lock, verbose=verbose, concurrent=concurrent, deploy=deploy, pre=pre)
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/cli.py", line 1311, in do_init
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/cli.py", line 1102, in do_lock
    pre=pre
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/utils.py", line 557, in resolve_deps
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/utils.py", line 507, in actually_resolve_reps
    resolved_tree.update(resolver.resolve(max_rounds=PIPENV_MAX_ROUNDS))
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/patched/piptools/resolver.py", line 102, in resolve
    has_changed, best_matches = self._resolve_one_round()
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/patched/piptools/resolver.py", line 200, in _resolve_one_round
    for dep in self._iter_dependencies(best_match):
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/patched/piptools/resolver.py", line 274, in _iter_dependencies
    for dependency in self.repository.get_dependencies(ireq):
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/patched/piptools/repositories/pypi.py", line 153, in get_dependencies
    result = reqset._prepare_file(self.finder, ireq)
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/patched/pip/req/req_set.py", line 523, in _prepare_file
    abstract_dist.prep_for_dist()
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/patched/pip/req/req_set.py", line 134, in prep_for_dist
    self.req_to_install.run_egg_info()
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/patched/pip/req/req_install.py", line 438, in run_egg_info
    command_desc='python setup.py egg_info')
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/patched/pip/utils/__init__.py", line 667, in call_subprocess
    cwd=cwd, env=env)
  File "/usr/lib/python3.6/subprocess.py", line 709, in __init__
    restore_signals, start_new_session)
  File "/usr/lib/python3.6/subprocess.py", line 1344, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: '"/usr/bin/python3"': '"/usr/bin/python3"'
```

OddBloke:
This fixes #1240.

The shellquote call was added for every platform in https://github.com/pypa/pipenv/commit/908b41df75f6306a427a198d506b3b5f9ebef676, but this won't work on, at least, Linux. I've taken an educated guess at this being required on Windows as the subprocess module has [documentation explaining how Windows argument sequences are treated differently](https://docs.python.org/2/library/subprocess.html#converting-argument-sequence).

I don't have access to a Windows environment, so I can't confirm whether or not my supposition is correct.

### DIFF
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -516,7 +516,7 @@ def actually_resolve_reps(deps, index_lookup, markers_lookup, project, sources, 
             ),
             err=True)
 
-        click.echo(crayons.blue(e))
+        click.echo(crayons.blue(str(e)))
 
         if 'no version found at all' in str(e):
             click.echo(crayons.blue('Please check your version specifier and version number. See PEP440 for more information.'))
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -535,7 +535,7 @@ def resolve_deps(deps, which, which_pip, project, sources=None, verbose=False, p
     markers_lookup = {}
 
     python_path = which('python', allow_global=allow_global)
-    backup_python_path = shellquote(sys.executable)
+    backup_python_path = sys.executable
 
     results = []
 
--- a/tests/test_pipenv.py
+++ b/tests/test_pipenv.py
@@ -450,8 +450,12 @@ tablib = "<0.12"
             with open(p.pipfile_path, 'w') as f:
                 contents = """
 [packages]
-requests = {git = "https://github.com/requests/requests.git", editable = true}
-"oslo.utils" = "==1.4.0"
+pypa-docs-theme = {git = "https://github.com/pypa/pypa-docs-theme", editable = true}
+
+# This version of requests depends on idna<2.6, forcing dependency resolution
+# failure
+requests = "==2.16.0"
+idna = "==2.6.0"
                 """.strip()
                 f.write(contents)
             c = p.pipenv('install')
--- a/tests/test_pipenv.py
+++ b/tests/test_pipenv.py
@@ -440,6 +440,26 @@ tablib = "<0.12"
             assert 'tablib' in p.lockfile['default']
 
 
+    @pytest.mark.e
+    @pytest.mark.install
+    @pytest.mark.vcs
+    @pytest.mark.resolver
+    def test_editable_vcs_install_in_pipfile_with_dependency_resolution_doesnt_traceback(self):
+        # See https://github.com/pypa/pipenv/issues/1240
+        with PipenvInstance() as p:
+            with open(p.pipfile_path, 'w') as f:
+                contents = """
+[packages]
+requests = {git = "https://github.com/requests/requests.git", editable = true}
+"oslo.utils" = "==1.4.0"
+                """.strip()
+                f.write(contents)
+            c = p.pipenv('install')
+            assert c.return_code == 1
+            assert 'FileNotFoundError' not in c.out
+            assert 'FileNotFoundError' not in c.err
+
+
     @pytest.mark.run
     @pytest.mark.install
     def test_multiprocess_bug_and_install(self):
--- a/tests/test_pipenv.py
+++ b/tests/test_pipenv.py
@@ -456,8 +456,8 @@ requests = {git = "https://github.com/requests/requests.git", editable = true}
                 f.write(contents)
             c = p.pipenv('install')
             assert c.return_code == 1
-            assert 'FileNotFoundError' not in c.out
-            assert 'FileNotFoundError' not in c.err
+            assert "Your dependencies could not be resolved" in c.err
+            assert 'Traceback' not in c.err
 
 
     @pytest.mark.run


techalchemy:
@OddBloke thanks for the PR and the relevant documentation.  Ideally this wouldn't even be falling back to the backup python but for whatever reason it is in your case.  I am not so sure it would be necessary to shellquote on windows-- we aren't shellquoting our [main which function](https://github.com/pypa/pipenv/blob/v9.0.1/pipenv/cli.py#L1431) which is used to look for python:

```python
    def which(command, location=None, allow_global=False):
        if location is None:
            location = project.virtualenv_location

        if not allow_global:
            if os.name == 'nt':
                p = find_windows_executable(os.path.join(location, 'Scripts'), command)
            else:
        else:
            if command == 'python':
                p = sys.executable

    return p
```

I've tried this on windows now with and without the shellquoting and it seems to work fine both ways so I'm under the impression that python escapes the representation of the executable properly when necessary (I even tried in cases where the executable path has spaces in it, both still work),

This looks good to me.

pipenv/utils.py
```diff
@@ -535,7 +535,9 @@ def resolve_deps(deps, which, which_pip, project, sources=None, verbose=False, p
     markers_lookup = {}
 
     python_path = which('python', allow_global=allow_global)
-    backup_python_path = shellquote(sys.executable)
+    backup_python_path = sys.executable
+    if os.name == 'nt':
```
techalchemy: My only caveat is that we don't currently use shellquoting for `sys.executable` when we invoke it in other places, even for windows, so I believe you can skip this alternative logic -- I think you are probably assuming it was added for a reason, but in this case I think it was just added as a mistake :)

tests/test_pipenv.py
```diff
@@ -440,6 +440,25 @@ def test_windows_pinned_pipfile(self):
...
+    @pytest.mark.install
+    @pytest.mark.vcs
+    def test_editable_vcs_install_in_pipfile_with_dependency_resolution_doesnt_traceback(self):
+        # See https://github.com/pypa/pipenv/issues/1240
+        with PipenvInstance() as p:
+            with open(p.pipfile_path, 'w') as f:
+                contents = """
+[packages]
+requests = {git = "https://github.com/requests/requests.git", editable = true}
+"oslo.utils" = "==1.4.0"
```
techalchemy: So what actually is the dependency conflict here? Does `oslo.utils==1.4.0` conflict with requests? perhaps you can pin requests to a specific `ref` to ensure we keep having the same conflict? I'm fairly confident we will be changing resolver logic in the future so we may want to keep that in mind as it may invalidate this test (i.e. if we change the way resolver failures are handled/python version hacking operates).

tests/test_pipenv.py
```diff
@@ -440,6 +440,25 @@ def test_windows_pinned_pipfile(self):
             assert 'tablib' in p.lockfile['default']
 
 
+    @pytest.mark.e
+    @pytest.mark.install
+    @pytest.mark.vcs
```
techalchemy: Would you be so kind as to add the `resolver` mark as well, so we can track this down when we do make substantial changes to the resolver?

tests/test_pipenv.py
OddBloke: The dependency conflict is entirely contained within `oslo.utils`; it depends on a specific version of pbr, and the latest version of one of its dependencies depends on a different version of pbr.

So, really, it's just there to force a dependency resolution failure so that we end up on the broken code path, the second resolution attempt.

pipenv/utils.py
OddBloke: Cool, I'll drop the if.

tests/test_pipenv.py
OddBloke: Will do.

OddBloke:
@techalchemy Changes made. :slightly_smiling_face:

tests/test_pipenv.py
nateprewitt: Hrmm, so just because this state exists now doesn’t mean it will necessarily exist in the future. When it’s a package outside of our control, I’d prefer we avoid implicit testing conditions.

Is it possible for us to find package(s) we can pin at exact versions which will cause this failure mode?

tests/test_pipenv.py
```diff
@@ -440,6 +440,26 @@ def test_windows_pinned_pipfile(self):
...
+                contents = """
+[packages]
+requests = {git = "https://github.com/requests/requests.git", editable = true}
+"oslo.utils" = "==1.4.0"
+                """.strip()
+                f.write(contents)
+            c = p.pipenv('install')
+            assert c.return_code == 1
+            assert 'FileNotFoundError' not in c.out
+            assert 'FileNotFoundError' not in c.err
```
nateprewitt: Can we get a positive assertion here? Right now this could fail and produce a trace back for any reason except FileNotFoundError. That’s contradictory to what the test claims to be checking. What are we expecting to happen in this case?

tests/test_pipenv.py
```diff
@@ -440,6 +440,26 @@ def test_windows_pinned_pipfile(self):
             assert 'tablib' in p.lockfile['default']
 
 
+    @pytest.mark.e
```
nateprewitt: This may already be a convention we established but ‘e’ seems really vague and could mean at least two things in this context (editable or the error object). If this is already used, @techalchemy we may want a note about modifying these to a more explicit value.

pipenv/utils.py
```diff
@@ -535,7 +535,7 @@ def resolve_deps(deps, which, which_pip, project, sources=None, verbose=False, p
     markers_lookup = {}
 
     python_path = which('python', allow_global=allow_global)
-    backup_python_path = shellquote(sys.executable)
+    backup_python_path = sys.executable
```
nateprewitt: I’m not positive this was added on a whim, this is specifically to handle paths with spaces in them. I’d like to get testing around that area before we remove this.

pipenv/utils.py
```diff
@@ -516,7 +516,7 @@ class PipCommand(pip.basecommand.Command):
             ),
             err=True)
 
-        click.echo(crayons.blue(e))
+        click.echo(crayons.blue(str(e)))
```
nateprewitt: Was there a reason relating to this PR for wrapping the error object with str?

tests/test_pipenv.py
OddBloke: Yep, this was on the test that I copied for this one; my assumption was "editable", so I carried it over.

pipenv/utils.py
OddBloke: I did some noodling on how to do such testing, but didn't really come up with a great way of doing it. That said, my WORKON_PATH change in #1245 will probably give us a route to producing virtualenvs (and therefore `backup_python_path`s) with paths of our choosing, so I'll have a play around with that.

(I did some local playing around with subprocess (on Linux), and it didn't seem like whitespace (or even quotes) in paths needed to be escaped to work.)

pipenv/utils.py
OddBloke: This line, as was, was causing the actual error I was seeing in the terminal to be replaced by another one (something along the lines of `__str__ returned a NoCandidateFound`); it only occurred in testing, so this worked around the problem.

pipenv/utils.py
techalchemy: @nateprewitt it's going to be really hard to keep writing extra layers of tests to do specific things that trick the resolver into failing to the secondary resolver and such if we are just trying to test this... this call is functionally identical to the line above it if `allow_global=True`, so we can possibly change this code to say:

```python
backup_python_path = which('python', allow_global=True)
```

And keep the escaping logic (if necessary) in that one place.

The path just gets passed as a string to the hacked python version, it's never used as a shell input (and therefore preserves spaces and doesn't require escaping).  I tested this by hand, but I have no idea how we would confirm that the resolver still works with a python path with spaces unless we can run an appveyor instance with a python path with spaces in it.  In general do you think it's better to shell escape things at all possible points? Or should we try to keep that logic relegated only to the points where we are passing things to the shell?

pipenv/utils.py
OddBloke: This is the traceback I was seeing:

```
  File "/home/daniel/.local/bin/pipenv", line 11, in <module>
    sys.exit(cli())
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 722, in __call__
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 1066, in invoke
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 895, in invoke
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 535, in invoke
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/cli.py", line 1865, in install
    do_init(dev=dev, allow_global=system, ignore_pipfile=ignore_pipfile, system=system, skip_lock=skip_lock, verbose=verbose, concurrent=concurrent, deploy=deploy, pre=pre)
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/cli.py", line 1311, in do_init
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/cli.py", line 1102, in do_lock
    pre=pre
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/utils.py", line 545, in resolve_deps
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/utils.py", line 519, in actually_resolve_reps
    click.echo(crayons.blue(e))
  File "/home/daniel/.local/lib/python3.6/site-packages/pipenv/vendor/click/utils.py", line 221, in echo
    message = text_type(message)
TypeError: __str__ returned non-string (type NoCandidateFound)
```

tests/test_pipenv.py
OddBloke: I've tightened up the assertions here, so they match the test name a bit better.

tests/test_pipenv.py
OddBloke: My latest commit switches to using requests, idna and the PyPA Sphinx theme (as it has no dependencies). It also adds a comment explaining why we're choosing these particular versions.

tests/test_pipenv.py
techalchemy: @nateprewitt if there is a convention of marking things with `e` it's possible someone just did that for convenience to isolate a test to run locally, we should be cognizant of that in future PRs and try to get it cleaned up

PR #1258, opened 2018-01-02
Fixed broken links to fancy install methods

dragon788:
Outside of the TOC blocks it was referring to install.html instead of /install/ and resulted in broken links.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -52,6 +52,7 @@ Pipenv is a python package and so can be installed using ``pip`` as you would ex
 ::
 
     $ pip install pipenv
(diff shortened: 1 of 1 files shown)


docs/index.rst
```diff
@@ -52,14 +52,13 @@ Pipenv is a python package and so can be installed using ``pip`` as you would ex
 ::
 
     $ pip install pipenv
-    ✨🍰✨
```
nateprewitt: Would you mind putting these back?

nateprewitt:
Thanks again  @dragon788!

Issue #1150, opened 2017-12-02
Options to improve Python runtime autodiscovery on Windows


ncoghlan:
Pipenv currently has two Python runtime discovery and/or installation strategies:

1. Search `PATH` (in `pipenv.cli.find_a_system_python`), adding any `pyenv` environments to the end before it does so
2. Install it with `pyenv` (in `pipenv.cli.ensure_python`)

This misses a few potential possibilities for improvement on Windows:

1. Discovery based on the Windows launcher configuration file (https://docs.python.org/3/using/windows.html#customization)
2. Discovery based on the Windows registry (https://github.com/zooba/pep514tools#usage)
3. Supporting `conda` based automatic Python installation in addition to `pyenv` installation
4. Supporting @uranusjr's https://github.com/uranusjr/pythonup-windows (and its macOS counterpart)

Actually doing this would probably requiring refactoring the autodiscovery code though, such that the discovery & installation strategy used by `ensure_python` can be chosen at import time based on the platform.

uranusjr:
#841 :(

uranusjr:
The launcher also looks for some environment variables, so that should probably be considered as well. I’m pretty interested in working on this feature, but probably can’t put much real effort until the end of the year.

ncoghlan:
@uranusjr Cool! I'm actually thinking it may make more sense to start by refactoring the existing "Python finding" code out of `pipenv.cli` and into a separate `pipenv.pythonfinder` helper module.

The reason I suggest that is because the way the current `pyenv` integration works is as follows:

1. There's an `activate_pyenv` step that adds all installed pyenv Python's to `PATH` in the current process
2. There's the usual `find_a_system_python` step to actually locate an installed Python
3. There's then a fallback to ask `pyenv` to *install* the requested Python, if it isn't found otherwise

That gives us 3 operations to abstract away:

* building an inventory of "available Python runtimes" (current implementation: add `pyenv` executables to PATH)
* searching the Python runtime inventory for a requested version (current implementation: scan PATH for suitably named files)
* requesting installation of a particular version if nothing appropriate is found (current implementation: run `pyenv` in a subprocess)

So I'm picturing an API where the usage looks something like this in `cli.ensure_python`:

    # Existing code to figure out `python`
    path_to_python = pipenv.pythonfinder.get_runtime_path(python)
    if path_to_python is None and PIPENV_SHOULD_AUTO_INSTALL:
[...]

PR #1259, opened 2018-01-02
[WIP] Refactor Python lookups and Add PEP 512 Finder for Windows

uranusjr:
Tries to kickstart #1150

* Move system_which() to utils so the finders can use it.
* Find pyenv versions with a different method. Instead of adding them all into PATH, resolve their names directly to find a best match.
* Install the finders into pipenv.pythonfinders based on environment configuration. Currently this simply install pyenv if available, and then the PATH lookup (always).
* Implement PEP 512 finder on Windows.
* Move automatic pyenv installation into the finder submodule.

TODO:

* Add tests for finders.

### DIFF
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -28,23 +28,22 @@ from pipreqs import pipreqs
 from blindspin import spinner
 from urllib3.exceptions import InsecureRequestWarning
 from .project import Project
+from .pythonfinders import find_python, PythonInstallationNotFoundError
 from .utils import (
     convert_deps_from_pip, convert_deps_to_pip, is_required_version,
     proper_case, pep423_name, split_file, merge_deps, resolve_deps, shellquote, is_vcs,
     python_version, suggest_package, find_windows_executable, is_file,
     prepare_pip_source_args, temp_environ, is_valid_url, download_file,
-    get_requirement, need_update_check, touch_update_stamp
+    get_requirement, need_update_check, touch_update_stamp, system_which
 )
 from .__version__ import __version__
 from . import pep508checker, progress
 from .environments import (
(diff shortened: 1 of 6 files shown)


uranusjr:
Added PEP 514 finder. The current implementation only supports CPython installations (under reg key `Python\PythonCore`). It looks for per-user, machine-wide (native), and machine-wide-32-bit-on-64-host Python installations, in that order, and call them one by one to find an installation matching the specified version.

uranusjr:
I’ve moved what I can and I *think* everything is working.

techalchemy:
Sorry to leave this hanging for so long! I know most of us have been busy due to holidays and catch up. I will give this a review but I’m excited about the refactor and the prospect of cleaning more things up generally. Thanks for putting this all together, I know it was a substantial effort!

uranusjr:
I’ve been testing this with my own installation, and there are a behaviour that might worth discussing. Currently (both on master and here) a Python is chosen with a first-match-wins strategy. Say I want Python 3. The finder first look for one in pyenv (or registry on Windows), and then in PATH. If the user, say, installed only old Pythons in pyenv, there’s no way to get to the new Pythons in PATH.

The alternative would be to collect *all* possible Pythons, so we can choose the best possible match. This is what the Python launcher (`py.exe`) does, but I wonder if it is necessary here.

kennethreitz:
I'm very apprehensive to touch this code, as it's working well now.

kennethreitz:
this looks solid

techalchemy:
I am less apprehensive because our codebase is becoming difficult to work with until we do some refactoring. I am apprehensive about reading this and doing a thorough enough review on it. I will make time shortly.

kennethreitz:
I'd like to leave things as they are, for now. I really like our current implementation.

PR #1262, opened 2018-01-03
Make Appveyor Pass Again™

uranusjr:
This should fix the Windows CI… or so I hope.

Also corrected a test failure on Mac because of system symlinks.


uranusjr:
Oops, so #1244 also addresses this… I guess it should be merged instead. dac4ba7 is still relevant though.

PR #1266, opened 2018-01-03
Remove kr_easter_egg

Kwpolska:
Since pipenv is an official PyPA project, and is recommended by packaging.python.org, it should not discriminate against or promote any specific packages, since that is unprofessional and may be discouraging to some (i.e. yours truly).

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -1556,11 +1556,6 @@ def format_pip_output(out, r=None):
     return out
 
 
-# |\/| /\ |) [-   ]3 `/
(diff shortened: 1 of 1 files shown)


kennethreitz:
will consider it, but this will be done by the project, but by a contributor.

dmtucker:
Just curious, why does the submitter matter?

nedbat:
FYI: https://github.com/pypa/pipenv/commit/9a0aceee2f334b624f56c6b20ebeace4a8dcbeb9

PR #1287, opened 2018-01-11
replace netloc by hostname when adding trusted-host arg to pip

karec:
Fixing a bug when using a combination of user / password in index url and `verify_ssl = false` in Pipfile

Example Pipfile to reproduce : 

```
[[source]]

name = "pypi"
verify_ssl = true

[[source]]
name = "private"
url = "https://user:password@myrepo/simple"
verify_ssl = false


[dev-packages]



[packages]

mypackage = "*"


[requires]

python_version = "3.5"

```

Using `pipenv lock --verbose` with this give us the following output

```
Using pip: -i https://pypi.python.org/simple --extra-index-url https://user:password@myrepo/simple --trusted-host user
```

### DIFF
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -439,7 +439,7 @@ def prepare_pip_source_args(sources, pip_args=None):
 
         # Trust the host if it's not verified.
         if not sources[0].get('verify_ssl', True):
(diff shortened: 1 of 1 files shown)


nateprewitt:
Yep, this seems good to me. We’re actually in the process of doing the same thing inRequests right now. Thanks!

Issue #1273, opened 2018-01-06
/tmp fills up with pipenv-*-requirement.txt files


SethMichaelLarson:
Temporary directory fills up with many temporary requirements.txt files.
Using Pipenv frequently results in hundreds of files in `/tmp`.

##### Describe your environment

1. Ubuntu 17.10
1. Python version: `Python 3.6.2`
1. Pipenv version: `pipenv, version 8.2.6`

##### Expected result

Pipenv cleans up these files after completing successfully.

##### Actual result

Pipenv leaves the temporary files.

##### Steps to replicate

Run these steps in an empty directory

```bash
$ rm -rf /tmp/pipenv-*
$ ls -Al /tmp | grep pipenv | wc -l # 0
$ pipenv install pytest
$ ls -Al /tmp | grep pipenv | wc -l # 2
```

nateprewitt:
Yep, this is a known issue and we keep closing the tickets around it. I’d like to use this as the tracking ticket going forward. The issue here is we’re using TemporaryFiles mkstemp instead of using the object as a context.

There may be other areas that need cleanup around this but those are the cases generating the largest volume. Anyone who would like to look into fixing this is welcome. We just need to make sure the solution is cross platform compatible.

Thanks @SethMichaelLarson for getting this issue logged.

kennethreitz:
fixed

PR #1289, opened 2018-01-12
Revert "Clean up temporary files created during installation"

techalchemy:
Reverts pypa/pipenv#1233

- This breaks everything on windows

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -734,7 +734,7 @@ def do_install_dependencies(
     """"Executes the install functionality."""
 
     def cleanup_procs(procs, concurrent):
-        for c, cleanups in procs:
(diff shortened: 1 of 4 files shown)


nateprewitt:
Yeah, this is why we weren’t merging anything till Appveyor was running properly. Was there any issue open with the previous PR that we can open back up to track this?

nateprewitt:
#1273 will be the tracking issue for this. Thanks for catching this @techalchemy!

uranusjr:
I haven’t tried, but why does it break on Windows? Is it because Windows have issues deleting files with handled held by processes?

techalchemy:
@uranusjr that is correct. However I have a rework that won't require a full revert i think

PR #1296, opened 2018-01-13
WIP: Wait for pip before unlinking tmp requirements

techalchemy:
- Wait for delegator/subprocess to finish interacting with pip before
deleting tmpfile resources created for this purpose
- Use `NamedTemporaryFiles` for win32 tmpfiles
- Allows graceful opening and closing, flushing of files
- Offload handling to a context manager that can either close or delete
tmpfiles on exit
- Allow `delete=False` to be passed to context mgr for persistence
- Files will always be closed


techalchemy:
* I only know  this gets appveyor to run in the first place, so i'm not sure how else it is broken.

tests/test_pipenv.py
```diff
@@ -27,6 +27,7 @@ def __init__(self, pipfile=True, chdir=False):
         self.path = tempfile.mkdtemp(suffix='project', prefix='pipenv')
         self.pipfile_path = None
         self.chdir = chdir
+        self.prev_umask = os.umask(0077)
```
uranusjr: Should this be `0o077`? Python 3 requires `0o` for octal numbers.

OddBloke:
https://github.com/pypa/pipenv/blob/master/pipenv/patched/piptools/repositories/pypi.py#L78 seems the most likely candidate for the `tmp...source` failures.

OddBloke:
Oh, in fact, they all appear to be related to `tmp...source`, so I suspect the backported/vendored `TemporaryDirectory` implementation in https://github.com/pypa/pipenv/blob/master/pipenv/patched/piptools/_compat/tempfile.py is probably at fault.

uranusjr:
`tempfile.TemporaryDirectory` [was cleaned up substantially in 3.4](https://bugs.python.org/issue19077). Maybe we can try using the new implementation instead? I have no idea if this is a good idea. Maybe @ncoghlan can provide some insight.

techalchemy:
@uranusjr we’d have to backport for 2.7 

@OddBloke can we confirm that this is a resolution related issue somehow? I am on windows today so I can just experiment with that. I’ll be on IRC in a bit as well

uranusjr:
@techalchemy I know, I mean maybe we can backport the 3.4+ version to 2.7. From reading the bug report I get the impression it is a logical rewrite, not cleanup based on some upgrades of other parts in new Python versions, and can be safely used in 2.7.

OddBloke:
As far as I could tell, the cleanup is mostly just switching to use rmtree
(rather than a separate implementation). So we may also want/need to
backport rmtree, in case that's the fix.

Of course, this assumes that this works well anywhere... :-P

OddBloke:
@techalchemy, I'm on South Africa time this week, so probably won't be
around on IRC, unfortunately. :-(

OddBloke:
It occurs to me that we see very similar errors across all versions of Python suggests that a backport of `TemporaryDirectory` isn't likely to resolve the issues.

ncoghlan:
The cleanup in issue 19077 mainly related to temporary directory cleanup on interpreter shutdown - it used to reference lots of module globals, so it could fail if the interpreter had already run the "set all module level attributes to None" step for any of the modules it used.

That's potentially a problem for the PypiRepository code that @OddBloke pointed out, as I can't see anywhere that actually calls "repository.clear_caches()" explicitly prior to shutdown (at least, not in a quick code search for the method name).

However, I don't think it's the issue the tests are hitting on Appveyor. Rather, it looks to me like they're hitting a variant of https://bugs.python.org/issue14243#msg288504, where something is attempting to unlink a temporary file that the test process still has open (or vice-versa: the test might be attempting to unlink a file that a pipenv object still has open). That works fine on \*nix style filesystems, but Windows will complain about it (it requires that all open handles be closed before it will let you remove the file).

techalchemy:
@ncoghlan I agree with you that it is almost certainly related to unlinking open FDs, my current efforts around this include reimplementing the explicit attempt to finalize weakrefs that was included in the `tempfile.TemporaryDirectory` implementation as well as explicitly terminating process groups recursively using psutil.

The pip codebase itself acknowledges that repository indexes are cloned in read-only mode, but our calls to `pip install` are done via `delegator` which uses `popen` with `shell=True`. On windows that is known to leak file descriptors as it spawns an entirely new process group.

I have a reimplementation of all of these things but instead of ensuring an empty directory *during* the process, I am getting the tests passing by ensuring that the directories are _completely removed_ at the end of the process.  Here is a proof of concept, I'll have a working patch shortly:

```python
    def test_editable_vcs_install(self):
        my_temp_dir = None
        with PipenvInstance() as p:
            with temp_environ():
                with TemporaryDirectory(prefix='pipenv', suffix='tmp') as tmp_dir:
                    os.environ['TMPDIR'] = '{}'.format(tmp_dir)
                    my_temp_dir = '{}'.format(tmp_dir)
                    other_temp_dir = '{}'.format(p.path)
                    c = p.pipenv('install -e git+https://github.com/requests/requests.git#egg=requests')
                    assert c.return_code == 0
                    assert 'requests' in p.pipfile['packages']
                    assert 'git' in p.pipfile['packages']['requests']
                    assert 'editable' in p.pipfile['packages']['requests']
                    assert 'editable' in p.lockfile['default']['requests']
                    assert 'chardet' in p.lockfile['default']
                    assert 'idna' in p.lockfile['default']
                    assert 'urllib3' in p.lockfile['default']
                    assert 'certifi' in p.lockfile['default']
        assert not os.path.exists(my_temp_dir)
        assert not os.path.exists(other_temp_dir)
```

OddBloke:
@techalchemy Could we have those two new context managers effectively folded in to the __enter__ and __exit__ of `PipenvInstance`?  We'll want these wrapped around every test run, I think?

techalchemy:
Yeah I was just experimenting there I folded them in before I went to bed, will get something up when I get in

PR #1298, opened 2018-01-14
Missing comma in __all__ tuple?

cclauss:


nateprewitt:
Hey @cclauss thanks for finding this. [crayons](https://github.com/kennethreitz/crayons) is an external library that we're vendoring. You'll want to propose the patch there, and then we'll get the update version of crayons merged here. Thanks!

nateprewitt:
You know what, it looks like this was introduced in our patched version. I'm going to say this is good. Thanks for contributing, @cclauss!

cclauss:
@nateprewitt  This change seems to be reversed out again at https://github.com/pypa/pipenv/blob/master/pipenv/patched/crayons.py#L21  Was something wrong with it?

nateprewitt:
The branch was force rebase back to 9.0.1 last week and we’re missing a lot of stuff. If you wouldn’t mind making a note in #1324 that this needs to be readded, that would be super helpful. Thanks!

PR #1299, opened 2018-01-14
Avoid an attribute error when attempting to install failed dependencies

joealcorn:
`pip_install` returns a tuple, it seems like we only want the first value in this instance, similar to when installing them the [first time around](https://github.com/pypa/pipenv/blob/6292d87875ead27e95cda2f2375e4e684ac7e889/pipenv/cli.py#L820)
Without this patch an `AttributeError` is raised:

```
  ☤  ▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉ 0/55 — 00:00:00
  File "/Users/joe/.venvs/test/bin/pipenv", line 11, in <module>
  File "/Users/joe/git/pipenv/pipenv/vendor/click/core.py", line 722, in __call__
  File "/Users/joe/git/pipenv/pipenv/vendor/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/Users/joe/git/pipenv/pipenv/vendor/click/core.py", line 1066, in invoke
  File "/Users/joe/git/pipenv/pipenv/vendor/click/core.py", line 895, in invoke
  File "/Users/joe/git/pipenv/pipenv/vendor/click/core.py", line 535, in invoke
  File "/Users/joe/git/pipenv/pipenv/cli.py", line 1899, in install
    do_init(dev=dev, allow_global=system, ignore_pipfile=ignore_pipfile, system=system, skip_lock=skip_lock, verbose=verbose, concurrent=concurrent, deploy=deploy, pre=pre)
  File "/Users/joe/git/pipenv/pipenv/cli.py", line 1314, in do_init
    skip_lock=skip_lock, verbose=verbose, concurrent=concurrent)
  File "/Users/joe/git/pipenv/pipenv/cli.py", line 864, in do_install_dependencies
    if c.return_code != 0:
AttributeError: 'tuple' object has no attribute 'return_code'
```

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -851,7 +851,7 @@ def do_install_dependencies(
                 index = index.split()[0]
 
             # Install the module.
-            c = pip_install(
+            c, _ = pip_install(
                 dep,
                 ignore_hashes=ignore_hash,
                 allow_global=allow_global,


OddBloke:
@joealcorn What packages were you attempting to install when you were seeing this?

(I'd like to write a reproducer test, but I'm not sure how to trigger the condition.)

joealcorn:
The failures were actually because I had made a fresh virtualenv that had outdated `pip` and `setuptools` versions. I can dig into it more when I'm home if need be.

Maybe we could patch `_pip_install` to raise an exception under test or upload a broken package to pypi for just this reason?

kennethreitz:
That sounds like a good plan!

PR #1301, opened 2018-01-15
Clean up temporary files created during tests

OddBloke:


techalchemy:
I can see what you’re doing but I’m on mobile now so I can’t fully comment. The basic issue appears to relate to leaky file descriptors when using shell based invocations, and particularly for editable installs. If you look at the app store faulures on my WIP branch you might see that it can’t delete those

OddBloke:
@techalchemy I think the Appveyor failure is due to my previous change, not this one. I'll have a look at fixing that in a separate branch.

techalchemy:
What I mean is that if you want to clean up editable installed git repositories you are going to need to remove whatever you pass to `—src` which, on windows, will cause you permissions errors when you try even if you traverse the tree and set write bits. 

This is because `popen_spawn` leaks open file descriptors on windows even when the process itself is terminated. So while tests may pass the goal of cleaning up temp resources is not yet being accomplished.

OddBloke:
@techalchemy Ah, OK, so you're saying that even once the other failures are addressed, I will still need to do something different here because I'll hit the same problem?

I'll close this out until the other issue has been resolved, so I can shamelessly steal^W^Wreuse the solution there.

PR #1315, opened 2018-01-17
Fix AttributeError on pipenv uninstall --verbose

hynek:
Current code tries to run .format on the return code of click.echo which is always None.


hynek:
Those CI failures look very much unrelated.

kennethreitz:
They are :)

kennethreitz:
✨🍰✨

PR #1331, opened 2018-01-21
Updates cli.py with proper requirements filename

sobolevn:
In some systems `Requirements.txt` != `requirements.txt` and this line could possibly bring confusion to users.
Since we are only searching for actual `requirements.txt` in `find_requirements()`: https://github.com/pypa/pipenv/blob/cbcc89bee73e5af2e34252deb686ad7578e14c61/pipenv/utils.py#L1121

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -307,7 +307,7 @@ def ensure_pipfile(validate=True):
 
         # If there's a requirements file, but no Pipfile...
         if project.requirements_exists:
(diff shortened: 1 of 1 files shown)


sobolevn:
Btw, do we really need this line (`if 'requirments.txt'`) in `find_requirments`: https://github.com/pypa/pipenv/blob/cbcc89bee73e5af2e34252deb686ad7578e14c61/pipenv/utils.py#L1120 ?

kennethreitz:
nice catch.

PR #1337, opened 2018-01-23
Upload wheel

methane:
When installing from source package, `pipenv` script imports `pkg_resources` and
it makes the command startup slower. ([ref](https://dev.to/methane/how-to-speed-up-python-application-startup-time-nkf))

Uploading wheel package to PyPI will save many people from  this pitfall.


kennethreitz:
We don't do wheels for a reason.

PR #1269, opened 2018-01-05
Link to OWASP Top 10 from docs

ncoghlan:

PR #1236, opened 2017-12-22
More thorough checks before resolving files

techalchemy:
- Fixes #1225
- Add checks for Wheel files and archive files specifically
- Adopts egg-fragment parsing for local file paths to allow naming of
filesystem paths as well
- Adds these checks both to the `get_requirement()` function and the
`is_installable_file()` function

PR #1235, opened 2017-12-22
Properly parse urls as Link objects

techalchemy:
- Fixes #1121
- Adds a bunch of tests for `pipenv.utils.get_requirement()`

PR #1204, opened 2017-12-13
Fix: quote src for safer handling

bradsbrown:
Wrap the `src` path in quotes for safer handling of paths that include special characters (discovered when using a path that contained parens).

PR #1354, opened 2018-01-24
9.x rebuild

### LINKED ISSUE #1002: Pipenv doesn't resolve --system python correctly.

### LINKED ISSUE #1324: Catalog PRs post-9.0.1
It looks like 9.0.3 erased all commits post-9.0.1 which means we’ve lost some relevant PRs. We need to go back through our PR list and determine which ones we need to get added back. Preferably a maintainer can just cherrypick them onto master.

I’ll assign this to myself, but if someone else wants to pick this up, please feel free to do so.

### LINKED ISSUE #1329: Issues with python_version marker in pyenv install
Using pipenv 9.0.3 on python 3.6.4 the following failed:
```sh
$ pipenv --three install "enum==0.4.6; python_version < '3.4'"
…
Installing enum==0.4.6; python_version < '3.4'…
⠧
Adding enum==0.4.6; python_version < '3.4' to Pipfile's [packages]…
  File "/bin/pipenv", line 11, in <module>
    sys.exit(cli())
  File "/venvs/pipenv/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 722, in __call__
  File "/venvs/pipenv/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/venvs/pipenv/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 1066, in invoke
  File "/venvs/pipenv/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 895, in invoke
  File "/venvs/pipenv/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 535, in invoke
  File "/venvs/pipenv/lib/python3.6/site-packages/pipenv/cli.py", line 1934, in install
  File "/venvs/pipenv/lib/python3.6/site-packages/pipenv/cli.py", line 1093, in do_lock
    deps = convert_deps_to_pip(project.packages, project, r=False, include_index=True)
  File "/venvs/pipenv/lib/python3.6/site-packages/pipenv/utils.py", line 742, in convert_deps_to_pip
    if not deps[dep]['version'] == '*':
TypeError: string indices must be integers
```
This leads to `enum = "==0.4.6; python-version < '3.4'"` being in the `Pipfile`'s packages section although I think it should be `enum = {version = "==0.4.6", python_version = "< '3.4'"}`. There is same outcome when using `--two`.

I also tried using a requirements file instead, the attempt at [dependency resolution](https://github.com/pypa/pipenv/issues/1277#issuecomment-356153687) highlighted that the [`enum`](https://pypi.python.org/pypi/enum) isn't intended for compatibility (I should have used [`enum-compat`](https://pypi.python.org/pypi/enum-compat/0.0.2)):
```
$ pipenv --three install --requirements requirements.txt
…
Requirements file provided! Importing into Pipfile…
Updated Pipfile.lock (2c7e63)!
Installing dependencies from Pipfile.lock (2c7e63)…
An error occurred while installing enum==0.4.6! Will try again.
  🐍   ▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉ 1/1 — 00:00:00
Collecting enum==0.4.6 ▉▉▉▉▉▉▉▉▉▉▉▉▉▉ 0/1 — 00:00:00
  Using cached enum-0.4.6.tar.gz
    Complete output from command python setup.py egg_info:
      File "<string>", line 1, in <module>
      File "/share/virtualenvs/test-hxkKlP5o/lib/python3.6/site-packages/setuptools/__init__.py", line 5, in <module>
        import distutils.core
      File "/share/virtualenvs/test-hxkKlP5o/lib/python3.6/distutils/__init__.py", line 4, in <module>
        import imp
      File "/share/virtualenvs/test-hxkKlP5o/lib/python3.6/imp.py", line 27, in <module>
        import tokenize
      File "/share/virtualenvs/test-hxkKlP5o/lib/python3.6/tokenize.py", line 33, in <module>
        import re
      File "/share/virtualenvs/test-hxkKlP5o/lib/python3.6/re.py", line 142, in <module>
        class RegexFlag(enum.IntFlag):
    AttributeError: module 'enum' has no attribute 'IntFlag'
    
    ----------------------------------------

Command "python setup.py egg_info" failed with error code 1 in /tmp/pip-build-0uqv6ezh/enum/

  ☤  ▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉ 0/1 — 00:00:00
```
This leads to `enum = "==0.4.6"` ending up in the `Pipfile`'s (same with `--two` but it installs).

It seems like the issues are:
 * `--requirements` isn't pick picking up marker requirements
 * requirements added as args aren't parsed correctly

techalchemy:
Included in this PR are the following previous pull requests which were reverted due to breakages caused by lost revision history at some point in between 9.0.0 and 9.0.3 (the last released PR seems to have been on 12/14/15)

- Fixes #1324 

### Merges
- #1098 - Pass   `allow_global` to `resolve_deps`
- #1178 - Improve requirements testing
- #1186 - Add support for `~=` specifier   check in paths
- #1193 - Pass `clear` flag to lock when   updating
- #1200 - Fix backup resolver
- #1204 - Quote `PIP_SRC` variable input   to subprocess
- #1221 - Windows drive case   normalization
- #1235 - Parse urls as links
- #1236 - Resolve local files into   pipfile
- #1241 - Don't shellquote backup python   executable
- #1242 - dont use requirements.txt when   filename provided
- #1244 - ensure_project drive   normalization
- #1250 - Update pyenv python versions
- #1258 - Fix broken links to fancy   install methods
- #1268 - Update fails if freeze has   comments
- #1269 - Link to OWASP top 10 from docs
- #1275 - Add reference to a 3rd party   plugin for tox
- #1278 - Fix bad link in `pipenv --envs`   output
- #1282 - Fix uninstall error in verbose   mode
- #1287 - replace netloc by hostname when   adding trusted-host arg to pip

In addition the following changes are included:

- Parse URIs into `pip.Link` objects separately
- Pin failing test to `docker==2.7`
- Swap from ansible to alembic for testing -- ansible is not supported on windows and is causing failures
- Test for string presence before attempting string comparisons when converting dependencies to pip format
- Fixes #1329 and potentially other marker-resolving issues
- This will likely close a bunch of opened and reopened issues
- Fixes #1002 

Additional commit 4a4ee77 introduces a custom `rmtree` function to help handle appveyor test failures related to filesystem permissions:

#### Add custom rmtree function to handle windows rms
    
- Only warn on failure to remove a tree
- This should mitigate a large portion of our windows test failures
- After sufficiently substantial time investment I am currently of the opinion that there is no real elegant way to handle this (I have tried programmatically finding all processes with open file handles, terminating them with a 500ms retry and a 15s timeout)
- Hopefully this gets appveyor passing


nateprewitt:
This really should be broken out into at least three separate PRs. We need to make sure the tests for 9.x actually work without changes, then add back the necessary commits from 9.2, then we should add new functionality. I don't know if anyone can review something of this size properly.

techalchemy:
@nateprewitt the tests for 9.x were never passing on appveyor, and we have lost most of the revision history for master.  I re-merged things roughly in the order in which PRs were accepted as far as I can tell, and since appveyor was failing both before and after those merges it made sense to me to get the tests passing first and then figure out how to split things up if necessary.  I can split the test changes out now that I have them, and then we can rebase this onto those once we are in agreement, if you think that makes sense.  Is that  good starting point?

FYI nothing here passed on appveyor prior to the changes I am introducing which is why I thought it would be valuable to have those changes on here. All of my changes relate to error handling when removing temporary resources on windows essentially

nateprewitt:
Closing this out since I think we've got the issues resolved in separate PRs now. Thanks again for spearheading this @techalchemy!

PR #1357, opened 2018-01-25
Fix installation dead links with good RST linkage

davidfischer:
These links are 404-ing on docs.pipenv.org. This PR fixes that!

http://www.sphinx-doc.org/en/stable/markup/inline.html#cross-referencing-arbitrary-locations


techalchemy:
Hey @davidfischer this is being re-fixed in #1354 so I'm going to go ahead and close this, thanks for the PR though!

davidfischer:
No sweat. I'm glad it's getting handled.

PR #1358, opened 2018-01-26
[WIP] Parse markers from pip requirements

Code0x58:
I only just saw the update on #1329 where it was mentioned on a PR two days ago saying the issue is fixed :sob: *goes to bed*

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -29,6 +29,7 @@ from pipreqs import pipreqs
 from blindspin import spinner
 from urllib3.exceptions import InsecureRequestWarning
 from pip.req.req_file import parse_requirements
+from pip.req.req_install import InstallRequirement
 from click_didyoumean import DYMCommandCollection
 from .project import Project
 from .utils import (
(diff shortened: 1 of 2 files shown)


techalchemy:
I am only claiming it's fixed, that doesn't mean it actually is -- historically speaking 

Of note is that we actually used to use `pip.req.InstallRequirement.from_line` but moved away from it because it isn't really suitable, or it wasn't.  I think we may have reimplemented most of its functionality by now anyway except we don't build wheels or `setup.py` files during the process of parsing install requirements

Code0x58:
Thanks for the insight. It sounds like what you get from avoiding `InstallRequirement` (at least `.from_line`) makes it worth it, I think I used it because I saw it was the type of `package` in `pipenv.cli.import_requirements(…)`.

After a good night's sleep, I am happy to drop this PR and look at adding tests if needed once your PR is in.

techalchemy:
@Code0x58 sorry for being a bit unresponsive, happy to dialogue more on some of this once we get the rest of the old PRs merged in and check whether things are working properly

We will definitely need to refactor some of our parsing code in the near future

PR #1361, opened 2018-01-26
Revised master

nateprewitt:
This merges our commit history from 9.0.1 until 2018-01-02 which was the last date we had a complete commit tree. The tests run here and we're back to a semi-stable state.

Things that are still missing: #1097, #1268, #1269, #1275, #1278, #1282, #1287, #1298

I excluded #1204 and #1178 because I'm not positive we wanted to merge them in the first place. That should probably be reviewed though.


techalchemy:
@nateprewitt thanks for tracking this down... wasn’t #1097 merged by then? I.e are you excluding it because it causes failures?

As for #1178 it still looks ok to me but I’m ok with reviewing further or leaving it for a larger cleanup effort

#1204 should likely be passed through our own shellquote function instead of using raw quotes

Will take a closer look later but this seems to make sense. I can split out the testing changes into a separate PR

techalchemy:
FYI also still missing are #1221, #1235, #1236, #1241, #1242, #1244 some of which were relatively important

nateprewitt:
#1097 was several commits so I hadn't moved it over yet.
#1221 broke the windows build and I thought we'd decided it shouldn't have been merged as it was originally written.

I think the rest of those were all merged for 9.0.2. I haven't reviewed the list of missing commits thoroughly, just what state was up to Jan 2nd. I just want to get us into a known state with tests running before we do anything more.

techalchemy:
Ok I am in agreement then. Let me merge this and then I propose we take the following 3 actions:

* Open a PR re-merging the commits from #1097 
* Split the commits affecting temporary file cleanup from #1354 and put them in a separate WIP PR (since they are now functional) to discuss whether we want to do this
* See if tests pass for the remaining items on #1354 (which contains #1221 as well as the fix for its test breakages)

PR #1364, opened 2018-01-27
Temporary File Cleanup

techalchemy:
This is a small piece of what might need to be a bigger cleanup of some of our backend processes.  This aims to provide the capacity to cleanly handle temporary filesystem resources, and also aims to use more carefully those resources which we do generate.

Broad fixes to testing infrastructure
- Add custom rmtree function to handle windows rms
- Only warn on failure to remove a tree
- This should mitigate a large portion of our windows test failures
- After sufficiently substantial time investment I am currently of the opinion that there is no real elegant way to handle this (I have tried programmatically finding all processes with open file handles, terminating them with a 500ms retry and a 15s timeout)
- Close and handle tempfiles in resolvers
- Only create tempfiles for non-editable deps
- set umask and `PIP_SRC` for test context
- Warn instead of throwing exceptions
- Backport TemporaryDirectory for python2/3 compatibility and reliance on custom rmtree error handling
- Vendor backport of `weakref.finalize` for `TemporaryDirectory.cleanup()`
- Add PyTest fixtures to wrap and cleanup `PIP_SRC` directory at the module level to avoid permissions errors on appveyor
- Add resource warnings for PY2 compat Lock instead of installing complex resolver test
- Add a special condition for allowing access errors

### DIFF
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -1341,9 +1341,9 @@ def pip_install(
         pip.logger.setLevel(logging.INFO)
 
     # Create files for hash mode.
-    if (not ignore_hashes) and (r is None):
-        r = tempfile.mkstemp(prefix='pipenv-', suffix='-requirement.txt')[1]
-        with open(r, 'w') as f:
+    if not package_name.startswith('-e ') and (not ignore_hashes) and (r is None):
+        fd, r = tempfile.mkstemp(prefix='pipenv-', suffix='-requirement.txt')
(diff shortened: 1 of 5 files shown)


nateprewitt:
Looks like we've still got some build issues here, ping me when we're ready to review. Thanks!

kennethreitz:
rebase required!

techalchemy:
waiting on #1372 (has a hidden test dependency) + needs a minor refactor before it can be merged

PR #1372, opened 2018-01-28
Requirements generation and resolver related post-9.0.1 PRs

### LINKED ISSUE #1329: Issues with python_version marker in pyenv install

techalchemy:
- Fixes #1329 

Reintroduces the following:
- #1241 (don't shellquote backup executable)
- #1235 (Properly parse URLs as link objects when generating requirements)
- #1236 (More thorough checks before resolving files)

### DIFF
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -308,17 +308,15 @@ def get_requirement(dep):
     # Strip extras from the requirement so we can make a properly parseable req
     dep, extras = pip.req.req_install._strip_extras(dep)
     matches_uri = any(dep.startswith(uri_prefix) for uri_prefix in SCHEME_LIST)
-    # Only operate on local, existing, non-URI formatted paths
-    if is_file(dep) and isinstance(dep, six.string_types) and not matches_uri:
+    # Only operate on local, existing, non-URI formatted paths which are installable
+    if is_file(dep) and isinstance(dep, six.string_types) and not matches_uri and is_installable_file(dep):
         dep_path = Path(dep)
-        # Only parse if it is a file or an installable dir
-        if (dep_path.is_file() and is_installable_file(dep_path.absolute())) or (dep_path.is_dir() and pip.utils.is_installable_dir(dep)):
-            dep_link = Link(dep_path.absolute().as_uri())
-            if dep_path.is_absolute() or dep_path.as_posix() == '.':
-                path = dep_path.as_posix()
-            else:
-                path = get_converted_relative_path(dep)
-            dep = dep_link.egg_fragment if dep_link.egg_fragment else dep_link.url_without_fragment
+        dep_link = Link(dep_path.absolute().as_uri())
+        if dep_path.is_absolute() or dep_path.as_posix() == '.':
+            path = dep_path.as_posix()
+        else:
+            path = get_converted_relative_path(dep)
+        dep = dep_link.egg_fragment if dep_link.egg_fragment else dep_link.url_without_fragment
     elif is_vcs(dep):
         # Generate a Link object for parsing egg fragments
         dep_link = Link(dep)
@@ -339,8 +337,8 @@ def get_requirement(dep):
             else:
                 req.uri = dep_link.url_without_fragment
     # If the result is a local file with a URI and we have a local path, unset the URI
-    # and set the path instead
-    elif req.local_file and req.uri and not req.path and path and not req.vcs:
+    # and set the path instead -- note that local files may have 'path' set by accident
+    elif req.local_file and path and not req.vcs:
         req.path = path
         req.uri = None
     elif req.vcs and req.uri and uri != req.uri:
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -282,6 +282,7 @@ def get_requirement(dep):
     :returns: :class:`requirements.Requirement` object
     """
     path = None
+    uri = None
     # Split out markers if they are present - similar to how pip does it
     # See pip.req.req_install.InstallRequirement.from_line
     if not any(dep.startswith(uri_prefix) for uri_prefix in SCHEME_LIST):
@@ -314,20 +315,19 @@ def get_requirement(dep):
 
     elif is_valid_url(dep) and matches_uri:
         dep_link = Link(dep)
+        uri = dep_link.url_without_fragment
         # Parse the requirement using just the dependency name version from the egg fragment
         # if possible. Then we can drop in the URI later.  This is how pip does it.
         dep = dep_link.egg_fragment if dep_link.egg_fragment else dep_link.url_without_fragment
-        if dep_link.egg_fragment:
-            path = dep_link.url_without_fragment
     req = [r for r in requirements.parse(dep)][0]
     # If the result is a local file with a URI and we have a local path, unset the URI
     # and set the path instead
-    if path and not req.path and not matches_uri:
+    if path and not req.path:
         req.path = path
         req.uri = None
         req.local_file = True
-    elif matches_uri and path and not req.uri:
-        req.uri = path
+    elif matches_uri and uri and not req.uri:
+        req.uri = uri
     if markers:
         req.markers = markers
     if extras:
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -297,23 +297,37 @@ def get_requirement(dep):
         markers = None
     # Strip extras from the requirement so we can make a properly parseable req
     dep, extras = pip.req.req_install._strip_extras(dep)
+    matches_uri = any(dep.startswith(uri_prefix) for uri_prefix in SCHEME_LIST)
     # Only operate on local, existing, non-URI formatted paths
-    if (is_file(dep) and isinstance(dep, six.string_types) and
-            not any(dep.startswith(uri_prefix) for uri_prefix in SCHEME_LIST)):
+    if is_file(dep) and isinstance(dep, six.string_types) and not matches_uri:
         dep_path = Path(dep)
         # Only parse if it is a file or an installable dir
         if dep_path.is_file() or (dep_path.is_dir() and pip.utils.is_installable_dir(dep)):
-            if dep_path.is_absolute() or dep_path.as_posix() == '.':
-                path = dep_path.as_posix()
-            else:
-                path = get_converted_relative_path(dep)
-            dep = dep_path.resolve().as_uri()
+            # Create pip Link objects for obtaining fragments and spotting wheels
+            dep_link = Link(dep_path.absolute().as_uri())
+            if dep_path.is_dir() or dep_link.is_wheel or is_archive_file(dep_path.as_posix()):
+                if dep_path.is_absolute() or dep_path.as_posix() == '.':
+                    path = dep_path.as_posix()
+                else:
+                    path = get_converted_relative_path(dep)
+                dep = dep_link.egg_fragment if dep_link.egg_fragment else dep_link.url_without_fragment
+
+    elif is_valid_url(dep) and matches_uri:
+        dep_link = Link(dep)
+        # Parse the requirement using just the dependency name version from the egg fragment
+        # if possible. Then we can drop in the URI later.  This is how pip does it.
+        dep = dep_link.egg_fragment if dep_link.egg_fragment else dep_link.url_without_fragment
+        if dep_link.egg_fragment:
+            path = dep_link.url_without_fragment
     req = [r for r in requirements.parse(dep)][0]
     # If the result is a local file with a URI and we have a local path, unset the URI
     # and set the path instead
-    if req.local_file and req.uri and not req.path and path:
+    if path and not req.path and not matches_uri:
         req.path = path
         req.uri = None
+        req.local_file = True
+    elif matches_uri and path and not req.uri:
+        req.uri = path
     if markers:
         req.markers = markers
     if extras:
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -542,7 +542,7 @@ def resolve_deps(deps, which, which_pip, project, sources=None, verbose=False, p
     markers_lookup = {}
 
     python_path = which('python', allow_global=allow_global)
-    backup_python_path = shellquote(sys.executable)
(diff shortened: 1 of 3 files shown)


techalchemy:
@nateprewitt once this is merged it should actually make the #1364 pass (after a rebase) -- I rebased it from the main 9.x rebuild branch and I believe it included a test that belonged in this branch (the test that is now failing over there).  This is the last set of changes we are still missing with an additional fix + test case to catch a regression over VCS requirements being included as paths

pipenv/utils.py
```diff
@@ -282,6 +284,14 @@ def get_requirement(dep):
     :returns: :class:`requirements.Requirement` object
     """
     path = None
+    uri = None
+    editable = False
+    dep_link = None
+    # check for editable dep / vcs dep
+    if isinstance(dep, six.string_types) and dep.startswith('-e '):
```
nateprewitt: Are we ever passing in non-str types here? We do another `startswith` check immediately below that will fail if that's the case.

pipenv/utils.py
```diff
@@ -297,23 +307,46 @@ def get_requirement(dep):
         markers = None
     # Strip extras from the requirement so we can make a properly parseable req
     dep, extras = pip.req.req_install._strip_extras(dep)
-    # Only operate on local, existing, non-URI formatted paths
-    if (is_file(dep) and isinstance(dep, six.string_types) and
-            not any(dep.startswith(uri_prefix) for uri_prefix in SCHEME_LIST)):
+    matches_uri = any(dep.startswith(uri_prefix) for uri_prefix in SCHEME_LIST)
+    # Only operate on local, existing, non-URI formatted paths which are installable
+    if is_file(dep) and isinstance(dep, six.string_types) and not matches_uri and is_installable_file(dep):
```
nateprewitt: So this block is getting a bit out of control. Let's try to simplify a little.

 If something is an "installable file" is it not inherently a "file"? Do we need both checks here?

Are we mutating `dep` to be a non-string above? If it is, are we certain whatever type it's being converted to is still a representation of an installable file?

pipenv/utils.py
```diff
@@ -297,23 +307,46 @@ def get_requirement(dep):
...
+        # Generate a Link object for parsing egg fragments
+        dep_link = Link(dep)
+        # Save the original path to store in the pipfile
+        uri = dep_link.url
+        # Construct the requirement using proper git+ssh:// replaced uris or names if available
+        dep = clean_vcs_uri(dep)
+    if editable:
+        dep = '-e {0}'.format(dep)
     req = [r for r in requirements.parse(dep)][0]
+    # if all we built was the requirement name and still need eveerything else
```
nateprewitt: everything*

pipenv/utils.py
```diff
@@ -297,23 +307,46 @@ def get_requirement(dep):
...
+            else:
+                req.uri = dep_link.url_without_fragment
     # If the result is a local file with a URI and we have a local path, unset the URI
-    # and set the path instead
-    if req.local_file and req.uri and not req.path and path:
+    # and set the path instead -- note that local files may have 'path' set by accident
+    elif req.local_file and path and not req.vcs:
         req.path = path
         req.uri = None
+    elif req.vcs and req.uri and uri != req.uri:
```
nateprewitt: I believe this was a special case that pip handles? Does it not apply to svn or mercurial? Is there are reason why pip won't just handle this for us when we pass it one format?

pipenv/utils.py
```diff
@@ -297,23 +307,46 @@ def get_requirement(dep):
...
+        if req.uri.startswith('git+ssh://') and uri.startswith('git+git'):
+            req.uri = req.uri.replace('git+ssh://', 'git+')
+            req.line = req.line.replace('git+ssh://', 'git+')
+    if editable and not req.editable:
```
nateprewitt: May be simpler to just do `req.editable = editable`, no? Is there a case where we'd want this mutating during parsing?

pipenv/utils.py
```diff
@@ -536,7 +569,7 @@ def resolve_deps(deps, which, which_pip, project, sources=None, verbose=False, p
     markers_lookup = {}
 
     python_path = which('python', allow_global=allow_global)
-    backup_python_path = shellquote(sys.executable)
+    backup_python_path = sys.executable
```
nateprewitt: Was this related to this PR, or something bundled in with merging the other missing commits?

pipenv/utils.py
```diff
@@ -627,7 +660,7 @@ def convert_deps_from_pip(dep):
     extras = {'extras': req.extras}
 
     # File installs.
-    if (req.uri or req.path or (os.path.isfile(req.name) if req.name else False)) and not req.vcs:
+    if (req.uri or req.path or (is_installable_file(req.name) if req.name else False)) and not req.vcs:
```
nateprewitt: `is_installable_file` should probably handle None and empty strings so we're not doing checks like this.

pipenv/utils.py
```diff
@@ -297,23 +307,46 @@ def get_requirement(dep):
...
+                req.uri = dep_link.url_without_fragment
     # If the result is a local file with a URI and we have a local path, unset the URI
-    # and set the path instead
-    if req.local_file and req.uri and not req.path and path:
+    # and set the path instead -- note that local files may have 'path' set by accident
+    elif req.local_file and path and not req.vcs:
         req.path = path
         req.uri = None
+    elif req.vcs and req.uri and uri != req.uri:
+        if req.uri.startswith('git+ssh://') and uri.startswith('git+git'):
```
nateprewitt: Can we run this on the URI above when we call `clean_vcs_uri` so `uri` and `dep` are both normalized? That way we're not doing the same process twice.

tests/test_pipenv.py
```diff
@@ -602,12 +626,12 @@ def test_specific_package_environment_markers(self):
     @pytest.mark.tablib
     def test_install_editable_git_tag(self):
         with PipenvInstance() as p:
-            c = p.pipenv('install -e git+git://github.com/kennethreitz/tablib.git@v0.12.1#egg=tablib')
+            c = p.pipenv('install -e git+https://github.com/kennethreitz/tablib.git@v0.12.1#egg=tablib')
```
nateprewitt: Why are we changing this? If the user types `git+git://` we should still be handling it the same.

tests/test_utils.py
```diff
@@ -250,3 +250,28 @@ def test_win_normalize_drive(self, input_path, expected):
     @pytest.mark.skipif(os.name == 'nt', reason='*nix file paths tested')
     def test_nix_normalize_drive(self, input_path, expected):
         assert pipenv.utils.normalize_drive(input_path) == expected
+
+    @pytest.mark.requirements
+    def test_get_requirements(self):
```
nateprewitt: Can we get some divisions into conceptual testing blocks and minor comments for what we're confirming/refuting here?

techalchemy:
Thanks for taking the time to look through this in such detail, @nateprewitt, it's nice to have extra scrutiny at this depth.  A lot of the specific questions you asked will kind of underscore much larger underlying issues with the complexity that has evolved around some seemingly harmless functions

To respond broadly:

Definitely agree it needs a refactor, but I didn't think this PR was the place for it.  I wanted to get this working again with the PRs that we had in 9.0.1 (which is mostly what this is, plus a small fix for vcs dependencies).  There is a massive amount of complexity built into things that should definitely be separated out, even I am struggling to keep up with a lot of this stuff now.  Given that we have no real standard for separation of functionality yet, I was reluctant to introduce it here

I want to note that without these changes currently, aspects of the resolver are broken (the related issues are linked in the referenced PRs). So while I believe we can and should refactor this code, I would push for a roadmap to include a refactor in a minor release potentially, while getting this functionality patched asap.  There have been quite a few reports related to these so I feel it would be good to fix the behavior

I'll go through and get the small stuff you mentioned, add some comments where things are unclear, and maybe we can go from there with some additional feedback from another maintainer

pipenv/utils.py
techalchemy: Not intentionally, no. This is a good catch

pipenv/utils.py
techalchemy: 1. I had a long answer typed up here until I reread this more carefully and realized you are right. And in fact I am precisely reimplementing the logic that differentiates the two as well.

2. No we never convert `dep` to be a non-string, that was why I introduced additional variables to split out relevant information as needed

pipenv/utils.py
techalchemy: ah yes, this is the root of all of my frustration with this code. We are roughly reimplementing the logic out of the pip codebase with minor adjustments.  We used to just directly call pip's version of the parser but it actually triggers a download or something? Can't remember what Kenneth said. If we passed this to pip it would be a different story, but we pass it to `requirements.parse()` which does absolutely no handling of anything

pipenv/utils.py
techalchemy: This was a PR, #1241 -- since the python path here is never actually used as an argument to a subprocess call, it doesn't make sense to shellquote it (and it was breaking something for someone)

pipenv/utils.py
techalchemy: this is mainly just a metadata alteration -- this way the requirement itself is generated correctly (see above comments about not using pip / having to do all of this stuff ourselves) but that means that a user who runs `pipenv install git+git@github.com:someuser/somerepo.git` will see a transformed url when they read their pipfile.  First we transform that to `git+ssh` notation (per pip's translation) -- this transforms it back to what they originally supplied on install.

tests/test_pipenv.py
techalchemy: Apparently `git+git://` is a deprecated protocol

tests/test_utils.py
techalchemy: done.

kennethreitz:
agreed with @techalchemy

techalchemy:
@nateprewitt I was able to address most of the things you mentioned and tests are now passing

This code still sucks and needs to be refactored

nateprewitt:
✨

PR #1394, opened 2018-02-03
Removes "check --style" command, closes #1340

### LINKED ISSUE #1340: Consider removing flake8 support
the less code, the better!

sobolevn:
I had a lot of troubles trying to run tests on my mac.
In the end I had not any luck running it.

Related issue: #1340 

Big thanks to @techalchemy for his help in trying to set up everything properly.
Btw, @techalchemy how do you think should I create a separate issue about tests failures?

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -167,20 +167,6 @@ Example::
 
 ✨🍰✨
 
-☤ Code Style Checking
----------------------
-
-Pipenv has `Flake 8 <http://flake8.pycqa.org/en/latest/>`_ built into it. You can check the style of your code like so, without installing anything::
-
-    $ cat t.py
-    import requests
-
-    $ pipenv check --style t.py
-    t.py:1:1: F401 'requests' imported but unused
-    t.py:1:16: W292 no newline at end of file
-
-Super useful :)
-
 ☤ Open a Module in Your Editor
 ------------------------------
 
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -2036,8 +2036,14 @@ def do_check(three=None, python=False, unused=False, style=False, args=None):
         args = []
 
     if style:
-        sys.argv = ['magic', project.path_to(style)] + list(args)
-        flake8.main.cli.main()
+        # TODO: remove `--style` argument completely until version 10.0.0
+        message = (
+            'Style-checking has been removed from Pipenv. '
+            'You can perform the checking without '
+            'Pipenv involved by adding Flake8 directly '
+            'to your dev-packages, and call `pipenv run flake8` instead.'
+        )
+        click.echo(crayons.red(message))
         exit()
 
     if unused:
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -23,7 +23,6 @@ import requests
 import pipfile
 import pipdeptree
 import semver
-import flake8.main.cli
 from pipreqs import pipreqs
 from blindspin import spinner
 from urllib3.exceptions import InsecureRequestWarning
@@ -2044,7 +2043,7 @@ def do_check(three=None, python=False, unused=False, style=False, args=None):
             'to your dev-packages, and call `pipenv run flake8` instead.'
         )
         click.echo(crayons.red(message))
-        exit()
+        exit(1)
 
     if unused:
         deps_required = [k for k in project.packages.keys()]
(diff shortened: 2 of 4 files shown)


techalchemy:
@sobolevn thanks for putting this together and yeah I would say it would be good to open an issue on that. Because of the nature of pipenv the tests will always be a struggle if you install in a virtualenv but we should document the process for people who want to develop

nateprewitt:
Hey @sobolevn, thanks for this! So I gave it a little more thought, and while this is a fairly broken feature, there may still be users relying on it. I'd like to do one more stable release some time this coming week (9.1.0) with a deprecation warning on `--style`. Then we'll merge this for either 9.2.0 or 10.0.

Would you mind putting together one more quick PR to add a `warnings.warn("<warning message>", DeprecationWarning)` to the `style` conditional?

sobolevn:
@nateprewitt ok, should I include any specific version into the deprecation message?

nateprewitt:
If we're following strict semantic versioning, we should do it for 10.0 but I don't think we really want to release another major version immediately after a previous one with few major changes. Due to the complaints around this and it not being maintainable, I think we may fuzz the rules a little bit and say that it will be deprecated in 9.1.0 and removed in 9.2.0.

@techalchemy @kennethreitz?

techalchemy:
@nateprewitt I would tend to agree especially considering it's another aspect which is broken across the virtualenv boundary for python version mismatches

uranusjr:
While I agree on the semantic versioning point, I think the essence of this feature makes it possible to justify an exception. The feature removal is very to ammend by either invoking the Flake8 command, or installling flake8 to the virtualenv, and run it instead. This should be quite straightforward to everyone already using this feature. For anyone having problems switching to Flake8 directly, their code probably already has something wrong with it (e.g. mismatched version syntaxes) that *should* be exposed, but escaped ddetection with `pipenv check --style`. There is really no downside forcing users to move away from this as soon as possible.

I would think it is enough to remove the linting feature outright, but keep the `--style` flag until the next major release to show a clear migration message. Something like

```
Style-checking has been removed from Pipenv. You can perform the checking without Pipenv involved by adding Flake8 to your dev-packages, and run "pipenv run flake8" instead.
```

nateprewitt:
@uranusjr I like the idea of keeping the `--style` flag intact until 10.0. That's probably the best way to get the information to people using the feature in its current state, and hopefully reduce issue churn from it being removed.

@sobolevn I'm rerunning the tests on #1395 and will merge once they pass. When that's done would you mind rebasing this onto current master and modifying it with @uranusjr's suggested warning? Then if we can add a TODO comment to remove it with 10.0, I think we'll be set. Thanks again for putting all of this together, it's really appreciated!

sobolevn:
@nateprewitt So, to clarify things:
1. We are merging #1395 as-is
2. I will rebase #1394 onto `master`
3. I am refactoring #1394 to still have `--style` flag, which does nothing but shows a warning from https://github.com/pypa/pipenv/pull/1394#issuecomment-362890813
4. I am adding a `TODO` to remove it completely until `10.0`
5. Pushing it here

Did I miss something?

nateprewitt:
@sobolevn yep, that sounds great. Thanks!

sobolevn:
@nateprewitt done.
I have closed https://github.com/pypa/pipenv/pull/1395 too.

kennethreitz:
Needs a rebase. 

I added a deprecation warning in https://github.com/pypa/pipenv/commit/192306cb24b70f822d32403981b2dbd3c05bf017.

sobolevn:
@kennethreitz I thought that https://github.com/pypa/pipenv/commit/192306cb24b70f822d32403981b2dbd3c05bf017#diff-ec74dda0078733c5b6755671cacf1de5R2028 and https://github.com/pypa/pipenv/commit/192306cb24b70f822d32403981b2dbd3c05bf017#diff-ec74dda0078733c5b6755671cacf1de5R2028 are mutually exclusive. 

Which one should we stick to?

kennethreitz:
you linked to the same diff twice :)

sobolevn:
Fixed.

1. https://github.com/pypa/pipenv/pull/1394/files#diff-ec74dda0078733c5b6755671cacf1de5R2038 2. https://github.com/pypa/pipenv/commit/192306cb24b70f822d32403981b2dbd3c05bf017#diff-ec74dda0078733c5b6755671cacf1de5R2028

kennethreitz:
updated it to say 10.0.0

PR #1397, opened 2018-02-04
lazy load pip; refactor cli

kennethreitz:
Signed-off-by: Kenneth Reitz <me@kennethreitz.org>


nateprewitt:
@kennethreitz I played around with this a couple months ago without success. I think you'll also need to move the imports in pipenv.utils since that will load all of pip when it gets imported.

nateprewitt:
@kennethreitz, sorry I didn't catch this in the last glance. It looks like `requirements` has the same issue in utils. Hopefully this will mitigate the problem for now and then we may want to look at refactoring out pip usage to a single interface. That way we can remove direct pip imports from the cli.py file and having imports in several places. I think we also have a fairly significant import time with urllib3 to just do warning suppression which might need a look.

PR #1422, opened 2018-02-12
Fix: shellquote src path

bradsbrown:
Re-work former PR https://github.com/pypa/pipenv/pull/1204 to get the fix back into the new `master` branch using the `shellquote` helper, as mentioned in https://github.com/pypa/pipenv/pull/1361#issuecomment-360797751.

### DIFF
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -1344,7 +1344,7 @@ def pip_install(
 
         # Don't specify a source directory when using --system.
         if not allow_global and ('PIP_SRC' not in os.environ):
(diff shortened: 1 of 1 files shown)


techalchemy:
@bradsbrown thanks for the PR, good catch! Travis had a transient failure so I restarted the test but this does get passed down so it makes sense to be quoting this as it will be a path and it will be passed as an argument to a subprocess call

bradsbrown:
the py3.4 Travis test _really_ doesn't like me for some reason, it appears.

PR #1426, opened 2018-02-14
Altered regex in dotenv.main.parse_dotenv to handle single quotes lik…

### LINKED ISSUE #1418: Update internal dotenv (or make it a dependency)
Variables in a `.env` file that have single-quoted spaces are not loaded into the environment. If double quotes are used, then the variables are loaded into the environment.

1. ubuntu 17.10
1. Python version: `$ python -V` - 3.6.4
1. Pipenv version: `$ pipenv --version` - 9.0.3

##### Expected result

All variables in `.env` would be loaded

##### Actual result

Variables in `.env` that are single-quoted and contain spaces are not loaded

##### Steps to replicate

1. Given the following `.env` file:

```
FOO='Hello World'
BAR='Test'
BAZ="Hello World"
```

2. Running `pipenv shell`
3. `FOO` will not be set in the environment.
4. This happens because the `FOO` line doesn't match the regex used (`'^\s*(\w*)\s*=\s*("[^"]*"|[^\s]*)\s*$'`) in the parsing algorithm. 

##### Additional information

I tracked this issue down by stepping through the dotenv loading code used by pipenv. When I found this line of code, I learned that only double quotes would allow spaces:

https://github.com/pypa/pipenv/blob/master/pipenv/patched/dotenv/main.py#L100

This parsing algorithm diverges a bit from the official dotenv logic:

https://github.com/theskumar/python-dotenv/blob/master/dotenv/main.py#L98

danielhones:
…e double quotes.  Fixes #1418

### DIFF
--- a/pipenv/patched/dotenv/main.py
+++ b/pipenv/patched/dotenv/main.py
@@ -10,6 +10,8 @@ from collections import OrderedDict
 
 __escape_decoder = codecs.getdecoder('unicode_escape')
 __posix_variable = re.compile('\$\{[^\}]*\}')
+__variable_declaration = re.compile('^\s*(\w*)\s*=\s*("[^"]*"|\'[^\']*\'|[^\s]*)\s*$',
+                                    flags=re.MULTILINE)
 
 
 def decode_escaped(escaped):
@@ -97,7 +99,7 @@ def dotenv_values(dotenv_path):
 
 def parse_dotenv(dotenv_path):
     with open(dotenv_path) as f:
-        for k, v in re.findall('^\s*(\w*)\s*=\s*("[^"]*"|[^\s]*)\s*$', f.read(), flags=re.MULTILINE):
+        for k, v in __variable_declaration.findall(f.read()):
             if len(v) > 0:
                 quoted = v[0] == v[len(v) - 1] in ['"', "'"]
 
--- a/pipenv/patched/dotenv/test_main.py
+++ b/pipenv/patched/dotenv/test_main.py
@@ -1,0 +1,85 @@
+import os
+from textwrap import dedent
+import unittest
+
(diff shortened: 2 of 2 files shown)


danielhones:
I've looked at the Travis build and saw that one job timed out and another job errored, which also happened in [PR 1427](https://github.com/pypa/pipenv/pull/1427) where the same jobs failed.  The failure on Python 3.6 was surprising because that's what I use on my local machine and the tests passed there.  Both the failed Travis and AppVeyor jobs printed this error message:

```
piptools.cache.CorruptCacheError: The dependency cache seems to have been corrupted.
Inspect, or delete, the following file:
  /home/travis/.cache/pipenv/depcache-py3.6.json  # for Travis
  C:\Users\appveyor\AppData\Local\pipenv\pipenv\Cache\depcache-py3.6.json  # for AppVeyor
```

Looking through the code I don't see how it would be caused by the changes on this branch, but I'm admittedly unfamiliar with the codebase.  I'm happy to troubleshoot this further, but I could use a pointer in the right direction if it's something wrong with the code change and not the build process or environment.

danielhones:
I'm not sure what happened with the Github/Travis integration, but a different Travis build ran for this PR and passed: https://travis-ci.org/pypa/pipenv/builds/342605827.  There's now a different error in the AppVeyor build:

```
>       assert 'Flask' in p['packages']
E       AssertionError: assert 'Flask' in {'flask': '*'}
```

Tracing through the code, it looks like this could happen if [this line](https://github.com/pypa/pipenv/blob/master/pipenv/utils.py#L862) in `recase_file` was hit due to a non-OK response when requesting the package from PyPi, and it would use the lower case "flask" as the key instead.  If that's the case then I think it should be okay to consider that test passing.

PR #1427, opened 2018-02-14
Add a utility method for setup.py commands.

phildini:
This commit adds a `pipenv setuppy` command that provides the functionality
of `pipenv run setup.py {command}`.

Running variations on `python setup.py {command}` is something that I end up doing a lot as a package maintainer. Right now, I either need to be doing all of my work in a `pipenv shell`, defeating part of the usefulness of pipenv for me, or I'm typing `pipenv run python setup.py` a lot. This hopefully makes it easier for package maintainers and those who are less requirements.txt-based to use pipenv more easily.

I'm open to suggestions for how to test, better document, or really change in any way. I want some form of this to exist so pipenv and setuptools can live in harmony, the actual syntax matters less to me.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -88,6 +88,7 @@ Other Commands
 
 - ``shell`` will spawn a shell with the virtualenv activated.
 - ``run`` will run a given command from the virtualenv, with any arguments forwarded (e.g. ``$ pipenv run python``).
+- ``setuppy`` will run a setup.py command from the virtualenv (e.g. ``$ pipenv setuppy test``).
 - ``check`` asserts that PEP 508 requirements are being met by the current environment.
 - ``graph`` will print a pretty graph of all your installed dependencies.
 
@@ -158,6 +159,7 @@ Fish is the best shell. You should use it.
       lock       Generates Pipfile.lock.
       open       View a given module in your editor.
       run        Spawns a command installed into the...
+      setuppy    Runs ``python setup.py {command}`` inside the virtualenv.
       shell      Spawns a shell within the virtualenv.
       uninstall  Un-installs a provided package and removes it...
       update     Uninstalls all packages, and re-installs...
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -308,6 +308,25 @@ def run(command, args, three=None, python=False):
     core.do_run(command=command, args=args, three=three, python=python)
 
 
+@click.command(
+    add_help_option=False,
+    short_help="Runs `python setup.py \{command\}` inside the virtualenv.",
+    context_settings=dict(
+        ignore_unknown_options=True,
+        allow_interspersed_args=False,
+        allow_extra_args=True
+    )
+)
+@click.argument('command')
+@click.argument('args', nargs=-1)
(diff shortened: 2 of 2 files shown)


uranusjr:
What is the benefit of this over `pipenv run python setup.py`, aside from being shorter?

phildini:
Here's my pattern of thinking, after watching @kennethreitz's talk about Python for Humans, his talk on Pipenv, and seeing this on the Pipenv homepage:

> Pipenv is a tool that aims to bring the best of all packaging worlds (bundler, composer, npm, cargo, yarn, etc.) to the Python world. Windows is a first–class citizen, in our world.

npm is, in many ways, a very user-friendly package manager, especially around running tests, running scripts, and publishing packages. `npm test` and `npm publish` and `npm start` are all common commands.

For most python package developers right now, `python setup.py test` and `python setup.py sdist && python setup.py bdist_wheel --universal` are the closest we get to such simple commands.

`pipenv` is amazing, I want to see it become _the_ defacto way to work with python packages, and I think that means adding convenience methods for common patterns.

I'm 100% open to the idea that there is a better way to accomplish what I'm suggesting. Maybe the pipfile should describe a `test` script? Maybe there's a better name than `setuppy`?

But today, trying to bridge the world of packages relying heavily on `setup.py` while using pipenv for working with those packages means I'm constantly typing `pipenv run python setup.py {command}`, and after a few dozen times typing those ten extra characters, I start to wonder how much we're asking of package maintainers in using pipenv.

Hence, this attempt at a solution.

techalchemy:
Hey @phildini let me start by saying thanks for the pull request. I’m glad you listened to one of Kenneth’s talks and were moved enough to get involved

So if you were to answer the question put forward by @uranusjr would you say that this PR does something beyond removing the word `python` from the pipenv invocation? We can discuss the merits and scope of pipenv as a whole but that belongs in an issue (typically where requests for API changes should start as well), but the proposed change isn’t focused on those things. Let’s keep things focused here: if the problem in question is that your invocation is too long, I’d probably nudge you toward Makefiles or invoke scripts rather than API changes

phildini:
I don't currently believe this is a problem only I am having. If I thought I was the only Python developer trying to use `setup.py` with `pipenv`, then I would happily write my own custom scripts and aliases to make this work locally. I believe that other Python package maintainers are having this same issue.

Maybe it doesn't but them as much to be verbose! That's totally possible.

It sounds like I'm being redirected to discussing this in an issue; if that's the consensus of the maintainers on the project feel free to close this and I'll open an issue. I often view a PR as an Issue + an example of how to fix it.

kennethreitz:
see pipenvlib

kennethreitz:
closing because this adds a new command, which is a big no-no

techalchemy:
see #1433

phildini:
#1433 definitely looks neat! 

@kennethreitz Thanks for the context on adding commands being bad. How would I have learned this outside of this PR? Is there a place I can document design philosophy so the next person has more of a sense of what the project is looking for? I read the docs, and the github wiki, and the contributor's guide, and didn't see anything.

Thanks!

PR #1433, opened 2018-02-17
scripts

kennethreitz:
Signed-off-by: Kenneth Reitz <me@kennethreitz.org>


kennethreitz:
that would be a dumb user move :)

kennethreitz:
i think we can stick with being pretty naive

nateprewitt:
Yeah, I’m not saying the behavior is wrong, I’m asking if we want documentation a.) showing scripts usage and b.) calling out that scripts takes precedence over virtualenv binaries.

PR #1440, opened 2018-02-17
add documentation to makefile

dschaller:


Makefile
```diff
@@ -1,12 +1,21 @@
...
+help:
+	@grep -E '^\.PHONY: [a-zA-Z_-]+ .*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = "(: |##)"}; {printf "\033[36m%-30s\033[0m %s\n", $$2, $$3}'
 
+.PHONY: run-tests ## Run unit tests
 run-tests:
 	pipenv run pytest tests
+
+.PHONY: init ## Initialize pipenv for development
 init:
 	python setup.py install
```
jtratner: aside - your description makes me think this should be `develop` rather than `install`

dschaller: Ah good point. I'll update the description to be something like `Initialize pipenv with dev dependencies`. Does that make more sense?

jtratner: not 100% sure of the original intent here :) I actually think your description is correct and `python setup.py install` --> `python setup.py develop`

PR #1442, opened 2018-02-17
Add instructions for conda plus pipenv

jtratner:
Figured it'd be a good idea to have a standard answer for "How do I use this with conda?"

Left out any commentary about conflicts between conda and Pipenv about requirements... hoping that's obvious to folks who'd use both.

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -78,6 +78,14 @@ Also useful for deployment is the ``--deploy`` flag::
 This will fail a build if the ``Pipfile.lock`` is out–of–date, instead of generating a new one.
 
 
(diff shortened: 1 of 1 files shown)


kennethreitz:
this is not how i'd recommend using conda and pipenv together.

jtratner:
gotcha - do you have a recommended way of doing it?

Maybe:

```
source activate myenvironment
pipenv install --python=$(which python)
```

would be helpful to know / have a recommended way to handle - actively exploring using pipenv + conda at work right now.

kennethreitz:
that second one, without the source activate part :)

jtratner:
@kennethreitz - ptal

PR #1447, opened 2018-02-18
MIsc. comment and doc typos

luzpaz:
Found via `codespell -q 3 --skip="./patched"`


jtratner:
Thanks for the PR!

Would you mind removing the changes from the pipenv/vendor directory?

While it’s nice to fix typos, since those are vendored libraries it’s
better to leave them as-is so they match their sources.

After that change - good to merge :)

luzpaz:
@jtratner 
reverted and squashed.

nateprewitt:
Thanks for this @luzpaz!

PR #1388, opened 2018-02-02
Updates `activate_virtualenv`, closes #1371

sobolevn:
`.` is actually a `source` alias. But it works better, since `source` is not supported in raw `sh`.
As it was discussed in https://github.com/pypa/pipenv/issues/1371 some systems do not have `bash` installed.
Prove link: https://askubuntu.com/questions/25488/what-is-the-difference-between-source-and

PR #1458, opened 2018-02-21
Update activate_virtualenv again for Fish 3.0+

uranusjr:
Fish 3 removes the `.` command in favour of `source`, so we need to adapt.

See https://github.com/pypa/pipenv/pull/1388#issuecomment-367353694


techalchemy:
Tests are not your friend today. The changes look good, any idea about the travis failures on 2.7?

uranusjr:
Is it related to #1314? What is the canonical way to denote packages in Pipfile now anyway, if any?

PR #1461, opened 2018-02-21
sketch of basic pipenv show impl

dradetsky:
Basic WIP sketch of addition of `pipenv show PKG_NAME` functionality, analogous the the command in pip.

I'm aware of some things that ought to be added to this to make it acceptable for merging. I made these changes as part of a quick experiment in determining whether I could easily do it before I opened an issue asking whether such a feature was welcome. Since I got this far, I figured I'd show my work.

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -390,6 +390,14 @@ def update(
     )
 
 
+@click.command(short_help="Show info about installed packages.")
+@click.argument('package_name', default=False)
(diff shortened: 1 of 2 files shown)


ncoghlan:
As per the discussion on #1462, we think it's better to improve the documentation around using `pipenv run pip ...` to access `pip` subcommands beyond those directly supported by `pipenv`.

PR #1464, opened 2018-02-22
Don't change working directory in pipenv shell

### LINKED ISSUE #1459: pipenv shell command changes directory to root of repo
When activating pipenv shell at some folder in the repo with a Pipfile at root it will activate the shell and move you to the folder containing the Pipfile

1. Linux/Mac OSX
1. Python 2.7.12
1. pipenv, version 9.0.3

##### Expected result

Stay in the same directory as you were when running ```pipenv shell```

##### Actual result

Directory is changed to the root of the repo

##### Steps to replicate

1. cd any number of levels deep into a repo with a Pipfile at the root
2. type ```pipenv shell```
3. Observe that working directory is now changed to the same on that contains the Pipfile

danielhones:
Fixes #1459.  This probably isn't the cleanest place to put the fix, but since the intent for `pipenv run` seems similar to how pipenv shell should behave, I added it to the same place.

### DIFF
--- a/pipenv/project.py
+++ b/pipenv/project.py
@@ -40,8 +40,8 @@ class Project(object):
         self._requirements_location = None
         self._original_dir = os.path.abspath(os.curdir)
 
-        # Hack to skip this during pipenv run, or -r.
-        if ('run' not in sys.argv) and chdir:
+        # Hack to skip this during pipenv run, -r, or pipenv shell.
+        if ('run' not in sys.argv) and ('shell' not in sys.argv) and chdir:
             try:
                 os.chdir(self.project_directory)
             except (TypeError, AttributeError):
--- a/tests/test_pipenv.py
+++ b/tests/test_pipenv.py
@@ -11,6 +11,7 @@ from pipenv.core import activate_virtualenv
 from pipenv.utils import temp_environ, get_windows_path, mkdir_p, normalize_drive
 from pipenv.vendor import toml
 from pipenv.vendor import delegator
+from pipenv.vendor.pexpect import pty_spawn
 from pipenv.project import Project
 try:
     from pathlib import Path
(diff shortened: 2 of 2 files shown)


techalchemy:
I agree this isn't pretty, it does seem to be where we handle this currently.  I am willing to go ahead with this with a test to prevent regression, but I'd be interested in hearing alternative approaches or reviewing refactors as well. It seems like the logic to change working directories during CLI interaction probably should be separate from the project itself (given our current structure)

danielhones:
Sure thing, this commit has tests to check the working directory after pipenv run and pipenv shell.  I was hoping they would also pass on Windows but it looks like they failed so I need to set up a Windows VM to work on them, as well as see why they failed on Travis.

As for alternative approaches to handling the directory change, I think a good start would be to move that part out of `Project.__init__` and into a separate method.  I'm still new to this codebase so I don't know when or why we would need the directory to change when initializing the object, but a few greps of the repo makes it look like Project(chdir=False) isn't used at all.

I have in mind something like this:

```python
# in pipenv/project.py:

class Project(object):
    """docstring for Project"""

    def __init__(self):
        super(Project, self).__init__()
        self._name = None
        self._virtualenv_location = None
        self._download_location = None
        self._proper_names_location = None
        self._pipfile_location = None
        self._requirements_location = None
        self._original_dir = os.path.abspath(os.curdir)

    def change_directory(self):
        try:
            os.chdir(self.project_directory)
        except (TypeError, AttributeError):
            pass
```

And if it would nice to use it this way, we could make a decorator for the core.py functions that need the directory changed like this:

```python
# in pipenv/core.py:

project = Project()

def change_directory(f):
    """Decorator for core functions that need the working directory
    changed to the project's root."""
    def call_f(*args, **kwargs):
        project.change_directory()
        return f(*args, **kwargs)
    return call_f
```

I think the major drawback of this is that it wouldn't be backwards compatible for someone using the Project class outside of this repo.

kennethreitz:
shell is intended to work this way.

keunhong:
> shell is intended to work this way.

This is such an unfortunate attitude

slhck:
I find myself constantly `cd`ing back in a nested folder hierarchy, and `cd -` does not work. I'd also like to ask about reconsidering at least having an *option* for `pipenv shell` to put the user back into the directory where they were before launching the command. Such as `pipenv shell --keep` to keep the `pwd`.

dashesy:
this is the most confusing aspect of `pipenv shell`! It certainly more looks like a bug

kenneth-reitz:
we can always revisit anything :)

piroux:
What are the blocking points to revisit this PR which would remove many daily unnecessary keystrokes for a lot of people ?

btonasse:
Can we reconsider this? This forced change of directory is infuriating

Issue #1455, opened 2018-02-21
pipenv graph json yields stacktrace, pipenv graph merely exits


mdaniel:
- [X] Be sure to check the existing issues, [both open and closed](https://github.com/pypa/pipenv/issues?utf8=%E2%9C%93&q=graph+json)

_Describe the issue briefly here_

Running `pipenv graph --json` without a `pipenv` virtualenv yields a very ugly stacktrace; `pipenv graph` yields no such stacktrace, just empty output

##### Describe your environment

1. macOS, 10.13.3
1. `Python 2.7.14; Python 3.6.4`
1. `pipenv, version 9.0.3` (also tested with 9.1.0 as `--update` did more than I expected it to)

##### Expected result

```
$ pipenv graph --json
[]
```

##### Actual result

The ticket template requested `--verbose` output, but:
```
$ pipenv graph --json --verbose
Error: no such option: --verbose
```

```
$ pipenv graph --json
Traceback (most recent call last):
  File "/usr/local/Cellar/pipenv/9.0.3/libexec/bin/pipenv", line 11, in <module>
    load_entry_point('pipenv==9.0.3', 'console_scripts', 'pipenv')()
  File "/usr/local/Cellar/pipenv/9.0.3/libexec/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/Cellar/pipenv/9.0.3/libexec/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/local/Cellar/pipenv/9.0.3/libexec/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/Cellar/pipenv/9.0.3/libexec/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/Cellar/pipenv/9.0.3/libexec/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/Cellar/pipenv/9.0.3/libexec/lib/python3.6/site-packages/pipenv/cli.py", line 2422, in graph
    for d in simplejson.loads(c.out):
[...]

PR #1466, opened 2018-02-22
fix graph command for missing virtualenv

dschaller:
Fixes https://github.com/pypa/pipenv/issues/1455

I went with the approach of bypassing the build a dependency tree when a virtualenv does not exist for optimization purposes. The other approach would have been to build a dependency tree and conditionally switch to valid empty JSON after evaluating the empty output.

### DIFF
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -2194,26 +2194,33 @@ def do_graph(bare=False, json=False, reverse=False):
     if reverse:
         flag = '--reverse'
 
+    if not project.virtualenv_exists:
+        click.echo(
+            u'{0}: No virtualenv has been created for this project yet! Consider '
+            u'running {1} first to automatically generate one for you or see'
+            u'{2} for further instructions.'.format(
+                crayons.red('Warning', bold=True),
+                crayons.green('`pipenv install`'),
+                crayons.green('`pipenv install --help`')
+            ), err=True
+        )
+        sys.exit(1)
+
+
     cmd = '"{0}" {1} {2}'.format(
         python_path,
         shellquote(pipdeptree.__file__.rstrip('cdo')),
         flag
     )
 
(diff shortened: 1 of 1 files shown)


techalchemy:
With the changes we pushed in 9.1.0 we are sort of staking out a position on providing interaction with non-pipenv-managed (and specifically non-virtual) python environments (i.e. we do one-way interactions with a pre-resolved graph unless in a virtualenv). I wonder if we shouldn't simply provide an error if there is no project found with instructions to install?

dschaller:
👍 when you say instructions to install do you mean the existence of a pipfile?

techalchemy:
@dschaller I'm thinking maybe like a warning to stderr like the one found [here](https://github.com/pypa/pipenv/blob/7402765c0ea1775d9f6dad1cf11329fff27b1d6a/pipenv/core.py#L702), except just indicate that no virtualenv was detected and that the user should create one with `pipenv install` or see `pipenv install --help` for instructions, we used to have text along those lines somewhere

dschaller:
@techalchemy mind taking a look at the changes?

pipenv/core.py
```diff
@@ -2194,6 +2194,19 @@ def do_graph(bare=False, json=False, reverse=False):
     if reverse:
         flag = '--reverse'
 
+    if not project.virtualenv_exists:
+        click.echo(
+            u'{0}: No virtualenv has been created for this project yet! Consider '
+            u'running {1} first to automatically generate a one for you or see'
```
nateprewitt: I think we need to remove "a" from "generate a one for you". We could probably shorten it to just "generate one".

dschaller:
@nateprewitt updated :D

techalchemy:
Rerunning the appveyor test to make sure the 3.6 failure was just a transient one, then we can go ahead with this

Issue #1462, opened 2018-02-21
I miss `pip show`


dradetsky:
I've recently started using pipenv. I like it, but I miss having `pip show` around. Doing `pip shell ; pip show PKG_NAME` makes me unhappy.

I've started a branch to add this command, although it occurs to me to ask whether this is considered a bad feature to add for some reason, such as because there's another canonical way to get this info that I didn't know about or something.

My WIP PR is #1461. I do plan to make it less crappy, although not before hearing that the basic idea is welcome.

##### Describe your environment

1. Arch Linux
1. Python version: `3.6.4`
1. Pipenv version: `9.0.3`

dradetsky:
For example, I would 

- add a `more_packages` arg to the click cmd, after figuring out how that works, and get info for multiple arguments at once.
- steal more ideas from `pip_install` to make the implementation of `pip_show` more robust.
- add some documentation of the command in various places (advice in re where to find all these places would be totes awesome; is there anything besides `format_help` and `docs/`?)

uranusjr:
You can `pipenv run pip show PACKAGE`. I’m personally -1 to this.

dradetsky:
-1 to using `pipenv run` or -1 to what i'm proposing?

fwiw, I didn't actually know that the `pipenv run ...` solution existed; maybe the answer is a PR which adds "using pip stuff" to dox

uranusjr:
I’m -1 to `pipenv show`, to more generally, any top-level subcommand that is essentially a shortcut to `pipenv run`.

uranusjr:
(Sorry for the extra comment, I submitted too early.)

You’re not the first, and certainly won’t be the last trying to coming up with suggestions due to unawareness to `pipenv run`, and this (to me) shows that the documentation probably needs some re-writing. The `run` subcommand is not present in the Usages section of README, and is only mentioned in a relatively obscure place. Restructuring the documentation a little might be useful. Would it be a possibility @kennethreitz?

dradetsky:
I think I agree with both your -1 and the proposition that the dox need work.

ncoghlan:
+1 for explicitly calling out `pipenv run pip <subcommand>` for getting package level info out of an environment with first running `pipenv shell`. As @uranusjr notes, we don't want to be in the business of shadowing all of `pip`'s subcommands in `pipenv`.

PR #1467, opened 2018-02-22
document pip commands from pipenv

dradetsky:
Explicitly point out in first page of docs that you can use `pipenv run pip ...` to invoke pip utility commands.

I previously made #1461, which was sensibly rejected on the grounds that it wasn't actually needed. Somebody pointed out in #1462 that I'm not the first person to make this mistake. The proposed change is perhaps not the strongest possible guarantee that nobody will make the mistake again, but also could help. Maybe. It's at any rate extremely unlikely to hurt.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -151,6 +151,9 @@ Fish is the best shell. You should use it.
        Install a local setup.py into your virtual environment/Pipfile:
        $ pipenv install -e .
 
+       Use a lower-level pip command to get info about a package:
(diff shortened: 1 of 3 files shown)


dradetsky:
I decided to add the examples in more places, since the first one I added is easy to miss

dradetsky:
I considered adding a whole little section about this to `basics.rst`, but I thought that might be overkill.

kennethreitz:
thanks!

PR #1481, opened 2018-02-24
add extra indexes from pip config files in Pipfile gen

dschaller:
This is just an idea but it would be useful if we automatically parsed pip config files and added the extra indexes as sources in the generated Pipfile.

Feedback is welcomed and greatly appreciated :)


kennethreitz:
love this! i implemented it myself at one point, but it got lost.

PR #1493, opened 2018-02-24
Convert dependency name to PEP 423 name when looking for it in lockfile

greysteil:
Currently, running `pipenv lock --keep-outdated` with a `Pipfile` with `Requests = "*"` blows up because the name is normalised in the lockfile.

Also fixes a typo.


kennethreitz:
thanks!

PR #1495, opened 2018-02-25
Reinstate docs for Homebrew installation method

richardcooper:


kennethreitz:
this isn't the proper way to do links in RST, but i'll allow it :)

Issue #1503, opened 2018-02-26
All requirements are commented out after `pipenv lock -r`


MarcDufresne:
Since the update to 10.X all our `requirements.txt` files get commented out because of this specific piece of code: https://github.com/pypa/pipenv/blob/master/pipenv/core.py#L802

It wants the dependencies to include `--hash` but generating the lock file does not populate the hashes. And in our situation we can't use hashes because our team is a mix of different linux distros and macOS, so hashes are useless.

I suggest adding a flag to the `lock` command to ignore the `'--hash' not in dep[0]` requirement.

Something like:

```python
if '--hash' not in dep[0] and not ignore_pip_hash_mode:
```

##### Describe your environment

1. OS version: macOS 10.13, Ubuntu 16.04, Arch Linux x86_64 Linux 4.15.5-1-ARCH
1. Python version: 3.6.3, 3.6.4
1. Pipenv version: v10.1.0

##### Expected result

Normal requirement file, with no requirement being commented out.

```
alembic==0.9.7
arrow==0.12.1
mako==1.0.7
markupsafe==1.0
munch==2.2.0
pymysql==0.7.11
python-dateutil==2.6.1
python-editor==1.0.3
pyyaml==3.12
six==1.11.0
sqlalchemy==1.1.15
ujson==1.35
```

##### Actual result

Commented out requirements

```
# alembic==0.9.7
# arrow==0.12.1
# mako==1.0.7
# markupsafe==1.0
# munch==2.2.0
# pymysql==0.7.11
# python-dateutil==2.6.1
# python-editor==1.0.3
# pyyaml==3.12
# six==1.11.0
# sqlalchemy==1.1.15
# ujson==1.35
```

##### Steps to replicate

Pipfile:
```
[[source]]
url = "https://our.custom.devpi.com/simple"
verify_ssl = true

[requires]
python_version = "3.6"

[packages]
alembic = ">=0.9,<0.10"
arrow = ">=0.12.0"
Munch = "*"
PyYAML = ">=3.12,<4"
sqlalchemy = ">=1.1,<1.2"
ujson = "*"
PyMySQL = ">=0.7,<0.8"

[dev-packages]
pytest = ">=3.0,<3.1"
pytest-cov = ">=2.4,<2.5"
flake8 = "*"
yapf = "*"
```

Commands:
```bash
$ pipenv lock -r
```

This works fine on 9.0.3, but not on 10.1.0. 
[...]

PR #1504, opened 2018-02-27
Always try to generate a hash for packages in Pipfile.lock

jtratner:
Even for custom package indexes!

I tested this against Counsyl's internal devpi instance and it worked perfectly :) 

That said, I worry this would cause more issues than it would fix (because it will *always* generate hashes even if the index does not present them), but it'd be really *nice* to do it :) .  Perhaps we could put this behind an environment variable and let folks opt-in to it and kick the tires.

Would fix issue in #1503.

### Evidence this works

Pipfile like this:

```
›› cat Pipfile
[[source]]

url = "https://internal-devpi/my/index/+simple"
verify_ssl = true
name = "counsyl-pypi"

[packages]

internal-package = "*"
```

installs with hashes

```
Installing 'internal-package --hash=sha256:somehash'
$ "/Users/jtratner/.virtualenvs/example-LTUvl4NZ/bin/pip" install   --verbose  --no-deps  -r /var/folders/fm/sjgpzb856kld04jvq82bxrcw0000gp/T/pipenv-KBTpdr-requirement.txt --require-hashes -i https://internal-devpi/my/index/+simple --exists-action w
Installing 'six==1.11.0 --hash=sha256:70e8a77beed4562e7f14fe23a786b54f6296e34344c23bc42f07b15018ff98e9  --hash=sha256:832dc0e10feb1aa2c68dcc57dbb658f1c7e65b9b61af69048abc87a2db00a0eb'
$ "/Users/jtratner/.virtualenvs/example-LTUvl4NZ/bin/pip" install   --verbose  --no-deps  -r /var/folders/fm/sjgpzb856kld04jvq82bxrcw0000gp/T/pipenv-bi05kM-requirement.txt --require-hashes -i https://internal-devpi/my/index/+simple --exists-action w
```

and generates requirements.txt without comments - woo!

### DIFF
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -391,16 +391,20 @@ def resolve_deps(deps, which, which_pip, project, sources=None, verbose=False, p
 
                     for release in cleaned_releases[version]:
                         collected_hashes.append(release['digests']['sha256'])
-
                     collected_hashes = ['sha256:' + s for s in collected_hashes]
 
-                    # Collect un-collectable hashes.
-                    if not collected_hashes:
-                        collected_hashes = list(list(resolver.resolve_hashes([result]).items())[0][1])
 
                 except (ValueError, KeyError, ConnectionError):
                     if verbose:
                         print('Error fetching {}'.format(name))
+
+            # Collect un-collectable hashes (should work with devpi).
+            if not collected_hashes:
+                try:
+                    collected_hashes = list(list(resolver.resolve_hashes([result]).items())[0][1])
+                except (ValueError, KeyError, ConnectionError, IndexError):
+                    if verbose:
+                        print('Error generating hash for {}'.format(name))
 
             d = {'name': name, 'version': version, 'hashes': collected_hashes}
 
--- a/HISTORY.txt
+++ b/HISTORY.txt
@@ -1,5 +1,6 @@
 next:
  - Assume `PIPENV_VENV_IN_PROJECT` if `./.venv/` already exists.
+ - Use and generate hashes for PyPI mirrors and custom indexes.
 10.1.0:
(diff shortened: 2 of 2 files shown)


pipenv/utils.py
```diff
@@ -391,17 +391,21 @@ def resolve_deps(deps, which, which_pip, project, sources=None, verbose=False, p
...
                     if verbose:
                         print('Error fetching {}'.format(name))
 
+            # Collect un-collectable hashes (should work with devpi).
+            if not collected_hashes:
+                try:
+                    collected_hashes = list(list(resolver.resolve_hashes([result]).items())[0][1])
+                except (ValueError, KeyError, ConnectionError, IndexError):
+                    if verbose:
+                        print('Error generating hash for {}'.format(name))
```
kennethreitz: click.echo

kennethreitz: `click.echo('{0}: Error generating hash for {1}'.format(crayons.red('Warning', bold=True), name))`

kennethreitz: format indexes (e.g. `{1}`) are also required for older pythons

techalchemy:
This is clever - not a full solution but a good fallback imo.  @kennethreitz I can get with you later on this but I’ve been back and forth a bit with @jtratner on the ultimate goal of getting the resolver to grab this stuff

kennethreitz:
this looks like it'll be good to have in place either way

techalchemy:
Yeah agree. Definitely an improvement

jtratner:
Thanks for the cleanup @kennethreitz!

One aside - this will probably make installs slower for custom PyPIs because the hash generation requires hashing the entire tarball or wheel. (which obv isn't the end of the world)

PR #1507, opened 2018-02-27
Fix inline comments causing install error

serialx:
### Environment
 * pipenv, version 10.1.0
 * Python 3.6.4 (Homebrew)
 * macOS High Sierra, version 10.13.3

### Steps to reproduce
1. Create a simple Pipfile with inline comments followed by a whole-line comment within packages:
```
[[source]]

verify_ssl = true
name = "pypi"


[packages]

# Django related packages
django = ">=2.0"
pillow = "*"  # necessary for django ImageField
# OAuth / Security related packages
django-oauth-toolkit = "*"


[requires]

python_version = "3.6"
```
2. Run `pipenv install rules`

### Expected behaviour
`rules` package is installed, and added to Pipfile

### Resulting behaviour
Fatal error with message:
```
$ pipenv install rules
Using /usr/local/bin/python3.6m to create virtualenv…
⠋Running virtualenv with interpreter /usr/local/bin/python3.6m
Using base prefix '/usr/local/Cellar/python3/3.6.4_2/Frameworks/Python.framework/Versions/3.6'
New python executable in /Users/serialx/.local/share/virtualenvs/api-k66jWLAd/bin/python3.6
Also creating executable in /Users/serialx/.local/share/virtualenvs/api-k66jWLAd/bin/python

Virtualenv location: /Users/serialx/.local/share/virtualenvs/api-k66jWLAd
Installing rules…
Collecting rules
Installing collected packages: rules
Successfully installed rules-1.3

Adding rules to Pipfile's [packages]…
  File "/usr/local/Cellar/pipenv/10.1.0/libexec/bin/pipenv", line 11, in <module>
    load_entry_point('pipenv==10.1.0', 'console_scripts', 'pipenv')()
  File "/usr/local/Cellar/pipenv/10.1.0/libexec/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 722, in __call__
  File "/usr/local/Cellar/pipenv/10.1.0/libexec/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/local/Cellar/pipenv/10.1.0/libexec/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 1066, in invoke
  File "/usr/local/Cellar/pipenv/10.1.0/libexec/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 895, in invoke
  File "/usr/local/Cellar/pipenv/10.1.0/libexec/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 535, in invoke
  File "/usr/local/Cellar/pipenv/10.1.0/libexec/lib/python3.6/site-packages/pipenv/cli.py", line 197, in install
    selective_upgrade=selective_upgrade
  File "/usr/local/Cellar/pipenv/10.1.0/libexec/lib/python3.6/site-packages/pipenv/core.py", line 1875, in do_install
    project.add_package_to_pipfile(package_name, dev)
  File "/usr/local/Cellar/pipenv/10.1.0/libexec/lib/python3.6/site-packages/pipenv/project.py", line 507, in add_package_to_pipfile
    p = self._pipfile
  File "/usr/local/Cellar/pipenv/10.1.0/libexec/lib/python3.6/site-packages/pipenv/project.py", line 292, in _pipfile
    p_section[norm_key] = p_section.pop(key)
  File "/usr/local/Cellar/pipenv/10.1.0/libexec/lib/python3.6/site-packages/pipenv/patched/prettytoml/elements/table.py", line 114, in pop
    v = self[key]
  File "/usr/local/Cellar/pipenv/10.1.0/libexec/lib/python3.6/site-packages/pipenv/patched/prettytoml/elements/abstracttable.py", line 58, in __getitem__
    raise KeyError
KeyError
```

### Possible solution to this error
I've debugged the code and found the root cause is when using `pop()` method in prettytoml package. For prettytoml to delete an element, it finds the starting point and finds the nearest `NewlineElement` and deletes the whole section from starting point to the newline element. The problem is that newlines with preceding comment token is unified into a single `CommentElement` entity. Thus prettytoml mistakenly deletes the next line of the comment. In which, this case it would be `django-oauth-toolkit = "*"`.

I've propose a simple patch to also find `CommentElement` when trying to find the following newline element.

~The side effect of this patch is that, since all comment line metadata is deleted, the newly contructed Pipfile has comments in the wrong lines:~ The comment related code-path seems to be unrelated to this patch.

```
[[source]]

verify_ssl = true
name = "pypi"


[packages]

# Django related packages
django = ">=2.0"# necessary for django ImageField
# OAuth / Security related packages
pillow = "*"
django-oauth-toolkit = "*"
rules = "*"


[requires]

python_version = "3.6"
```

### DIFF
--- a/pipenv/patched/prettytoml/elements/traversal/__init__.py
+++ b/pipenv/patched/prettytoml/elements/traversal/__init__.py
@@ -1,6 +1,6 @@
 from prettytoml import tokens
 from prettytoml.elements import common
-from prettytoml.elements.metadata import PunctuationElement, NewlineElement
+from prettytoml.elements.metadata import PunctuationElement, NewlineElement, CommentElement
 from prettytoml.elements.traversal import predicates
 
 
@@ -61,7 +61,7 @@ class TraversalMixin:
         """
         Returns the index of the following newline element after the given index, or -Infinity.
         """
-        return self.__find_following_element(index, lambda e: isinstance(e, NewlineElement))
+        return self.__find_following_element(index, lambda e: isinstance(e, NewlineElement) or isinstance(e, CommentElement))
 
     def _find_following_comment(self, index):
         """


serialx:
Actually, having one more look at the patch side-effect, unpatched pipenv 10.1.0 already breaks all the comments in the Pipfile. So I think this patch is *not* causing the comment side-effect.

kennethreitz:
comments shouldn't be a problem.

techalchemy:
But currently they are? I read something else about this

serialx:
@kennethreitz By removing all the comments in the reproduction example, pipenv install runs fine. I've pretty much debugged the whole code including prettytoml, and the comment parsing and element deleting(pop) of prettytoml is the definite problem here.

serialx:
BTW, taking a look at the libraries like prettytoml... It's been abandoned for like three years. I think we should avoid using this and look for alternatives.

kennethreitz:
TOML hasn't changed :)

I made some heavy modifications to it.

uranusjr:
I did some debugging and came up with some interesting discoveries. So I put a couple of prints around `p_section[norm_key] = p_section.pop(key)`, like this:

```python
print(key, p_section)
p_section[norm_key] = p_section.pop(key)
print(key, p_section)
```

and got this output:

```
django {'django': '>=2.0', 'pillow': '*', 'django-oauth-toolkit': '*'}
django {'django': '>=2.0', 'pillow': '*', 'django-oauth-toolkit': '*'}
pillow {'pillow': '*', 'django-oauth-toolkit': '*', 'django': '>=2.0'}
pillow {'django': '>=2.0', 'pillow': '*'}
django-oauth-toolkit {'django': '>=2.0', 'pillow': '*'}
```

So for some reason the prettytoml removes `django-oauth-toolkit` when popping `pillow`. I haven’t looked into the implementation of `pop`, but wants to record this finding first.

uranusjr:
I’ve opened #1535 to detail the problem. I’m going to close this because the fix is unlikely to be correct.

PR #1508, opened 2018-02-27
Use hyphens, not en dashes

anseljh:
En dashes are great, but these should be hyphens. :)

References:
* http://www.chicagomanualofstyle.org/qanda/data/faq/topics/HyphensEnDashesEmDashes/faq0002.html
* https://typographyforlawyers.com/hyphens-and-dashes.html


uranusjr:
You’re correct.

PR #1510, opened 2018-02-28
Updated docs - basic usage example pipenv workflow

Darkle:
I'm new to python so I didn't know 'exit' was the thing to type to deactivate the pipenv shell, so I thought I would add that to the basic usage doc.


kennethreitz:
thanks!

PR #1512, opened 2018-02-28
Correct csh syntax for activate_virtualenv

vishwin:
The `.` operator in csh executes a directory, which is the wrong behaviour here. At best, when pipenv is installed system-wide (preferably via the operating system package manager), the result is a permission denied error from the shell.

As csh uses the `source` command, reflect that here.


vishwin:
Github's direct editing is causing more problems wrt indenting at the very least. Closing and opening a new PR when this is sorted without using said direct editing.

PR #1513, opened 2018-02-28
Correct csh syntax in activate_virtualenv

vishwin:
In `csh`, the `.` operator executes a directory, which is the wrong behaviour here. At best, when `pipenv` is installed system-wide (preferably via the operating system package manager), the result is a permission denied error from the shell.

As `csh` uses the `source` command, reflect that here.


kennethreitz:
omg we should just always use source

techalchemy:
There was a relevant reason to switch, but yeah we should make sure we fall back if necessary. Maybe just try the alias and use source otherwise

PR #1514, opened 2018-02-28
Remove tmpfile after pip_install() finishes.

mteiste:
The hotfix for tmp files did not address pipenv-<hash>-requirement.txt files that were left behind when pip_install was called with r==None.


uranusjr:
I imagine this would require some safe guard around the unlink call in case Windows complains. But @techalchemy is the expert here.

techalchemy:
Yeah it’s not this simple unfortunately. On mobile, more review in a bit.

pipenv/core.py
```diff
@@ -1398,6 +1402,9 @@ def pip_install(
                 crayons.normal('pipenv lock', bold=True)
             )
         )
+        # Unlink possible tmp file created by us
+        if tmp_created:
+            os.unlink(tmp_created)
```
techalchemy: Overall we can never guarantee that this handle will be closed, especially on windows, and we can't even guarantee that we will still have permission to delete it depending on the circumstances -- especially if there was an error of some kind during resolution during which time we are kicking off `subprocess` calls which, on windows, are known to leak open file handles

mteiste: OK. 

BTW the tests were not failing due to this change, at the time they were failing also without the change :). Second, this was tested on windows. Third, the tmp_created was set (and therefore unlink() attempted) only if the function was called with r == None (e.g. the file was created by this call to  pip_install() itself not if it was created outside the pip_install(). The r or the fd was never returned either directly or undirectly (as far as I could see) and the file was written / accessed via "with statement" (that change seemed to have been made bit earlier to guarantee that the fd does get closed), if the unlink() fails due to errors like file handle leakage issues in windows subprocess handling, I cannot comment, but I would imagine we've lost at that time anyways :)

techalchemy:
@mteiste thanks for your input, I didn't make my decision based on test failures but rather based on specific concerns about the implementation around removal of temporary file objects which are owned by forked process groups on windows. I closed this PR because you actually have no way of knowing if the file descriptor is actually in use based simply on your own context manager, because forked processes (e.g. pip) are explicitly passed those resources.

More exhaustively, processes forked in `pip_install` are not all returned or completed there, which is one of many reasons why simply calling `os.unlink` is not a safe operation on windows. This code is not the sole owner of the file, so when the `unlink` inevitably fails due to another process reading it (that is sufficient to cause `unlink` to fail on windows), there is no wait or retry here.

>  if the unlink() fails due to errors like file handle leakage issues in windows subprocess handling, I cannot comment, but I would imagine we've lost at that time anyways :)

Windows is a first class citizen, so this is unfortunately not going to work as an approach. I'm sorry about this pull request, but I didn't close it out of spite or ignorance, there is legitimately a lot of nuance underlying this that I just finished sorting out with our tests. I am far from an expert but I do know that the naive approach doesn't work in this case

mteiste:
You have to admit the "I know where you are coming from was quite condescending" :) Just so you know where I actually come from here is how I arrived in the naive solution.

1) this function only creates the file if it is called with r == None, hence no unlinking is ever done unless the file was specifically created by this function itself.
2) first unlink is just before sys.exit(1) so if that fails it is not an issue
3) second unlink was done after we have processed all the sources and executed all the pip commands. [What apparently lead me to astray was the assumption that checking for return_code of delegator.run() also means that the subprocess has finished and when the loop has finished no more pip commands requiring the existence of the tmp file will be launched]

So, did the change, run tests -> failed -> reverted the change -> run tests -> same failures ergo not likely to have been introduced by the change.

So initially based on these assumptions the naive solution looked like the simplest, hence to me the preferred one (and has been working just fine on my environment (which is windows by the way :) ) but that is anecdotal not real proof and I stand corrected that basically my assumption 3 is faulty rendering the "fix" unusable. If you had not told me that the processes using the tmpfile are not guaranteed to have finished at the time pip_install() returns I would've argued that surrounding the latter unlink() in try...except would've been enough to make this safe, but alas that seems not to be the case, so I stop arguing and let you forget this patch attempt ever existed :).

Anyways, I thank you for the education and hope the issue gets a proper fix at some point in the future.

techalchemy:
Forking commands into the background which themselves fork commands into the background isnt something you can assume is going to be a completed task just because the function returned. I said I know where you’re coming from because I do, I just happened to have worked on this in the context of our test runner which regularly had failures on windows due to concurrency issues. 

As far as hoping for a fix, we’ll probably merge #1538 today which relies on a backport of `TemporaryDirectory` and `weakref.finalize()` to handle removal of these files. Exception handling here would be an alternative but the implementation of `finalize` allows `TemporaryDirectory` objects to handle removal more cleanly, plus this will fall back on our own `ResourceWarning` automatically if it fails rather than raising an exception at all. Thanks again for the PR, and for the engagement on this, hopefully this helps explain where I’m coming from!

PR #1517, opened 2018-02-28
Treat conda environments as a virtual environment

jtratner:
This PR treats conda environments as a virtual environment. (could potentially remove the entire docs section on conda now).

Open questions:

1. do we want to add a test case for this?
2. is it still worth noting in docs that your requirements can get out of sync?

### why conda is useful

Conda is a great tool for installing hard-to-build software and especially hard-to-build python packages that may or may not have wheels (most of them do, but if you want versions of, say, numpy and pandas compiled against a different BLAS library, it's way easier to do with conda + a little easier to maintain cross-platform).

### conda is like a virtualenv

Currently pipenv treats conda environments as if they need to have their own virtualenv within them; but each conda environment actually maintains a separate set of Python site-packages.

### DIFF
--- a/HISTORY.txt
+++ b/HISTORY.txt
@@ -1,3 +1,5 @@
+next:
+ - Treat conda environments as virtualenvs and install into conda env's site-packages.
 10.1.2:
  - Bugfix.
 10.1.1:
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -81,7 +81,10 @@ This will fail a build if the ``Pipfile.lock`` is out–of–date, instead of ge
 ☤ ``pipenv`` and ``conda``
 --------------------------
 
-To use Pipenv with a Conda–provided Python, you simply provide the path to the Python binary::
+Pipenv will automatically detect conda environments and install into the specific environment's
+system site packages. (however if you use conda to install specific python packages, you'll need to
+pin those packages to the same version in your Pipfile)If you want to create an isolated virtualenv
+within that environment, pass pipenv the path to the anaconda python::
 
     $ pipenv install --python=/path/to/anaconda/python
 
--- a/pipenv/environments.py
+++ b/pipenv/environments.py
@@ -47,7 +47,8 @@ PIPENV_USE_SYSTEM = False
 if 'PIPENV_ACTIVE' not in os.environ:
     if 'PIPENV_IGNORE_VIRTUALENVS' not in os.environ:
         PIPENV_VIRTUALENV = os.environ.get('VIRTUAL_ENV')
-        PIPENV_USE_SYSTEM = bool(os.environ.get('VIRTUAL_ENV'))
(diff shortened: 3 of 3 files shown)


jtratner:
potential gotchas:

1. some conda environments may not have python installed 😱 
2. breaks pipenv run (but... hopefully you're aware of this if using conda with pipenv)

uranusjr:
I generally agree with @techalchemy, but this seems like a good hack if the user is aware of the pitfalls. Would it be equally functional if the user [automatically sets `VIRTUAL_ENV` when activating a Conda environment](https://stackoverflow.com/questions/31598963)? Put a file with (assuming Bash)

```
export VIRTUAL_ENV="$CONDA_DEFAULT_ENV"
```

in `activate.d`, and unset in `deactivate.d`. I don’t personally user Anaconda, but that seems plausible…?

kennethreitz:
I have a problem with this — CONDA_DEFAULT_ENV will always be set.

kennethreitz:
right?

jtratner:
> I have a problem with this — CONDA_DEFAULT_ENV will always be set.

Good question - but actually no! (and this is why I think `CONDA_DEFAULT_ENV` is essentially equivalent to `VIRTUAL_ENV`)

The standard set up for conda is to just add the conda python binary to your path, looks like this:

```
jtratner@C13101jtratner$ (git::master) which python
/Users/jtratner/miniconda3/bin/python
jtratner@C13101jtratner$ (git::master)echo $CONDA_DEFAULT_ENV

jtratner@C13101jtratner$ (git::master)
```

If you activate without an environment specified, it dumps you into root environmment

```
jtratner@C13101jtratner$ (git::master) source activate
(root) jtratner@C13101jtratner$ (git::master) which python
/Users/jtratner/miniconda3/bin/python
```

However, you can pick an environment, and this switches to something really
similar to virtualenv:

```
jtratner@C13101jtratner$ (git::master) source activate myenv
(myenv) jtratner@C13101jtratner$ (git::master) which python
/Users/jtratner/miniconda3/envs/myenv/bin/python
```

jtratner:
@uranusjr

>> export VIRTUAL_ENV="$CONDA_DEFAULT_ENV"
> in activate.d, and unset in deactivate.d. I don’t personally use Anaconda, but that seems plausible…?

That's a little complicated for a non-technical user to manage.  Many users just follow conda docs and that's it, so would be nice to be able to support conda within pyenv with a similar workflow :)

kennethreitz:
i think these two just don't want to play nicely together

uranusjr:
The solution is probably making Conda gain this functionality instead. Interestingly, I just read the Conda source, and it seems [they are actually implementing a `shell` subcommand](https://github.com/conda/conda/blob/aedf26c/conda/cli/main.py). I couldn’t find anywhere in the documentation, but maybe we’re witness something…?

PR #1523, opened 2018-03-02
Update marker os_name to platform_version in docs

jxltom:
According to doc, marker ```os_name``` is ```os.name``` as a PEP 508 specifier.

In python2, ```os.name``` has following values: 'posix', 'nt', 'java'
In python3, it has following values: 'posix', 'nt', 'os2', 'ce', 'java', 'riscos'

I believe using ```platform_version = "== 'Windows'"``` is more appropriate.

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -46,7 +46,7 @@ Here's an example ``Pipfile``, which will only install ``pywinusb`` on Windows s
 
     [packages]
     requests = "*"
-    pywinusb = {version = "*", os_name = "== 'windows'"}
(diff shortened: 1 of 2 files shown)


kennethreitz:
this is straight out of the pep 508 docs

jxltom:
@kennethreitz But ```os_name = "== 'nt'``` in Windows in Python3 rather than ```os_name = "== 'windows'```....

```platform_version = "== 'Windows'"``` should be used

uranusjr:
Well if both of you are correct, the PEP 508 docs are wrong. Maybe we should fix that first. @kennethreitz can you provide s pointer where those came from?

PR #1531, opened 2018-03-03
fix PYENV_ROOT-based pyenv detection

zigg:
I noticed that Pipenv wasn't detecting my pyenv installation (just a straight-up `brew install pyenv` on my Mac) and offering to install the Python version requested by my `Pipfile`.

I believe this is because `PYENV_INSTALLED` was checking the `PYENV_ROOT` environment variable instead of the result of the `PYENV_ROOT` default get; `expanduser` also returns even invalid paths so it's important to check for existence of the pyenv root with `exists`.

### DIFF
--- a/pipenv/environments.py
+++ b/pipenv/environments.py
@@ -89,7 +89,7 @@ PIPENV_DONT_USE_PYENV = os.environ.get('PIPENV_DONT_USE_PYENV')
 
 PYENV_ROOT = os.environ.get('PYENV_ROOT', os.path.expanduser('~/.pyenv'))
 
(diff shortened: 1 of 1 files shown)


kennethreitz:
i think you didn't follow the installation instructions properly ;)

zigg:
Ah, okay. Just for the record, I'd deleted the shims from my startup (step 3 [here](https://github.com/pyenv/pyenv#basic-github-checkout)) because I didn't think I needed the shell integration anymore. 

But if you do have pyenv's shell integration enabled, Pipenv will indeed correctly find it.

Sorry about that.

kennethreitz:
excellent!

PR #1533, opened 2018-03-03
Fix dependency list type mismatch

tekulvw:
When the dependency list got sorted in 6261122d (current line 828) you'd start getting `TypeError` when the code on line 823 executed because it changed the dependency tuple to a dependency list.


techalchemy:
@tekulvw can you throw in a test for this also? We have been making pretty substantial changes and it'd be nice to avoid regressing this again

techalchemy:
P.s. thanks for the first-time contribution, you kind of dove in head first there :)

PR #1536, opened 2018-03-03
Fix __delitem__ in prettytoml’s table element

uranusjr:
Two bugs involved here.

First of all, `__delitem__` uses `_find_following_newline` to find where the line to delete ends, which in turns looks for the next closest newline element. But if the previous line ends with a comment element (which implies a newline), it will also be deleted. This would be a problem if the comment is inline.

```
foo=1  # comment
bar=2
```

In this minimal example, the elements in play are

```python
AtomicElement('foo')
PunctuationElement('=')
AtomicElement(1)
WhitespaceElement('  ')
CommentElement('# comment')
AtomicElement('bar')
PunctuationElement('=')
AtomicElement(2)
```

When deleting `foo`, it starts from `AtomicElement('foo')` and looks for the next newline, but does not consider `CommentElement`, ending up deleting all elements. This fix replaces `_find_following_newline` with `_find_following_line_terminator`, which does consider both newline and comment.

Also there is a logic error in _find_following_line_terminator that prevents it from even working. The two -inf/inf clauses should serve as guards when either or both indexes are infinity, but they were flip-flopped and not working at all. This change doesn't seem to affect existing tests (at least not breaking them more than they already were).

### DIFF
--- a/pipenv/patched/prettytoml/elements/table.py
+++ b/pipenv/patched/prettytoml/elements/table.py
@@ -105,7 +105,7 @@ class TableElement(abstracttable.AbstractTable):
         preceding_newline = self._find_preceding_newline(begin)
         if preceding_newline >= 0:
             begin = preceding_newline
-        end = self._find_following_newline(begin)
+        end = self._find_following_line_terminator(begin)
         if end < 0:
             end = len(tuple(self._sub_elements))
         self._sub_elements = self.sub_elements[:begin] + self.sub_elements[end:]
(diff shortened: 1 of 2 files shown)


techalchemy:
@uranusjr this appears to break most of the tests, any thoughts?

uranusjr:
Because I’m dumb, basically 🙄 Seriously though it’e because my `inf` logic is as messed up as the old one. I’ve push a change to rectify this (if I’m actually correct this time).

techalchemy:
lol :p thanks for the fix

Issue #1552, opened 2018-03-05
Failure to parse ipywidgets extras_require


intrinseca:
It seems that pipenv is getting confused by the requirements in ipywidgets (https://github.com/jupyter-widgets/ipywidgets/blob/master/setup.py).

```
extras_require = setuptools_args['extras_require'] = {
    ':python_version<"3.3"' : ['ipython>=4.0.0,<6.0.0'],
    ':python_version>="3.3"': ['ipython>=4.0.0'],
    'test:python_version=="2.7"': ['mock'],
    'test': ['nose'],
}
```

It declares a conflict between IPython > 6.0.0 and ipywidgets even though my environment is Python 3.6.

##### Describe your environment

1. OS Type: Windows Server 2012 (though I expect I could reproduce on Windows 10 x64)
1. Python version: Python 3.6.4
1. Pipenv version: version 11.0.2

##### Expected result

pipenv allows IPython > 6.0.0 to be installed as Python is >= 3.3.

##### Actual result

Incidentally, I had to add `print(c.out)` to `utils.venv_resolve_deps` to get it to print the output from the venv rather than just eating it, so I could work out which packages were going wrong. 

```
Warning: Your dependencies could not be resolved. You likely have a mismatch in your sub-dependencies.
  You can use $ pipenv install --skip-lock to bypass this mechanism, then run $ pipenv graph to inspect the situation.
Could not find a version that matches ipython<6.0.0,>6.0.0,>=4.0.0
Tried: 0.10, 0.10.1, 0.10.1, 0.10.2, 0.10.2, 0.11, 0.11, 0.12, 0.12, 0.12.1, 0.12.1, 0.13, 0.13, 0.13.1, 0.13.1, 0.13.2,
 0.13.2, 1.0.0, 1.0.0, 1.1.0, 1.1.0, 1.2.0, 1.2.0, 1.2.1, 1.2.1, 2.0.0, 2.0.0, 2.0.0, 2.1.0, 2.1.0, 2.1.0, 2.2.0, 2.2.0,
 2.2.0, 2.3.0, 2.3.0, 2.3.0, 2.3.1, 2.3.1, 2.3.1, 2.4.0, 2.4.0, 2.4.0, 2.4.1, 2.4.1, 2.4.1, 3.0.0, 3.0.0, 3.0.0, 3.1.0,
[...]

PR #1557, opened 2018-03-05
Fix verbose display

intrinseca:
In investigating #1552 I thought I'd try and put in a proper fix - this way the verbose output is printed whether there is an error or not, and I also traced back the `verbose` flag being dropped up the call stack.


kr-mercury:
Can one of the admins verify this patch?

kr-mercury:
Can one of the admins verify this patch?

kennethreitz:
@kr-mercury build

kr-mercury:
Can one of the admins verify this patch?

PR #1558, opened 2018-03-05
Fix --selective-upgrade compatibility with Python 2

gzzo:
`list.copy()` is equivalent to `list[:]`, but the former is only available in Python 3.

### DIFF
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -1840,7 +1840,7 @@ def do_install(
     # Support for --selective-upgrade.
     if selective_upgrade:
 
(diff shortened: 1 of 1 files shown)


SmileyChris:
Whoops, just reported this as #1559

Reviewing the code, actually copying the list seems redundant - just `enumerate(package_list)` should be fine.

uranusjr:
The loop changes itself at line 1851. It does not matter here, but in general it is always recommended you loop through a copy if you ever do that, just to be safe. This looks good to me and I’ll merge as-is.

Issue #1547, opened 2018-03-05
cannot install google-cloud-datastore


david-gang:
Be sure to check the existing issues, both open and closed.

Describe the issue briefly here.

##### Describe your environment

1. OS Type ubuntu 16.0.4
1. Python version: `$ python -V` 3.5.2 (python of virtualenv will be different)
1. Pipenv version: `$ pipenv --version` 11.0.2

##### Expected result
Pipfile.lock should be created

##### Actual result

When possible, provide the verbose output (`--verbose`), especially for locking and dependencies resolving issues.
```
pipenv install --python=3.6.4 --verbose
Creating a virtualenv for this project…
Using /home/david/.pyenv/versions/3.6.4/bin/python3.6m to create virtualenv…
⠋Running virtualenv with interpreter /home/david/.pyenv/versions/3.6.4/bin/python3.6m
Using base prefix '/home/david/.pyenv/versions/3.6.4'
New python executable in /home/david/.local/share/virtualenvs/test-FTV5fq3F/bin/python3.6m
Also creating executable in /home/david/.local/share/virtualenvs/test-FTV5fq3F/bin/python
Installing setuptools, pip, wheel...done.

Virtualenv location: /home/david/.local/share/virtualenvs/test-FTV5fq3F
Pipfile.lock not found, creating…
Locking [dev-packages] dependencies…
Locking [packages] dependencies…

Warning: Your dependencies could not be resolved. You likely have a mismatch in your sub-dependencies.
  You can use $ pipenv install --skip-lock to bypass this mechanism, then run $ pipenv graph to inspect the situation.
Could not find a version that matches futures>=3.2.0
Tried: 0.2.python3, 0.1, 0.2, 1.0, 2.0, 2.1, 2.1.1, 2.1.2, 2.1.3, 2.1.4, 2.1.5, 2.1.5, 2.1.6, 2.1.6, 2.2.0, 2.2.0, 3.0.0, 3.0.1, 3.0.2, 3.0.3, 3.0.4, 3.0.5, 3.1.0, 3.1.1
```

##### Steps to replicate
here is my pipfile
```
[[source]]

verify_ssl = true
url = "https://pypi.python.org/simple"
name = "pypi"


[dev-packages]

yapf = "*"
pylint = "*"
mypy = "*"
rope = "*"
coverage = "*"
responses = "*"


[packages]

requests = "*"
google-cloud-datastore = "*"
google-cloud-storage = "*"
falcon = "*"
geopy = "*"
[...]

PR #1564, opened 2018-03-06
dont add dependency if the constraint does not evaluate

david-gang:
This should fix https://github.com/pypa/pipenv/issues/1547.
It at least fixed in my local environment. I will be happy to get every comment. 

I would like to hear how to add unit tests for this.


kennethreitz:
this appears to break all the things

PR #1569, opened 2018-03-06
Publish universal wheels when uploading to pypi

jtratner:
This ought to work - checked that these two commands generated what I expected:

```
›› rm -rf dist && time python setup.py sdist &> sdist.txt && time python setup.py bdist_wheel &> bdist.txt
python setup.py sdist &> sdist.txt  1.76s user 0.38s system 99% cpu 2.154 total
python setup.py bdist_wheel &> bdist.txt  1.10s user 0.45s system 97% cpu 1.581 total
›› ls dist
pipenv-11.0.2-py2.py3-none-any.whl pipenv-11.0.2.tar.gz
```

and thus `twine upload dist/*` should be good enough. (successfully installed pipenv from the wheel too)

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,3 +1,6 @@
+[bdist_wheel]
+universal = 1
+
 [flake8]
 exclude = .git,__pycache__,docs/,pipenv/vendor/,get-pipenv.py,setup.py
 ignore =
--- a/setup.py
+++ b/setup.py
@@ -57,6 +57,7 @@ class UploadCommand(Command):
 
         self.status('Building Source distribution…')
         os.system('{0} setup.py sdist'.format(sys.executable))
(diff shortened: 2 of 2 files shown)


kennethreitz:
universal wheels mess up out distribution. I can't remember the details, but this is not an oversight.

jtratner:
@kennethreitz - gotcha (figured I'd just put up a PR rather than issue => PR).  Are you open to posting source-only wheels specialized to python version? (i.e., just building python 3 wheels would also be nice)

kennethreitz:
i'm def open to that

kennethreitz:
let's target building a python3 only wheel.

I really need to remember what the wheel issue was. it was a no-go, show-stopper, for sure.

Issue #1247, opened 2017-12-23
Running `pipenv --completion` is very slow


bdusell:
Running `pipenv --completion` is consistently slow (it takes about 1 second to run), even after the command has been run multiple times and everything has been loaded into memory.

This means including `eval $(pipenv --completion)` in a `.bashrc` file makes the user wait for a second until the shell prompt comes up.

##### Describe your environment

1. OS: Kubuntu 17.04
1. Python version: 2.7.13
1. Pipenv version: 9.0.1

My shell is bash.

##### Expected result

The command shouldn't take so long to run.

##### Actual result

When running in bash, the command takes a whole second to run.

```
$ time pipenv --completion

_pipenv_completion() {
    local IFS=$'\t'
    COMPREPLY=( $( env COMP_WORDS="${COMP_WORDS[*]}" \
                   COMP_CWORD=$COMP_CWORD \
                   _PIPENV_COMPLETE=complete-bash $1 ) )
    return 0
}

complete -F _pipenv_completion -o default pipenv


real    0m0.987s
user    0m0.888s
sys     0m0.080s
```

##### Steps to replicate

Run `time pipenv --completion` in a bash shell.

P.S. This is a great project.

mimischi:
Same here using `zsh`.

**Describe your environment**
1. OS: Arch Linux (4.14.8-1-ARCH)
2. Python version: 3.6.3
3. Pipenv version: 9.0.1

**Result**

```bash
$ time pipenv --completion 

#compdef pipenv
_pipenv() {
  eval $(env COMMANDLINE="${words[1,$CURRENT]}" _PIPENV_COMPLETE=complete-zsh  pipenv)
}
if [[ "$(basename ${(%):-%x})" != "_pipenv" ]]; then
  autoload -U compinit && compinit
  compdef _pipenv pipenv
fi

pipenv --completion  1.20s user 0.09s system 99% cpu 1.301 total
```

eromoe:
much more slow ..
```
pipenv --completion  2.64s user 0.29s system 99% cpu 2.954 total
```

Xeite:
I got an same issue

1. OSX sierra
2. Python 2.7.14
3. pipenv 9.0.1

`simple_tools!$ time pipenv --completion                                                                                            1.03s [master●●] ~/Projects/python/simple_tools

#compdef pipenv
_pipenv() {
[...]

PR #1592, opened 2018-03-07
Move handling of --completion earlier to make it way faster

anowlcalledjosh:
More details in issue #1247, but basically:

- `pipenv.core` takes about 400ms to import
- why do that before checking if `--completion` was passed?

This takes `pipenv --completion` from taking 1.1 seconds down to only 0.6 seconds on my machine (I think it can still go faster, hopefully I'll have time to explore that more in a couple of weeks, but this seemed like an easy win).

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -46,6 +46,19 @@ def cli(
     python=False, help=False, update=False, py=False,
     site_packages=False, envs=False, man=False, completion=False
 ):
+    if completion:  # Handle this ASAP to make shell startup fast.
+        if PIPENV_SHELL:
+            os.environ['_PIPENV_COMPLETE'] = 'source-{0}'.format(PIPENV_SHELL.split(os.sep)[-1])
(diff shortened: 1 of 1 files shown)


kennethreitz:
this won't work. we can't shell out the rest of our running to a subprocess.

anowlcalledjosh:
Hm? This is still working exactly how it was before, just that completion is now handled before doing the expensive imports. All I've done is move a section of code a few lines upwards.

kennethreitz:
you moved literally all of our code into a delegator call that will always exit 0.

kennethreitz:
oh, i see. interesting.

kennethreitz:
i misread

kennethreitz:
thanks!

anowlcalledjosh:
Thank *you*, you're the one who came up with and maintains this whole project :)

I would like to get rid of that delegator call if possible, I might have a go at that this evening.

Issue #1600, opened 2018-03-08
ImportError: cannot import name '_psutil_windows'


sdementen:
Issue on windows on 11.1.3 via appveyor (see logs on
https://ci.appveyor.com/project/sdementen/piecash/build/1.0.372/job/uogfacdy8m9y902x#L151)

kennethreitz:
#1587. Please check existing issues before opening new ones.

techalchemy:
Fixed & released

Issue #1587, opened 2018-03-07
Traceback on Windows 10 involving _psutil_windows


brettcannon:
Be sure to check the existing issues, both open and closed.

Describe the issue briefly here.

##### Describe your environment

1. OS Type: Windows 10
1. Python version: `$ python -V`: 3.6.4
1. Pipenv version: `$ pipenv --version`: 11.1.2

##### Expected result

No traceback

##### Actual result

```
PS C:\Users\brcan\Desktop\pipenv-test> py -3 -m pipenv install
Creating a virtualenv for this project…
Traceback (most recent call last):
  File "C:\Users\brcan\AppData\Local\Programs\Python\Python36\lib\runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "C:\Users\brcan\AppData\Local\Programs\Python\Python36\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "C:\Users\brcan\AppData\Local\Programs\Python\Python36\lib\site-packages\pipenv\pew\__main__.py", line 1, in <module>
    import pew
  File "C:\Users\brcan\AppData\Local\Programs\Python\Python36\lib\site-packages\pew\__init__.py", line 3, in <module>
    from . import pew
  File "C:\Users\brcan\AppData\Local\Programs\Python\Python36\lib\site-packages\pew\pew.py", line 39, in <module>
    import psutil
  File "C:\Users\brcan\AppData\Local\Programs\Python\Python36\lib\site-packages\pipenv\vendor\psutil\__init__.py", line 126, in <module>
    from . import _pswindows as _psplatform
  File "C:\Users\brcan\AppData\Local\Programs\Python\Python36\lib\site-packages\pipenv\vendor\psutil\_pswindows.py", line 16, in <module>
    from . import _psutil_windows as cext
ImportError: cannot import name '_psutil_windows'

Virtualenv location:
Creating a Pipfile for this project…
Creating a virtualenv for this project…
Traceback (most recent call last):
  File "C:\Users\brcan\AppData\Local\Programs\Python\Python36\lib\runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "C:\Users\brcan\AppData\Local\Programs\Python\Python36\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "C:\Users\brcan\AppData\Local\Programs\Python\Python36\lib\site-packages\pipenv\pew\__main__.py", line 1, in <module>
[...]

PR #1605, opened 2018-03-08
Vendor pew under script name `pewtwo`

techalchemy:
- Breaks dependency on psutil
- Continued work on #1587 and #1600


techalchemy:
The windows failure seems to be related to a parser error for the tap files + one error with the test marked as `deploy`, which passes locally for me.  I am pretty sure this is good to go

uranusjr:
Is the script needed at all? I get the impression everything it does is available via `python -m pipenv.pew`.

kennethreitz:
We've essentially forked pew at this point.

techalchemy:
`pipenv.pew` now calls this version of pew, and when we use the delegator direct call approach we call `pewtwo`

PR #1609, opened 2018-03-08
Don't assume grandparent process is the shell

### LINKED ISSUE #1601: pipenv shell failing due to SyntaxError: invalid syntax
When trying to run pipenv shell it fail to do so with error message below:
pipenv shell --verbose
  File "<string>", line 1
    c:\python27\python.exe -c "from pew.pew import prevent_path_errors; prevent_path_errors()"
     ^
SyntaxError: invalid syntax

##### Environment

1. Windows 7
2. Python 2.7.14
3. pipenv, version 11.1.4

##### Steps to replicate

just try to run pipenv shell.

techalchemy:
- Windows only (pew implementation bug)
- Traverse process tree until we find a known shell
  like cmd or powershell currently
- Fixes #1601

### DIFF
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -2028,6 +2028,7 @@ def do_uninstall(
 
 
 def do_shell(three=None, python=False, fancy=False, shell_args=None):
+    from pipenv.patched.pew import pew
 
     # Ensure that virtualenv is available.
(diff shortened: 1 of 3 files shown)


kennethreitz:
dont' forget to update history!

PR #1615, opened 2018-03-08
Bump VS Code support to no longer be listed as in-progress

brettcannon:
And since VS Code's integration is live, if there's any desire to update the instructions to point to VS Code instead of Sublime whose support is still in-development, let me know and I can send a PR for that.


techalchemy:
Thanks! Will defer to @kennethreitz on the second part of the question

kennethreitz:
\o/ ✨🍰✨

@brettcannon that'd be great.

brettcannon:
@kennethreitz Awesome! I opened Microsoft/vscode-python#988 to track this on our end so we don't forget about it. 😃

PR #1623, opened 2018-03-08
Use VS Code as the example editor for EDITOR usage

brettcannon:
Now that VS Code has pipenv support out of the box, might at well use an example that doesn't require any extra setup 😉


kennethreitz:
this pains me a bit (as a huge subl user), but i'll accept it :)

brettcannon:
@kennethreitz Hopefully I will convert you to a huge VS Code user 😉

kennethreitz:
well writing the sublime plugin is proving… frustrating, to say the least. so maybe :)

techalchemy:
As a VSCode user this satisfies me greatly

Issue #1620, opened 2018-03-08
Support `py` as an acceptable Python intpreter


brettcannon:
On Windows, Python is not put on `PATH` by default. Instead the `py` command is made available which operates somewhat like `pipenv --three` where a single command handles executing different Python versions. It would be great if pipenv picked up on `py` -- or Python versions installed by reading the Windows registry -- instead of only looking for `python` on `PATH`.

kennethreitz:
Someone had a working proposal in progress for this at some point, but it got lost in the shuffle. 

the tl;dr is basically — it heavily complicates our codebase, and ideally the "proper" way to install Pythons currently is to add them to your PATH. 

That's the answer we're going with *for now*. If you're open to contributing elegant PEP 397 support to our codebase in a way that doesn't totally overhaul what we're already doing, I'm open to it.

brettcannon:
And I totally understand and I realize that Windows is totally bucking the trend compared to UNIX and thus proper support is a cost that has to be weighed.

Looping in @zooba in case he has interest in this. I also wonder if this would be easier if `py` had a `--path` or something to emit the path to the interpreter so that something equivalent in PowerShell to ```py -3 -m pipenv --version `py -3 --path` ``` would work.

kennethreitz:
that'd be wonderful.

techalchemy:
@uranusjr and @ncoghlan were working on this as far as I knew, one of them will remember which PR we had

uranusjr:
#1150 and #1259. My approach is not really using py.exe, but reimplements the whole PEP 514 logic, which is probably too radical.

Re-thinking this right now it is probably doable with just one more subprocess call in Pipenv. Instead of using the executable’s path directly, we can do something like this ([in `ensure_python`](https://github.com/pypa/pipenv/blob/2dc06d4/pipenv/core.py#L550):

```python
path_to_python = …
path_to_python = subprocess.check_output([path_to_python, '-c', 'import sys; print(sys.executable)']).strip()
return path_to_python
```

to make sure we’re getting the path of the real executable (`python.exe`) instead of the shim (`py.exe`).

zooba:
As the PEP 514 author, I don't think it's too radical :)

See https://github.com/zooba/pep514tools if you want code that is free-as-in-MIT which you are welcome to simply take and vendor/reuse. I'm not really maintaining it, but only because nobody is using it and asking for changes.

techalchemy:
[...]

PR #1627, opened 2018-03-09
Ask py.exe for its opinion when choosing a Python

uranusjr:
Hopefully a viable alternative to #1620. In addition to looking for executables in PATH, also asks py.exe (if there is one) for its opinion when choosing an executable. Windows users would be able to use `--two`, `--three`, and `--python=<VERSION>`; this will be passed on to py.exe and ask if it can find any matching executable.

Non-Windows users will not be affected because there wouldn't be a `py.exe` to start with.

### DIFF
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -353,40 +353,77 @@ def ensure_pipfile(validate=True, skip_requirements=False):
             project.write_toml(p)
 
 
+def find_python_from_py(python):
+    """Find a Python executable from on Windows.
+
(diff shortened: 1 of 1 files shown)


uranusjr:
Wait, don’t merge yet, I need to fix this on Python 2.

uranusjr:
OK, fixed. Maybe this should be re-written to use delegator instead?

kennethreitz:
that would be more uniform with the codebase :)

PR #1668, opened 2018-03-11
Prep for warehouse release (ETA: EOM)

kennethreitz:
This pull request changes the defaults for PyPI, for when the URL officially changes later this month.


uranusjr:
Maybe also a good idea to detect the old URL from previously generated Pipfile, and print a courtesy note to prompt the user to switch?

kennethreitz:
The old version will continue to work for a long time.

uranusjr:
It seems pypi.org has left beta.

techalchemy:
@auvipy we will merge it, but there is no rush

PR #1729, opened 2018-03-14
Restore .env after testing

uranusjr:
Not sure if this is a good idea. The tests keep overriding my `.env` during development, so here’s a fixture to backup/restore it after being overwritten during tests.

### DIFF
--- a/tests/test_pipenv.py
+++ b/tests/test_pipenv.py
@@ -41,6 +41,23 @@ def pip_src_dir(request):
     return request
 
 
+@pytest.fixture()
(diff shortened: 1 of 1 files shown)


kennethreitz:
I was wondering what was going on here.

techalchemy:
Maybe we can add one to delete .venv

kennethreitz:
let's just not create it in the first place

uranusjr:
I do plan to look into that as well.

kennethreitz:
✨🍰✨

Issue #1651, opened 2018-03-10
Make a licensing note somewhere about `pipenv check`


ncoghlan:
While writing https://github.com/pyupio/safety-db/pull/2261, I realised that commercial redistributors of `pipenv` should take close note of the fact that `pipenv check` relies on a CC-BY-NC-SA CVE database maintained by pyup.io: https://github.com/pyupio/safety-db

This means that commercial redistributors of `pipenv` need to choose between:

* disabling `pipenv check`;
* patching `pipenv` to use a different vulnerability database (e.g. one they maintain themselves); or
* paying `pyup.io` for a commercial usage license

kennethreitz:
We actually embed a commercial license within the project, to ensure users get up-to-date security notifications.

Likely against TOS, but I’m one to ask forgiveness before permission. I know they’re Pipenv fans, so I’m sure they wouldn’t mind.

kennethreitz:
Er, not a license, an API key.

kennethreitz:
@pyupio any concern here? Happy to make immediate changes if so. :)

ncoghlan:
#1652 added a Sphinx note to https://pipenv.readthedocs.io/en/latest/advanced/#detection-of-security-vulnerabilities pointing out that while we're confident the "I'm a non-commercial user" case and the "We have our own pyup.io license" case are both fine, there are otherwise quite a few legal grey areas here, so commercial redistributors and end users will need to make their own assessment of the legal risks.

kennethreitz:
👍

jayfk:
> @pyupio any concern here? Happy to make immediate changes if so. :)

No, the use case perfectly fine! The reason the database is CC-BY-NC-SA is to get funding off of corporations who want to include it in their (closed) security products.

Having a little info text when running `pipenv check` should also wipe out all concerns about attribution. Something like `using Safety DB from YY-MM-DD` should work. This is, however, not a requirement.

If a licensing expert thinks that both licenses are incompatible, I'd be happy to relicense it for pipenv or work on another solution.

ncoghlan:
@jayfk Thanks for the clarification!

The murkiness I was worried about was that while `pipenv` itself is non-commercial, it definitely has commercial users, and potentially commercial redistributors as well (although I'm not personally aware of any of the latter yet - I just expect it may happen at some point in the future). 

I think between your comment above, and the note in the docs pointing out that commercial users and redistributors may want to follow up with pyup.io themselves, it's all good for now :)

[...]

PR #1730, opened 2018-03-14
Revise `pipenv check` caveat for pyup.io feedback

ncoghlan:
Jannis Gebauer of pyup.io let us know that he's definitely fine
with commercial *use* of the `pipenv check` feature, so it's
only commercial redistributors of `pipenv` that may need to
take a closer look at the licensing situation, and perhaps
talk to `pyup.io` directly.

Follow-up on #1651

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -186,11 +186,12 @@ Example::
 
 .. note::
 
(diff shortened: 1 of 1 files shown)


kennethreitz:
✨🍰✨

PR #1733, opened 2018-03-14
Add section on distinction between setup.py and Pipfile

ksonj:
I've prepared a small section on how Pipfile/pipenv differs from the available mechanisms in setup.py (`install_requires` etc). This is something I've struggled with repeatedly in the past and in essence this is what has crystallized to me from what I've read and experienced over the last few years of handling python projects. I'm not totally confident this is actually the Right Way™ to do things. If people disagree I'd be happy about some feedback.


kennethreitz:
this is fantastic.

kennethreitz:
✨🍰✨

Issue #1666, opened 2018-03-11
running the tests creates a .venv in the repo dir


kennethreitz:

kennethreitz:
this is driving me crazy

PR #1742, opened 2018-03-14
Fix venv creation during tests

uranusjr:
Turns out #1666 and #1729 actually have the same root cause, and fixing one resolves the other. Changes made in #1729 are reverted because they are not needed anymore.

Also added some extra tests that helped me locate this problem.

### DIFF
--- a/tests/test_pipenv.py
+++ b/tests/test_pipenv.py
@@ -127,19 +127,22 @@ class TestPipenv:
     def test_pipenv_venv(self):
         with PipenvInstance() as p:
             p.pipenv('--python python')
-            assert p.pipenv('--venv').out
+            venv_path = p.pipenv('--venv').out.strip()
+            assert os.path.isdir(venv_path)
 
     @pytest.mark.cli
     def test_pipenv_py(self):
         with PipenvInstance() as p:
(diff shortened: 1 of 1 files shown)


tests/test_pipenv.py
```diff
@@ -712,10 +700,9 @@ def test_shell_nested_venv_in_project(self, pypi):
 
     @pytest.mark.run
     @pytest.mark.dotenv
-    @pytest.mark.usefixtures('backup_dotenv')
     def test_env(self):
 
-        with PipenvInstance(pipfile=False) as p:
```
uranusjr: This is the actual fix.

tests/test_pipenv.py
```diff
@@ -712,10 +700,9 @@ def test_shell_nested_venv_in_project(self, pypi):
...
+        with PipenvInstance(pipfile=False, chdir=True) as p:
```
uranusjr: This is the actual fix.

kennethreitz:
✨🍰✨

tests/test_pipenv.py
kennethreitz: ✨🍰✨

PR #1749, opened 2018-03-15
Update `pipenv check` docs for Safety-DB removal

ncoghlan:
We removed the embedded copy of Safety-DB, so there's
no longer any copyright concern about the CC-BY-NC-SA
license, but it does mean `pipenv check` may end up throttled
eventually as all requests to the backend API use a common
key.


kennethreitz:
✨🍰✨

PR #1751, opened 2018-03-15
Reorder dependencies passed to setuptools's setup

giuscri:
Workaround for https://github.com/pypa/setuptools/issues/498

Otherwise `python setup.py develop` will fail.


giuscri:
ping

techalchemy:
@giuscri I have never encountered a problem with this. Can you explain the issue?

PR #1754, opened 2018-03-15
Cache parsed pipfile

jtratner:
On a (390+ line) Pipfile, it takes ~5s to parse the entire thing :O.  Pipenv has to parse the pipfile repeatedly and all over the place, so caching the contents speeds things up dramatically when trying to do `pipenv lock` (at least in this case).

This PR makes a little cache based upon file location + md5sum of contents for the pipfile (and the hashing is pretty fast here), which takes ~0.09ms on a cache hit. Given that the cache key is based on the file contents, should be completely fine to do (only possible issue is if parsed_pipfile gets mutated).  I tried returning a copy of the object, but `contoml` throws super weird exceptions if you do that, so I added a helper to clear the pipfile cache in places where it *might* be mutated (even tho I think it's likely overly defensive).

### DIFF
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -291,6 +291,7 @@ def ensure_pipfile(validate=True, skip_requirements=False):
     if validate and project.virtualenv_exists and not PIPENV_SKIP_VALIDATION:
         # Ensure that Pipfile is using proper casing.
         p = project.parsed_pipfile
+        p.clear_pipfile_cache()
         changed = ensure_proper_casing(pfile=p)
         # Write changes out to disk.
         if changed:
--- a/pipenv/project.py
+++ b/pipenv/project.py
@@ -1,5 +1,4 @@
 # -*- coding: utf-8 -*-
-import copy
 import json
 import os
 import re
@@ -304,7 +303,7 @@ class Project(object):
             parsed = self._parse_pipfile(contents)
             _cache.pipfile_cache[cache_key] = parsed
         # deepcopy likely unnecessary but why not avoid bugs?
-        return copy.deepcopy(_cache.pipfile_cache[cache_key])
+        return _cache.pipfile_cache[cache_key]
 
     def _parse_pipfile(self, contents):
         # If any outline tables are present...
--- a/pipenv/project.py
+++ b/pipenv/project.py
@@ -294,6 +294,9 @@ class Project(object):
 
     @property
     def parsed_pipfile(self):
+        """Parse Pipfile into a TOMLFile and cache it
+
+        (call clear_pipfile_cache() afterwards if mutating)"""
         # Open the pipfile, read it into memory.
         with open(self.pipfile_location) as f:
             contents = f.read()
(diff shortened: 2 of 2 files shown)


jtratner:
One thing remaining to test: `Pipfile` with unicode characters!  Will try to add a test for that later.

jtratner:
I think this is good to go now, it also passes all tests locally for me. 

Perf:

`pipenv lock` takes:
master - 2 minutes + (got bored of waiting - had not even gotten to resolving dev packages yet)
this branch - 45s


Got rid of the hashing because I realized it was (1) a false economy and (2) made str/bytes stuff confusing between python 2 and 3. I didn't add a unicode test because it turns out pipfiles with Unicode appear to be broken on master (something to do with TOML parsing - not sure if worth investigating).

kennethreitz:
why a thread local?

jtratner:
> why a thread local?

wasn't sure about multiprocessing / multithreading issues (i.e., if there was some parallel installation somewhere that might be impacted by it). Maybe being overly cautious?

Easy to change it to a global :)

I think I first tried storing it on the project object but that gets regenerated over and over.

techalchemy:
This is a neat idea -- how can we be sure we don't inadvertently miss something by keeping the pipfile cached?  What if we have eager updating in place and a new lock would update things? Should we have some kind of cache invalidation?

Locking itself isn't done in parallel afaik, probably why it's slow.

jtratner:
One thing: it reads the pipfile in each time and uses the contents as a
cache key. So only issue is if you mutate the pipfile, don’t write it to
disk, and also don’t clear the pipfile cache.

I think I caught every location with pipfile mutation (and nearly all of
them immediately persist the pipfile to disk) so pretty safe :)

So each project.parsed_pipfile call reads the file from disk and then uses
contents as cache key.

techalchemy:
@jtratner I saw that, I meant a timeout of some kind

jtratner:
How would a timeout matter/help? Ie be faster than reading the pipfile from
disk each time?

On my Mac this takes ~9ms per call to parsed_pipfile.

@kennethreitz - do you want me to switch it to normal global?

jtratner:
Also given that it’s either process-local or thread-local I think it’s
going to be pretty short-lived.

But definitely open to suggestions :)

dvf:
I suggest rather using Python's builtin (LRU cache)[http://book.pythontips.com/en/latest/function_caching.html] for this.

dvf:
Also, is this really necessary? If we're taking ~5s to parse the entire file, then maybe we should look at the cause of that.

techalchemy:
Re: cache timeout, I misread the intention. I have no real sense of how much of a problem this currently causes for the average use case, but if I’m not mistaken we already cache the lockfile. If the performance gain is potentially as dramatic as you say, it seems like a good idea. 

I would be curious about the impact on smaller Pipfiles. I haven’t managed to sit down and give this a full review but I might try it out on some of my own stuff tomorrow and get back to you

jtratner:
@techalchemy - I can profile it on a smaller pipfile (tho I went the dumb
route and just slapped a timing decorator on it)


Note that it *must* be as fast or faster because both implementations read
in the entire pipfile each time, the only extra expense (if everything was
a cache miss) is a dict lookup.

@dvf - do we want to add the lru cache backport as a dependency? I can
probably just whip up an equivalent decorator to work on python 2 (esp
since the cache eviction doesn’t matter to us)

uranusjr:
@jtratner A common pattern I find is to simply make the LRU cache decorator a non-op on Python 2. Things would still work, just not as good on Python 2.

kennethreitz:
There's no need for an LRU cache here. @jtratner if you switch this to a normal global I think we'll be good to go.

jtratner:
switched to normal global + rebased on top of master to make sure it all still passes

kennethreitz:
:cake:

Issue #1753, opened 2018-03-15
SyntaxError: invalid syntax


pglombardo:
When running `pipenv` I'm getting a syntax error.  `python -m pipenv.help` also results in a Syntax error.

This might be due to pipenv although I'm not completely convinced.  Any direction, hints or ideas would be appreciated

Python 3.5.2 (via pyenv)
pip 9.0.1 from /Users/pglombardo/.pyenv/versions/3.5.2/lib/python3.5/site-packages (python 3.5)
pipenv (11.8.0) installed via pip
MacOS Sierra

------------

##### Expected result

No syntax errors

##### Actual result

Stacktrace output:

```bash
$ python -m pipenv.help                                                                                                                                                     [17:06:06]
Traceback (most recent call last):
  File "/Users/pglombardo/.pyenv/versions/3.5.2/lib/python3.5/runpy.py", line 184, in _run_module_as_main
    "__main__", mod_spec)
  File "/Users/pglombardo/.pyenv/versions/3.5.2/lib/python3.5/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/Users/pglombardo/.pyenv/versions/3.5.2/lib/python3.5/site-packages/pipenv/help.py", line 9, in <module>
    from .core import project, system_which, find_python_in_path, python_version
  File "/Users/pglombardo/.pyenv/versions/3.5.2/lib/python3.5/site-packages/pipenv/core.py", line 24, in <module>
    import pipdeptree
  File "/Users/pglombardo/.pyenv/versions/3.5.2/lib/python3.5/site-packages/pipenv/vendor/pipdeptree.py", line 16, in <module>
    import pip9
  File "/Users/pglombardo/.pyenv/versions/3.5.2/lib/python3.5/site-packages/pipenv/vendor/pip9/__init__.py", line 28, in <module>
    from pip9.vcs import git, mercurial, subversion, bazaar  # noqa
  File "/Users/pglombardo/.pyenv/versions/3.5.2/lib/python3.5/site-packages/pipenv/vendor/pip9/vcs/subversion.py", line 9, in <module>
    from pip9.index import Link
  File "/Users/pglombardo/.pyenv/versions/3.5.2/lib/python3.5/site-packages/pipenv/vendor/pip9/index.py", line 31, in <module>
    from pip9.wheel import Wheel, wheel_ext
[...]

Issue #693, opened 2017-09-25
ImportError: No module named packages.urllib3.exceptions


cs01:
```bash
>> sudo pip install pipenv --upgrade
Requirement already up-to-date: pipenv in /usr/local/lib/python2.7/dist-packages
Requirement already up-to-date: virtualenv in /usr/local/lib/python2.7/dist-packages (from pipenv)
Requirement already up-to-date: pew>=0.1.26 in /usr/local/lib/python2.7/dist-packages (from pipenv)
Requirement already up-to-date: pip>=9.0.1 in /usr/local/lib/python2.7/dist-packages (from pipenv)
Requirement already up-to-date: pip-tools>=1.9.0 in /usr/local/lib/python2.7/dist-packages (from pipenv)
Requirement already up-to-date: requests>2.0.0 in /usr/local/lib/python2.7/dist-packages (from pipenv)
Requirement already up-to-date: flake8 in /usr/local/lib/python2.7/dist-packages (from pipenv)
Requirement already up-to-date: pathlib; python_version == "2.7" in /usr/local/lib/python2.7/dist-packages (from pew>=0.1.26->pipenv)
Requirement already up-to-date: shutilwhich; python_version == "2.7" in /usr/local/lib/python2.7/dist-packages (from pew>=0.1.26->pipenv)
Requirement already up-to-date: backports.shutil-get-terminal-size; python_version == "2.7" in /usr/local/lib/python2.7/dist-packages (from pew>=0.1.26->pipenv)
Requirement already up-to-date: pythonz-bd>=1.10.2 in /usr/local/lib/python2.7/dist-packages (from pew>=0.1.26->pipenv)
Requirement already up-to-date: virtualenv-clone>=0.2.5 in /usr/local/lib/python2.7/dist-packages (from pew>=0.1.26->pipenv)
Requirement already up-to-date: setuptools>=17.1 in /usr/local/lib/python2.7/dist-packages (from pew>=0.1.26->pipenv)
Requirement already up-to-date: click>=6 in /usr/local/lib/python2.7/dist-packages (from pip-tools>=1.9.0->pipenv)
Requirement already up-to-date: six in /usr/local/lib/python2.7/dist-packages (from pip-tools>=1.9.0->pipenv)
Requirement already up-to-date: first in /usr/local/lib/python2.7/dist-packages (from pip-tools>=1.9.0->pipenv)
[...]

PR #1758, opened 2018-03-16
Improve resolver path-patching

uranusjr:
Add the parent site-packages directory to sys.path *after* built-in ones, and only when it does not already exist.

This should vastly improve the resolver's robustness against hostile Python environments, such as #1753. It might help with issues like #693 as well, but I'm not sure (and not interested in figuring out either).

### DIFF
--- a/pipenv/__init__.py
+++ b/pipenv/__init__.py
@@ -19,7 +19,7 @@ try:
 except Exception:
     pass
 from .cli import cli
-from .import resolver
+from . import resolver
 
 if __name__ == '__main__':
     cli()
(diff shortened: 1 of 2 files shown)


uranusjr:
BTW why is `__init__.py` importing the resolver anyway? Can it be removed outright?

techalchemy:
This looks like it passed the tests that normally pass, so IMO this is probably good to go.

I'm not sure why init imports the resolver, but can you get this back in sync w/ master?

uranusjr:
done

PR #1763, opened 2018-03-16
pipenv.help output command formatting fixes

QasimK:
* The `<details>` tag did not always close itself
* GitHub seems to require an additional line break after HTML to respect Markdown formatting again

Both were encountered when I was creating an issue


techalchemy:
Thanks for the pull request, always appreciate the help!

PR #1768, opened 2018-03-16
Make white only for python 3 dev-packages

jtratner:
I think white only is available for Python 3+ (with no versions for python 2). This was causing test runner to fail for me locally.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -12,7 +12,7 @@ click = "*"
 pytest-pypy = {path = "./tests/pytest-pypi", editable = true}
 pytest-tap = "*"
 stdeb = {version="*", sys_platform="== 'linux'"}
-white = {version="*"}
+white = {version="*", markers="python_version >= '3.0'"}
 
 
 [packages]


jtratner:
I actually have no idea whether (practically) this should be an error or not. I.e. should pipenv "know" that there's only python 3+ for dep and just skip it?

techalchemy:
@jtratner in order for black to be installed at all we need to add:

```ini
[pipenv]

allow_prereleases = true
```

Since it's still in alpha.  Also I think black (the subdependency) is only available for 3.6

jtratner:
Don’t 100% understand what you mean - but feel free to put in whatever
works and/or not merge this - I just want to be able to run the test suite
on 2.7 ;)

uranusjr:
This PR is correct IMO. Whether we need to add the prerelease flag is another issue, but without the Python 3-only flag dev installation simply doesn’t work on Python 2.

uranusjr:
As of the detecting Python 3-only dependencies, it is absolutely possible (via classifiers), but I don’t think it’s particularly useful, or even reliable. Classifiers are a crapshoot except for the most well-managed packages because most authors just don’t bother (and don’t have time) to set them very well. Auto detection is destined to be faulty. Better to stay explicit.

Issue #1688, opened 2018-03-12
Allow Pipfile to consume environment variables


dvf:
It would be awesome if `Pipfile` could be parsed to allow for environment variables.

While it's great that `pipenv` supports private indexes, it's annoying if those indexes require credentials—it's bad practice to store them in source control:

```
[[source]]

url = "https://{{ MY_USERNAME }}:{{ MY_PASSWORD }}@my_custom_repo.jfrog.io/simple"
verify_ssl = true
name = "pypi"
```

Happy to contribute a PR for this if this gets some thumbs.

kennethreitz:
:+1:

gsemet:
@kennethreitz what do you think about making this section able to inherits values from a global setting file, such as /etc/pipenv.conf (a-la /etc/pip.conf), so all Pipfile within a private organisation can use the same internal repository (nexus, artifactory, other,...)

kennethreitz:
no

techalchemy:
@gsemet I believe pipfiles will use sources specified in your pip.conf now though

kennethreitz:
When we create a pipfile for you, it will be populated with sources from pip.conf.

gsemet:
not sure, we use our internal nexus cache in pip.conf:
```
$ cat /etc/pip.conf
[global]
trusted-host = nexus.mycompagny.fr
index = http://nexus.mycompagny.fr/repository/pypi/pypi
index-url = http://nexus.mycompagny.fr/repository/pypi/simple
```

when executing pip install on its own, it uses this server.

when creating a pipfile:
```
$ pipenv install request
...
$ cat Pipfile                                                                                                                                                                                              
[[source]]
name = "pypi"
verify_ssl = true
url = "https://pypi.python.org/simple"
...```

kennethreitz:
we pick up user pip conf i think, not global pip conf.

kennethreitz:
here's the code: https://github.com/pypa/pipenv/blob/master/pipenv/project.py#L419

dvf:
That sounds potentially dangerous if you’re storing credentials in pip.conf. 

@kennethreitz, I was reading another issue where you referred to the Pipfile as the source of truth. I think this is a great idea and should stay that way.

kennethreitz:
we could detect high-entropy strings automatically (or just urlparse).

kennethreitz:
@dvf pipfile is def the source of truth.

kennethreitz:
[...]

PR #1769, opened 2018-03-16
Allow values in Pipfile to consume Environment Variables

dvf:
Relying on `os.path.expandvars()` to inject environment variables in strings. Works on Windows, back-supports Python 2.7.

See https://github.com/pypa/pipenv/issues/1688

I'm not 100% sure that this is the correct place to do this injection.

Some thoughts:
- The Pipfile hash should be determined after values have attempted to be inserted.
- Pipfile shouldn't be intelligent by trying to differentiate env vars from regular characters—if an env var can't be inserted, the string should remain intact.
- As it stands, Pipenv's vendor folder contains some version of this repository. I don't fully understand the reasons behind this, maybe we could pin it as a requirement instead? We could then separate tests.

Mirrored Pipfile PR: https://github.com/pypa/pipfile/pull/105

### DIFF
--- a/pipenv/vendor/pipfile/api.py
+++ b/pipenv/vendor/pipfile/api.py
@@ -62,6 +62,24 @@ class PipfileParser(object):
     def __repr__(self):
         return '<PipfileParser path={0!r}'.format(self.filename)
 
+    def inject_environment_variables(self, d):
+        """
+        Recursively injects environment variables into TOML values
+        """
+
+        if not d:
+            return d
+
+        for k, v in d.items():
+            if isinstance(v, str):
+                d[k] = os.path.expandvars(v)
+            elif isinstance(v, dict):
+                d[k] = self.inject_environment_variables(v)
+            elif isinstance(v, list):
+                d[k] = [self.inject_environment_variables(e) for e in v]
+
+        return d
+
     def parse(self):
         # Open the Pipfile.
         with open(self.filename) as f:
@@ -78,8 +96,11 @@ class PipfileParser(object):
         config = {}
         config.update(default_config)
 
+        # Deserialize the TOML, and parse for Environment Variables
+        parsed_toml = self.inject_environment_variables(toml.loads(content))
+
         # Load the Pipfile's configuration.
-        config.update(toml.loads(content))
+        config.update(parsed_toml)
 
         # Structure the data for output.
         data = {
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -302,6 +302,27 @@ To prevent pipenv from loading the ``.env`` file, set the ``PIPENV_DONT_LOAD_ENV
 
     $ PIPENV_DONT_LOAD_ENV=1 pipenv shell
 
+☤ Support for Environment Variables
+-----------------------------------
+
+``pipenv`` supports the usage of environment variables in values. For example:
+
+    [[source]]
+    url = "https://${PYPI_USERNAME}:${PYPI_PASSWORD}@my_private_repo.example.com/simple"
+    verify_ssl = true
+    name = "pypi"
+
+    [dev-packages]
+
+    [packages]
+    requests = {version="*", index="home"}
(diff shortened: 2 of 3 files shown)


jtratner:
can you add a test that fails when an environment variable is not defined?

jtratner:
(looks like it doesn't error just doesn't expand  in python 3.6)

```
>>> os.path.expandvars('${A}')
'${A}'
```

dvf:
This seems correct to me, would you expect it to throw an error?

jtratner:
I'm concerned about something error about not connecting that turns out to be not being able to connect to `$USERNAME:$PASSWORD@mypypi.org` or something like that.

Way nicer to error and say `USERNAME not defined!` :)

Also how does this impact hashing? Which parts of the pipfile are allowed to have env vars in them?

jtratner:
(but I'm def not BDFL here :) )

pipenv/vendor/pipfile/api.py
```diff
@@ -62,6 +62,24 @@ def __init__(self, filename='Pipfile'):
     def __repr__(self):
         return '<PipfileParser path={0!r}'.format(self.filename)
 
+    def inject_environment_variables(self, d):
```
techalchemy: Generally anything in the `vendor` directory needs to be kept pristine, any modifications would require this package to move into the `patched` directory

More significantly you should simultaneously PR this change back up to the pipfile repository as well since we are not the only project which uses pipfiles

dvf:
After some thought I don't think this is ready. Would appreciate some clarity here.

Some thoughts:

- The `Pipfile` hash should *not* be affected by the values of the env vars, only the keys. This means the hash will have to be calculated before injection.
- **Disagree with @jtratner regarding raising errors:** `Pipfile` shouldn't be intelligent by trying to differentiate env vars from regular characters—if an env var can't be inserted, the string should remain the same.
- The reasons for having a vendor folder aren't clear to me if they haven't been patched: Why not just add `Pipfile` as a requirement?

jtratner:
@techalchemy - maybe only the sources section should allow env vars? And
then everywhere else it’s just interpreted literally? Or alternatively you
add another key that are the vars to substitute?

Generally, not sure where it would ever make sense to have `$` as a name.

If hash is calculated before env var substitution then should it be
disallowed in the packages sections?

dvf:
@techalchemy, I'm confused by your concern about package hashes. Maybe I'm missing something, but it would be weird if package hashes depended on the `Pipfile` itself. 

> Generally, not sure where it would ever make sense to have `$` as a name.

It's bad to make assumptions about values: `$` is an allowable character for an URL.

jtratner:
> `$` is an allowable character

Good point! Hmm...

dvf:
Thanks for everyone's interest. 
I'm taking the contributors' advice and opening https://github.com/pypa/pipfile/pull/105 instead.

dvf:
@kennethreitz @jtratner @techalchemy do you guys mind taking a peek at https://github.com/pypa/pipfile/pull/105

techalchemy:
We vendor things to isolate them from system dependencies. This is a common practice.

dvf:
I'm re-opening after taking suggestions and making the mirrored Pipfile https://github.com/pypa/pipfile/pull/105

pipenv/vendor/pipfile/api.py
dvf: I did that here: https://github.com/pypa/pipfile/pull/105

kennethreitz:
:cake:

kennethreitz:
Made some changes to this — primarily, env vars are not propogated to the pipfile.lock, as they are here. They are also expanded at runtime. We want to keep secrets out of source control.

njgrisafi:
It would be nice if we can have environment variables loaded throughout the entire Pipfile. E.g.
```
[dev-packages]
my-package = {path="${PATH_TO_REPO}", editable = true}
```

Issue #1773, opened 2018-03-17
prepare_pip_source_args doesn't properly handle URLs w/ user & password


asherf:
When the I use a custom source in the Pipfile that is protected by a user & password. 

The [logic in prepare_pip_source_args](https://github.com/pypa/pipenv/blob/2282e7045f08c0794973d296b11ea174e7fda48a/pipenv/utils.py#L268) tries to manually parse ParseResult.netloc (returned from urlparse) this results in the username being passed as the `--trusted-host` when `verify_ssl` is set to false in the Pipfile.

For example, if one of the sources in the Pipfile is `https://user:password@custom.example.com/simple`  
it will result calling `resolver.py` with the wrong params (passed via the PIPENV_PACKAGES environment variable) and called from venv_resolve_deps
the value of the ` --trusted-host` will be the username instead of the host name.

I am adding a [PR](https://github.com/pypa/pipenv/pull/1774) with tests to demonstrate & fix the issue.

PR #1774, opened 2018-03-17
fix prepare_pip_source_args not properly handle user & pw in repo URLs

asherf:
When the I use a custom source in the Pipfile that is protected by a user & password. 

The [logic in prepare_pip_source_args](https://github.com/pypa/pipenv/blob/2282e7045f08c0794973d296b11ea174e7fda48a/pipenv/utils.py#L268) tries to manually parse ParseResult.netloc (returned from urlparse) this results in the username being passed as the `--trusted-host` when `verify_ssl` is set to false in the Pipfile.

For example, if one of the sources in the Pipfile is `https://user:password@custom.example.com/simple`  
it will result calling `resolver.py` with the wrong params (passed via the PIPENV_PACKAGES environment variable) and called from venv_resolve_deps
the value of the ` --trusted-host` will be the username instead of the host name.

https://github.com/pypa/pipenv/issues/1773

### DIFF
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -265,7 +265,7 @@ def prepare_pip_source_args(sources, pip_args=None):
                     pip_args.extend(
                         [
                             '--trusted-host',
-                            urlparse(source['url']).netloc.split(':')[0],
+                            urlparse(source['url']).hostname,
                         ]
                     )
     return pip_args
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -358,3 +358,38 @@ twine = "*"
         assert not git_reformat.local_file
         # Test regression where VCS uris were being handled as paths rather than VCS entries
         assert git_reformat.vcs == 'git'
+
+    @pytest.mark.utils
+    @pytest.mark.parametrize(
+        'sources, expected_args',
+        [
+            ([{'url': 'https://test.example.com/simple', 'verify_ssl': True}],
+             ['-i', 'https://test.example.com/simple']),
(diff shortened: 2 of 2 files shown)


asherf:
I think one of the builds fail due to an error unrelated to my changes...
http://mercury.home.kennethreitz.org:8080/job/pipenv/201/PYTHON=CPython-2.7.14/console
```An error occurred while installing white==0.1.2! Will try again.
Collecting white==0.1.2 

  Could not find a version that satisfies the requirement white==0.1.2 (from -r /tmp/pipenv-tnzIc6-requirements/pipenv-eU03nG-requirement.txt (line 1)) (from versions: )
No matching distribution found for white==0.1.2 (from -r /tmp/pipenv-tnzIc6-requirements/pipenv-eU03nG-requirement.txt (line 1))

/var/lib/jenkins/workspace/pipenv/PYTHON/CPython-2.7.14/pipenv/utils.py:1233: ResourceWarning: Implicitly cleaning up <TemporaryDirectory '/tmp/pipenv-tnzIc6-requirements'>
  warnings.warn(warn_message, ResourceWarning)
Build step 'Virtualenv Builder' marked build as failure```

techalchemy:
Hey @asherf thanks for the PR! I was looking at our handling of this within the last few days as I was skimming some of our codebase and I was pretty skeptical that we were handling this. I haven’t had a chance to look over the changes here (been away from my desk all day) but I’ll get through the backlog tomorrow!

Thanks again!

PR #1777, opened 2018-03-18
Pass clear flag to resolver

puzan:
Fix clear for lock command. Clear flag was missed on resolver execution.


techalchemy:
This looks good to me. On mobile so will need to check CI results from home before merging

techalchemy:
Thanks for the contribution!

PR #1778, opened 2018-03-18
better cleanup_toml

giuscri:
Before this commit Pipfile had a more
than neeeded newlines between sections.

Currently generated Pipfile is
```
[[source]]

verify_ssl = true
name = "pypi"


[packages]



[dev-packages]



[requires]

python_version = "3.6"
```
while after this commit it would become
```
[[source]]
verify_ssl = true
name = "pypi"

[packages]

[dev-packages]

[requires]
python_version = "3.6"
```


techalchemy:
I think this was intentional but I’m not sure. There was an attempt at this in the past which broke pipfile parsing

Issue #1728, opened 2018-03-14
Docker build is failing


iddan:
Docker build is failing.

<details><summary>$ python -m pipenv.help output</summary>
Pipenv version: `'11.7.1'`

Pipenv location: `'/usr/local/lib/python3.6/site-packages/pipenv'`

Python location: `'/usr/local/opt/python3/bin/python3.6'`

Other Python installations in `PATH`:

  - `2.6`: `/usr/bin/python2.6`
  - `2.6`: `/usr/bin/python2.6`
  - `2.7`: `/usr/local/bin/python2.7`
  - `2.7`: `/usr/local/bin/python2.7`
  - `2.7`: `/usr/bin/python2.7`
  - `3.6`: `/usr/local/bin/python3.6m`
  - `3.6`: `/usr/local/bin/python3.6`

  - `2.7.10`: `/usr/bin/python`
  - `2.7.14`: `/usr/local/bin/python2`
  - `3.6.2`: `/usr/local/bin/python3`

PEP 508 Information:

```
{'implementation_name': 'cpython',
 'implementation_version': '3.6.2',
 'os_name': 'posix',
 'platform_machine': 'x86_64',
 'platform_python_implementation': 'CPython',
 'platform_release': '16.7.0',
 'platform_system': 'Darwin',
 'platform_version': 'Darwin Kernel Version 16.7.0: Thu Jan 11 22:59:40 PST '
                     '2018; root:xnu-3789.73.8~1/RELEASE_X86_64',
 'python_full_version': '3.6.2',
 'python_version': '3.6',
 'sys_platform': 'darwin'}
```

System environment variables:

  - `CAML_LD_LIBRARY_PATH`
  - `MANPATH`
  - `TERM_PROGRAM`
  - `ANDROID_HOME`
  - `SHELL`
  - `TERM`
  - `TMPDIR`
  - `PERL5LIB`
  - `Apple_PubSub_Socket_Render`
  - `TERM_PROGRAM_VERSION`
  - `TERM_SESSION_ID`
  - `OCAML_TOPLEVEL_PATH`
  - `USER`
  - `SSH_AUTH_SOCK`
  - `__CF_USER_TEXT_ENCODING`
  - `OPAMUTF8MSGS`
  - `PATH`
  - `PWD`
  - `LANG`
  - `XPC_FLAGS`
  - `XPC_SERVICE_NAME`
  - `SHLVL`
  - `HOME`
  - `LOGNAME`
  - `_`
[...]

PR #1793, opened 2018-03-19
Add git to Dockerfile for git installs

jtratner:
@kennethreitz - addresses comment in #1728 about git installs failing


kennethreitz:
:sparkles: :cake: :sparkles:

Issue #1801, opened 2018-03-20
--update option disappeared in version 11.8.3 ?


dimitris-k:
<details><summary>$ python -m pipenv.help output</summary>

Pipenv version: `'11.8.3'`

Pipenv location: `'/root/.local/lib/python3.5/site-packages/pipenv'`

Python location: `'/usr/bin/python3'`

Other Python installations in `PATH`:

  - `2.7`: `/usr/bin/python2.7`
  - `2.7`: `/usr/bin/python2.7`
  - `3.5`: `/usr/bin/python3.5m`
  - `3.5`: `/usr/bin/python3.5`

  - `2.7.13`: `/usr/bin/python`
  - `2.7.13`: `/usr/bin/python2`
  - `3.5.3`: `/usr/bin/python3`

PEP 508 Information:

```
{'implementation_name': 'cpython',
 'implementation_version': '3.5.3',
 'os_name': 'posix',
 'platform_machine': 'x86_64',
 'platform_python_implementation': 'CPython',
 'platform_release': '4.9.0-6-amd64',
 'platform_system': 'Linux',
 'platform_version': '#1 SMP Debian 4.9.82-1+deb9u3 (2018-03-02)',
 'python_full_version': '3.5.3',
 'python_version': '3.5',
 'sys_platform': 'linux'}
```

System environment variables:

  - `LESS_TERMCAP_us`
  - `PYTHONDONTWRITEBYTECODE`
  - `LC_CTYPE`
  - `PWD`
  - `LANG`
  - `USER`
  - `XDG_RUNTIME_DIR`
  - `PIP_PYTHON_PATH`
  - `_`
  - `OLDPWD`
  - `SSH_CONNECTION`
  - `LESS_TERMCAP_me`
  - `XDG_SESSION_ID`
  - `TERM`
  - `LESSOPEN`
  - `MAIL`
  - `SHLVL`
  - `LESS_TERMCAP_ue`
  - `LESSCLOSE`
  - `LESS_TERMCAP_so`
  - `SSH_AUTH_SOCK`
  - `SSH_TTY`
  - `PATH`
  - `HOME`
  - `SHELL`
  - `LESS_TERMCAP_mb`
  - `LOGNAME`
  - `LS_COLORS`
  - `PYTHONUNBUFFERED`
  - `SSH_CLIENT`
  - `LESS_TERMCAP_se`

Pipenv–specific environment variables:

Debug–specific environment variables:

[...]

PR #1803, opened 2018-03-20
remove --update option from the documentation

paraschas:
I've removed the --update option from the documentation, as it was recently removed from pipenv.


paraschas:
Closes #1801

PR #1813, opened 2018-03-21
Fix problem with incorrect import path

mmasztalerczuk:
At this moment tests don't work. The error is:

> Traceback:
> tests/test_vendor.py:11: in <module>
>     from pipenv.vendor.pipfile.api import PipfileParser
> E   ModuleNotFoundError: No module named 'pipenv.vendor.pipfile.api'


kennethreitz:
thanks

Issue #1188, opened 2017-12-10
exit vs deactivate


sobolevn:
A proper way to exit from the `pipenv shell` is by `exit` command.
But I am used to `deactivate` through all these years.

And whenever I accidentally use `deactivate` I have to close my terminal window and reopen it again.

Is there a way to fix this somehow?
Maybe some way to recover from the wrong command?
Or disallow `deactivate`?

uranusjr:
I believe you can still `exit` after accidentally using `deactivate`, can you not? I do like the idea of overwriting `deactivate` to say something like “Hey, you might want to use `exit` instead”, like what Python does when you enter `exit` without the parentheses.

kennethreitz:
no

sobolevn:
@uranusjr thanks for your advice! That's a good workaround.

@kennethreitz what's your vision of this problem? Should users just get used to it?

kennethreitz:
@sobolevn basically yes :)

PR #1814, opened 2018-03-21
suggest `exit` after deactivated pipenv shell

giuscri:
https://github.com/pypa/pipenv/issues/1188


kennethreitz:
I think our current implementation is fine. Thank you, though!

PR #1812, opened 2018-03-21
Resolve editable packages on the local filesystem

techalchemy:
- Check for editable pipfile entries when running `pipenv install`
- Check only when lockfile or virtualenv doesnt exist but pipfile does
- If so, first install the editable dependencies, then do resolution
- Required to accommodate `ireq.get_dist()` which we use for setup.py
  parsing
- Fixes #1782

PR #1815, opened 2018-03-21
Fix detection of editable packages

uranusjr:
Introduced in #1812. The values aren’t always dict, but can be strings… Also fix the `editable_dev_packages` call that is probably also wrong.

### DIFF
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -1868,7 +1868,7 @@ def do_install(
     if project.pipfile_exists and (
         not project.lockfile_exists or not project.virtualenv_exists
     ):
-        section = project.editable_packages if not dev else project.dev_editable_packages
+        section = project.editable_packages if not dev else project.editable_dev_packages
         for package in section.keys():
             converted = convert_deps_to_pip(
                 {package: section[package]}, project=project, r=False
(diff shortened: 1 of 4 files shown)


techalchemy:
Hm yeah I keep forgetting people hand generate pipfiles.

pipenv/core.py
```diff
@@ -1868,7 +1868,7 @@ def do_install(
     if project.pipfile_exists and (
         not project.lockfile_exists or not project.virtualenv_exists
     ):
-        section = project.editable_packages if not dev else project.dev_editable_packages
+        section = project.editable_packages if not dev else project.editable_dev_packages
```
techalchemy: Shhh this never happened

tests/test_pipenv.py
```diff
@@ -294,6 +294,24 @@ def test_install_without_dev(self, pypi):
             c = p.pipenv('run python -c "import tablib"')
             assert c.return_code == 0
 
+    @pytest.mark.cli
+    @pytest.mark.install
+    def test_install_without_dev_section(self, pypi):
+        with PipenvInstance(pypi=pypi) as p:
+            with open(p.pipfile_path, 'w') as f:
+                contents = """
```
jakul: I've got a little tip for you, apologies if you're already aware of it.

You can use `textwrap.dedent` to make this read more nicely in the source code:

```
                     contents = dedent("""
                          [packages]
                          tablib = "*"
                     """)
```

PR #1817, opened 2018-03-21
Pipfile.lock hashes back to determinitic order

### LINKED ISSUE #1529: Hash order is not deterministic
The order of hashes in both Pipfile.lock and --requirements output is not deterministic: every `pipenv update` shuffles them around which makes running an update and checking for changes pointless.

1. OS Type macOS High Sierra
1. Python version: `$ python -V` 3.6.4
1. Pipenv version: `$ pipenv --version` 10.1.2

NB I’m using devpi as a PyPI mirror.

##### Expected result

The hashes should be in a stable order. Easiest would be to just sort them alphabetically.

##### Actual result
 
The order changes with every run.

##### Steps to replicate

Install a bunch of packages, run update. Watch hashes permutate.

### LINKED ISSUE #1664: sort hashes in Pipfile.lock
I quickly searched the closed issues and don't seem to find something similar:

Is it possible to make pipenv sort the hashes before writing them to the Pipfile.lock file so as there are fewer changes to the file in each installation of a new module?

Thanks for the great project.

rconradharris:
The intent has been for hashes to be in a stable order to minimize diff
noise. Commit 1fa3e2a567da779b0c081ac28893aca3353c0674 caused a
regression here by converting to a set and back.

Fixes #1529
Fixes #1664

### DIFF
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -496,7 +496,7 @@ def resolve_deps(
             except (ValueError, KeyError, ConnectionError, IndexError):
                 if verbose:
                     print('Error generating hash for {}'.format(name))
-            collected_hashes = list(set(collected_hashes))
+            collected_hashes = sorted(set(collected_hashes))
             d = {'name': name, 'version': version, 'hashes': collected_hashes}
             if index:
                 d.update({'index': index})
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -1070,7 +1070,7 @@ def do_lock(
             {dep['name']: {'version': '=={0}'.format(dep['version'])}}
         )
         # Add Hashes to lockfile
-        lockfile['develop'][dep['name']]['hashes'] = sorted(dep['hashes'])
+        lockfile['develop'][dep['name']]['hashes'] = dep['hashes']
         # Add index metadata to lockfile.
         if 'index' in dep:
             lockfile['develop'][dep['name']]['index'] = dep['index']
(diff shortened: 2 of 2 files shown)


pipenv/utils.py
```diff
@@ -496,7 +496,7 @@ def resolve_deps(
             except (ValueError, KeyError, ConnectionError, IndexError):
                 if verbose:
                     print('Error generating hash for {}'.format(name))
-            collected_hashes = list(set(collected_hashes))
+            collected_hashes = list(sorted(set(collected_hashes)))
```
techalchemy: Is there ever a time when `sorted()` doesn't return a list? Can't we drop the `list()` call here?

techalchemy:
Thanks for the pr! One minor clarification and then I think we're good

pipenv/utils.py
rconradharris: That's a good point.

While we're fixing this we might want to remove the other `sorted` calls that were added here https://github.com/pypa/pipenv/commit/4d8e01b80f1560ce0fc5535f2fc04bca8cf30927 since they don't end up having an effect on the final result.

uranusjr: `sorted` is guarenteed to return a list, according to the docs. This might change in future Python releases (anything might, technically), but that would be highly unlikely.

rconradharris:
Few changes:

1. Just use `sorted` to convert to list
2. Remove unnecessary `sorted` calls that were added earlier that no longer have an effect since we sort at the last second
3. Typofix: `determinitic`  -> `deterministic` in title

uranusjr:
LGTM, waiting for CI to finish.

uranusjr:
Grr, CI passed, but the report got stuck. Need admin to merge.

techalchemy:
Waiting on CI, then merging

PR #1821, opened 2018-03-22
More documentation

uranusjr:
The diff is not helpful because I changed the line ending (was on Windows and did not notice). I added three sections:

* Your dependencies could not be resolved (additional content on how to find and manually delete cache)
* ValueError: unknown locale: UTF-8
* /bin/pip: No such file or directory


uranusjr:
Note to self: Add another section for #1827 and #1828?

techalchemy:
I'll hold off on merging for those if you like

PR #1823, opened 2018-03-22
test_complex_lock is not passing if python interpreter is not 2.7

mmasztalerczuk:
When I run tests I have a problem with `test_complex_lock`. The reason is `apscheduler` package which for python 3> doesn't need `funcsigs` and `futures` dependency. 
All other tests are working with python 3>

### DIFF
--- a/tests/test_pipenv.py
+++ b/tests/test_pipenv.py
@@ -249,8 +249,10 @@ class TestPipenv:
             c = p.pipenv('install apscheduler')
             assert c.return_code == 0
             assert 'apscheduler' in p.pipfile['packages']
-            assert 'funcsigs' in p.lockfile[u'default']
-            assert 'futures' in p.lockfile[u'default']
+
+            if p.pipfile['requires']['python_version'] == "2.7":
+                assert 'funcsigs' in p.lockfile[u'default']
+                assert 'futures' in p.lockfile[u'default']
 
     @pytest.mark.dev
     @pytest.mark.run


techalchemy:
This test should always pass. It always passes on Linux. Funcsigs should always be in the lockfile, it should just have markers as a key as well. Are you running tests yourself? What os?

mmasztalerczuk:
Yes, this test always passes but only if the version of intepreter is 2.7. Because if I use interpreter 3> for running of test , the test is downloading version `apschedular` which doesn't have funcsigs and futures (https://pypi.python.org/pypi/APScheduler ->  futures; python_version == "2.7"
funcsigs; python_version == "2.7" )

I don't know if somebody is running tests on python 3>, but after this changes all tests will be passed.

I have Archlinux, default python 3.6

techalchemy:
Yes I understand that the test passes running python3 pipenv into a 2.7 environment. The test case was my idea. 

The goal of pipenv is to have a resolver that when you run it on python 3, resolving a python 3 environment, it also includes optional packages that would only be installed on python 2. The goal of pipenv is not to arbitrarily force tests to pass

This is why I said before that these tests should always pass as-is. Not because I forgot that apscheduler resolves funcsigs for python 2, but because we rewrote the resolver specifically around this test case. 

I am going to close this because this test exists for a reason

mmasztalerczuk:
Ok, thanks for an explanation. So, I will look why these packages are not installed then.

Issue #1757, opened 2018-03-15
Markers now override unconditional dependencies resulting in a broken environment


jacebrowning:
This is a carryover from a discussion on https://github.com/pypa/pipenv/issues/1659 and the simplest way I could find to replicate this behavior.

With this `Pipfile` on macOS and pipenv 11.8.0:

```
[[source]]

url = "https://pypi.python.org/simple"
verify_ssl = true
name = "pypi"

[requires]

python_version = "3.6"

[packages]

arrow = "*"
coverage-space = "*"
faker = "*"
fire = "*"
freezegun = "*"
inflection = "*"
jupyter = "*"
pillow = "~=4.3"
pycodestyle = "~=2.3.1"
pylint = "~=1.7.5"
pytest = "~= 3.3"
pytest-cov = "*"
pytest-describe = "*"
pytest-expecter = "~= 1.1"
pytest-html = "*"
pytest-json-report = "~= 0.6"
pytest-mock = "*"
pytest-ordering = "*"
pytest-random = "*"
splinter = "*"
tldextract = "*"
watchdog = "*"
```

```
$ pipenv install

Creating a virtualenv for this project…
Using /Users/Browning/.pyenv/versions/3.6.3/bin/python3.6m (3.6.3) to create virtualenv…
...
Installing dependencies from Pipfile.lock (1ba60a)…
Ignoring colorama: markers 'sys_platform == "win32"' don't match your environment
Ignoring funcsigs: markers 'python_version < "3.0"' don't match your environment
  🐍   ▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉ 88/88 — 00:00:26
To activate this project's virtualenv, run the following:
 $ pipenv shell
```

which results in a broken environment:

```
$ pipenv run coverage.space
Traceback (most recent call last):
  File "/Users/Browning/.local/share/virtualenvs/colorama-demo-Q5bNGJGk/bin/coverage.space", line 7, in <module>
    from coveragespace.cli import main
  File "/Users/Browning/.local/share/virtualenvs/colorama-demo-Q5bNGJGk/lib/python3.6/site-packages/coveragespace/cli.py", line 24, in <module>
    import colorama
ModuleNotFoundError: No module named 'colorama'
```

despite `coverage-space` directly requiring it:

```
$ pipenv graph --reverse
...
colorama==0.3.7
[...]

PR #1825, opened 2018-03-22
Add tests for environment marker precedence logic

uranusjr:
This confirms #1757 is fixed. There is one potential discussion here, however.

Currently if your dependencies specify conflicting environment markers, Pipenv will only choose one of them, as demonstrated in the `test_specific_package_environment_markers` test case. I would instead argue Pipenv should try to aggregate the conditions, i.e. with two markers `sys_plarform=='win32'` and `os_name=='splashwear'`, the lockfile should contain something like

```
"markers": "sys_plarform=='win32' or os_name=='splashwear'"
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ python:
 env:
   - TEST_SUITE='dotvenv or check or unused or requirements'
   - TEST_SUITE='complex'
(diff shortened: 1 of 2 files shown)


techalchemy:
Thanks, waiting on CI and then merging :D

PR #1826, opened 2018-03-22
Normalize the package names in `_pipfile`

frostming:
Remove the `dict()` so that the mutation will affect `pfile`, `parsed_pipfile` will keep unchanged until the file is overwritten.

This fixes some issues when packages names don't comply to PEP423, see the tests.

And also a issue of `contoml` found when debugging


techalchemy:
Er you’re gonna have to describe the issue with contoml specifically

techalchemy:
Also note that we mock all of pypi so if you’re installing a package we don’t have mocked you need to add it _everywhere_ it needs to be mocked. So you should use a package we have mocked

frostming:
@techalchemy I don't quite understand

1. `items` returns a `dictitems` object and hence can't be connected with a list. The issue about `items` doesn't relate to the normalization, I put this in its own commit.
2. What packages are mocked? I picked one from `PYPI_VENDOR_DIR`, am I getting wrong?

techalchemy:
> items returns a dictitems object and hence can't be connected with a list. The issue about items doesn't relate to the normalization, I put this in its own commit.

I asked you to describe what the bug is with contoml.  This doesn't sound like a bug in contoml, it sounds like a problem with the approach you are taking to normalization.

frostming:
It is triggered when trying to print the contoml parsed object, which will call object.items() inside. The bug(illegal concatenating) lies there.

Please take a look at the diff

frostming:
@techalchemy I thought the `patched` directory was to host vendor libraries with some customizations, otherwise they will go to `vendor` dir. So the bug fixing about the library isn't included?

If that is true, then I agree with your workaround. The fix of `items` won't affect tests but only debugging so just leave it.

pipenv/project.py
```diff
@@ -353,7 +353,7 @@ def _pipfile(self):
         # mutation time!
         self.clear_pipfile_cache()
         for section in ('packages', 'dev-packages'):
-            p_section = dict(pfile.get(section, {}))
+            p_section = pfile.get(section, {})
             for key in list(p_section.keys()):
                 # Normalize key name to PEP 423.
                 norm_key = pep423_name(key)
```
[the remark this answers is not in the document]
frostming: After I looked into the `contoml/file.py`, I find that using `pop` here won't break anything. Thanks anyway. New modification is committed.

frostming:
The issue can be reproduced in this way:

1. Write `some_package` into `Pipfile` and do `pipenv install`
2. Do `pipenv install some-package`. pipenv thinks it is not included in `Pipfile` and add a duplicate line `some-package` into it.

But also it was an obvious coding fault: mutate a dict without ref to the original object. I don't see why this should be rejected.

techalchemy:
@frostming please have patience. I was not writing a summary for your benefit but for the benefit of others when I point them to review. I already agreed with you that this is a bug.

> But also it was an obvious coding fault: mutate a dict without ref to the original object. I don't see why this should be rejected.

I don't have any plans to reject this --  I only mean that we have to consider how fixing this bug might impact people who are using pipenv in production environments.  This is a large project in use by tens- to hundreds- of thousands of systems.  This change alters the representation of package names in existing pipfiles, even if it alters it to be correct. That might have unforeseen consequences, which would impact whether this gets merged as part of a major release or not.

pipenv/project.py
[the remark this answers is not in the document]
techalchemy: oh @frostming sorry it took me so long to see this here -- with this change, does pipenv properly normalize the pipfile again?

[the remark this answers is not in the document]
frostming: According to my search result of the code base, normalization will happen when Pipfile is changed: add package/remove package/add index

pipenv/core.py
```diff
@@ -288,7 +288,6 @@ def ensure_pipfile(validate=True, skip_requirements=False):
     if validate and project.virtualenv_exists and not PIPENV_SKIP_VALIDATION:
         # Ensure that Pipfile is using proper casing.
         p = project.parsed_pipfile
-        p.clear_pipfile_cache()
```
frostming: This line was buggy but was never hit because `PIPENV_SKIP_VALIDATION` is always true

pipenv/core.py
```diff
@@ -290,6 +289,7 @@ def ensure_pipfile(validate=True, skip_requirements=False):
                 err=True,
             )
             project.write_toml(p)
+			project.clear_pipfile_cache()
```
techalchemy: Please use spaces!!!

techalchemy:
I had some time to look into this.  Broadly, this behavior is the goal:

- Normalize package names with pep423 etc when _we_ write them to the pipfile (e.g. someone calls `pipenv install` from the command line)
- Normalize all package names written to the lockfile
- Return the normalized name for comparison against other packages in the index
- *Don't* normalize anything we didn't write -- if someone manually writes a lockfile, leave the names alone including capitalization etc

This isn't a criticism of your PR or your code, I'm not sure it necessarily applies. I am laying out the needs of this part of the code so that we can work with some clarity -- thanks for your patience and continued effort around this, I know this section of our code is not that pretty and we really do appreciate the help

tests/test_pipenv.py
```diff
@@ -539,6 +539,43 @@ def test_sequential_mode(self, pypi):
...
             assert c.return_code == 0
 
+    @pytest.mark.install
+    @pytest.mark.run
+    def test_pep423_name_install(self, pypi):
+        with PipenvInstance(pypi=pypi) as p:
+            with open(p.pipfile_path, 'w') as f:
+                contents = """
+[packages]
+python_dateutil = "*"
```
jtratner: can you add a test case that has an unnormalized name and confirm that it stays unnormalized and comments are preserved when you install a new package that doesn't overlap?

pipenv/project.py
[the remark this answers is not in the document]
jtratner: this might cause all comments in the file to be wiped out and re-normalize all names in the Pipfile (but hopefully we can add a unit test to check that doesn't happen)

frostming:
@techalchemy Sorry for the messing with TABs, fixed
With the fix of this PR, package names will be normalized when pipfile is written(install, uninstall). Moreover, all names are PEP423ed and recased when imported from a requirements.txt

@jtratner This case doesn't apply to this PR. But your proposal is perhaps better and there is another discussion for it https://github.com/pypa/pipenv/issues/1855#issuecomment-376402240

Thanks all

tests/test_pipenv.py
jtratner: can easily check this by adding to a couple more lines to the pipfile

```
# Pre-comment - I should totally be preserved
requests = "*"  # Inline comment - should definitely be preserved too!
```

jtratner:
@frostming - I'm just concerned that popping out all of the keys will remove comments, so would be great if you could update your test case to confirm that comments are not lost.

frostming:
@jtratner Please review

jtratner:
nice! This actually makes me realize another case (which wouldn't have to preserve the comments) - and I'm not sure whether it's covered in test cases or what current behavior is.

```
python_dateutil = "==1.5"
```

then you say `pipenv install python_dateutil`

should this change to?

```
python_dateutil = "*"
```

or stay the same?

Similarly, if you specify a version, should the pipfile be updated?

(@techalchemy - thoughts?)

frostming:
IMO "*" shouldn't overwrite. But another version specifier ">=", "<", "==" should overwrite that in Pipfile.

techalchemy:
The easy solution is to just rewrite the line without the comment, imo, but I dont know how hard it is to preserve the comment and just change the version

pipenv/project.py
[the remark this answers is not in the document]
techalchemy: @jtratner @frostming this is not actually desired behavior -- we don't want to rewrite the case of the entire pipfile every time someone installs a single package, we only want to normalize the one they are installing

frostming:
In contoml, you can change the value of a key while keeping the inline comment.

And if you delete the key, the inline comment will remain at the same line to become a line comment.

pipenv/project.py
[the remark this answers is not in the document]
frostming: Yup, so would you mind I change the behavior here, together with recasing issue #1855 ?

frostming:
@techalchemy @jtratner I refactored the recasing/normalization part a bit so that package names will be normalized only when user does `pipenv install`

Also fixes #1855 and covers it in test suites

frostming:
Oops, multiprocess is broken, needs more coding

techalchemy:
I don’t think the multiprocessing issue is your fault — our tests are behaving strangely right now

techalchemy:
ok sorry please update with master and we can resume actually working on stuff

pipenv/patched/contoml/file/file.py
```diff
@@ -231,7 +231,7 @@ def has_anonymous_entry():
         if has_anonymous_entry():
             return items
         else:
-            return list(items) + [('', self[''])]
+            return items + [('', self[''])]
```
uranusjr: Is this correct? I read the original issue description and it seems the old was correct; this would make the concat fail again because `items` is a `dict_items`, not list.

uranusjr: Maybe a test is needed for this.

frostming: That was a mistake when rebasing, revert it now. Thanks

techalchemy:
At a glance this looks okay to me, thanks a bunch for taking care of this! I would like another set of eyes just to verify this too before I merge so @uranusjr @ncoghlan if one of you guys wouldn't mind

uranusjr:
The normalisation part looks good to me. I don’t quite understand the caching part and won’t comment.

This and my previous script entry change makes me think we’d probably remove direct access to `parsed_pipfile` and wrap operations with functions instead. But that can be covered in a later refactor. This PR is good as-is.

pipenv/project.py
```diff
@@ -631,6 +625,10 @@ def add_package_to_pipfile(self, package_name, dev=False):
             p[key] = {}
         package = convert_deps_from_pip(package_name)
         package_name = [k for k in package.keys()][0]
+        name = self.get_package_name_in_pipfile(package_name, dev)
+        if name and name != package_name:
+            # Replace the packge name
+            del p[key][name]
```
ncoghlan: If I'm understanding this part of the change correctly, I don't think it does what we want in the case that @jtratner mentions in https://github.com/pypa/pipenv/pull/1826#issuecomment-376758741, where `Pipfile` contains a stricter version specifier than `project_name="*"` and the install command is just `pipenv install project_name`.

The simplest fix I can see would be to just leave the denormalised name alone in the case where it's already present - the change to normalise keys on lookup in  `get_package_name_in_pipfile` will handle such cases.

tests/test_pipenv.py
```diff
@@ -590,6 +590,55 @@ def test_sequential_mode(self, pypi):
...
+            c = p.pipenv('install requests')
+            assert c.return_code == 0
+            assert 'python_DateUtil' in p.pipfile['packages']
+            c = p.pipenv('install python_dateutil')
+            assert c.return_code == 0
+            assert 'python-dateutil' in p.pipfile['packages']
+            assert 'python_DateUtil' not in p.pipfile['packages']
+            contents = open(p.pipfile_path).read()
+            assert '# Pre comment' in contents
+            assert '# Inline comment' in contents
```
ncoghlan: Another couple of test cases are needed to check that:

1. the version constraint is preserved for the `python_DateUtil = "==1.5"` case if no version constraint is given on the command line

2. the version constraint is replaced if an explicit version constraint is given on the command line (including resetting it to a wildcard dependency with `pipenv install 'python_DateUtil=*'`)

frostming:
@ncoghlan Thanks for your inputs of great help!
>No need to reconvert package_name on every iteration - it can be converted once before the loop.

Fixed

>The simplest fix I can see would be to just leave the denormalised name alone in the case where it's already present - the change to normalise keys on lookup in get_package_name_in_pipfile will handle such cases.

Agree, see my latest commit.

> Another couple of test cases are needed to check that:
> 1. the version constraint is preserved for the python_DateUtil = "==1.5" case if no version constraint is given on the command line

Changed the tests a bit

> 2. the version constraint is replaced if an explicit version constraint is given on the command line (including resetting it to a wildcard dependency with pipenv install 'python_DateUtil=*')

Tests added. `install package=*` is neither supported by pip nor pipenv, we can improve this later. I don't do it in this PR.

techalchemy:
@frostming I think as long as using an actual pin (e.g. `package==x.y`) replaces the correct entry it is probably good. I know this wound up being a complicated pr so thank you for your patience here, this actually has the potential to break a lot of stuff, but some of that is currently broken, so disentangling what behaviors we actually _want_ was difficult. I’ll give it another look when I’m off mobile but otherwise I think we might be good!

tests/test_pipenv.py
```diff
@@ -590,6 +590,58 @@ def test_sequential_mode(self, pypi):
...
+Requests = "==2.14.0"   # Inline comment
+"""
+                f.write(contents)
+
+            c = p.pipenv('install')
+            assert c.return_code == 0
+
+            c = p.pipenv('install requests')
+            assert c.return_code == 0
+            assert 'requests' not in p.pipfile['packages']
```
techalchemy: I just want to say I am really happy with this, you did an amazing job organizing this PR and keeping the logic straight with all of the different requests and rewrites.  The fact that this assert statement works kind of blows my mind.

techalchemy:
Final update to make sure we aren't colliding with other changes and then merging, this looks good.

jtratner:
Wooo!

PR #1840, opened 2018-03-23
Detect .venv at project root when in subdirectory

jtratner:
Also remove `PIPENV_SHELL_COMPAT` which appears to be no longer used.

(confirmed test case fails without this change and passes with it).

Before:

```
$ PIPENV_VENV_IN_PROJECT=1 pipenv install
$ pipenv --venv
/Users/jtratner/example7/.venv
$ mkdir subdir
$ cd subdir
$ pipenv --venv # :(
No virtualenv has been created for this project yet!
```

After:

```
$ pipenv --venv
/Users/jtratner/example7/.venv
$ mkdir subdir
$ cd subdir
$ pipenv --venv
/Users/jtratner/example7/.venv
```

The full test suite doesn't pass locally for me right now so I can't tell if I've broken anything else.

### DIFF
--- a/HISTORY.txt
+++ b/HISTORY.txt
@@ -1,7 +1,7 @@
-latest:
- - Fix bug in detecting .venv at project root when in subdirectories.
 11.9.1:
  - Resolve editable packages on the local filesystem.
+ - Ensure lock hash does not change based on injected env vars.
+ - Fix bug in detecting .venv at project root when in subdirectories.
 11.9.0:
  - Vastly improve markers capabilities.
  - Support for environment variables in Pipfiles.
--- a/HISTORY.txt
+++ b/HISTORY.txt
@@ -1,3 +1,5 @@
+latest:
+ - Fix bug in detecting .venv at project root when in subdirectories.
 11.9.1:
  - Resolve editable packages on the local filesystem.
 11.9.0:
--- a/pipenv/environments.py
+++ b/pipenv/environments.py
@@ -14,8 +14,6 @@ PIPENV_PYTHON = os.environ.get('PIPENV_PYTHON')
 # Create the virtualenv in the project, instead of with pew.
 PIPENV_VENV_IN_PROJECT = bool(
     os.environ.get('PIPENV_VENV_IN_PROJECT')
-) or os.path.isdir(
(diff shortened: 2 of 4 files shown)


HISTORY.txt
```diff
@@ -1,5 +1,7 @@
 11.9.1:
  - Resolve editable packages on the local filesystem.
+ - Ensure lock hash does not change based on injected env vars.
```
jtratner: put this here because I'm assuming there will be a 11.9.1 release that will be pushed out, so best to put it here?

gsemet:
I usually do a symbolic link from ~/.venv to the pipenv virtualenv in my makefiles, so that ~./venv points out automatically to the current venv done my pipenv. Will this keeps it working?

jtratner:
@gsemet - presumably? This just changes the behavior when you're in a subdirectory of the folder wit the Pipfile. Doesn't change anything when you're in the current directory.

(But that said, it sounds like you could simplify your Makefiles by setting `PIPENV_VENV_IN_PROJECT=1` - that way pipenv would *build* it's virtualenv at `$PROJECT/.venv`)

gsemet:
I didn’t knew it thanks !

pipenv/environments.py
```diff
@@ -55,9 +53,6 @@
 PIPENV_SKIP_VALIDATION = True
 # Tells Pipenv where to load .env from.
 PIPENV_DOTENV_LOCATION = os.environ.get('PIPENV_DOTENV_LOCATION')
-# Use shell compatibility mode when using venv in project mode.
-if PIPENV_VENV_IN_PROJECT:
-    PIPENV_SHELL_COMPAT = True
```
uranusjr: Not sure why this was here in the first place (need to ask @kennethreitz), but some additional checks are likely required in `do_shell` if you delete this.

jtratner:
@uranusjr

I thought so too - but `PIPENV_SHELL_COMPAT` isn’t/wasn’t actually *used*
anywhere (at least not to my grepping). I think it was added a while back
and then got removed at some point, but without removing it from
environments.py

uranusjr:
I see. Thanks for the clarification!

PR #1842, opened 2018-03-24
Fix `shlex.split()` call to check for posix first

techalchemy:
- Currently it assumes posix complience and uses `posix=True`
- This breaks when parsing on windows

### DIFF
--- a/pipenv/project.py
+++ b/pipenv/project.py
@@ -369,8 +369,9 @@ class Project(object):
     @property
     def scripts(self):
         scripts = self.parsed_pipfile.get('scripts', {})
+        posix = os.name == 'posix'
         for (k, v) in scripts.items():
-            scripts[k] = shlex.split(v, posix=True)
+            scripts[k] = shlex.split(v, posix=posix)
         return scripts
 
     def update_settings(self, d):
--- a/pipenv/project.py
+++ b/pipenv/project.py
@@ -373,7 +373,7 @@ class Project(object):
         posix = os.name == 'posix'
         _scripts = {}
         for (k, v) in scripts.items():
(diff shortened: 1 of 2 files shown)


techalchemy:
Hm still broken actually

uranusjr:
I tried to write some unit tests around this and hit another bug

```toml
[scripts]
printfoo = "python -c \"print('foo')\""
```

This won’t print anything on Windows. `pipenv run python -c "print('foo')"` works fine. I plan to investigate, but want to put this up first.

uranusjr:
p.s. This issue makes me want to introduce an array-based syntax to script entries

```toml
[scripts]
printfoo = ["python", "-c", "print('foo')"]
```

It would make quoting a lot easier, especially for cross-platform projects.

uranusjr:
Okay so there So a couple of findings. (All of the followings assume Windows.)

1. Pipenv does not handle non-ASCII entries well *at all*.
2. Some additional parsing is required to handle quoting well. This works:

```toml
[scripts]
foo = "python -c print('foo')"
```

but this does not

```toml
[scripts]
foo = "python -c \"print('foo')\""
```

because `shlex.split` does not handle unquoting. This would be a problem if there are paths need escaping.

---

Both of these are pretty difficult to tackle (I think), and will need to be handled in a seperate issue. They are also edge cases and not critical though. This PR fixes common use cases.

jtratner:
This one should really never have worked:

```
[scripts]
foo = "python -c print('foo')"
```

at least not on POSIX

uranusjr:
@jtratner Yeah, it works only on Windows cmd (because `'` is not a quote). I fixed this in #1857.

PR #1845, opened 2018-03-25
Lint + test fixes

jtratner:
1. Fix up a test case that was never passing (but should have) <-- clearly wasn't running in CI (and I missed it as a last minute change right before merging)
2. Remove unused / duplicate imports
3. whitespace fixups

Now when you run flake8, mostly good:

```
$ flake8 tests/*.py pipenv/*.py
tests/test_utils.py:371:14: E131 continuation line unaligned for hanging indent
tests/test_utils.py:382:12: E131 continuation line unaligned for hanging indent
tests/test_utils.py:384:13: E122 continuation line missing indentation or outdented
pipenv/__init__.py:22:1: F401 '.resolver' imported but unused
pipenv/cli.py:15:1: F403 'from .environments import *' used; unable to detect undefined names
pipenv/cli.py:137:12: F405 'PIPENV_SHELL' may be undefined, or defined from star imports: .environments
pipenv/cli.py:140:27: F405 'PIPENV_SHELL' may be undefined, or defined from star imports: .environments
pipenv/cli.py:201:16: F405 'PIPENV_USE_SYSTEM' may be undefined, or defined from star imports: .environments
pipenv/core.py:2177:17: E211 whitespace before '('
```

### DIFF
--- a/tests/test_pipenv.py
+++ b/tests/test_pipenv.py
@@ -1,14 +1,12 @@
 import os
-from pkg_resources import parse_version
 import re
-import tempfile
 import shutil
 import json
 import pytest
 import warnings
 from pipenv.core import activate_virtualenv
 from pipenv.utils import (
-    temp_environ, get_windows_path, mkdir_p, normalize_drive, rmtree, TemporaryDirectory
+    temp_environ, get_windows_path, mkdir_p, normalize_drive, TemporaryDirectory
 )
 from pipenv.vendor import toml
 from pipenv.vendor import delegator
@@ -374,7 +372,6 @@ tablib = "*"
             assert 'requests' in p.lockfile['develop']
             assert 'flask' in p.lockfile['develop']
 
-
             c = p.pipenv('uninstall --all-dev')
             assert c.return_code == 0
             assert 'requests' not in p.pipfile['dev-packages']
@@ -383,7 +380,6 @@ tablib = "*"
             assert 'pytest' not in p.lockfile['develop']
             assert 'tpfd' in p.pipfile['packages']
             assert 'tpfd' in p.lockfile['default']
-
 
             c = p.pipenv('run python -m requests.help')
             assert c.return_code > 0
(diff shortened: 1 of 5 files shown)


tests/test_pipenv.py
```diff
@@ -1120,7 +1112,6 @@ def test_clean_on_empty_venv(self, pypi):
             c = p.pipenv('clean')
             assert c.return_code == 0
 
-
     @pytest.mark.install
     def test_environment_variable_value_does_not_change_hash(self, pypi, monkeypatch):
```
jtratner: is there some reason this test failure didn't bubble up from CI?

techalchemy: I'm kind of confused about why there is a need to use monkeypatch here.   Is that actually preferable over using the `temp_environ` context manager? If so, why?

jtratner:
I didn’t know about it :p can change it

tests/test_pipenv.py
frostming: It doesn't have a mark that is included in TEST_SUITE env
You may need to add `@pytest.mark.run` if it is proper for your case

jtratner:
rebased and got rid of monkeypatch, looks like somebody else fixed the actual test case (again, was last minute change-function-and-merge)

uranusjr:
I think the monkeypatch fixture is needed, otherwise the environs won’t be restored between tests, will they?

jtratner:
> I think the monkeypatch fixture is needed, otherwise the environs won’t be restored between tests, will they?

@uranusjr - I think the `temp_environ()` handles it

uranusjr:
Ah you’re right.

techalchemy:
FYI I have no actual objection to monkeypatch just noting that we are using the context manager in most other places and I’m not familiar with the tradeoffs. But the context manager will restore the original environment anytime the context is exited

techalchemy:
At a glance this seems uncontroversial. I’ll double check in the morning and merge. What was the tool you mentioned? Asking for a friends codebase at work ;)

techalchemy:
OK verified all of these, they are good to go

PR #1820, opened 2018-03-22
Add command descriptions in doc

frostming:
Update `sphinx-click` to the master branch

Closes #1819 

Preview
image

PR #1846, opened 2018-03-26
update Pipfile.lock

frostming:
- Update sphinx-click to latest version, #1820 
- Update other developing tools


jtratner:
are you sure this pipfile.lock is still good? I'd be concerned about rebasing with `Pipfile.lock` change.

frostming:
@jtratner Yes it is

PR #1850, opened 2018-03-26
Only consider supported ext when fetching metadata

uranusjr:
I don’t dare messing with pip internals, so I chose the least disruptive route. When trying to look for a package to resolve, pip (notpip, actually) should only look for package formats that it can actually handle. An extra flag is passed to signal this need.

### DIFF
--- a/tests/test_pipenv.py
+++ b/tests/test_pipenv.py
@@ -880,11 +880,17 @@ import records
 
     @pytest.mark.lock
     def test_lock_ignore_eggs(self, pypi):
-        with PipenvInstance(pypi=pypi) as p:
+        """Ensure locking works with packages provoding egg formats.
+
+        See https://github.com/pypa/pipenv/issues/1849.
+        TODO: This hits the real PyPI. What happens if RandomWords somehow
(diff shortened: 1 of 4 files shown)


tests/test_pipenv.py
```diff
@@ -878,6 +878,23 @@ def test_activate_virtualenv_no_source(self):
 
         assert command == '{0}/bin/activate'.format(venv)
 
+    @pytest.mark.lock
+    def test_lock_ignore_eggs(self, pypi):
+        """Ensure locking works with packages provoding egg formats.
+
+        See https://github.com/pypa/pipenv/issues/1849.
+        TODO: This hits the real PyPI. What happens if RandomWords somehow
+        changes its uploads? Need a better way to test this.
```
techalchemy: WE mock the rest of pypi, just mock it (or I can if you don't mind me pushing to this branch)

uranusjr: I’m not sure what I need to do here, so sure, go for it.

uranusjr:
Experimenting an alternative fix.

PR #1853, opened 2018-03-27
Set -o pipefail so errors propagate through piping

uranusjr:
This prevents the pipe syntax in the $TAP_OUTPUT branch from masking the underlying pytest failure.

Reference: http://redsymbol.net/articles/unofficial-bash-strict-mode/


techalchemy:
Thanks for taking care of this, maybe we will catch errors now

PR #1856, opened 2018-03-27
Misc. typos

luzpaz:
Found via `codespell -q 3`

### DIFF
--- a/run-tests.sh
+++ b/run-tests.sh
@@ -37,7 +37,7 @@ else
 
 		echo "Clearing Caches…"
 		rm -fr ~/Library/Caches/pip
-		rm -fr ~/Libary/Caches/pipenv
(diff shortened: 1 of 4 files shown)


run-tests.sh
```diff
@@ -37,7 +37,7 @@ else
 
 		echo "Clearing Caches…"
 		rm -fr ~/Library/Caches/pip
-		rm -fr ~/Libary/Caches/pipenv
+		rm -fr ~/Library/Caches/pipenv
```
luzpaz: please review

uranusjr: This fix is correct.

techalchemy:
Thanks for the PR, will merge from a PC since this client can’t override test failures

PR #1860, opened 2018-03-27
Combine RUN apt-get calls in Dockerfile

### LINKED ISSUE #1859: possible docker build issue
In your [Dockerfile](https://github.com/pypa/pipenv/blob/master/Dockerfile) you have `RUN apt update` and `RUN apt install ...` in separate lines. This is not recommended, since docker's caching can cause old versions of your packages to get installed.

From the [docs](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run):
> Always combine RUN apt-get update with apt-get install in the same RUN statement. For example:
> ```
>    RUN apt-get update && apt-get install -y \
>        package-bar \
>        package-baz \
>        package-foo
> ```
> Using apt-get update alone in a RUN statement causes caching issues and subsequent apt-get install instructions fail.

don-swept:
- with the calls in separate lines, there can be caching issues
- changed `apt` to `apt-get` to stop instability warnings

Fixes #1859

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,11 @@
 FROM ubuntu:17.10
 
 # -- Install Pipenv:
-RUN apt update
-RUN apt install software-properties-common python-software-properties -y
-RUN add-apt-repository ppa:pypa/ppa -y
-RUN apt update
-RUN apt install git pipenv -y
+RUN apt-get update \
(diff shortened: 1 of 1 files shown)


techalchemy:
Thanks for the PR!

Issue #1824, opened 2018-03-22
Is it possible to list which virtualenvs are managed by pipenv?


pfmoore:
In my `~/.virtualenvs` folder, I have a number of environments - some are managed by `pew` and others by `pipenv`. I can't easily tell which is which (although my guess is that the ones ending in `-<random letters>` are managed by pipenv), nor do I appear to be able to identify which project directory is associated with the pipenv-managed venvs.

Basically, I'm looking for a way to do housekeeping on old or experimental projects - identify and tidy up no longer used virtual environments, determine whether it's OK to delete a project directory or if it will leave a venv behind, etc. My `~` directory is on a limited-space SSD, so I'd prefer to keep it clean.

techalchemy:
Hey @pfmoore this is an idea we’ve kicked around a bit in the past but I believe we decided against. We’ve been primarily cutting down on features to focus on the basics lately

Given the use case this may be worth revisiting. @kennethreitz will have to weigh in

pfmoore:
Hi @techalchemy thanks for the information.

Looking at the code, it seems that the env name is based on a hash of the full pathname of the project. So it's not reversible - which means the only way to make that link would be by adding a pointer back to the project directory from the venv. That's a shame :-(

Also, that algorithm means that there's a potential issue - I hesitate to describe it as an out and out bug, but I'd be interested in what you think - on case-insensitive filesystems, where changing the case of the project directory name would break the link to the venv, even though logically the directory is the same. Certainly I normalise the case of directories on occasion, so it's not completely out of the question that this happens. I know that the venv is "throwaway" in the sense that a quick `pipenv install` should rebuild it, but again it has the potential to leave unwanted venvs behind. In my experience, Windows users in particular would expect to be able to change the case of a directory name without it mattering, so while it's not technically an OS-dependent issue, it probably does relate to user expectations on different operating systems...

techalchemy:
@pfmoore two points of response:

> the only way to make that link would be by adding a pointer back to the project directory from the venv

That would be the only solution to this particular problem, agreed.  I believe that is what `pew setproject` does  which was one consideration for implementation of this, the other was to just set a `PROJECT_HOME` variable or something like that and hash paths there to check for virtualenvs.  The alternative would be to write some pipenv-specific file.  I think both of these are pretty equally fine, possibly the second option is preferable.  Thoughts?

> on case-insensitive filesystems, where changing the case of the project directory name would break the link to the venv, even though logically the directory is the same

[...]

PR #1861, opened 2018-03-27
Associate project dir with new virtual environment

meshy:
This adds a flag to `pew` that lets it know which directory to switch to when activating an environment with `pew workon foo`.

I really miss this feature from `virtualenvwrapper`, and I'd love still use this workflow, so if I've gone about this the wrong way, please let me know, and I'll see what I can do to make it right.

This fits the workflow I'm used to, but I understand it might impact other people's workflows.

~~Because of that, this may be blocked waiting for the PR on `pew` to add an [option to prevent the directory changing when running `pew workon`](https://github.com/berdario/pew/pull/180), as that will offer a way to restore the existing behaviour.~~
**Update:** That's been merged now.

I wasn't sure where to add tests for this particular functionality, but I'm not entirely familiar with pytest, so I might have missed something.

If there's a place you think is appropriate for the tests, or a better way to achieve this, please let me know.


techalchemy:
Hey @meshy thanks for putting htis together, sorry for the delay! We had some issues with our testing infrastructure recently and we are just getting that sorted.  Once that's straightened out I'd say we can merge this as a bolt-on for now. We can sort out the testing bit if we ever actually build out the feature in pipenv.

/cc @pfmoore #1824

meshy:
Hey @techalchemy, no worries on the delay :)

I'm delighted you're happy with the concept of merging this. Thanks very much!

techalchemy:
plz update onto master and we can resume work

meshy:
No worries. I've just rebased, and repushed.

meshy:
Thanks for pointing me in the direction of Pipes -- it does look very nice, but I'm concerned about moving to a project that's only a little over a day old. I'll keep an eye on it though.

At the moment, my work-around for this is to make the following call immediately after creating the virtualenv:

```
pipenv install --three
echo $(pwd) > $(pipenv --venv)/.project
```

erinxocon:
@meshy can you pull from master and rebase?

uranusjr:
The tests failures are unrelated, but I think it’s best to fix them first before merging. We can’t have a release until everything’s back working anyway…

gtalarico:
> FWIW, this PR solves the issue that the Pipes author mentions in their follow-up tweet.
> The --link approach is kind of nasty but I could not think of another way to detect & tie envs to project dirs.

Hi @meshy -  How would this PR solve the issue? 
Pipes needs a way to from a pipenv folder, know which directory the env was created from.
Will the additional pew flags create a `.project` file in the env folder?

meshy:
@gtalarico Hey!

> Will the additional pew flags create a `.project` file in the env folder?

Yes, exactly `:)`

meshy:
Thanks!

Issue #1862, opened 2018-03-28
Conda-specific compatibility note in documentation


acdha:
I was attempting to help some visiting astronomers start using pipenv and ran into two problems with the instructions:

https://github.com/pypa/pipenv/blob/master/docs/advanced.rst#-pipenv-and-conda

The easy problem was that `/path/to/anaconda…` will not work if you simply copy and paste it but it's possible to miss that because while it prints “The path /foo/bar/baaz (from --python=/foo/bar/baaz) does not exist” that is buried under a lengthy traceback with the last thing people see being “TypeError: 'NoneType' object is not subscriptable”. Do you think it'd be worth a pull-request which would just end with the first message?

The more time-consuming problem was actually a Conda problem where the virtualenv creation failed with an error message early in the startup about being unable to `import _io` in the tempfile stdlib module, which was fixed at some point but hadn't been updated on their system. `conda update python virtualenv` fixed the problem but wasn't the first thing they were jumping to try. Should I contribute a brief note for the docs suggesting this more obviously?

techalchemy:
@acdha that would be greatly appreciated -- that particular `TypeError` pretty much always means we couldn't find whatever your executable was, the additional information is not very helpful I would agree.

As far as the conda integration, I am not too familiar with the typical conda workflow but if you think this is a common issue I'd probably track it in the normal conda integration notes, whereas if you think it's less common it might belong in the troubleshooting documentation.

Thanks for letting us know about this and for helping out! These kinds of things always fall to the wayside but are incredibly important

acdha:
I'll work on a pull request for the missing executables.

Re:Conda docs, I'm not sure I'd say it's common — I think it's the kind of situation where most people don't hit it but the ones who do can get really confused trying to search for a solution. I'll take a stab at proposing a doc update.

acdha:
As an aside, it feels like the `@click.option` stack in cli.py has too many instances where common options are repeated. Is there a plan to split those out or should I take a stab at it?

techalchemy:
@acdha the current plan has been focused on fixing testing infrastructure so I'm pretty sure this hasn't made it to anyone's priority list.  I am very far from a click expert so my involvement with the stack in `cli.py` is basically minimal, if you're willing and able to consolidate we'd appreciate it for sure

Karthik-PM:
can i help with this ?

matteius:
@Karthik-PM That would be great!

azizHakim:
[...]

PR #1864, opened 2018-03-28
Require “--python” values to exist

acdha:
This makes it easier for someone to realize when they have provided an invalid Python path without missing that message due to the sizable virtualenv traceback which follows it. I first encountered this when someone copy-and-pasted the example from the Conda documentation which has `--python=/path/to/anaconda/python`.

See #1862

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -67,6 +67,18 @@ def setup_verbose(ctx, param, value):
     return value
 
 
+def validate_python_path(ctx, param, value):
+    # Validating the Python path is complicated by accepting a number of
+    # friendly options: the default will be boolean False to enable
+    # autodetection but it may also be a value which will be searched in
+    # the path or an absolute path. To report errors as early as possible
+    # we'll report absolute paths which do not exist:
+    if isinstance(value, (str, bytes)):
+        if os.path.isabs(value) and not os.path.isfile(value):
+            raise click.BadParameter('Expected Python at path %s does not exist' % value)
+    return value
+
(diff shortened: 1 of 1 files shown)


uranusjr:
This is still WIP so I might be commenting too early, but aside from absolute path `--python` also accepts version-like names (e.g. `3.6`, `2.7.10`). They need to be taken into consideration when validating the argument.

acdha:
@uranusjr Yes — I ended up with a more conservative approach of only validating things which appear to be absolute paths to avoid that. It's worked in limited local testing but I'm still waiting for the test suite to finish running.

techalchemy:
hey @acdha when you get a second can you update w/ current master so we can get tests running against this?

techalchemy:
p.s sorry for the delay, tests were very broken (also sorry for the enormous pull you are going to have to do)

techalchemy:
If you could revisit this at some point it would certainly make a good addition!

acdha:
Sorry about the delays — I'm on parental leave the moment and don't have too much brainpower left for open-source 😀 

I rebased it against master and the tests are passing now on Travis so I removed the WIP prefix

techalchemy:
Congrats and thanks a bunch!

P.S. We actually met at a conference at the LoC a long time ago, Back then I was working as the 'Curator of Executable Content' on an archive project, small world!

acdha:
Ah, small world indeed! If you're ever near LC, feel free to stop by for
coffee.

Chris

PR #1865, opened 2018-03-28
Pass allow_prereleases setting to lock

### LINKED ISSUE #1835: --pre behaves weirdly in various commands
Be sure to check the existing issues (both open and closed!).

Describe the issue briefly here.

- `pipenv lock` should respect `allow_prereleases` in existing Pipfile
- `pipenv install packagename` fails to pass `--pre` either from command line or from Pipfile into `pipenv lock`, causes the command to fail during locking
  - Either we pass `--pre` or have` allow_prereleases` in the Pipfile, as long as there's package name in the cli arguments, it does not work, because of this code fails to passing `pre=pre`, to my guess: https://github.com/pypa/pipenv/blob/master/pipenv/core.py#L2005
- `pipenv update --pre --outdated` not respecting the `--pre` flag, removing `--outdated` works

------------

##### Expected result

None of the things above works when there's a prerelease dependency in the Pipfile, for example, `proto-google-datastore-v1beta3`

##### Actual result

They should all work (IMO, at least for the last two)

##### Steps to replicate

Pipfile is as follows:

```
[[source]]
verify_ssl = true
name = "pypi"

[packages]
"proto-google-datastore-v1beta3" = "*"

[dev-packages]
ipdb = "*"

[requires]
python_version = "3.6"

[pipenv]
allow_prereleases = true
```

Running any of the 3 commands above should fail with this Pipfile with the error

```
Could not find a version that matches proto-google-datastore-v1beta3
Tried: 0.0.2rc0, 0.0.3rc0, 1.0.0b0, 1.0.0b1, 1.0.0b2
```

frostming:
Fixes #1835 

I didn't find a prerelease package in mocked pypi so I uploaded one.

### DIFF
--- a/tests/test_pipenv.py
+++ b/tests/test_pipenv.py
@@ -955,6 +955,25 @@ requests = {git = "https://github.com/requests/requests", egg = "requests"}
 
     @pytest.mark.lock
     @pytest.mark.requirements
+    def test_lock_with_prereleases(self, pypi):
+
+        with PipenvInstance(pypi=pypi) as p:
+            with open(p.pipfile_path, 'w') as f:
+                contents = """
+[packages]
+sqlalchemy = "==1.2.0b3"
+
+[pipenv]
+allow_prereleases = true
+                """.strip()
+                f.write(contents)
+
+            c = p.pipenv('lock')
+            assert c.return_code == 0
+            assert p.lockfile['default']['sqlalchemy']['version'] == '==1.2.0b3'
+
+    @pytest.mark.lock
+    @pytest.mark.requirements
     @pytest.mark.complex
     @pytest.mark.maya
     def test_complex_deps_lock_and_install_properly(self, pypi):
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -503,9 +503,6 @@ def lock(
 
     # Ensure that virtualenv is available.
     core.ensure_project(three=three, python=python)
-    # Load the --pre settings from the Pipfile.
-    if not pre:
-        pre = core.project.settings.get('pre')
     if requirements:
         core.do_init(dev=dev, requirements=requirements)
     core.do_lock(
@@ -739,9 +736,6 @@ def update(
                 crayons.white('.', bold=True),
             )
         )
-        # Load the --pre settings from the Pipfile.
-        if not pre:
-            pre = core.project.settings.get('pre')
         core.do_lock(
             verbose=verbose, clear=clear, pre=pre, keep_outdated=keep_outdated
         )
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -1011,6 +1011,8 @@ def do_lock(
     from notpip._vendor.distlib.markers import Evaluator
     allowed_marker_keys = ['markers'] + [k for k in Evaluator.allowed_values.keys()]
     cached_lockfile = {}
+    if not pre:
+        pre = project.settings.get('allow_prereleases')
     if keep_outdated:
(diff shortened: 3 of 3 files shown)


tests/test_pipenv.py
```diff
@@ -934,6 +934,25 @@ def test_complex_lock_with_vcs_deps(self, pip_src_dir, pypi):
             assert 'python_dateutil' not in lock['default']
             assert 'python_dateutil' not in lock['develop']
 
+    @pytest.mark.lock
+    @pytest.mark.requirements
+    def test_lock_with_prereleases(self, pypi):
+
+        with PipenvInstance(pypi=pypi) as p:
+            with open(p.pipfile_path, 'w') as f:
+                contents = """
```
gsemet: what do you think about using dedent to keep a nice pythonic flow?

frostming: In fact TOML is insensitive of indents, I think we can just add them to make it prettier

techalchemy:
Thanks again, keep em coming. Sorry we are so slow right now just trying to get tests passing again before we start merging — really appreciate all the work you’re putting in! It’s a huge help

techalchemy:
OK, before we merge this I have one small item: currently if you `pipenv install --pre` or `pipenv install somepackage --pre` we simply run `project.update_settings({'allow_prereleases': pre})` -- Similarly, we should simply do this in `cli.py` so that when we query it later we have access to accurate info. 

As it stands with your changes, `pipenv lock --pre` won't ever generate a lockfile with prereleases in it unless it was previously set in the Pipfile.  

@ncoghlan does the behavior I described seem like what you would expect?

frostming:
> Similarly, we should simply do this in `cli.py` so that when we query it later we have access to accurate info.
Do you mean we should update settings in `cli.py` with respect of `pre` flag and `core.py` always read the `allow_prereleases` from `Pipfile`? In that way `core.py` won't need to get `pre` passed in at all.

> `pipenv lock --pre` won't ever generate a lockfile with prereleases in it unless it was previously set in the Pipfile.

Not really, `do_lock` also reads the `pre` passed in.

ncoghlan:
I think having this behave the same way as `--keep-outdated` makes sense, where the command line setting overrides the `Pipfile` setting, makes sense, so +1 from me for the PR (pending a rebase and passing CI).

However, it does make me wonder if we may want to expand the `_meta` section in lock files to capture some additional information:

* `locked_at`: when the `Pipfile.lock` was generated (while it could be argued that this hinders file reproducibility, my view would be that it would enhance it, since you'd know that to get the same pinned versions, you'd have to ignore releases after that date & time)
* `kept_versions`: which versions in the current lock file have been retained from a previous version of the lock file by `--keep-outdated` (a list of package names would suffice, since the exact versions are down in the package entries)
* `prereleases_allowed`: boolean setting

@techalchemy Do you think it would make sense to file a separate issue for that "Lock file reproducibility" idea? I don't think it's an urgent addition, at all, but I do think it could be helpful to third party auditing tools (and potentially even pipenv's own self-tests)

techalchemy:
@ncoghlan I do think that would make sense, from an audit standpoint it would be very clear then what was changing from update to update which adds a lot of value.  And yeah, from a development perspective that's usually something I'm trying to track down with a debugger or a diff so it would improve my own workflow.  Thanks for the input as always!

frostming:
Rebase done. Is there anything else I need to do?

techalchemy:
No I just wanted to merge the 3.5 update first, but tests are failing now for reasons that are unclear

PR #1869, opened 2018-03-28
use textwrap.dedent in unit tests

gsemet:
keep the python flow

Signed-off-by: Gaetan Semet <gaetan@xeberon.net>


gsemet:
nop pbl, i can make the rebase later

techalchemy:
I see where you're coming from but there are a lot of conflicts and I'm not sure this is the cleanup we need at the moment

PR #1874, opened 2018-03-29
Remove extraneous is_vcs check

jtratner:
@techalchemy - as we discussed previously - this removes some code that can never be executed.

1. is_vcs requires (req.path or req.uri) to be truthy.
2. (req.path or req.uri) would be true in first if statement

Thus if you have:

```
if (req.uri or req.path or ...) and not req.vcs
elif req.vcs or is_vcs(req.path)
```

if `is_vcs(req.path)` is Truthy, then the first if clause must be
truthy, so it's unnecessary.

Also add a quick test to confirm that originating purpose for that check
is now handled otherwise.

### DIFF
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -535,8 +535,8 @@ def convert_deps_from_pip(dep):
         # Add --editable if applicable
         if req.editable:
             dependency[req.name].update({'editable': True})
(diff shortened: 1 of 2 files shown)


jtratner:
np @techalchemy - thanks for the quick review :)

PR #1875, opened 2018-03-29
Fix scripts quoting

jtratner:
Better shell quoting

1. Only use split() when we actually come in from CLI (possibly never
   need to do that at all) - e.g. 
```
[scripts]
docs = "bash -c \"cd docs && make html\""
```
now works!

2. Add some additional test cases to cover quoting.
3. Better error message in case of missing executable pulled from a run
script.

E.g.:

```
[scripts]
random = "myexecutable a b c"
myexecutable = "echo 5"
```

```
Error: the command myexecutable (from random) could not be found within PATH.
```

vs. previous, which had a slightly disingenous error message.

```
Error: the command myexecutable could not be found within PATH or Pipfile's [scripts].
```

### DIFF
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -2184,20 +2184,30 @@ def inline_activate_virtualenv():
         )
 
 
+def _construct_run_command(project, command, args):
+    """Construct a command to run concatenating project and args.
+
+    ``command`` and/or the script from the ``Pipfile`` will be inserted before args.
+
+    Returns (executable, args for executable)"""
+    args = list(args or [])
+    # Script was found…
+    if command in project.scripts:
+        prefix_argv = project.scripts[command]
+    else:
+        # Separate out things that were passed in as a string.
+        prefix_argv = list(command.split())
+    executable, prefix_argv = prefix_argv[0], prefix_argv[1:]
+    for __c in reversed(prefix_argv):
+        args.insert(0, __c)
+    return executable, args
+
+
 def do_run(command, args, three=None, python=False):
     # Ensure that virtualenv is available.
     ensure_project(three=three, python=python, validate=False)
     load_dot_env()
-    # Script was found…
-    if command in project.scripts:
-        command = ' '.join(project.scripts[command])
-    # Separate out things that were passed in as a string.
-    _c = list(command.split())
-    command = _c.pop(0)
-    if _c:
-        args = list(args)
-        for __c in reversed(_c):
-            args.insert(0, __c)
+    executable, args = _construct_run_command(project, command, args)
     # Activate virtualenv under the current interpreter's environment
     inline_activate_virtualenv()
     # Windows!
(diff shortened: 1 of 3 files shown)


jtratner:
This might need some skipping on windows - @uranusjr / @techalchemy wdyt?

uranusjr:
> This might need some skipping on windows

The test cases provided here should work on Windows. If they don’t, it’s a bug.

pipenv/core.py
```diff
@@ -2184,44 +2184,66 @@ def inline_activate_virtualenv():
...
+    ``command`` and/or the script from the ``Pipfile`` will be inserted before args.
+
+    Returns (executable, args for executable)"""
+    args = list(args or [])
+    # Script was found…
+    if command in project.scripts:
+        prefix_argv = project.scripts[command]
+    else:
+        # Separate out things that were passed in as a string.
+        prefix_argv = list(command.split())
```
frostming: maybe `shlex.split` in case  we pass in complex command?

pipenv/core.py
```diff
@@ -2184,44 +2184,66 @@ def inline_activate_virtualenv():
...
+                click.echo(
+                    '{0}: the command {1} could not be found within {2} or Pipfile\'s {3}.'
+                    ''.format(
+                        crayons.red('Error', bold=True),
+                        crayons.red(command),
+                        crayons.normal('PATH', bold=True),
+                        crayons.normal('[scripts]', bold=True),
+                    ),
+                    err=True,
+                )
```
frostming: These are good, but why not Windows?

frostming:
@jtratner, @uranusjr , FYI, The tests can't run through on my Windows machine. It turns out to work when I change to `shlex.split(command, posix=True)`. This was changed in #1842, I don't know by what concern the `posix` flag is changed. I can see in the discussion this change didn't fix things.

So can we just revert the posix flag back to `True`?

pipenv/core.py
frostming: Okay, I think no one would ever do `pipenv run "foo 'bar bla'".

techalchemy:
It's possible we can still slap `shlex.split()` in here, i believe the posix flag just determines whether we treat the input as a list or not

pipenv/core.py
uranusjr: Why? Windows also uses `PATH` for lookup, and the error message applies as well on Windows to me.

uranusjr: Ah, I see what you mean, sorry. The Windows branch needs the same error messages as well. Right.

pipenv/core.py
uranusjr: If they do, it should be parsed as-is. I mean, maybe they actually have a command called `"foo 'bar bla'"`, who knows.

jtratner:
In this case, I think it'd be much easier to have separate `run_windows` and `run_posix` functions (and tests!) rather than attempt to do some kind of cross-platform wizardry.  It's a little hard to figure out how windows *ought* to work as well.

More generally: if you dump a line like `bash -c "something && another_thing 'with quotes'"` in a Pipfile, is it supposed to work cross-platform?

I'd argue for "let's do best effort but you probably should dump everything in a script file otherwise"

tests/test_pipenv.py
```diff
@@ -1167,11 +1168,13 @@ def test_environment_variable_value_does_not_change_hash(self, pypi):
     def test_scripts_basic(self):
         with PipenvInstance(chdir=True) as p:
```
techalchemy: Pretty much this entire test is invalid for windows -- you can't escape double quotes on windows, period. So There is no way to make this test pass on windows.

jtratner: yeah I think you should just skip the second half or split it in two - not super familiar with windows

PR #1896, opened 2018-04-03
Hash cache

jtratner:
Re-use pip's safe file cache to store hash values for packages.

Set up is:

1. only hash if the remote provides a checksum/hash in URL (e.g. `https://pypi.python.org/mypkg/whatever-path.whl#md5=....`)
2. Store hash in a path including the checksum/hash in the URL (again, only if valid)
3. always return a hash

I will remove print statements before merging.

Gotta think about how to test this too...

### DIFF
--- a/pipenv/patched/piptools/repositories/pypi.py
+++ b/pipenv/patched/piptools/repositories/pypi.py
@@ -55,9 +55,10 @@ class HashCache(SafeFileCache):
             hash_value = self.get(location.url)
         if not hash_value:
             hash_value = self._get_file_hash(location)
+            hash_value = hash_value.encode('utf8')
         if can_hash:
             self.set(location.url, hash_value)
-        return hash_value
+        return hash_value.decode('utf8')
 
     def _get_file_hash(self, location):
         h = hashlib.new(FAVORITE_HASH)
--- a/pipenv/patched/piptools/repositories/pypi.py
+++ b/pipenv/patched/piptools/repositories/pypi.py
@@ -14,6 +14,7 @@ from notpip.wheel import Wheel
 from notpip.req.req_install import InstallRequirement
 from pip9._vendor.packaging.requirements import InvalidRequirement
 from pip9._vendor.pyparsing import ParseException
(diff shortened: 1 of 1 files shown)


jtratner:
Reduces warm lock time from 13.5 minutes (where hash component took 12-13 minutes) to 1.3 minutes! (of which the hash component takes 30s)

techalchemy:
@ncoghlan would you mind giving this a once over for security concerns? This is a sizeable performance gain

techalchemy:
@jtratner @vphilippon I see that pip-tools has brought in some local repository hashing upstream, is this going to be impacted by that at all if we bring it over?

see https://github.com/jazzband/pip-tools/blob/master/piptools/repositories/local.py#L67-L78 --

```python
    def get_hashes(self, ireq):
        key = key_from_req(ireq.req)
        existing_pin = self.existing_pins.get(key)
        if existing_pin and ireq_satisfied_by_existing_pin(ireq, existing_pin):
            hashes = existing_pin.options.get('hashes', {})
            hexdigests = hashes.get(FAVORITE_HASH)
            if hexdigests:
                return {
                    ':'.join([FAVORITE_HASH, hexdigest])
                    for hexdigest in hexdigests
                }
        return self.repository.get_hashes(ireq)
```

jtratner:
> I see that pip-tools has brought in some local repository hashing
upstream, is this going to be impacted by that at all if we bring it over?

Good point. Aside from making the diff more complicated, I think it’s fine.

It’s slightly obscured, but the hash key is based on the full URL including
the fragment (and is only cached if the fragment provided some kind of
checksum) so it’s effectively like caching on a web ETag. Ie the server
would have to be lying to us about the hash, but we are already trusting
the server to give us the right stuff.

jtratner:
Once you merge in the fix tests branch I’ll clean this up :)

techalchemy:
plz rebase :D

techalchemy:
I merged in master since it was doable automatically although I see now you want to clean this up first so I'll just leave it

PR #1902, opened 2018-04-03
Man page section about `tox.ini` updated (no need for `pipenv run`)

vlcinsky:
As `pipenv` runs in `command` section in `tox` activated virtualenv,
there is no need to run `pipenv run <some_command>` and commands can
be run directly as `<some_command>`. This also relaxes the requirement
for passing `HOME` environment variable into `pipenv run` call.


techalchemy:
Uh I don't really know anything about tox, someone else should really review this...

techalchemy:
@vlcinsky are you knowledgeable w.r.t. tox? I actually tried to configure it for pipenv while fixing testing (was trying to see if I could make appveyor test runs speed up with detox or something like that) but I wound up just slowing things down dramatically...

vlcinsky:
@techalchemy I feel very comfortable with `tox` on localhost and I did few CI integrations with it (.travis.yml). But distributed runs using `detox` is not the area of my expertise.

I am afraid there are numerous moving parts (concept of `tox.ini` defined test on it's own, `detox` configuration, willingness of hosting system to run parallel processes ...).

I would first review basic test (either using `tox` or `run-tests.sh` or whatever for possible problems.

techalchemy:
I made a few changes because I want to keep our instructions and documentation consistent, `pipenv run` will work correctly even in an activated virtualenv and keeping this consistent reduces the risk of confusion -- I don't think we need to go into detail about when you can and can't skip it

Otherwise, looks good, merging

vlcinsky:
@techalchemy removal of `pipenv run` removed real motivation of my PR.

You might be aware of common misconception and confusion about "starting virtualenv" - after learning about activating virtualenv, people do not understand, it only simplifies calls to underlying python interpreter within that virtualenv and mistakenly expect some special magic. This results in attempts to run activation script before using scripts from the virtualenv even though it would be sufficient to simply call the python interpreter using absolute path.

The situation in `tox.ini` is similar, `tox` user is used to install some python libraries into virtualenv and then simply use it. Adding `pipenv run <command>` gives false impression, that commands installed via `pipenv` shall be magically run via `pipenv run` even though it is not true.

Imagine a test suite, which has couple of test cases, where are called python package provided CLI commands. Do we expect them to be modified to `pipenv run <command>` variant?

Definitely not, but man page is currently contributing to such misunderstanding.

PR #1916, opened 2018-04-05
Fix Python version parser for 2.7.0

uranusjr:
At this point I think it is better to just parse the whole thing with regex instead. I read some past commits on this part of code and found d00d33be. @eli-b can you help me test if Miniconda’s version output works with this regex pattern?


techalchemy:
rebase => master plz

techalchemy:
This seems good, merging

PR #1909, opened 2018-04-04
Scripts and test fixes

uranusjr:
Extracted from the requirements refactor branch. @techalchemy you’ll probably get a ton of conflicts when you try to merge that branch…

1. Update lock file so things can run on Appveyor.
2. Add an extra job to Travis to make sure all tests are run.
3. Enable `takes-forever` regardless of branch (so I can know if the tests are really fixed). This can be reverted after merging this PR.
4. Add Appveyor configuration (current build status: https://ci.appveyor.com/project/uranusjr/pipenv).
5. Refactor the script parser to fix things (with tests).
6. Mock maya to make the test pass.

PR #1922, opened 2018-04-05
Fix tests and scripts

techalchemy:
- Fix tests & scripts parsing
- Add appveyor script
- Add missing pypi mocks
- Update Lockfile
- Switch to pathlib2 as fallback


pipenv/project.py
```diff
@@ -388,9 +389,22 @@ def settings(self):
...
+        except KeyError:
+            return False
+
+    def build_script(self, name, extra_args=None):
+        try:
+            script = Script.parse(self.parsed_pipfile['scripts'][name])
+        except KeyError:
+            script = Script([name])
+        except ValueError:
+            raise ValueError('invalid script entry {0!r}'.format(name))
```
jtratner: The only way you can get here is like this, right?

```
[scripts]
tests = "bash ./run-tests.sh"
blah = ""
```

This ends up looking kinda gross to the user:

```
Loading .env environment variables…
  File "/Users/jtratner/pipenv/pipenv/project.py", line 400, in build_script
    script = Script.parse(self.parsed_pipfile['scripts'][name])
  File "/Users/jtratner/pipenv/pipenv/cmdparse.py", line 22, in parse
    return cls(value)
  File "/Users/jtratner/pipenv/pipenv/cmdparse.py", line 14, in __init__
    raise ValueError('invalid script')
ValueError: invalid script

During handling of the above exception, another exception occurred:

  File "/Users/jtratner/.pyenv/versions/3.6.4/bin/pipenv", line 11, in <module>
  File "/Users/jtratner/pipenv/pipenv/vendor/click/core.py", line 722, in __call__
  File "/Users/jtratner/pipenv/pipenv/vendor/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/Users/jtratner/pipenv/pipenv/vendor/click/core.py", line 1066, in invoke
  File "/Users/jtratner/pipenv/pipenv/vendor/click/core.py", line 895, in invoke
  File "/Users/jtratner/pipenv/pipenv/vendor/click/core.py", line 535, in invoke
  File "/Users/jtratner/pipenv/pipenv/cli.py", line 599, in run
    core.do_run(command=command, args=args, three=three, python=python)
  File "/Users/jtratner/pipenv/pipenv/core.py", line 2252, in do_run
    do_run_posix(command, args)
  File "/Users/jtratner/pipenv/pipenv/core.py", line 2214, in do_run_posix
  File "/Users/jtratner/pipenv/pipenv/project.py", line 404, in build_script
    raise ValueError('invalid script entry {0!r}'.format(name))
ValueError: invalid script entry 'blah'
```

Usability-wise, perhaps a better feedback would look like:

```
$ pipenv run blah
Can't run *blah* - it's an empty script.
$ echo $?
1
```

I'm not 100% clear the extra indirection is worth it here, but I understand where you're coming from :)

pipenv/utils.py
```diff
@@ -572,6 +571,10 @@ def convert_deps_from_pip(dep):
             dependency[req.name] = extras
             if specs:
                 dependency[req.name].update({'version': specs})
+        if hasattr(req, 'markers'):
+            if isinstance(dependency[req.name], six.string_types):
```
jtratner: just for my knowledge, how do you hit this edge?

tests/test_pipenv.py
```diff
@@ -1261,12 +1275,14 @@ def test_scripts(self):
...
-            assert executable == 'cmd'
-            assert argv == ['arg1', 'a', 'b']
+
+            project = Project()
+            script = project.build_script('multicommand')
+            assert script.command == 'bash'
+            assert script.args == ['-c', 'cd docs && make html']
+            script = project.build_script('appendscript', ['a', 'b'])
+            assert script.command == 'cmd'
+            assert script.args == ['arg1', 'a', 'b']
```
jtratner: nice this part looks a lot better :)

uranusjr:
This contains #1909 right? I’ll rework the script-parsing part mentioned about a little.

pipenv/utils.py
techalchemy: Lol. So above this we basically allow a pipfile entry to wind up in the state of `name = “==x.y”` if all we have is a version. This piece just dict-ifies that case if we have markers to add.

Markers is not always an attribute on a requirement however. I tried a few other approaches without success

techalchemy:
@uranusjr I'm not sure it contained _all_ of #1909 so you may want to verify that

I just reorganized and squashed everything

techalchemy:
FYI I will be merging this when it finishes unless there is an objection. I am open to suggestions about dealing with the extremely large artifacts we are now mocking.

uranusjr:
I cherry-picked changes not included previous. Fingers crossed. (If this does not work I’ll revert and work on them later.)

tsiq-oliverc:
I'd be worried that keeping the matrix in `appveyor.yml` up to date (and in sync what's specified in `.travis.yml`) is going to prove tricky, leading to inevitable gaps in CI coverage.

Is there any way to reduce the copy-pasta in the config?  (I'm unfamiliar with Appveyor, but will take a look at their docs.)

uranusjr:
The situation is indeed bad, but unfortunately I don’t know of any bullet-proof solutions :(

tsiq-oliverc:
I had a quick look at the docs, and it seems that one may be able to leverage [multiple matrix dimensions](https://www.appveyor.com/docs/build-configuration/#build-matrix) with environment variables.  That approach would at least reduce a big chunk of copy-paste.

I'm not at a computer this afternoon, but I may try experimenting with this later.

techalchemy:
Please don't change the appveyor config, it was very difficult to get it working.

tests/test_pipenv.py
```diff
@@ -282,6 +282,7 @@ def test_basic_install(self, pypi):
 
     @pytest.mark.complex
     @pytest.mark.lock
+    @pytest.mark.skip(reason='Does not work')
```
techalchemy: This test needs to pass. I will work on this

tsiq-oliverc:
Sure.  But you understand the concern, I guess?  There's a ton of copy-paste in there, which is *definitely* going to break silently at some point in future.  IIRC there's already been a recent incident of "tests are silently failing on CI", so the paranoia (as someone who would like to be able to rely on Pipenv in a user context) isn't completely unwarranted.

I realise I'm moaning a lot, so as a minor constructive suggestion - could the `TEST_SUITE`s be simplified?  For example, could they be named e.g. `a`, `b`, `c`, `d` in the CI configs, and the mappings specified elsewhere (e.g. in `run_tests.sh`)?  This would also allow for a programmatic / more robust way to deal with the `not` case.

uranusjr:
@tsiq-oliverc I’m all for it if there is a better solution. I couldn’t find one.

techalchemy:
In this PR I want tests passing consistently so that we can merge other PRs

Making the appveyor config more agreeable to you personally is not high on my priority list, that is not a personal qualm with you but a literal allocation of resources issue. Could it be configured better? Maybe. I tried many different arrangements and talked to several people about it, and I also consulted the appveyor configs of about 50 large projects and I couldn't find any reference materials that had a better organization system

My ideal case is kind of fine the way things are, which is that we catch everything in the last integration test anyway, so I'm just struggling to see the problem i guess.

tsiq-oliverc:
@techalchemy - For sure, I concur that getting the tests back to green is a priority.  I'd be proposing to submit a PR once this has merged (in any case, a PR against a PR from a GitHub fork is too complex for me to handle!)

I suppose what I'm saying is that as complexity / entropy / duplication grows, the probability of divergence / bugs approaches 1.  That's just the reality I'm familiar with.  And CI is the last line of defence, so there's nothing further that could catch it in this case.

But given there's no direct way in Appveyor to clean this up, my proposed PR would be something like factoring out the duplicated lists of test markers (in both Travis and Appveyor config) to a single source of truth, and then also eliminating the duplication due to the `not` list.

techalchemy:
I was hoping to find something like this in pytest but I couldn't find a suitable marker and I am just sick of tinkering with this crap to be perfectly honest, every small change cascades like 8 new failures. But yeah I was thinking of making a single list for each of the 4 other runs and then combining them in a separate place for the "not" use case

I'm gonna just merge this one for now sowe can start merging other things.  There's still a test in here that we have skipped, but it needs to be fixed. That one is gonna suck

Issue #1910, opened 2018-04-04
Unable to install/rm/install version controlled package


altendky:
Full commands and output: https://gist.github.com/altendky/165b025c6c2ca3d0a2ef84ae542e36bc

In reality I am having issues with my own repos but I decided to test with something more official and it showed the same issues.  I would expect the following sequence of commands to work on Windows but it does not.  It does work in Linux.

```
pipenv install -ve git+https://github.com/pypa/pipenv#egg=pipenv
pipenv --rm
pipenv install -v
```
Instead of working I get an error from the last one including the following (full log in gist above).

```
Command "git config --get-regexp remote\..*\.url" failed with error code 1 in C:\Users\FSTAB\.virtualenvs\test7-QkksyU_j\src\pipenv
```

Note that at the end the `pipenv` source directory is nearly empty and it's `.git` directory is as well.

```
PS C:\Users\FSTAB\desktop\test7> dir C:\Users\FSTAB\.virtualenvs\test7-QkksyU_j\src\pipenv
PS C:\Users\FSTAB\desktop\test7> dir C:\Users\FSTAB\.virtualenvs\test7-QkksyU_j\src\pipenv\.git


    Directory: C:\Users\FSTAB\.virtualenvs\test7-QkksyU_j\src\pipenv\.git


Mode                LastWriteTime         Length Name
----                -------------         ------ ----
d-----         4/4/2018  10:47 AM                objects
```

altendky:
I observed this with some version reporting 11.9.0 first (might have been from pypi, maybe not...) and then as I recall I installed from master and then created this report which claims 11.9.1, but since I didn't `-e`... no record.  Just to be sure I just now did a `-e` install of 8a67a21d61a2383253fe0dd5e7a8d79d51d30d2d and a short log.

https://gist.github.com/altendky/105157700c3d54b6d0b303da1a73cdad

altendky:
After `--rm` there's still `<env>\src\pipenv\.git\objects\pack\pack-a86bebb39dc5850b462e6af478af45590e09a22d.idx` and `.pack` left.  Everything else is deleted.  I could see how this would cause Git to have issues.  I'll try to dig a bit more.

altendky:
In my experience, this isn't sufficient on Windows.  Deleting directories is somehow actually pretty hard.

https://github.com/pypa/pipenv/blob/8a67a21d61a2383253fe0dd5e7a8d79d51d30d2d/pipenv/core.py#L171

[...]

PR #1927, opened 2018-04-06
Use util.rmtree() in cleanup_virtualenv()

altendky:
Issue #1910


pipenv/core.py
```diff
@@ -168,7 +169,7 @@ def cleanup_virtualenv(bare=True):
         click.echo(crayons.red('Environment creation aborted.'))
     try:
         # Delete the virtualenv.
-        shutil.rmtree(project.virtualenv_location, ignore_errors=True)
+        rmtree(project.virtualenv_location)
```
altendky: I did specifically remove `ignore_errors=True` because I would much rather get an exception than silently leave stuff behind that may cause trouble in the future.

I did the test below with the `pip-selfcheck.json` file open so it couldn't be deleted.

```
PS C:\epc\t\812\testy> ..\pipenv\venv\scripts\pipenv --rm
Removing virtualenv (C:\Users\sda\.virtualenvs\testy-7ppH9Nsf)…
[WinError 145] The directory is not empty: 'C:\\Users\\sda\\.virtualenvs\\testy-7ppH9Nsf'
```

uranusjr: The thing is, if we don’t silent the errors, the tests won’t even pass consistently on Windows, because how file handles are leaked everywhere in Python. If you really want this information, print out a warning. Don’t let it fail here.

altendky: It was already being caught and `echo()`ed but I'll push a version that I think will output a bit more inline with the rest of pipenv messaging.

techalchemy:
Looks great, if only our actual merge could get CI to pass...

techalchemy:
rebase => master plz

altendky:
@techalchemy You explicitly need a rebase? Or is the GitHub 'Update branch' button enough?

techalchemy:
oh I don't need you to specifically rebase just get it current by whatever means you like

altendky:
Yeah, just haven't gotten to my computer... But I can click a button on a website from my phone.  :]

techalchemy:
Thanks for this!

PR #1935, opened 2018-04-07
Simplify CI test orchestration

tsiq-oliverc:
Rather than splitting tests up into all these test suites, just run everything in one go.  Not only is it simpler, it's quicker too ([6:30](https://travis-ci.org/tsiq-oliverc/pipenv/builds/363621909) vs. [13:00](https://travis-ci.org/tsiq-oliverc/pipenv/builds/363608076) for the non-takes-forever tests on Travis, and ~23 vs. ~53 minutes for Appveyor).

~~As the `cli` tests were previously only run for Py3, and at least some of them fail for Py2, I've annotated them with `@py3_only`.~~

~~I've also simplified `run-tests.sh`, as it's no longer called from CI (maybe it could be removed entirely, as it seems quite scary?  Ditto for `run-tests.bat`).~~

I suspect that as the markers (other than `install`) are no longer used, a lot of them could be removed now.


techalchemy:
as a FYI I am seeing about generally speeding up some tests over in #1936

uranusjr:
As for markers, how about splitting them into different files? `test_pipenv.py` is becoming too large and unwieldy for me. I can work on that if this is acceptable.

techalchemy:
prolly makes sense.  I am so over tests

tsiq-oliverc:
@techalchemy - Yeah, I'm confused as to why the CLI tests are only being running for Py3, but that's currently the case on master - the "integration" stage in Travis is not being run as part of the Py2/3 matrix.

So in that sense, I'm not sure this PR makes things any worse?

tsiq-oliverc:
I've just run the `cli` tests locally in a Python 2.7.14 Docker container, and these are the results:

```
tests/test_pipenv.py::TestPipenv::test_pipenv_where FAILED                                                     [  6%]
tests/test_pipenv.py::TestPipenv::test_pipenv_venv FAILED                                                      [ 13%]
tests/test_pipenv.py::TestPipenv::test_pipenv_py FAILED                                                        [ 20%]
tests/test_pipenv.py::TestPipenv::test_pipenv_rm FAILED                                                        [ 26%]
tests/test_pipenv.py::TestPipenv::test_pipenv_graph FAILED                                                     [ 33%]
tests/test_pipenv.py::TestPipenv::test_pipenv_graph_reverse FAILED                                             [ 40%]
tests/test_pipenv.py::TestPipenv::test_pipenv_check FAILED                                                     [ 46%]
tests/test_pipenv.py::TestPipenv::test_venv_envs FAILED                                                        [ 53%]
tests/test_pipenv.py::TestPipenv::test_bare_output FAILED                                                      [ 60%]
tests/test_pipenv.py::TestPipenv::test_help FAILED                                                             [ 66%]
tests/test_pipenv.py::TestPipenv::test_man PASSED                                                              [ 73%]
tests/test_pipenv.py::TestPipenv::test_install_parse_error PASSED                                              [ 80%]
tests/test_pipenv.py::TestPipenv::test_install_without_dev_section FAILED                                      [ 86%]
tests/test_pipenv.py::TestPipenv::test_deploy_works FAILED                                                     [ 93%]
tests/test_pipenv.py::TestPipenv::test_clean_on_empty_venv FAILED                                              [100%]
```

So it looks like a few of them *do* pass under Py2.  We could remove the `@py3_only` from those ones, and then we'd at least have explicit documentation of which ones are broken in Py2?

Could then imagine a separate effort / PR to fix those tests (because I agree it's bizarre that basic CLI functionality is failing tests!).

tsiq-oliverc:
@uranusjr - FWIW, I strongly agree that `test_pipenv.py` is way too big and complex.  Let me know if I can help with that.

tsiq-oliverc:
Further clues regarding CLI tests:

- Running with `-s`, I learn that they fail due to a complaint that `ordereddict` is not installed.
- And indeed, it's not listed in `Pipfile.lock`.
- Running `pipenv install` doesn't change that.
- It's listed as `ordereddict; python_version<"3.0"` in `setup.py`.
- If I nuke `Pipfile.lock` and then re-run `pipenv install`, it *does* get added to the lockfile, but without a marker.

Sounds like there could be an issue with processing markers?

techalchemy:
It shouldn’t be an issue with I took it out of the Pipfile because it gets installed in the package installation. I can’t tell why this is working in other branches but not in yours. 

I run the test suite against both versions of python locally regularly. They both regularly pass. See also: appveyor, which runs this suite against python 2 and 3. This isn’t really a valid argument. Testing needs to work. These tests need to run. This is a testing issue. For your case I’m assuming you didn’t run `pipenv run pip install -e .` because that would pick it up due to pathlib2? Or one of the other dependencies. 

As far as the time I think the number we care about is how many actual minutes it takes and how quickly we can isolate failures. There’s a lot of overhead to spinning up parallel environments so you’re cutting some time that way for sure. But using those numbers we are moving from 23 mins to 16 under these tests (but not doing integration tests on python 2). 

Also, I should caution you against all the changes to the run tests files— these are necessary for when the tests also build against Kenneth’s CI served and he is highly likely to change them all back to what they were if we make a bunch of changes for no reason. 

Markers are used for organization. Removing them for no reason seems kind of silly. We could however try to organize them. 

My point is: you seem to be in a hurry to take an axe to things. Don’t. Let’s focus on functionality. If we can stop calling run_tests from Travis, that’s fine, but revert the changes to it and leave the file alone. Same for the appveyor script. These improvements are significant and the work is hugely appreciated, but let’s save the reorganization work for another pr

tsiq-oliverc:
@techalchemy - TL;DR - It's a good thing you pushed back, you forced me to find the issue here.  The CLI tests now run and pass for Py2.

I hadn’t noticed [this line][1] is subtly different to [this line][2], so convinced myself that the CLI tests weren't being run for Py2 on Travis.  They’re actually buried inside this final test suite, not in the "integration" section (so the Py3 equivalents were actually being run twice!)

**However:** with the parallelism set to `auto`, there were frequent test failures.  I’ve wound this down to 4; things are still fast, but pass consistently.  This strongly implies there’s hidden coupling between the test cases, but this PR doesn’t change that.

[1]: https://github.com/pypa/pipenv/blob/master/appveyor.yml#L40
[2]: https://github.com/pypa/pipenv/blob/master/.travis.yml#L12


> these are necessary for when the tests also build against Kenneth’s CI served[r?]

I guess I'm unfamiliar with the history / current status of this project, so what does this mean in practice?


> If we can stop calling run_tests from Travis, that’s fine, but revert the changes to it and leave the file alone

I've reverted `run-tests.sh`.  (Though it seems redundant to keep unused logic around.)


> We could however try to organize them.

Yeah, I think that will come for free with @uranusjr's suggested file-split.


> Same for the appveyor script.

I haven't reverted this, given things are now working again.  My belief is the simpler config makes it much easier to see we're running the same tests on Appveyor / Travis / Py2 / Py3.  LMK what you think.

uranusjr:
@tsiq-oliverc It should be quite simple to split up the tests with some Pytest help (mainly turning `PipenvInstance` into a fixture). I want to wait until #1936 is done, however, because they totally conflict each other and I’ll need to redo everything again.

tsiq-oliverc:
@uranusjr - Yup, makes sense.

techalchemy:
Yeah sorry for the typo, I was on mobile this morning. The `run_tests.sh` script runs against a server in Kenneth's house (and that's why we were piping output to the tap stream for Jenkins). What it means in practice is there is a server that this runs against and you shouldn't change it if we aren't calling it.

techalchemy:
Your comment got me thinking, so in #1936 I split out python 2  and python 3 tests in different build stages.  This has some big speedups and hasn't failed on me yet (13 mins actual clock time) -- but comes at the cost of a complicated `travis.yml` file.  Before we go forward I want to discuss the tradeoffs

- I have some other test speedups in that branch which may affect the speed
- Looking over the total runtime of this PR it is clearly the better option.  I'd say you can separate out the python 2 and python 3 integration step and I'll include that in a comment

.travis.yml
```diff
@@ -20,20 +17,20 @@ install:
...
 
 jobs:
   include:
-    - stage: integration
-      env: TEST_SUITE='cli'
     - stage: takes-forever
       env:
         - TEST_SUITE='install'
         - PYTEST_ADDOPTS='--cache-clear'
         - RUN_INTEGRATION_TESTS=1
```
techalchemy: Overall you can just add here so the section looks like this:

```yml
jobs:
  include:
    - stage: takes-forever
      python: "3.6"
      env:
        - TEST_SUITE='install'
        - CACHE_NAME='INSTALL-1'
        - PYTEST_ADDOPTS='--cache-clear'
        - RUN_INTEGRATION_TESTS=1
    - stage: takes-forever
      python: "2.7"
      env:
        - CACHE_NAME='INSTALL-2'

Comment (techalchemy) [reply .travis.yml]: I couldn't get the above to work for some reason but I have no idea why, it seems like it should work fine to me... so maybe scratch that?

**Comment by techalchemy** [comment]
Okay sorry for the comment spam -- I like what you've done here as it definitely simplifies things.  This is my final input on this PR (and I'd make the changes myself but this is 100% happening because of your changes here so I feel you should get credit in the commit log too) -- 

- Axe the build stages completely.
- Axe the tap stream output completely.

We gain the speedups of parallelizing but lose the (hidden) cost incurred by overparallelizing because I believe travis is just killing jobs that take too long (plus some of these marks only run 2 tests and the cost of spinning up a machine for them is incredibly high).

Again **thank you**.  This is way simpler and much faster.

Reference: https://travis-ci.org/techalchemy/pipenv/jobs/363950542 (9 minutes of total real-world runtime)

travis,yml:

```yml
sudo: false
dist: trusty
cache: pip
language: python
python:
  - "3.6"
  - "2.7"
env:
  global:
    - PYPI_VENDOR_DIR='./tests/pypi/'
    - GIT_ASK_YESNO='false'
  matrix:
    - TEST_SUITE='not install'
      CACHE_NAME=$TRAVIS_PYTHON_VERSION-TEST
    - TEST_SUITE='install'
      RUN_INTEGRATION_TESTS='1'
      PYTEST_ADDOPTS='--cache-clear'
      CACHE_NAME=$TRAVIS_PYTHON_VERSION-INSTALL

# command to install dependencies
install:
  - "pip install --upgrade pip"
  - "pip install -e . --upgrade --upgrade-strategy=only-if-needed"
  - "pipenv install --deploy --system --dev"

# command to run the dependencies
script:
  - 'if [[ -n "$RUN_INTEGRATION_TESTS" ]]; then rm -fr ~/.cache/pip; fi'
  - 'export PYPI_VENDOR_DIR="$(pwd)/tests/pypi/"'
  - 'pip install -e "$(pwd)" --upgrade'
  - 'pipenv install --dev'
  - 'pipenv run time pytest -v -n 4 --ignore="pipenv/vendor" --ignore="pipenv/patched" -m "$TEST_SUITE" tests'
```

tsiq-oliverc:
@techalchemy - Yeah, I was thinking about the cost/benefit of making the `install` tests unconditional, and it looks like it's a good trade-off!  (In any case, the total runtime is dominated by Appveyor.)

I've pushed a change along the lines of your suggestion, which seems to run fine.  Note that I've removed the cache entirely.  The Travis runtime is dominated by the "install" tests, so I *think* that adding the cache back into the "non install" tests wouldn't help.  But I don't feel strongly about this.

techalchemy:
(I suspect the duplicate install lines aren’t actually necessary but I wasnt too sure)

techalchemy:
This is awesome, thanks a bunch. Will review and merge when I get into the office

techalchemy:
This looks good, merging! :tada:

PR #1936, opened 2018-04-08
Test speedups

techalchemy:
Speeding up some tests


techalchemy:
This makes roughly no difference on travis but saves about 5 minutes of runtime on appveyor as is

PR #1946, opened 2018-04-10
Windows VEnv Case normalization

### LINKED ISSUE #1925: Detecting virtualenv location in windows is unreliable when using powershell.
PS C:\Users\N N> cd work
PS C:\Users\N N\work> pipenv --venv
No virtualenv has been created for this project yet!
PS C:\Users\N N\work> cd ..
PS C:\Users\N N> cd .\Work\
PS C:\Users\N N\Work> pipenv --venv
C:\Users\N N\.virtualenvs\Work-DPz4PjCb
PS C:\Users\N N\Work>

Seems to work fine when using cmd.exe as the shell. Not a big thing, but slightly annoying.

techalchemy:
Normalize casing on windows virtualenv names

- Fixes #1925

### DIFF
--- a/pipenv/project.py
+++ b/pipenv/project.py
@@ -11,6 +11,7 @@ import hashlib
 
 import contoml
 import delegator
+from pipenv.vendor.first import first
 import pipfile
 import pipfile.api
 import toml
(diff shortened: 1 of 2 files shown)


techalchemy:
When I was mulling this over the only thing I could come up with was to start from the base virtualenv directory and lowercase both sides of the equation to match before getting the actual path (but we might just lowercase that somewhere, I wasn't sure)

uranusjr:
It probably be best to lowercase the whole path if we can afford to throw away all old virtualenvs.

techalchemy:
I don't think we can afford that though, so we need an alternative

uranusjr:
Maybe it’s better to use `pathlib.Path.resolve()` instead.

techalchemy:
Fine with me but we need to check for existence first on windows, pathlib does not like it if you try to hand it a path that is invalid or doesn’t exist as I recall

uranusjr:
I plan to add it to `pipfile.Pipfile.find()`. If it ever finds us a non-existent file we’re toasted anyway. lol

techalchemy:
+1 from me then

techalchemy:
Feel free to make the changes if you like

uranusjr:
I added the resolve call, and changed the case-insensitive lookup a little so it loops through all possible envs, instead of only choosing the first. This should help with some edge cases with some sacrafice of speed, but with the resolve call it wouldn’t be reached in most cases anyway.

techalchemy:
Looks good to me

PR #1958, opened 2018-04-12
Mention scripts section in doc

frostming:
Just add the missing instructions for `scripts` to doc

Please review


uranusjr:
Reminder: remove the script mention in the wiki after merging this (because it wouldn’t be hidden anymore).

frostming:
Any advice? We do have a test case in the tests, are we going to refer to the test file or specific line in the documentation?

techalchemy:
Nope if we dont' have a good example lets just go with what we have here (I was just wondering if we have more reference material)

PR #1960, opened 2018-04-12
Split tests

uranusjr:
As promised, split tests into modules. I add some Pytest hooks to perform cross-module setup of environs, and implement `needs_internet` as a marker instead so it can be used in multiple files.

Similarly, `PipenvInstance` is wrapped in a fixture so we can simply name it in the arguments.

🤞


uranusjr:
Asteroid seems stuck. Any objections on this? I want to merge if possible because this would cause conflicts to all new test contributions.

techalchemy:
gotta give me a chance to review :p

uranusjr:
Sure, go head 😄 I’ll leave this to you then.

techalchemy:
seems fine at a glance, probably want to organize it in a fundamentally different way at some point but that's another project imo... if you can confirm you didn't remove any tests i'd say just hit the button when you're ready, I'll review more carefully later

uranusjr:
No tests removed I promise. I actually un-skipped one because it’s working.

PR #1962, opened 2018-04-12
Requirements Refactor (ready to merge, after conflict resolution)

techalchemy:
Full refactor of requirements parsing

- See commits


tests/unit/test_utils.py
```diff
@@ -108,7 +109,7 @@ def test_convert_from_pip(expected, requirement):
     ),
 ])
 def test_convert_from_pip_vcs_with_extra(expected, requirement):
-    assert pipenv.utils.convert_deps_from_pip(requirement) == expected
+    assert pipenv.requirements.PipenvRequirement.from_line(requirement).as_pipfile() == expected
```
uranusjr: You can merge this feature into `DEP_PIP_PAIRS` to make sure the new parser can catch the extra part.

uranusjr: nvm I’ll just push the change here.

altendky:
@techalchemy, thanks for the help on IRC.  This seems to be working to solve my issue (local `-e .` install causing remote vcs link in Pipfile.lock causing a request that failed due to being a private repo and pip not being able to get the password from me...  phew).  Now I can get started writing my tests.  :]  maybe...

techalchemy:
@ncoghlan I've been sitting on this refactor for awhile, to summarize it essentially provides a compatibility layer between pipenv/pipfile style requirements and pip-style requirements.  I wrote it in frustration when I had to fix a bug with the current codebase which breaks anytime someone touches it.

I've been contemplating breaking this into a library of its own, which I did with the last thing I had an idea like this for (https://github.com/techalchemy/pythonfinder) to allow for potential expansion independent of pipenv -- the API looks something like this:

```python
>>> req = PipenvRequirement.from_line(requirement)
>>> req.as_pipfile()
{'req': {'key': 'value'}}

>>> req = PipenvRequirement.from_pipfile(name, index, {'key': 'value', 'key2': 'value2'})
>>> req.as_line()
-e asdf.whatever#egg=name
```

The object itself offers access to hashes, index, name, vcs, etc just depending on what type of requirement you pass.  Check out `requirements.py` if you want, review if you have time, but general thoughts always appreciated

Will be around on synchronous communication for another hour or two => slack/irc/fb if you have anything general

ncoghlan:
@techalchemy Breaking out low level building blocks to make them easier to compose independently seems like a good idea to me. (As one example: it would be cool if `pip-compile` could accept `Pipfile` as input directly, and this would be a useful stepping stone towards enabling that)

kennethreitz:
Please resolve merge conflicts.

Issue #1961, opened 2018-04-12
Possible legal problems with vendor/patched libraries


hroncok:
# tl;dr

Pipenv probably violates couple of free software licenses by not shipping their text. It is also probably "infected" by GPL.

# Details

Hello, we (@mcyprian, me and the Fedora Python SIG @fedora-python) are trying to finally package pipenv for Fedora, so user can just do `sudo dnf install pipenv`. The Fedora package review request is in [Red Hat Bugzilla 1564500](https://bugzilla.redhat.com/show_bug.cgi?id=1564500).

While we are trying to unbundle (unvendor?) most of the 3rd party libraries shipped with pipenv, we are in a bit of hurry so we decided to leave the libraries that are

 * not yet packaged for Fedora,
 * or patched

bundled for now. The first category is a TODO for future, the second will probably remain bundled forever. 

As part of the review of the Fedora package, the reviewer is obligated to check whether the package is licensed with [approved free software/content license](https://fedoraproject.org/wiki/Licensing:Main?rd=Licensing#Good_Licenses) and whether the licensing information for the package is correct.

## Missing licenses

This is where I found out that all the vendored 3rd party libraries are shipped without their LICENSE/COPYING/etc. files and the [NOTICES](https://github.com/pypa/pipenv/blob/master/NOTICES) file is shipped instead.

> The contents of the vendor and patched directories are subject to different licenses
> than the rest of this project. Their respective licenses can be looked
> up at pypi.python.org.

This is unacceptable for Fedora\* and IMHO should not be acceptable for @pypa either. Most of the libraries are licensed with licenses that **require** the license text to be shipped. See MIT:

> The above copyright notice and this permission notice shall be included in all
> copies or substantial portions of the Software.

Or BSD:

> Redistributions of source code must retain the above copyright notice, this
>  list of conditions and the following disclaimer.

This is similar for most of the permissive licenses. *You should not strip the license file, in fact you need to ship it with the code unconditionally.* You may put the license text and copyrights inside another file, such as NOTICES, but a link to pypi is IMHO not enough.

\* This is currently only my opinion. It has not yet been reviewed by the Fedora legal team.

## Copyleft

[...]

PR #1970, opened 2018-04-12
Add licenses for 3rd party libraries

hroncok:
See https://github.com/pypa/pipenv/issues/1961

This is a work in progress.


techalchemy:
Thank you for all your help with this, it is greatly appreciated

hroncok:
I belive this can be merged as is and the rest can be figured out in https://github.com/pypa/pipenv/issues/1961

PR #1976, opened 2018-04-12
Check whether dependency exists in old lockfile during keep-updated lock

greysteil:
Currently, if a new dependency has been added to the `Pipfile` and `pipenv lock --keep-outdated` is run a `KeyError: '<dependency-nam>'` error is raised. Conditioning on whether the dependency is already in the lockfile fixes this.


techalchemy:
seems so obvious when you make a pr out of it :tada:

Issue #1977, opened 2018-04-12
pipenv sync unexpected behavior when Pipfile missing


pikeas:
Pipfile:
```
[[source]]
url = "https://pypi.python.org/simple"
verify_ssl = true
name = "pypi"

[packages]
aioredis = "*"

[dev-packages]
ipython = "*"

[requires]
python_version = "3.6"
```

Generate Pipefile.lock for the above, delete Pipfile, and then:
```
$ pipenv sync
...
$ cat Pipfile.lock
{
    "_meta": {
        "hash": {
            "sha256": "4e55147db217bb4120f6e68cb8cad7bc37011457441ce0eb9d97308315625834"
        },
        "pipfile-spec": 6,
        "requires": {
            "python_version": "3.6"
        },
        "sources": [
            {
                "name": "pypi",
                "url": "https://pypi.python.org/simple",
                "verify_ssl": true
            }
        ]
    },
    "default": {},
    "develop": {}
}
```

##### Expected result

pipenv installs the packages in Pipfile.lock.

##### Actual result

pipenv installs no packages and replaces Pipfile.lock with a new file containing no packages.

frostming:
I can send a PR to fix this quickly but I would like to go further into the missing features of `pipenv sync`, such as dry_run, dont_upgrade. It seems far from complete. Thanks

techalchemy:
There are some bugs in here that should get sorted out, so that would be appreciated?

butla:
@techalchemy I think this should be reopened. When I put Pipfile.lock without the Pipfile in a Docker image and run `pipenv sync` in the directory with Pipfile.lock I'm getting: `Error: Pipfile.lock is missing! You need to run $ pipenv lock first.`. When I copy the Pipfile as well the problem disappears.

marlonjan:
Just had the same issue as @butla. Took me a while to understand that the error message was incorrect. Pipfile was missing, Pipfile.lock was present, and I got 
```Error: Pipfile.lock is missing! You need to run $ pipenv lock first.```

techalchemy:
Sorry about that, error message seems to be pretty clearly wrong on this one!

Issue #1974, opened 2018-04-12
pipenv install --skip-lock  produces error and fails to install dependencies "You likely have a mismatch in your sub-dependencies"


cpoptic:
Be sure to check the existing issues (both open and closed!).

Describe the issue briefly here.
Running pipenv install   produces error:

> Pipfile.lock not found, creating…
Locking [dev-packages] dependencies…
Locking [packages] dependencies…

Warning: Your dependencies could not be resolved. You likely have a mismatch in your sub-dependencies.
  You can use $ pipenv install --skip-lock to bypass this mechanism, then run $ pipenv graph to inspect the situation.
Could not find a version that matches basemap
Tried: (no version found at all)

I then follow instructions and run >>

pipenv install --skip-lock
Installing dependencies from Pipfile…
An error occurred while installing time! Will try again.
An error occurred while installing calendar! Will try again.
An error occurred while installing codecs! Will try again.
An error occurred while installing json! Will try again.
An error occurred while installing sys! Will try again.
An error occurred while installing gzip! Will try again.
An error occurred while installing string! Will try again.
An error occurred while installing glob! Will try again.
An error occurred while installing re! Will try again.
An error occurred while installing matplotlib.pyplot! Will try again.
  ================================ 35/35 - 00:00:05
An error occurred while installing basemap! Will try again.
Installing initially–failed dependencies…
Collecting time --------------------- 0/11 - 00:00:00

  Could not find a version that satisfies the requirement time (from -r C:\Users\Chris\AppData\Local\Temp\pipenv-gnbsva_j-requirements\pipenv-2r7msx_f-requirement.txt (line 1)) (from versions: )
No matching distribution found for time (from -r C:\Users\Chris\AppData\Local\Temp\pipenv-gnbsva_j-requirements\pipenv-2r7msx_f-requirement.txt (line 1))

     ================================ 0/11 - 00:00:01

It seems to get hung up looking for "time" dependency, which is strange.

When I check the C:\Users\Chris\AppData\Local\Temp directory, there is no folder named "pipenv-gnbsva_j-requirements", so I'm guessing it's looking for a requirements.txt file in a folder that no longer exists.  How to fix this?

Please run `$ python -m pipenv.help`, and paste the results here.

Pipenv version: `'11.10.0'`

Pipenv location: `'C:\\ProgramData\\Anaconda3\\lib\\site-packages\\pipenv'`

Python location: `'C:\\ProgramData\\Anaconda3\\python.exe'`

[...]

Issue #1852, opened 2018-03-26
Feature request: `pipenv install --index myindex mypackage`


patrick-jones:
I would like to request a command line option for specifying the source during `pip install`. For example, given this Pipfile (from [the docs](https://docs.pipenv.org/advanced/#specifying-package-indexes)):

```
[[source]]
url = "https://pypi.python.org/simple"
verify_ssl = true
name = "pypi"

[[source]]
url = "http://pypi.home.kennethreitz.org/simple"
verify_ssl = false
name = "home"
```

I would like to be able to invoke something like:

`pipenv install --index home mypackage`

Which would install from my "home" index and add this into my Pipfile:
```
[packages]
mypackage = {version="*", index="home"}
```

techalchemy:
Makes sense, although if it's a package that isn't on pypi and your index is in your pipfile already it may work just fine to leave the index off

uranusjr:
Is this technically possible though? While there is `--extra-index-url` in pip, I don’t think you can specify which index to use for each dependency, only the installation as a whole.

jakul:
The example at https://docs.pipenv.org/advanced/#specifying-package-indexes specifies a different index for each requirement, so Pipenv must already support it.

techalchemy:
Yeah good point, doesn’t this already work?

techalchemy:
Try `pip install -e git+https://github.com/pypa/pipenv.git@1973-extra-index-urls#egg=pipenv` and let me know if it works

PR #1980, opened 2018-04-13
Add `--extra-index-url` from all extra indexes

### LINKED ISSUE #1973: pipenv install --skip-lock broken with multiple sources
Considering the following Pipfile

```
[[source]]
verify_ssl = true
name = "pypi"

[[source]]
url = "https://pypi.example.com/simple"
verify_ssl = true
name = "internal"

[packages]
myprivatepackage = {index = "internal", version = "==1.0.0", editable = true}
uWSGI = "==2.0.17"
unittest-xml-reporting = "*"
coverage = "*"

[dev-packages]

[requires]
python_version = "2.7"
```
---

```bash
pipenv install --verbose --skip-lock                                                                                                                                                         1 ↵
Installing dependencies from Pipfile…
Installing 'uwsgi==2.0.17 --extra-index-url https://private.example.com/simple'

WARNING: Invalid requirement, parse error at "'--extra-'"
ABORTING INSTALL... You will have to reinstall any packages that failed to install.
You may have to manually run pipenv lock when you are finished.
```

```
pipenv, version 11.9.0
```

I believe this might have been introduced in https://github.com/pypa/pipenv/pull/1706

cc @techalchemy

techalchemy:
Handle indexes, extra indexes, uncached sources

- Handle extra-index-urls when resolving
- Handle extra-index-url when using `--skip-lock`
- Parse index arguments when installing individual packages
- Translate index aliases to urls
- Always include extra indexes when installing a packages
- `get_source()` falls back to `parsed_pipfile['source']` for sources when
    not present in the lockfile (#1994)
- Include index and extra-index-url arguments in `pipenv lock -r` output
- Fixes #1973, #1974, #1852, #1977, #1994

### DIFF
--- a/pipenv/project.py
+++ b/pipenv/project.py
@@ -654,7 +654,7 @@ class Project(object):
             if name:
                 source = [s for s in sources if s.get('name') == name]
             elif url:
-                source = [s for s in sources if s.get('url') in url]
+                source = [s for s in sources if url.startswith(s.get('url'))]
             if source:
                 return first(source)
 


techalchemy:
@uranusjr I want to favor this over #1950 but the `get_source` method in `pipenv.Project` only checks the cached lockfile, which doesn't work if you add a new source to the pipfile that hasn't been locked yet (i.e. you use `--skip-lock`) -- thoughts?

techalchemy:
@uranusjr @jtratner @ncoghlan @frostming would appreciate if at least one of you can look this over before it gets merged so I can go ahead and cut a bugfix release

pipenv/project.py
```diff
@@ -628,15 +633,35 @@ def sources(self):
...
+        def find_source(sources, name=None, url=None):
             if name:
-                if source.get('name') == name:
-                    return source
-
+                source = [s for s in sources if s.get('name') == name]
             elif url:
-                if source.get('url') in url:
-                    return source
+                source = [s for s in sources if s.get('url') in url]
```
uranusjr: I know this was the case before, but is it a good idea to use `in` here, or should we use `startswith` instead?

pipenv/project.py
```diff
@@ -628,15 +633,35 @@ def sources(self):
...
+            if source:
+                return source[0]
```
uranusjr: Isn’t this essentially what `first` does? I don’t dislike this (I never liked the `first` library, in fact), but maybe we should be consistent here.

tests/integration/test_install_uri.py
```diff
@@ -92,3 +92,32 @@ def test_install_editable_git_tag(PipenvInstance, pip_src_dir, pypi):
...
+
+[packages]
+pytz = "*"
+six = {{version = "*", index = "pypi"}}
+
+[dev-packages]
+            """.format(os.environ.get('PIPENV_TEST_INDEX')).strip()
+            f.write(contents)
+        c = p.pipenv('install pytz --index pypi')
+        assert c.return_code == 0
```
uranusjr: A few problems:

* Comment seems wrong. A lot of tests below have the wrong comment, it seems? None of them are using a Git repo afaict.
* This is not future-proof. If we add pytz to the mock PyPI (we might) this cannot detect an incorrect `--index` implementation.
* Probably don’t want to use `dict.get`. If `PIPENV_TEST_INDEX` is not set this needs to fail loudly.

uranusjr:
Forgot to mention in the review—`split_index` and `split_extra_index` seem a bit too similar to me. Is it possible to refactor the logic out?

pipenv/project.py
techalchemy: probably, it seemed odd when I was reimplementing it but I kind of copied what was there... I can't really imagine why we'd use `in` vs `startswith`

pipenv/project.py
techalchemy: if it's imported I'll swap it over, in the spirit of speeding up startup time (which primarily was coming from indirect loading of the project module) I was trying to avoid extra imports

tests/integration/test_install_uri.py
techalchemy: oh yeah I think I've been copy pasting a bunch, I can do some comment cleanup

- I agree about the package, we can use something that's only available on the test index like the one I am using in the other test

pipenv/project.py
uranusjr: `git blame` shows @kennethreitz and it dates all the way back to the original implementation (v2.8.7!). Maybe no particular reason, just to hammer out something working first.

pipenv/core.py
```diff
@@ -1367,6 +1367,7 @@ def pip_install(
     pre=False,
     selective_upgrade=False,
     requirements_dir=None,
+    extra_indexes=None,
```
jtratner: wdyt about, rather than making this type polymorphic, enforcing that `extra_indexes` must always be a list and  when calling this function on line 841, ensuring that it's always a list (or having split_extra_index always return a list of elements)

pipenv/core.py
```diff
@@ -1415,51 +1416,63 @@ def pip_install(
...
-        if r:
-            with open(r) as f:
-                if '--hash' not in f.read():
-                    ignore_hashes = True
+        if extra_indexes:
+            if isinstance(extra_indexes, six.string_types):
+                extra_indexes = [extra_indexes,]
+            for idx in extra_indexes:
+                try:
+                    extra_src = project.find_source(idx).get('url')
```
jtratner: what exactly is going on here? how would you hit this case?

tests/integration/test_lock.py
```diff
@@ -150,3 +150,66 @@ def test_complex_lock_deep_extras(PipenvInstance, pypi):
...
+@pytest.mark.skip_lock
+@pytest.mark.index
+@pytest.mark.needs_internet
+@pytest.mark.install  # private indexes need to be uncached for resolution
+def test_private_index_skip_lock(PipenvInstance):
+    # This uses the real PyPI since we need Internet to access the Git
+    # dependency anyway.
+    with PipenvInstance() as p:
+        with open(p.pipfile_path, 'w') as f:
+            contents = """
```
jtratner: since these two are the same, how do you feel about using a constant and/or a fixture:

```
@pytest.mark.install # private indexes need to be uncached for resolution
@pytest.fixture
def PrivateIndexPipenvInstance(PipenvInstance):
    # This uses the real PyPI since we need Internet to access the Git
    # dependency anyway.
    with PipenvInstance() as p:
             f.write(contents)
        yield p
```

that way you don't have to repeat the `@pytest.mark.install` as much and you can slim down the test to make it easier to read.

jtratner:
```
def test_private_index_skip_lock(PrivateIndexPipenvInstance):
    p = PrivateIndexPipenvInstance
    c = p.pipenv('install --skip-lock')
    assert p.pipenv('install --skip-lock')
```

or if you want to keep the convention that you call PipenvInstances in a with statement, can do:

```
def PrivatePipenvIndexInstance(PipenvInstance):
     @contextlib.contextmanager
     def inner():
         with PipenvInstance() as p:
             # write stuff
             yield p
     return inner
```

tests/integration/test_project.py
```diff
@@ -0,0 +1,106 @@
...
+        sources = [
+            ['pypi', 'https://pypi.python.org/simple'],
+            ['testindex', os.environ.get('PIPENV_TEST_INDEX')]
+        ]
+        for src in sources:
+            name, url = src
+            source = [s for s in project.pipfile_sources if s.get('name') == name]
+            assert source[0]['name'] == name
+            assert source[0]['url'] == url
+            source = project.get_source(name=name)
```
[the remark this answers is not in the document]
techalchemy: so the `pipfile_sources` property hits the pipfile directly, exclusively.  The `get_source` method (which I didn't write and wasn't sure about) appears to consult the lockfile, and when I happened upon it it _only_ consulted the lockfile.  `find_source` is a helper method to deal with #1852 so it simply saves reimplementing the same calls/checks a bunch in order to figure out if we have a named source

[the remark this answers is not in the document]
techalchemy: However yes this can probably be done the way you described

pipenv/core.py
techalchemy: I think that makes sense but I want to stay out of the weeds on that logic until I get this bugfix released, so lets plan for 11.11.0

pipenv/core.py
techalchemy: If you have multiple indexes in your pipfile but you don't specify anything special when you run `pipenv install`, this makes sure we use all of the indexes you specify as extra index urls

techalchemy:
Ok mostly changed stuff, didn't rewrite all of the testing w/ the fixtures and refactor points per @jtratner but did refactor them a bit to be more concise & got them passing again

PR #1982, opened 2018-04-13
Purge strict-rfc3339 and timestamp from prettytoml

uranusjr:
Address https://github.com/pypa/pipenv/issues/1961#issuecomment-381006805

It turns out we don’t even need a library. `isoformat` with a minor tweak produces a perfectly RFC 3339-compatible output.


uranusjr:
`strict_rfc3339` is also used in a prettytoml test, but we’re not running it anyway.

techalchemy:
thanks for sorting this out also

uranusjr:
I should have written some tests for this, but feel free to merge first since I can’t do it until tomorrow.

techalchemy:
would love a test for this before merging I agree

uranusjr:
Tests added. I added a dev dependency to pytz so we can test the offset part more easily.

PR #1983, opened 2018-04-13
Unify the references to Pipenv in doc

frostming:
This minor patch unifies the references to Pipenv in the documentation.
When we mention Pipenv, we refer to the project rather than the CLI command. Further inputs to the doc should also follow the convention.

See also: requests's doc


docs/advanced.rst
```diff
@@ -102,8 +102,8 @@ Also useful for deployment is the ``--deploy`` flag::
 This will fail a build if the ``Pipfile.lock`` is out–of–date, instead of generating a new one.
 
 
-☤ ``pipenv`` and ``conda``
---------------------------
+☤ Pipenv and Conda
+------------------
 
 To use Pipenv with a Conda–provided Python, you simply provide the path to the Python binary::
```
uranusjr: This might need some more modifications. [Conda](https://conda.io) is a package manager + environment manager, and does not provide Python distributions. Anaconda does.

frostming: Conda means both anaconda and miniconda?

uranusjr: I don’t have an official reference from Continuum, but [Jake VanderPlas makes a clear distinction](http://jakevdp.github.io/blog/2016/08/25/conda-myths-and-misconceptions/#Myth-#1:-Conda-is-a-distribution,-not-a-package-manager).

frostming:
Good enough, how about this time

techalchemy:
Hey this looks good to me, thanks for cleaning this up!

PR #1986, opened 2018-04-13
Expand env vars in Pipfile

mbachry:
Documentation says it's possible to use environment variables in source urls (https://docs.pipenv.org/advanced/#injecting-credentials-into-pipfiles-via-environment-variables), but it doesn't seem to be the case at the moment. Here's a fix that works for me.

### DIFF
--- a/pipenv/help.py
+++ b/pipenv/help.py
@@ -45,13 +45,13 @@ def main():
     for key in os.environ:
         print('  - `{0}`'.format(key))
     print('')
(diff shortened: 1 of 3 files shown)


jtratner:
Yeah I think it’s still an open issue :-/

techalchemy:
Closing this in favor of integrated version that takes advantage of new functionality, thanks for the PR though and sorry for the delayed feedback!

PR #1998, opened 2018-04-16
I believe this tox config is more correct this way

sfermigier:


techalchemy:
Hi. Thanks for the contribution. I know this might make sense to you but generally it’s helpful to provide context or open an issue discussing what you think the problem is. As it stands I only know that you have deleted a line of documentation and stated that you think it is more correct. I’m not really a tox expert so I need some more context

sfermigier:
Sorry @techalchemy 

It's a bit hard to explain, except by saying that this line is both not needed, and if it was needed, it's in the wrong place and would most probably result in a bug.

Let's try anyway: `{[testenv]deps}` is used to reference the `deps` variable from the `testenv` section above to sort of "inherit" from it.

A correct idiom would be:

```
     [testenv]
     deps = pipenv
 
     [testenv:flake8-py3]
     deps = 
         {[testenv]deps}
         additional dependencies (of which there aren't)
     commands=
         pipenv install --dev
         pipenv run flake8 --version
         pipenv run flake8 setup.py docs project test
```

Except it's not needed since there is no additional dependency to specify in the `flake8-py3` environment.

techalchemy:
@sfermigier thanks for the explanation and for your patience :) Seems likely this was an artifact of some kind so it makes no sense to have this line here.  Thanks for putting this together!

PR #2016, opened 2018-04-17
Fix `pipenv install` to successfully use pip9

techalchemy:


uranusjr:
Wait, I don’t think this would work if `system` is false? `which('python', allow_global=system)` would return the virtualenv Python, which would not be able to locate `pipenv.vendor.pip9`.

techalchemy:
@uranusjr doesn’t matter, it runs with a temporary environment that adds the vendor directory to the python path.

techalchemy:
Tracking pypa/pip#5251 in case pip 10.0.1 is something we can delay this release for and revert this hackery

PR #2031, opened 2018-04-20
Do not destroy lock early

### LINKED ISSUE #1888: Delete Pipfile.lock as late as possible
When updating Pipfile.lock, the old version is deleted before anything happens.  If the update process fails (maybe because one Ctrl+C's it because of a mistake, or it's [taking too long][3], or [it crashes][4], etc.), then one is left with no lockfile.

Could the behaviour be changed so that the [lockfile deletion][1] is performed **immediately** before the [serialisation of the new content][2]?  (FWIW, I'd be happy to attempt a PR for this.)

I guess two rationales for the current behaviour:

1. Your `Pipfile.lock` should be in Git.  Yes, but it still sucks if you're doing a sequence of new package installs/etc. and haven't committed your work.
2. The update of the `Pipfile.lock` should be consistent with the current state of the virtualenv.  I'd buy that, but the current behaviour (no lockfile) doesn't strike me as better :wink:.

[1]: https://github.com/pypa/pipenv/blob/master/pipenv/core.py#L1026
[2]: https://github.com/pypa/pipenv/blob/master/pipenv/core.py#L1173
[3]: https://github.com/pypa/pipenv/issues/1785
[4]: https://github.com/pypa/pipenv/issues/1837

------------

##### Steps to replicate

1. Start with an existing Pipenv environment (i.e. you have a working `Pipfile`/`Pipfile.lock`).
2. Run `pipenv install <any_dep_that_takes_a_long_time>`.
3. Hit Ctrl+C before it's finished.

frostming:
It was introduced in https://github.com/pypa/pipenv/commit/719656481064dfdc7fb4d9e77a6f23ac7709a087, but I don't see what it has to do with other parts.

Remove `destroy_lockfile`, let's see whether this revision fails testing.

Fixes #1888


techalchemy:
Thanks for taking care of this, I think this is good to go

techalchemy:
@frostming this _mostly_ looks good, sorry... shouldn't you still destroy the lockfile before writing the new one?

I have one implementation question and I don't actually know the answer (/cc @ncoghlan because I'm sure this has been defined somewhere)

If we have been instructed to `lock` and we have a lockfile, but the locking process fails, is it better to leave the current lockfile in place? This PR is intended to stop us from deleting the lockfile before we find out of we can actually perform a lock.  It seems fine to me since we will still detect if the lockfile is out of date, I just don't want to run the risk of people thinking 'oh i have a lockfile, things are fine'

uranusjr:
The lock file is opened (way down in the locking process) with the `w` flag, which truncates the file, so technically no deletion is needed. This would fail with a more obscure exception if the existing lock file is not writable, but judging from existing issues I’m guessing this never happens.

ncoghlan:
I don't think it's been written down anywhere specifically for `pipenv` before, but "Don't destroy the old output until you've determined that you can create the new output" is a pretty common design principle in any tool that supports incremental builds, so I agree it's a good policy for `pipenv` as well.

techalchemy:
Cool, so we really shouldn’t just blanket overwrite this honestly... we should do a temporary rename, then write the lockfile and remove the old one or in case of exceptions rename it back.

ncoghlan:
Thinking about it a little more, I realised preserving the old lock file is even more important than I first thought: if we accidentally truncate it due to an error, then `--keep-outdated` and `--selective-upgrade` won't work properly any more. (I assume that's what I had in mind in likening it to incremental build support, I just hadn't consciously realised how apt that analogy actually was)

techalchemy:
The commits from here are included in #2077 so closing this one for now

PR #2034, opened 2018-04-21
Make sure to consider existing .venv directory

### LINKED ISSUE #1993: pipenv shell does not reuse .venv in project if PIPENV_VENV_IN_PROJECT is not set
```
> echo %PIPENV_VENV_IN_PROJECT%
%PIPENV_VENV_IN_PROJECT%

> pipenv run python -V
Python 3.6.5

> pipenv shell
ERROR: Environment '...' does not exist. Create it with 'pew new ...'
```

uranusjr:
Fix #1993. I made it more general than the issue, so the .venv directory
is honoured in all situations.

### DIFF
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -53,7 +53,6 @@ from .environments import (
     PIPENV_COLORBLIND,
     PIPENV_NOSPIN,
     PIPENV_SHELL_FANCY,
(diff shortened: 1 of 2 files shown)


techalchemy:
seems good, merging

PR #2035, opened 2018-04-22
Update vendored dependencies

techalchemy:
- Includes tooling for automatically installing vendored and patched dependencies
- Also includes patches to rebuild this from scratch
- Tested locally with a completely blank slate
- Also automatically pulls license files (`parse.py` comes from my copy where I broke it out from the code)


techalchemy:
With this PR, as long as you have `pipenv/vendor/vendor.txt` and `pipenv/patched/patched.txt` containing requirements in them along with `tasks/vendoring/patches/vendor/*.patch` and `tasks/vendoring/patches/patched/*.patch`, the entire vendor and patched directories can be rebuilt from scratch with the versions specified in those requirements files and the patches in the corresponding patch subdirectory. In addition, you can rebuild `safety.zip`.  As long as you have `invoke` installed:

```console
$ inv vendoring.update
$ inv vendoring.update-safety
```

Completely rebuilds the vendor directories and re-downloads licenses.  For packages that don't include licenses in their PyPI uploads, you can specify the download link manually in `tasks/vendoring/__init__.py`.

Issue #2040, opened 2018-04-23
Dependencies for locally installed distributions are not being taken into account


mvaled:
Installing locally defined distributions doesn't properly process (and lock) their dependencies.  Steps to reproduce:

1. Create a new project `penv2_bug`

2. Create `penv2_bug/foo/setup.py`, with content:

```python
from setuptools import setup

setup(
    name='foo',
    version='1.0',
    py_modules=['foo'],
    install_requires=[
        'xoutil>=1.8,<2.0',
    ]
)
```

3. Create the module `foo.py` with some content.

4. Run `pipenv install ./foo`.

The generated Pipfile.lock does not contain the resolved dependency for 'xoutil', but something like:

```json
{
    "_meta": {
        "hash": {
            "sha256": "314553368cd3db718a99e462140b44582aa76eda0d250c13ec9a9df36b8eae5f"
        },
        "pipfile-spec": 6,
        "requires": {
            "python_version": "3.6"
        },
        "sources": [
            {
                "name": "pypi",
                "url": "https://pypi.python.org/simple",
                "verify_ssl": true
            }
        ]
    },
    "default": {
        "2cac671": {
            "path": "./foo"
        }
    },
    "develop": {}
}
```

The Pipfile shows:

```toml
[[source]]
url = "https://pypi.python.org/simple"
verify_ssl = true
name = "pypi"

[packages]
"2cac671" = {path = "./foo"}

[requires]
python_version = "3.6"

[dev-packages]
```

<details><summary>$ python -m pipenv.help output</summary>

Pipenv version: `'11.10.0'`

Pipenv location: `'/home/manu/.local/lib/python3.5/site-packages/pipenv'`

Python location: `'/home/manu/.pyenv/versions/3.5.5/bin/python3.5'`

Other Python installations in `PATH`:

  - `2.7`: `/home/manu/.pyenv/shims/python2.7`
  - `2.7`: `/home/manu/.pyenv/shims/python2.7`
  - `2.7`: `/usr/bin/python2.7`
  - `3.5`: `/home/manu/.pyenv/versions/3.5.5/bin/python3.5m`
  - `3.5`: `/home/manu/.pyenv/versions/3.5.5/bin/python3.5`
  - `3.5`: `/home/manu/.pyenv/shims/python3.5`
  - `3.6`: `/home/manu/.pyenv/shims/python3.6m`
  - `3.6`: `/home/manu/.pyenv/shims/python3.6`
[...]

PR #2041, opened 2018-04-23
Note that ``-e`` impacts sub-dependencies in lock file.

mvaled:
Just a note in the docs to avoid others making the mistake I did in #2040.


techalchemy:
Thanks for this update, it's going to be very helpful I imagine!

looks good, merging!

PR #2052, opened 2018-04-25
Speed up lock by removing redundant freeze

SmileyChris:


techalchemy:
appreciate your digging into this (and PR'ing small bites as you go) -- you will find a lot of duplicated work, so try to be patient (we have a few refactors in progress at the moment)

as for specifically what is slow here, you can start by digging into the runtime of `pipenv-resolver --verbose --debug --clear <packagenames>`

SmileyChris:
Cheers, I noticed your other work in the same area.

Poking around that area, debugging a vcs issue you fixed already I noticed  the unnecessary double call to pip freeze (which is a pretty heavy time sink). That's all this PR addresses.

techalchemy:
Do you have a comparison of runtime with this change?

SmileyChris:
So it turns out the change is mostly negligible.

For a smallish project, I go from 9.00s before to 8.64 afters with a `pipenv lock`. That lines up with the approximate 500ms it takes to run `pip freeze` from within that project's virtualenv.

Still, it's duplicated and unnecessary work being done, so may as well clean it up :)

uranusjr:
I wouldn’t even call 0.5s negligible. It’s not exactly significant, but still worth patching.

techalchemy:
Yeah and I agree it's worth cleaning up, just wanted to know how to prioritize it.  Try out the latest version (just released) and let me know how much speedup that provides, if any.

Issue #2062, opened 2018-04-25
TOML parsing can't handle [packages."discord.py"]


greysteil:
Pipenv cannot install the following (valid) Pipfile:
```
[[source]]
url = "https://pypi.python.org/simple"
verify_ssl = true

[packages."discord.py"]
git = "https://github.com/Rapptz/discord.py.git"
```

<details><summary>$ python -m pipenv.help output</summary>

Pipenv version: `'11.10.0'`

Pipenv location: `'/usr/local/lib/python3.6/site-packages/pipenv'`

Python location: `'/usr/local/opt/python/bin/python3.6'`

Other Python installations in `PATH`:

  - `2.6`: `/usr/bin/python2.6`
  - `2.6`: `/usr/bin/python2.6`
  - `2.7`: `/usr/bin/python2.7`
  - `2.7`: `/usr/bin/python2.7`
  - `3.6`: `/usr/local/bin/python3.6m`
  - `3.6`: `/usr/local/bin/python3.6`
  - `3.6`: `/usr/local/bin/python3.6`

  - `3.6.5`: `/usr/local/bin/python`
  - `3.6.5`: `/usr/local/bin/python`
  - `2.7.10`: `/usr/bin/python`
  - `3.6.5`: `/usr/local/bin/python3`
  - `3.6.5`: `/usr/local/bin/python3`

PEP 508 Information:

```
{'implementation_name': 'cpython',
 'implementation_version': '3.6.5',
 'os_name': 'posix',
 'platform_machine': 'x86_64',
 'platform_python_implementation': 'CPython',
 'platform_release': '16.7.0',
 'platform_system': 'Darwin',
 'platform_version': 'Darwin Kernel Version 16.7.0: Wed Oct  4 00:17:00 PDT '
                     '2017; root:xnu-3789.71.6~1/RELEASE_X86_64',
 'python_full_version': '3.6.5',
 'python_version': '3.6',
 'sys_platform': 'darwin'}
```

System environment variables:

  - `TERM_PROGRAM`
  - `TERM`
  - `SHELL`
  - `CLICOLOR`
  - `TMPDIR`
  - `Apple_PubSub_Socket_Render`
  - `TERM_PROGRAM_VERSION`
  - `TERM_SESSION_ID`
  - `USER`
  - `SSH_AUTH_SOCK`
  - `__CF_USER_TEXT_ENCODING`
  - `LSCOLORS`
  - `PATH`
  - `PWD`
  - `EDITOR`
[...]

PR #2067, opened 2018-04-26
Fall back to toml for parsing errors in contoml

greysteil:
Doesn't fix https://github.com/pypa/pipenv/issues/2062 (it turns out that `toml` can't handle this case in the currently vendored version), but may fix other TOML parsing issues. 

I also tried this using `master` from `toml`, which required a few other changes to get working. Using the `master` branch *does* fix https://github.com/pypa/pipenv/issues/2062, so hopefully it will be released soon.

### DIFF
--- a/pipenv/project.py
+++ b/pipenv/project.py
@@ -11,6 +11,7 @@ from first import first
 import pipfile
 import pipfile.api
 import toml
+import prettytoml
 
 try:
     import pathlib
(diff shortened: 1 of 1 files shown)


greysteil:
Looks like `11.10.1` does indeed fix the parser bug in #2062. 🎉 I'm going to close this, but it's here if any other parser errors show up.

Issue #1847, opened 2018-03-26
Pipfile should explicitly specify an encoding


uranusjr:
Spun off from #1842.

Currently Pipenv does not handle non-ASCII characters very well. Pipfile is parsed with the platform’s default encoding; this becomes a problem for cross-platform projects if there are non-ASCII paths in the project, especially for the `scripts` section, but also for local dependencies’ `path` value.

A straightforward solution would be to strictly require Pipfiles to be encoded with UTF-8. This is in line with the best practice of the Python community in general, including core development. Pipenv would then use `io.open` instead of `open` to read Pipfiles instead (in `project.py`), so contents, including paths, are always read as text in the correct encoding.

techalchemy:
I think this makes a lot of sense, does anything break if we implement it right now?

uranusjr:
Yes, if you already have a platform-specific encoded Pipfile with non-ASCII stuff in it. But this is easily amendable by re-saving the Pipfile with UTF-8.

techalchemy:
How can we gracefully warn people about that when we encounter it?

uranusjr:
The parsing would almost always fail with `UnicodeDecodeError` if the encoding is not UTF-8, so it would probably be enough to catch that and emit a prettier error.

I doubt that would be necessary though. This would happen mainly on 1. Windows 2. POSIX systems with non-UTF-8 encoding. The latter is extremely unconventional these days (and I believe is not supported anyway). The former is probably next to non-existent, based on how many fundamental issues brettcannon (avoiding an @ here) was able to file when he adopted Pipenv through VS Code.

techalchemy:
Can we fall back to their default system encoding if we handle that exception?

uranusjr:
@jtratner Yeah it needs careful testing, but it can be pulled off. Many projects already did.

techalchemy:
@uranusjr @jtratner relevant link from Victor Stinner from yesterday regarding 3.7 which will by default open things in utf-8 mode: https://vstinner.github.io/python37-new-utf8-mode.html

vlcinsky:
Yes, encoding shall be agreed and not dependent on current default encoding. Otherwise we will repeat errors as in [distutils](https://bugs.python.org/issue32223) or in [pbr](https://bugs.launchpad.net/pbr/+bug/1745396) package.

For Python 2.x compatibility, this [ConfigParser snippet from PyMoTW](https://pymotw.com/2/ConfigParser/#unicode-configuration-data) could help.

[...]

PR #2069, opened 2018-04-26
Default to \n, retain consistent \r\n

altendky:
When writing the Pipfile and Pipfile.lock make an effort to retain
their existing newlines if consistent.  Default to \n (LF) for new
files and files with inconsistent line endings.


pipenv/project.py
```diff
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 import codecs
+import io
```
altendky: `io.open` was needed for consistent 2/3 behavior without using the deprecated `'U'` mode.  Given the discussion in #1847 it sounds like this may affect the encoding situation as well.

tests/integration/conftest.py
```diff
@@ -113,10 +113,14 @@ def pipfile(self):
 
     @property
     def lockfile(self):
-        p_path = os.sep.join([self.path, 'Pipfile.lock'])
+        p_path = self.lockfile_path
         with open(p_path, 'r') as f:
             return json.loads(f.read())
 
+    @property
+    def lockfile_path(self):
```
techalchemy: This is a bit inconsistent because the pipfile equivalent is called `pipfile_location`

altendky: In this particular class there is already `self.pipfile_path` above.  I'd think either leave this consistent within the class or change both?

pipenv/project.py
```diff
@@ -49,6 +51,16 @@ def _normalized(p):
     return normalize_drive(str(pathlib.Path(p).resolve()))
 
 
+DEFAULT_NEWLINES = u'\n'
+
+
+def preferred_newlines(f):
+    if isinstance(f.newlines, type(u'')):
```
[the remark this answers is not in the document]
uranusjr: I think `six.text_type` is the logically correct choice here.

techalchemy:
Ok @altendky now that we have vastly overcomplicated this, lockfile writing needs to use @uranusjr’s `atomic_open_for_write` contexmanager which uses my backport of `NamedTemporaryFile` to give you access to `io.open` and `newline` as arguments in atomic lockfile creation.

techalchemy:
Sorry I broke this briefly, just pushed a fix and merged it in here

pipenv/project.py
```diff
@@ -610,12 +623,31 @@ def write_toml(self, data, path=None):
                         )
                         data[section][package].update(_data)
             formatted_data = toml.dumps(data).rstrip()
+        if os.path.abspath(path) == os.path.abspath(self.pipfile_location):
```
mgedmin: Maybe use `os.path.samefile()` to handle case-insensitive filesystems better?

uranusjr: `abspath` does not normalise paths, however. We should do `normpath(abspath(...))`.

erinxocon: Good point

altendky:
@techalchemy, seems like py2 `io.open()` is handling the newlines ok so I'm digging a bit further.

```python
import io
import sys

contents = u'''\
blue
red
green
'''

print(sys.version_info)

for name, newline in (('rn', u'\r\n'), ('n', u'\n')):
    path = 'blah.{}.txt'.format(name)
    with io.open(path, 'w', newline=newline) as f:
        f.write(contents)

    with io.open(path) as f:
        f.read()

    print (name, newline, f.newlines)
```

```
PS C:\epc\t\812\pipenv.1847> .venv\scripts\python blah.py
sys.version_info(major=2, minor=7, micro=13, releaselevel='final', serial=0)
('rn', u'\r\n', u'\r\n')
('n', u'\n', u'\n')
```

techalchemy:
Two minor details, that's all I've got.  Huge thanks both to you and @uranusjr for getting all of this cleaned up, helping me figure out the incredibly specific details of backporting the functionality of `NamedTemporaryFile` from 3.x, and tracking down the ever impossible unicode/bytes issues.

With those changes (or a response indicating why not) this should be merged and get us passing again

PR #2073, opened 2018-04-27
Rename _PipenvInstance.pipfile_path to pipfile_location for consistency

altendky:
The inconsistency was noticed at https://github.com/pypa/pipenv/pull/2069#discussion_r184529995.  The PR that is not accepted first should be updated to avoid the original observation actually being out of place (`self.lockfile_path` -> `self.lockfile_location`).

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -47,7 +47,7 @@ class _PipenvInstance(object):
         self._path = TemporaryDirectory(suffix='-project', prefix='pipenv-')
         self.path = self._path.name
         # set file creation perms
-        self.pipfile_path = None
+        self.pipfile_location = None
         self.chdir = chdir
 
         if self.pypi:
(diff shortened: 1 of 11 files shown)


altendky:
@techalchemy if you change your mind about wanting this...  no biggy.  :]  It was only a couple minutes to search and replace.

techalchemy:
@altendky I suspect changing `lockfile_location` to `lockfile_path` is going to touch a lot less stuff ;) was there a reason you went with this one instead of that one as a change? I didn't look too hard at the options to be honest

altendky:
@techalchemy it still is `lockfile_location` over in #2069 which is consistent within the `_PipenvInstance` class, just not with the `Project class`.  Sounds like maybe I misunderstood one of your comments in IRC yesterday since I thought you wanted this.  I don't care too much myself.  :]  Feel free to reopen if you do actually want this.

PR #2080, opened 2018-04-27
Allow pyenv installs in non-interactive sessions in PIPENV_YES is set

greysteil:
I'd like to be able to use the pyenv auto-installs from within a non-interactive session to allow Dependabot to honour the Python version specified in a Pipfile.

If `PIPENV_YES` is set, it seems like a relatively strong indicator that `SESSION_IS_INTERACTIVE` logic should be ignored, so I don't think there's any need for an additional environment variable. Happy to add an explicit one if you would prefer, though.

### DIFF
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -461,7 +461,7 @@ def ensure_python(three=None, python=None):
         if not PYENV_INSTALLED:
             abort()
         else:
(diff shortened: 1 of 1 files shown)


techalchemy:
This is nice and easy for a fix! Beautiful

PR #2104, opened 2018-04-30
Backport `NamedTemporaryFile` for `io.open` support

techalchemy:
- Allows atomic locking to use `newline` argument


techalchemy:
In order to make the atomic locking work with #2069

uranusjr:
So `_compat.NamedTemporaryFile` is not really the same as `tempfile.NamedTempoeraryFile`, right? I kind of feel the module name should be called in another name. `_compat` feels like we’re simply shimming for the stdlib thing, but in reality it’s not the case.

PR #2101, opened 2018-04-30
Check whether inline tables are pinned during pipenv lock --keep-outdated flow

greysteil:
Fixes #2097.

Previously, we were passing an `InlineTableElement` to `is_pinned`, which always returns false for non-strings.

PR #2119, opened 2018-05-02
Implement __iter__ to make AbstractTable a mapping

uranusjr:
This supplements #2101 and makes the check more idiomatic.


techalchemy:
Nice. This stuff is all good. Will review shortly

Issue #2113, opened 2018-05-01
markers 'platform_python_implementation == "cpython"' don't match your environment when installing gevent


jfly:
When installing gevent using pipenv, greenlet is not being installed.

<details><summary>$ python -m pipenv.help output</summary>

Pipenv version: `'11.10.1'`

Pipenv location: `'/usr/lib/python3.6/site-packages/pipenv'`

Python location: `'/usr/bin/python'`

Other Python installations in `PATH`:

  - `2.7`: `/usr/bin/python2.7`
  - `2.7`: `/usr/bin/python2.7`
  - `2.7`: `/home/jeremy/.pyenv/shims/python2.7`
  - `2.7`: `/usr/bin/python2.7`
  - `3.6`: `/usr/bin/python3.6m`
  - `3.6`: `/usr/bin/python3.6`
  - `3.6`: `/home/jeremy/.pyenv/shims/python3.6`
  - `3.6`: `/usr/bin/python3.6`

  - `3.6.5`: `/usr/bin/python`
  - `3.6.5`: `/home/jeremy/.pyenv/shims/python`
  - `3.6.5`: `/usr/bin/python`
  - `2.7.14`: `/usr/bin/python2`
  - `2.7.14`: `/home/jeremy/.pyenv/shims/python2`
  - `2.7.14`: `/usr/bin/python2`
  - `3.6.5`: `/usr/bin/python3`
  - `3.6.5`: `/home/jeremy/.pyenv/shims/python3`
  - `3.6.5`: `/usr/bin/python3`

PEP 508 Information:

```
{'implementation_name': 'cpython',
 'implementation_version': '3.6.5',
 'os_name': 'posix',
 'platform_machine': 'x86_64',
 'platform_python_implementation': 'CPython',
 'platform_release': '4.16.5-1-ARCH',
 'platform_system': 'Linux',
 'platform_version': '#1 SMP PREEMPT Thu Apr 26 16:53:40 UTC 2018',
 'python_full_version': '3.6.5',
 'python_version': '3.6',
 'sys_platform': 'linux'}
```

System environment variables:

  - `LS_COLORS`
  - `LANG`
  - `LESS`
  - `FZF_DEFAULT_COMMAND`
  - `DISPLAY`
  - `PYENV_ROOT`
  - `OLDPWD`
  - `INVOCATION_ID`
  - `EDITOR`
  - `COLORTERM`
  - `PYENV_VIRTUALENV_INIT`
[...]

PR #2123, opened 2018-05-02
When formatting a requirement, only lowercase its name.

jfly:
This fixes https://github.com/pypa/pipenv/issues/2113.

This bug was introduced in
<https://github.com/jazzband/pip-tools/pull/452> as a band-aid fix to
<https://github.com/jazzband/pip-tools/issues/431>. Pipenv then copied
that code in <https://github.com/pypa/pipenv/commit/2553ebcbf7e2574c997a9164234e0adf077ffcac#diff-b56b95ccea8595a0f6f24ea753842976>, and inherited this latent bug.

Maybe the right fix is for pypa/packaging to lowercase the name? There's
a comment here
<https://github.com/pypa/packaging/blob/16.8/packaging/requirements.py#L86>
about normalizing the requirement's name, which might be what this is
referring to.

# Testing

Demo of how I tested this out:

## Before

Note the lowercase "cpython" in the output.

```bash
# Clear pipenv cache
$ rm -r /home/jeremy/.cache/pipenv/

# Test for bug
$ PIPENV_PACKAGES="gevent -i https://python-packages.honordev.com/build/dev/" "/home/jeremy/.local/share/virtualenvs/external-api-kWcvfjzK/bin/python" "/usr/lib/python3.6/site-packages/pipenv/resolver.py" --pre --verbose
RESULTS:
[{"hashes": ["sha256:0141b44a1eea5c3d2dec9e75da3bf0a51153bc6f9ef6e631a1e18460541d2ac1", "sha256:11108689372672000cf5393451cde5b4583b8394a747c1faf56ea5afcca687ef", "sha256:135c4cac0e3b35ba2497acb38ed739606cb2ca1acbec4c052fb4b7f527abfbfb", "sha256:1b75ddcdadce3174d8b0a5ea3f46af36274fe0275a47a73ee3e5be94c0471be5", "sha256:26878e0432cab7cdd85ed08caa0c277d38164a067ea2586c210a49948db9df28", "sha256:2b2be6b0991c17826f4f7a6ee0be3a62ae2825c92158d9a8ae567b399452a579", "sha256:42816718f9039c54f32a8bffe333a63e6e1cc9c673151dd07390b7e7d874bcf4", "sha256:6217c69fbac737d62b67bb07434a61a437373f4211f594ade9210f5a2d053607", "sha256:66574654c43c655beed59825b00e3a5c006c4d9a0b767e3a34ba256246e7df9d", "sha256:83127cd5bab985ad35f7523af32d32ce8f437834a30938b1f5a5b3ff76268eef", "sha256:86b6cf0a5f9d88ff85c6e9c595e5fdc8cb38249efe660cf6fbe0eeb50df241e9", "sha256:967c8df8192ab388e985e964f7c7608dc4a967c081c7b042228c56aeac276e78", "sha256:a140b720c0836e6121092f34649c750e7cff7fdda6b9169b001cac581246612a", "sha256:b3f41829ff9c2d2fde2c71a646b53b9c7f2b62908ee3601ad2d17eaa6c8b89b9", "sha256:c541e1d4f738c9760335a988b47527b0e816d62ce88eb09a61382e551b96dea5", "sha256:dbbe5285ce4d788b8b4643a7ee74544999537661b882b269a3a7bc1cff24b695", "sha256:e2808a40a5747c13e634edbdd42032dfe5f930a39eddd6f235703886cdf486bc", "sha256:e39617b8ae37b840e5b6c149ed5ce26cd5d0d2b348dd4410b1f584d7eac4b333", "sha256:ebeb962bd19c4297067e50c8786115900022564b90a91a13adc77296c6698df3", "sha256:f2e3f638e6bbede7295447c6ed8864ae151c45e85875996299996f4136d231df"], "version": "1.3b1", "name": "gevent"}, {"hashes": ["sha256:09ef2636ea35782364c830f07127d6c7a70542b178268714a9a9ba16318e7e8b", "sha256:0fef83d43bf87a5196c91e73cb9772f945a4caaff91242766c5916d1dd1381e4", "sha256:1b7df09c6598f5cfb40f843ade14ed1eb40596e75cd79b6fa2efc750ba01bb01", "sha256:1fff21a2da5f9e03ddc5bd99131a6b8edf3d7f9d6bc29ba21784323d17806ed7", "sha256:42118bf608e0288e35304b449a2d87e2ba77d1e373e8aa221ccdea073de026fa", "sha256:50643fd6d54fd919f9a0a577c5f7b71f5d21f0959ab48767bd4bb73ae0839500", "sha256:58798b5d30054bb4f6cf0f712f08e6092df23a718b69000786634a265e8911a9", "sha256:5b49b3049697aeae17ef7bf21267e69972d9e04917658b4e788986ea5cc518e8", "sha256:75c413551a436b462d5929255b6dc9c0c3c2b25cbeaee5271a56c7fda8ca49c0", "sha256:769b740aeebd584cd59232be84fdcaf6270b8adc356596cdea5b2152c82caaac", "sha256:ad2383d39f13534f3ca5c48fe1fc0975676846dc39c2cece78c0f1f9891418e0", "sha256:b417bb7ff680d43e7bd7a13e2e08956fa6acb11fd432f74c97b7664f8bdb6ec1", "sha256:b6ef0cabaf5a6ecb5ac122e689d25ba12433a90c7b067b12e5f28bdb7fb78254", "sha256:c2de19c88bdb0366c976cc125dca1002ec1b346989d59524178adfd395e62421", "sha256:c7b04a6dc74087b1598de8d713198de4718fa30ec6cbb84959b26426c198e041", "sha256:f8f2a0ae8de0b49c7b5b2daca4f150fdd9c1173e854df2cce3b04123244f9f45", "sha256:fcfadaf4bf68a27e5dc2f42cbb2f4b4ceea9f05d1d0b8f7787e640bed2801634"], "version": "0.4.13", "name": "greenlet", "markers": "platform_python_implementation == 'cpython'"}]
```

## After

Note the correctly capitalized "CPython" in the output:

```bash
# Clear pipenv cache
$ rm -r /home/jeremy/.cache/pipenv/

# Test for bug
$ PIPENV_PACKAGES="gevent -i https://python-packages.honordev.com/build/dev/" "/home/jeremy/.local/share/virtualenvs/external-api-kWcvfjzK/bin/python" "/usr/lib/python3.6/site-packages/pipenv/resolver.py" --pre
RESULTS:
[{"hashes": ["sha256:0141b44a1eea5c3d2dec9e75da3bf0a51153bc6f9ef6e631a1e18460541d2ac1", "sha256:11108689372672000cf5393451cde5b4583b8394a747c1faf56ea5afcca687ef", "sha256:135c4cac0e3b35ba2497acb38ed739606cb2ca1acbec4c052fb4b7f527abfbfb", "sha256:1b75ddcdadce3174d8b0a5ea3f46af36274fe0275a47a73ee3e5be94c0471be5", "sha256:26878e0432cab7cdd85ed08caa0c277d38164a067ea2586c210a49948db9df28", "sha256:2b2be6b0991c17826f4f7a6ee0be3a62ae2825c92158d9a8ae567b399452a579", "sha256:42816718f9039c54f32a8bffe333a63e6e1cc9c673151dd07390b7e7d874bcf4", "sha256:6217c69fbac737d62b67bb07434a61a437373f4211f594ade9210f5a2d053607", "sha256:66574654c43c655beed59825b00e3a5c006c4d9a0b767e3a34ba256246e7df9d", "sha256:83127cd5bab985ad35f7523af32d32ce8f437834a30938b1f5a5b3ff76268eef", "sha256:86b6cf0a5f9d88ff85c6e9c595e5fdc8cb38249efe660cf6fbe0eeb50df241e9", "sha256:967c8df8192ab388e985e964f7c7608dc4a967c081c7b042228c56aeac276e78", "sha256:a140b720c0836e6121092f34649c750e7cff7fdda6b9169b001cac581246612a", "sha256:b3f41829ff9c2d2fde2c71a646b53b9c7f2b62908ee3601ad2d17eaa6c8b89b9", "sha256:c541e1d4f738c9760335a988b47527b0e816d62ce88eb09a61382e551b96dea5", "sha256:dbbe5285ce4d788b8b4643a7ee74544999537661b882b269a3a7bc1cff24b695", "sha256:e2808a40a5747c13e634edbdd42032dfe5f930a39eddd6f235703886cdf486bc", "sha256:e39617b8ae37b840e5b6c149ed5ce26cd5d0d2b348dd4410b1f584d7eac4b333", "sha256:ebeb962bd19c4297067e50c8786115900022564b90a91a13adc77296c6698df3", "sha256:f2e3f638e6bbede7295447c6ed8864ae151c45e85875996299996f4136d231df"], "version": "1.3b1", "name": "gevent"}, {"hashes": ["sha256:09ef2636ea35782364c830f07127d6c7a70542b178268714a9a9ba16318e7e8b", "sha256:0fef83d43bf87a5196c91e73cb9772f945a4caaff91242766c5916d1dd1381e4", "sha256:1b7df09c6598f5cfb40f843ade14ed1eb40596e75cd79b6fa2efc750ba01bb01", "sha256:1fff21a2da5f9e03ddc5bd99131a6b8edf3d7f9d6bc29ba21784323d17806ed7", "sha256:42118bf608e0288e35304b449a2d87e2ba77d1e373e8aa221ccdea073de026fa", "sha256:50643fd6d54fd919f9a0a577c5f7b71f5d21f0959ab48767bd4bb73ae0839500", "sha256:58798b5d30054bb4f6cf0f712f08e6092df23a718b69000786634a265e8911a9", "sha256:5b49b3049697aeae17ef7bf21267e69972d9e04917658b4e788986ea5cc518e8", "sha256:75c413551a436b462d5929255b6dc9c0c3c2b25cbeaee5271a56c7fda8ca49c0", "sha256:769b740aeebd584cd59232be84fdcaf6270b8adc356596cdea5b2152c82caaac", "sha256:ad2383d39f13534f3ca5c48fe1fc0975676846dc39c2cece78c0f1f9891418e0", "sha256:b417bb7ff680d43e7bd7a13e2e08956fa6acb11fd432f74c97b7664f8bdb6ec1", "sha256:b6ef0cabaf5a6ecb5ac122e689d25ba12433a90c7b067b12e5f28bdb7fb78254", "sha256:c2de19c88bdb0366c976cc125dca1002ec1b346989d59524178adfd395e62421", "sha256:c7b04a6dc74087b1598de8d713198de4718fa30ec6cbb84959b26426c198e041", "sha256:f8f2a0ae8de0b49c7b5b2daca4f150fdd9c1173e854df2cce3b04123244f9f45", "sha256:fcfadaf4bf68a27e5dc2f42cbb2f4b4ceea9f05d1d0b8f7787e640bed2801634"], "version": "0.4.13", "name": "greenlet", "markers": "platform_python_implementation == 'CPython'"}]
```

# TODO

- [x] Report this upstream to pip-tools: https://github.com/jazzband/pip-tools/issues/660.
- [x] Is it possible to write a test for this? I don't know the existing test infrastructure well enough to know where such a test would go, or what it would look like.
- [ ] As you can see in my testing above, because of pipenv's caching, this problem won't simply go away with a release of a new version of pipenv (people will continue using buggy cache files). Do we have a mechanism for clearing the cache on a pipenv upgrade? Or will we just have to make a release note advising people to clear their pipenv caches?


techalchemy:
Thanks for the PR, I think I typed a comment about this on mobile but I have no idea if I sent t or if that was on the issue.

Probably related -- #2125

techalchemy:
@jfly ok, so first of all thanks a ton for digging into this. To answer your questions

1. We would want to upstream this, although I'm not sure if it's correct to only lowercase the requirement name (what if the user puts 'windows' for their `sys_platform', do we really want to be strict?) 
2. Test would look something like this and live in `tests/integration/test_install_markers.py`:

```python
@pytest.mark.markers
@flaky
def test_platform_python_implementation_marker(PipenvInstance, pypi):
    """Markers should be converted during locking to help users who input this incorrectly
    """
    with PipenvInstance(pypi=pypi) as p:
        import platform
            contents = """
[packages]
pytz = {{version = "*", platform_python_implementation = "== '{0}'"}}
            """.format(platform.python_implementation()).strip()
            f.write(contents)

        c = p.pipenv('install')
        assert c.return_code == 0

        assert p.lockfile['default']['funcsigs']['markers'] == "os_name == '{0}'".format(platform.python_implementation())
```
3. This approach gets around the caching problem because it leaves things uppercased -- then we can just update the lockfile code to do something like this (you need to do this in two places):

```python
    python_implementations = ('CPython', 'IronPython', 'Jython', 'PyPy',)
    for dep in results:
        # Add version information to lockfile.
        lockfile['develop'].update(
            {dep['name']: {'version': '=={0}'.format(dep['version'])}}
        )
        # Add Hashes to lockfile
        lockfile['develop'][dep['name']]['hashes'] = dep['hashes']
        # Add index metadata to lockfile.
        if 'index' in dep:
            lockfile['develop'][dep['name']]['index'] = dep['index']
        # Add PEP 508 specifier metadata to lockfile if dep isnt top level
        # or top level dep doesn't itself have markers
        if 'markers' in dep:
            if dep['name'] in dev_packages and not any(key in dev_packages[dep['name']] for key in allowed_marker_keys):
                continue
            implementation = dep['markers'].get('platform_python_implementation', '')
            if implementation:
                marker = first(impl for impl in python_implementations if lower(impl) == lower(implementation))
            else:
                marker = dep['markers']
            lockfile['develop'][dep['name']]['markers'] = marker
```

jfly:
> 1. We would want to upstream this ...

Ok, I've reported this to them in https://github.com/jazzband/pip-tools/issues/660.

> 1. ... what if the user puts 'windows' for their `sys_platform', do we really want to be strict?

If we want to be PEP 508 compliant, then yes, I think we do. If I'm understanding this correctly, then for the same reasons I mentioned in https://github.com/pypa/pipenv/issues/2113#issuecomment-385875585, I believe that since `sys_platform` is an environment marker, it should be compared using regular (case sensitive) Python string equality.

> 2. Test would look something like this and live in tests/integration/test_install_markers.py:

Thanks for the pointer. I messed around with that a bit, and came up with the following test:

```python
@pytest.mark.markers
@flaky
@pytest.mark.markers
@flaky
def test_platform_python_implementation_marker(PipenvInstance, pypi):
    """Markers should be converted during locking to help users who input this incorrectly
    """
    with PipenvInstance(pypi=pypi) as p:
            contents = """
[packages]
pytz = {{version = "*", platform_python_implementation = "== '{0}'"}}
            """.format(platform.python_implementation()).strip()
            f.write(contents)

        c = p.pipenv('install')
        assert c.return_code == 0

        #import pdb; pdb.set_trace()#<<<
        assert p.lockfile['default']['pytz']['markers'] == "platform_python_implementation == '{0}'".format(platform.python_implementation())
```

However, this test actually passes even without any changes to the Pipenv source code, so it's definitely not testing the bug I'm trying to fix =( I stepped through the Pipenv code during this test, and the reason it's not triggering the bug is because we end up on this line of resolver.py:

https://github.com/pypa/pipenv/blob/a16b1d06515b26db929e87905304ba00a9e1f33b/pipenv/patched/piptools/resolver.py#L279

and `self.repository.get_dependencies(ireq)` returns an empty array. To reproduce this bug, we'd need to go into the line which invokes `format_requirement`: https://github.com/pypa/pipenv/blob/a16b1d06515b26db929e87905304ba00a9e1f33b/pipenv/patched/piptools/resolver.py#L299.

I looked a bit into how to trigger that code, and the only solution I can think of would be to add a new package to the [`tests/pypi`](https://github.com/pypa/pipenv/tree/a16b1d06515b26db929e87905304ba00a9e1f33b/tests/pypi) directory. That new package could be something like gevent, which has a dependency on greenlet and uses the `platform_python_implementation` marker. Should I give that a shot?

> 3. This approach gets around the caching problem because it leaves things uppercased -- then we can just update the lockfile code to do something like this (you need to do this in two places):

Hmm... do you mean "it leaves things **lowercased**"? That looks like it could work, but it feels pretty silly to me to first lowercase a string, and then do a lookup later on to figure out it is actually supposed to be capitalized. It also doesn't do the right thing for other marker types such as sys_platform, right?

jfly:
Ping. Any thoughts on how to proceed with this? Thanks!

techalchemy:
Yeah sorry for the delay, feel free to add the requisite wheels and make sure we are handling this case properly but I'll probably merge this soon either way if tests still pass

techalchemy:
Thanks for all your hard work on this!

PR #2129, opened 2018-05-03
Optimisation: Prefer in list to in string

crispgm:
- `' '.join(sys.argv)` is unnecessary here
- in string can cause error. e.g. `--prefix` can get true for `do_pre`.


kennethreitz:
i did this for a reason — i don't remember what it was — but let's not change something that works.

Issue #1463, opened 2018-02-22
Add a `pipenv sync` subcommand (remove `update`)


ncoghlan:
(Breaking this out as the first step towards #1255)

`pipenv install` is currently overloaded to handle a number of different tasks, including both editing `Pipfile` to add new dependencies, and updating the current environment to match the contents of `Pipfile` and/or `Pipfile.lock`. This overloading currently leads to some quirky behaviours, like those described in #1137 (where `[packages]` and `[dev-packages]` are resolved independently of each other).

To start resolving this, `pipenv sync` would be a new subcommand that works *solely* based on `Pipfile.lock`: if that's missing, the command will fail, with a direction to run `pipenv lock` first.

The basic behaviour would be similar to the current `pipenv install --deploy --ignore-pipfile`, but with `[dev-packages]` handling adjusted as follows:

* when `--dev` is given, both `[packages]` and `[dev-packages]` will be installed, anything else will be uninstalled
* when `--no-dev` given, only `[packages]` will be installed, anything else will be uninstalled
* when neither is given, the default will be `--keep-dev`, which installs `[packages]`, ensures versions match for any already installed `[dev-packages]`, and uninstalls everything else

Other changes relative to `pipenv install`:

* no `--pre` argument (only uses exact versions from `Pipfile.lock`)
* no `--skip-lock` (only uses exact versions from `Pipfile.lock`)
* no `--ignore-pipfile` (only uses exact versions from `Pipfile.lock`)
* no `-r` or `-c` import options (only uses exact versions from `Pipfile.lock`)
* always aborts if `Pipfile.lock` is missing or out of date
* `--deploy` only adds the check of the Python version constraint
* no `--system` option (at least for now)

Options shared with `pipenv install`:

* `--three`/`--two`/`--python`: affect implicit venv creation as usual
* `--verbose`/`-h`/`--help`: work the same way as for any other subcommand

techalchemy:
Thanks for detailing this, I feel there are a lot of critical elements to untangle resolution that are kind of scattered across a few issues.  I think breaking this apart into smaller pieces and detailing a plan of action for each one will help guide both a badly needed refactor of certain elements as well as a reimplementation of some of this core logic.

If you are feeling ambitious, I wouldn't mind seeing some of these pieces in project form (though I might be the only one who likes to work that way), but from a very high-level we are talking about essentially:

- Creating environments
- Installing packages
  - 'Deploying' already-resolved dependency graphs
[...]

PR #2131, opened 2018-05-03
Check lockfile exists on sync, and don't update it

uranusjr:
Matching the behaviour mentioned in https://github.com/pypa/pipenv/issues/1463#issuecomment-386297896


erinxocon:
Thanks for the fix @uranusjr! If all tests go well I think this can be merged.

PR #2136, opened 2018-05-04
Only use `-e` for install when `editable=True`

### LINKED ISSUE #2121: Non-editable install from vcs does not work
I would like to install a package from Github in non-editable mode, mainly to not pull in dependencies. When I set `editable = false` in the `Pipfile`, and run `pipenv install`, the package is installed editable anyway, and dependencies are pulled in.

My pipenv version is 11.10.1

------------

##### Expected result

Using Pipfile:

```
[[source]]
verify_ssl = true
name = "pypi"

[packages]
slicerator = {git = "https://github.com/soft-matter/slicerator.git", editable = false, ref = "master"}

[requires]
python_version = "2.7"
```

I expect a non editable install of `slicerator`, without dependencies.

##### Actual result

```
 ROUND 1                           
Current constraints:
  slicerator from git+https://github.com/soft-matter/slicerator.git@master#egg=slicerator

Finding the best candidates:
  found candidate -e git+https://github.com/soft-matter/slicerator.git@master#egg=slicerator (constraint was <any>)

Finding secondary dependencies:

New dependencies found in this round:
  adding [u'six', '', '[]']
Removed dependencies in this round:
Unsafe dependencies in this round:
------------------------------------------------------------
Result of round 1: not stable

                          ROUND 2                           
Current constraints:
  six
  slicerator from git+https://github.com/soft-matter/slicerator.git@master#egg=slicerator

Finding the best candidates:
  found candidate six==1.11.0 (constraint was <any>)
  found candidate -e git+https://github.com/soft-matter/slicerator.git@master#egg=slicerator (constraint was <any>)

Finding secondary dependencies:
  six==1.11.0               requires -
------------------------------------------------------------
Result of round 2: stable, done

Updated Pipfile.lock (89400d)!
```

Resulting lockfile:

```
{
    "_meta": {
        "hash": {
            "sha256": "c7173951bea190534637732bfe1da99d65f79ac699656354358940dd1289400d"
        },
        "pipfile-spec": 6,
        "requires": {
            "python_version": "2.7"
        },
        "sources": [
            {
                "name": "pypi",
                "verify_ssl": true
            }
        ]
    },
    "default": {
        "six": {
            "hashes": [
                "sha256:70e8a77beed4562e7f14fe23a786b54f6296e34344c23bc42f07b15018ff98e9",
                "sha256:832dc0e10feb1aa2c68dcc57dbb658f1c7e65b9b61af69048abc87a2db00a0eb"
            ],
            "version": "==1.11.0"
        },
        "slicerator": {
            "editable": false,
            "git": "https://github.com/soft-matter/slicerator.git",
            "ref": "master"
        }
    },
    "develop": {}
}
```

##### Steps to replicate

Use the Pipfile above, then `pipenv install`.

This issue is part of #2068 I raised earlier.

techalchemy:
- Fixes #2121
- Add tests for regressions

### DIFF
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -717,7 +717,7 @@ def convert_deps_to_pip(deps, project=None, r=True, include_index=False):
         vcs = maybe_vcs[0] if maybe_vcs else None
         if not any(key in deps[dep] for key in ['path', 'vcs', 'file']):
             extra += extras
-        if not isinstance(deps[dep], Mapping):
+        if isinstance(deps[dep], Mapping):
             editable = bool(deps[dep].get('editable', False))
         # Support for files.
         if 'file' in deps[dep]:
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -657,6 +657,7 @@ def convert_deps_to_pip(deps, project=None, r=True, include_index=False):
     for dep in deps.keys():
         # Default (e.g. '>1.10').
         extra = deps[dep] if isinstance(deps[dep], six.string_types) else ''
+        editable = False
         extras = ''
         version = ''
         index = ''
@@ -712,6 +713,8 @@ def convert_deps_to_pip(deps, project=None, r=True, include_index=False):
         vcs = maybe_vcs[0] if maybe_vcs else None
         if not any(key in deps[dep] for key in ['path', 'vcs', 'file']):
             extra += extras
+        if not isinstance(deps[dep], six.string_types):
+            editable = str(deps[dep].get('editable', '')).lower() == 'true'
         # Support for files.
         if 'file' in deps[dep]:
             dep_file = deps[dep]['file']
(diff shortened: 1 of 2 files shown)


pipenv/utils.py
```diff
@@ -712,25 +713,21 @@ def convert_deps_to_pip(deps, project=None, r=True, include_index=False):
         vcs = maybe_vcs[0] if maybe_vcs else None
         if not any(key in deps[dep] for key in ['path', 'vcs', 'file']):
             extra += extras
+        if not isinstance(deps[dep], six.string_types):
+            editable = str(deps[dep].get('editable', '')).lower() == 'true'
```
uranusjr: Doesn’t TOML have native boolean? I thought you could

```python
editable = bool(deps[dep].get('editable', False))
```

Also it might be better to check for `Mapping` instead of “not string”.

tests/unit/test_utils.py
```diff
@@ -56,6 +56,7 @@
         {'requests': {
             'git': 'https://github.com/requests/requests.git',
             'ref': 'master', 'extras': ['security'],
+            'editable': 'false'
```
uranusjr: Why not a boolean `False`?

techalchemy:
I can’t reply on my app for some reason but I like your suggestions. I just want passing tests and code that doesn’t flag things as editable from the mere presence of the key. Feel free to fix this. I don’t actually know much about toml parsing as may be clear

Issue #1110, opened 2017-11-22
Can't lock private pypi packages when connection to https://pypi.org is iffy


helderco:
At least from my company network, I can't get a reliable connection to https://pypi.org:

```
$ curl https://pypi.org/ 
curl: (35) LibreSSL SSL_connect: SSL_ERROR_SYSCALL in connection to pypi.org:443 
```

This is causing `pipenv lock` to fail with a "Connection reset by peer", when pipenv is trying to grab hashes from the new warehouse API, for our private pypi packages.

##### Describe your environment

1. Mac OS X
1. Python version: 3.6.3
1. Pipenv version: 8.3.2

##### Expected result

I expect the lock to skip failed hash grabbing.

##### Actual result
```
. . .
Result of round 6: stable, done
Traceback (most recent call last):
  File "/Users/helder/.local/venvs/pipenv/lib/python3.6/site-packages/urllib3/connectionpool.py", line 601, in urlopen
    chunked=chunked)
  File "/Users/helder/.local/venvs/pipenv/lib/python3.6/site-packages/urllib3/connectionpool.py", line 346, in _make_request
    self._validate_conn(conn)
  File "/Users/helder/.local/venvs/pipenv/lib/python3.6/site-packages/urllib3/connectionpool.py", line 850, in _validate_conn
    conn.connect()
  File "/Users/helder/.local/venvs/pipenv/lib/python3.6/site-packages/urllib3/connection.py", line 326, in connect
    ssl_context=context)
  File "/Users/helder/.local/venvs/pipenv/lib/python3.6/site-packages/urllib3/util/ssl_.py", line 329, in ssl_wrap_socket
    return context.wrap_socket(sock, server_hostname=server_hostname)
  File "/usr/local/opt/pyenv/versions/3.6.3/lib/python3.6/ssl.py", line 407, in wrap_socket
    _context=self, _session=session)
  File "/usr/local/opt/pyenv/versions/3.6.3/lib/python3.6/ssl.py", line 814, in __init__
    self.do_handshake()
  File "/usr/local/opt/pyenv/versions/3.6.3/lib/python3.6/ssl.py", line 1068, in do_handshake
    self._sslobj.do_handshake()
  File "/usr/local/opt/pyenv/versions/3.6.3/lib/python3.6/ssl.py", line 689, in do_handshake
    self._sslobj.do_handshake()
ConnectionResetError: [Errno 54] Connection reset by peer

During handling of the above exception, another exception occurred:

[...]

PR #2145, opened 2018-05-05
Retry pypi.org requests on CI, or if an environment variable is set

greysteil:
Use PIPENV_MAX_RETRIES environment variable to determine how many times to retry requests to pypi.org. Default to 0 retries, unless running on CI, in which case default to 1 retry.

This was the suggested retry setup in the discussion on https://github.com/pypa/pipenv/issues/1110.

### DIFF
--- a/pipenv/environments.py
+++ b/pipenv/environments.py
@@ -23,6 +23,9 @@ PIPENV_COLORBLIND = bool(os.environ.get('PIPENV_COLORBLIND'))
 PIPENV_NOSPIN = bool(os.environ.get('PIPENV_NOSPIN'))
 # Tells Pipenv how many rounds of resolving to do for Pip-Tools.
 PIPENV_MAX_ROUNDS = int(os.environ.get('PIPENV_MAX_ROUNDS', '16'))
+# Specify how many retries Pipenv should attempt for network requests.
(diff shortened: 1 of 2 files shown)


techalchemy:
we discussed the CI check previously in the referenced issue (thanks for referencing that @greysteil) and after re-reading our previous discussion of this it seems like we just never got around to implementing it, this looks good to me

techalchemy:
FYI I'm actually pretty sure pip sets this to 3, so maybe we should just always default to that?

greysteil:
From the pip help output:

```
--retries <retries>         Maximum number of retries each connection should attempt (default 5 times).
```

Happy to change to a default of 5 for CI if you want to match that behaviour (the pip default if `--retries` isn't passed is 0).

techalchemy:
oh I see, that's a confusing bit of help text.  Lets just set it to one for now then, I assume you were hitting a bug and this fixed it? I think we should be pragmatic on this one and assume if it fixes the problem we can just go with the conservative option

greysteil:
👍 and yep, exactly - was tailing Dependabot's log and noticed it hit this on one dependency before succeeding in updating a subsequent one moments later.

Issue #915, opened 2017-10-16
pipenv shell starts a CMD shell when run from powershell


pfmoore:
I use powershell as my primary shell, but `pipenv shell` starts a CMD prompt.

##### Describe you environment

1. OS Type: Windows 7
1. Python version: `$ python -V` 3.6.2
1. Pipenv version: `$ pipenv --version` 8.2.6

##### Expected result

I'd expect the shell that was started to be the same as my current shell (i.e., powershell)

##### Actual result

The shell is always a CMD shell, as far as I can see there is no way to change this.

##### Steps to replicate

`pipenv shell`. It's possible to work around this by using `pipenv run powershell` but this doesn't set `$env:VIRTUAL_ENV` (which I believe is a known issue).

erinxocon:
@pfmoore this is actually a bug in `pew`.  I know because I introduced it :)  It will be fixed soon!  :)

pfmoore:
lol thanks for the clarification :-)

lllama:
@erinxocon Any idea when the fix will be released? (or if there's a work around for the time being?)

erinxocon:
@lllama no ETA yet, it's on my to do list, just been busy with work!  I'll try and expedite this as I'm starting to use windows more and more for work and this bothers me too.

lllama:
@erinxocon thanks for the reply. If anyone else is experiencing this, you can just launch powershell from within cmd and you'll still have paths and deps setup and installed.

lllama:
If anyone else is wondering about this, then it appears to have been fixed with version 11.1.8.

techalchemy:
Stealth-fixed in 2dc06d4a5cf547833b2faf1fd5f9e72922ce58a8

uranusjr:
@hstefan Cmder works perfectly for me with either cmd and Powershell, so this could be environment-dependent. Please do help improve it if possible!

techalchemy:
@hstefan you need to provide more details, such as which version of pipenv you are using.  For context, taking a stab at this would involve reimplementing the ctypes process tree traversal I implemented which initially resolved this.  You should let us know what you have set for environment variables, one of those might be causing issues (@uranusjr what is your `SHELL` set to)?

techalchemy:
@hstefan the output of `get-childitem ENV:` would be helpful

uranusjr:
Ah, I *think* I know what’s going on here. On certain code paths for Cmder you go through this function in Pew:

[...]

PR #2161, opened 2018-05-08
Rewrite most of the shell detection stuff (do not merge)

uranusjr:
The goal is to fix https://github.com/pypa/pipenv/issues/915#issuecomment-387138350


uranusjr:
Manually tested on cmd.exe, Powershell (pwsh.exe, but powershell.exe should work), and Bash in WSL. This needs more testers.

uranusjr:
Some explaination.

I did some process-digging and it turns out there is not a trace of `Cmder.exe` anywhere in the process tree when `_win_utils.get_shell()` is called, so I removed it from `SHELL_NAMES`. It is not useful anyway, since we really need to know not only we’re running Cmder, but also what shell it is spawning for us.

There are quite a few variants of subshell-spawning, each requiring a different call:

* Non-bash shells on non-Windows (simplest to handle).
* Bash (not including Msys Bash, which never worked).
* Bare cmd or Powershell (including pwsh)
* cmd in Cmder
* Powershell in Cmder

The logic gets complicated pretty quickly, so I decided to refactor them into five classes (plus an intermediate for the Cmder cases). `_detect_shell` now returns an instance of `Shell` (or its subclass) for `shell` to use.

pipenv/patched/pew/pew.py
```diff
@@ -148,79 +148,123 @@ def inve(env, command, *args, **kwargs):
...
-def fork_cmder(env, cwd):
-    shell_cmd = ['cmd']
-    cmderrc_path = r'%CMDER_ROOT%\vendor\init.bat'
-    if expandpath(cmderrc_path).exists():
-        shell_cmd += ['/k', cmderrc_path]
-    if cwd:
-        os.environ['CMDER_START'] = cwd
-    fork_shell(env, shell_cmd, cwd)
 
 def _detect_shell():
```
techalchemy: The decision tree in this function is kinda messy, wonder if we can clean this up any

techalchemy:
Overall these changes make sense.  I think we are maintaining 2 lists of windows compatible shells, we should consolidate those.  Otherwise I think we can schedule this for 11.11

uranusjr:
Note:

* [ ] Make sure this does not overwrite #2168.
* [ ] Generate patch.

pipenv/patched/pew/pew.py
uranusjr: Not sure if this is better.

uranusjr: Rewrote this a bit. Not sure if this makes more sense.

hstefan:
[Comment on #915](https://github.com/pypa/pipenv/issues/915#issuecomment-388134146)

Issue #2116, opened 2018-05-01
Feature: Add active environment name to PROMPT env var (Windows Support)


gtalarico:
On unix/fish, when a shell is started using `pipenv shell`, the `$VIRTUAL_ENV` variable can be used to provide a visual indicator on the shell. For Example, if the users prompt is ` (~/folder)  $`
once the shell is activated using `pipenv shell` the prompt becomes:
` (~/folder) (pipenvfolder-1234) $`
or similar. This currently works well with zsh/fish, etc and in general in the unix world is fairly easy to customize prompt.

On windows however, `%PROMPT% variable is used instead, and the variable is used by both cmd.exe and cmder console to add the virtual environment to the user's prompt.

Upon activating an environment, the environment's name is prepended to the `$PROMPT` var.
As far as I know, this is not customizable on cmd.exe, and is not easily customizable with cmder neither.

### cmd.exe behavior with virtualenv / virtualenv-wrapper
Active Environment is shown on prompt

![image](https://user-images.githubusercontent.com/9513968/39497845-16d6f64e-4d73-11e8-88c8-603a3c08457b.png)

### cmder behavior with virtualenv  / virtualenv-wrapper
Active Environment is shown on prompt

![image](https://user-images.githubusercontent.com/9513968/39497992-b03a709a-4d73-11e8-91cf-e825f7124a30.png)

### cmd and cmder with pipenv shell - No prompt indicator
Environment IS NOT shown anywhere on the prompt

![image](https://user-images.githubusercontent.com/9513968/39498041-f412e612-4d73-11e8-8892-54bc396bb2b2.png)

![image](https://user-images.githubusercontent.com/9513968/39498029-e8ae26a6-4d73-11e8-9e41-d2eaf806bd13.png)

# Proposed solution

For the pipenv environment to be shown by default on the built-in cmd.exe and other windows-friendly consoles,  pipenv could prepend the shell environment name to the PROMPT variable in the same way virtualenv does.

Virtualenv stores that old prompt in `_OLD_VIRTUAL_PROMPT` so it can restore the original on deactivate, but pipenv would not need that since the process is destroyed.

### Working example on cmd.exe
![image](https://user-images.githubusercontent.com/9513968/39498612-1264961c-4d77-11e8-9cf3-205b1e0e36f6.png)

### Working example on cmder
![image](https://user-images.githubusercontent.com/9513968/39498523-9a008fdc-4d76-11e8-8e70-b2870d280fa1.png)

[...]

PR #2167, opened 2018-05-09
Adds environment name to PROMPT on Cmd, Cmder,ConEmu

gtalarico:
This PR makes cmd.exe and cmder/conemu show the virtual environment's name 
in those shell's prompt.

This edit could also be in `fork_cmder()` , but that would not fix `cmd.exe`.
I understand the entire section might need some re-work and clean up but until then this should be an easy and contained fix that will make a lot of windows users happy.

image

PS: this is my first PR for a large project


gtalarico:
PS: This fixes https://github.com/pypa/pipenv/issues/2116

techalchemy:
Is this really all it takes...

techalchemy:
BTW we should upstream these to pew as well

techalchemy:
@uranusjr This change is pretty small, while the 11.11 changes are somewhat large in scope.  I really like the idea of getting these improvements and quality of life fixes done, if you think these are ok (and we get an upstream pr to pew opened) we can consider point releasing for these

uranusjr:
In general it is (`%PROMPT%` is essentially PS1 for cmd), but I wonder what consequences would it have on custom init scripts (`clink.lua` for example). Will it result in duplicate venv names, or put it in the wrong place? We’ll need to put this in the wild and see what happens. Yeah I’d say let’s do this.

techalchemy:
Seems like a good reason to get this out before a minor release, we can work through that stuff.

PR #2168, opened 2018-05-09
Escpace spaces on cmder shell path - Fixes #2115

### LINKED ISSUE #2115: Pipenv Shell raises error if shell's path contains space
This might be a windows only issue - Pipenv version 11.10.1
I am using cmder console

##### Steps to replicate
1. `set PIPENV_SHELL="c:\path_with space\enclosed\inquotes\shell"
2. `pipenv shell`

##### Expected result
Pipenv Shell is activated with the desired shell

##### Actual result
Pipenv Shell is activated with the default shell (cmd.exe)
> 'c:\path_with' is not regognized as an internal or external command

I use a portable cmder that is in a folder location that has spaces in its path.
To use Pipenv with cmder, I need to set `PIPENV_SHELL` to cmder's `init.bat` file
This files if there is a space in the folder path.

#### Work Around
Moving cmder's `init.bat` to a location with no spaces fixes it

image

gtalarico:
This PR fixes #2115 by escaping space characters windows-style `%CMDER_ROOT%`

The issue, which is detailed on #2115, happens when Cmder is running from a location which contains space characters.


uranusjr:
This looks good, but would cause conflicts with #2161 (because I rewrote too many things). I’d want to push this in a 11.10.x bugfix release (if there is), and remember to apply this in 11.11.

techalchemy:
I think I'd rather rework this for 11.11 if it still applies

techalchemy:
As an additional note for this pr and for #2167 you will need to add a patch to `tasks/vendoring/patches/patched/pew-some-description.patch` using `git diff -p pipenv/patched/pew/ > tasks/vendoring/patches/patched/pew-some-description.patch` where you fill n the description -- this is for our automated vendoring process.

gtalarico:
> I think I'd rather rework this for 11.11 if it still applies
👍 yes I am not proud of this 

> you will need to add a patch to 
done

PR #2178, opened 2018-05-11
Improving compatibility for bash in Windows

hstefan:


hstefan:
Related to #2161

uranusjr:
Wouldn’t this causes us to lose the old PATH?

hstefan:
I'll can do some thorough testing, but I'd be surprised if it did. 

Thec PR relies on bash to do the expansion when we generate the `export` line in our tmp rc, and we also source `.bashrc` if it exists, so if the user is doing anything different in there it should still work fine. But maybe I'm missing something, so please let me know if you have any particular concerns!

uranusjr:
Oh yeah you’re right, I missed the `:$PATH` part. Excellent.

Issue #2185, opened 2018-05-13
Why I can create only one virtualenv


jarrekk:
Hey, I installed `pipenv` with `brew install pipenv`, the version of Python and pipenv are:

- Python: 3.6.5
- pipenv: 2018.5.18

But now I can only create one virtualenv with python:3.6.5, no matter where I am:

```
~
•100% ➜ cd hello

~/hello
•100% ➜ pipenv --three
Creating a virtualenv for this project…
Using /usr/local/bin/python3 (3.6.5) to create virtualenv…
⠋Running virtualenv with interpreter /usr/local/bin/python3
Using base prefix '/usr/local/Cellar/python/3.6.5/Frameworks/Python.framework/Versions/3.6'
New python executable in /Users/jack/.local/share/virtualenvs/jack-JNXsLa7q/bin/python3.6
Also creating executable in /Users/jack/.local/share/virtualenvs/jack-JNXsLa7q/bin/python
Installing setuptools, pip, wheel...done.

Virtualenv location: /Users/jack/.local/share/virtualenvs/jack-JNXsLa7q

~/hello took 8s
•100% ➜ pipenv --rm
Removing virtualenv (/Users/jack/.local/share/virtualenvs/jack-JNXsLa7q)…

~/hello
•100% ➜ cd ../world

~/world
•100% ➜ pipenv --three
Creating a virtualenv for this project…
Using /usr/local/bin/python3 (3.6.5) to create virtualenv…
⠋Running virtualenv with interpreter /usr/local/bin/python3
Using base prefix '/usr/local/Cellar/python/3.6.5/Frameworks/Python.framework/Versions/3.6'
New python executable in /Users/jack/.local/share/virtualenvs/jack-JNXsLa7q/bin/python3.6
Also creating executable in /Users/jack/.local/share/virtualenvs/jack-JNXsLa7q/bin/python
Installing setuptools, pip, wheel...done.

Virtualenv location: /Users/jack/.local/share/virtualenvs/jack-JNXsLa7q

~/world took 7s
•100% ➜ cd ;pwd
/Users/jack
```

The two projects' virtualenv path are the same: `/Users/jack/.local/share/virtualenvs/jack-JNXsLa7q`.

IMO, the Python virtualenv should be located at:

- /Users/jack/.local/share/virtualenvs/**hello**-JNXsLa7q 
- /Users/jack/.local/share/virtualenvs/**world**-JNXsLa7q 

based on the video instructor at https://docs.pipenv.org/

[...]

PR #2186, opened 2018-05-13
Echo Pipfile location during init

uranusjr:
This should help the user know what's going on in situations like #2185.


techalchemy:
Why are these tests failing?

PR #2193, opened 2018-05-15
remove cacert.pem bundle

nateprewitt:
Now that we've vendored in a modern version of Requests, we don't want this anymore.


techalchemy:
Thanks Nate!

danieldaeschle:
I am waiting for release to fix this. When will it be released?

mystic-cg:
This problem still appeared.(Env: pip18.0, pipenv2018.7.1, win10, python3.6.4)

uranusjr:
@mystic-cg Uninstall and install Pipenv again. pip sometimes has problems removing existing files from the system when you upgrade a package. Nothing we can do here.

PR #2211, opened 2018-05-17
Clean unused sync options

### LINKED ISSUE #2010: Clean unused arguments/options of `pipenv sync`
There exist many arguments & options unused in `pipenv sync`/`do_sync`, which need to be cleaned up. But before doing that, I'd like to list them below to see whether they should be axed or preserved for future features.

- [x] `install`. Actually `install` command passed in, not used. **To be removed**
- [x] `dry_run`. Perhaps we should add this feature, listing the packages not updated or installed. **To be implemented**
- [x] `bare`  **Remain**
- [x] `dont_upgrade`. Same as `keep_outdated`? Need to redesign the current logic, **To be removed**
- [x] `user`. I don't know what does it mean.  **To be removed**
- [x] `clear`. Not used, dependency resolution isn't performed in sync.  **To be removed**
- [x] `unused`  **To be removed**

frostming:
Closes #2010 

`pipenv update --dry-run` will also output not installed packages, as what's done in non dry-run mode. Here is what it produces:
```bash
$ pipenv sync --dry-run
Package 'requests' out–of–date: '==2.14.0' installed, '==2.18.4' available.
Package 'chardet==3.0.4' is not installed.
Package 'urllib3==1.22' is not installed.
Package 'idna==2.6' is not installed.
Package 'certifi==2018.4.16' is not installed.
```

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -740,6 +740,9 @@ def check(
     default=None,
     help=u"List out–of–date dependencies.",
 )
+@option(
+    '--system', is_flag=True, default=False, help="System pip management."
+)
 @argument('package', default=False)
 @pass_context
 def update(
@@ -794,6 +797,7 @@ def update(
             python=python,
             bare=bare,
             verbose=verbose,
+            system=system,
             sequential=sequential,
         )
     else:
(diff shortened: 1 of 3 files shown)


frostming:
Does anyone know how to figure out what happened in the buildkite testing? I just got a 404.

techalchemy:
Sorry for the delay, we had some permissions issues -- I can give you an update on this once tests run again

kennethreitz:
this looks like a drastic change to our api, and i don't think it's what we want. i'd have to review more carefully — maintainers, feel free to re-open if needed.

frostming:
That's fine. BTW, the test failure is causes by **en-dash** in "out-of-date". I don't know whether it is more proper for en-dash instead of hyphen [in this case](http://www.thepunctuationguide.com/en-dash.html), just to throw it out to avoid similar failures in the future.

uranusjr:
“out-of-date” should most definitely use hyphens, not en dashes.

PR #2212, opened 2018-05-17
Update piptools (work in progress)

techalchemy:
- Step 1 in the compatibility steps for the manylinux / pip10 conversion
- Next step - drop vendored pip9, patch and vendor pip10


kennethreitz:
@techalchemy please put in the title if this is a WIP or not

PR #2214, opened 2018-05-18
note that Pipenv cannot publish packages

cs01:


techalchemy:
Does pip make similar clarifications? I’m not convinced we need to start stating all the things we don’t do

uranusjr:
People generally have a good understanding what pip does, and I find many having problems understanding what Pipenv is (I have no idea why they don’t, but they just don’t). Maybe it’s better to say Pipenv can be generally understood as *a nice wrapper to use Pipfile with pip and virtualenv*. This should give people a ballpark estimation of Pipenv’s goals.

cs01:
> Does pip make similar clarifications?

No it does not. But its description is "The PyPA recommended tool for installing Python packages" which does not really carry implications that it does.

Contrast that to what a person who encounters Pipenv for the first time might think:

> Python Development Workflow for Humans. 

"I do Python development and it sounds like this might be pretty all encompassing. I'm also a human, so I'll click the link".

> Pipenv — the tool for managing application dependencies from PyPA, free (as in freedom).

"Oh okay a tool for managing application dependencies. Right now I have dependencies in requirements.txt and setup.py and it's kind of a pain. Wait, this isn't just *a* tool, it's *the* tool. And it's free, like every other Python infrastructure tool out there , but it also mentions that for some reason."

> Pipenv is a tool that aims to bring the best of all packaging worlds (bundler, composer, npm, cargo, yarn, etc.) to the Python world. Windows is a first–class citizen, in our world.

"AMAZING! This is what I've been waiting for. The Pipfile probably looks like a package.json file or a Cargo file and allows me to publish to PyPI if I want, as well as generate a lock file. It's probably a well defined standard or generally accepted packaging format since Python is such a huge community. It must be a solved problem, and this must be it since it has 11k+ stars."

> It automatically creates and manages a virtualenv for your projects, as well as adds/removes packages from your Pipfile as you install/uninstall packages. It also generates the ever–important Pipfile.lock, which is used to produce deterministic builds.

"Nothing about unifying dependences from setup.py into Pipfile, but maybe they're just saving that for later. Or maybe that is done in this Pipfile thing? Those other tools have unified their dependencies, and this *is* dependencies for humans."

> The problems that Pipenv seeks to solve are multi-faceted:
You no longer need to use pip and virtualenv separately. They work together.
Managing a requirements.txt file can be problematic, so Pipenv uses the upcoming Pipfile and Pipfile.lock instead, which is superior for basic use cases.
Hashes are used everywhere, always. Security. Automatically expose security vulnerabilities.
Give you insight into your dependency graph (e.g. $ pipenv graph).
Streamline development workflow by loading .env files.

"Multi-faceted, huh? Sounds like it does a lot. Either way, I better use it, since it's *the* tool. I'm sure I can google around and figure out the best practice on how to publish with it, or at least have it generate setup.py files."

---

A reasonable person might expect this tool to be more than a nice wrapper around pip and virtualenv. And it is the responsibility of the developers to state all the things this tool doesn't do that a reasonable person might expect this tool to do. It makes things easier on the user since they don't waste their time trying to fit a square peg in a round hole, and on maintainers so they don't have to hear complaints about the tool.

I agree that 

> Pipenv: Python Development Workflow for Humans

would be better written as

> Pipenv: A convenient wrapper around pip and virtualenv

And it would be better off if 

> Pipenv is a tool that aims to bring the best of all packaging worlds (bundler, composer, npm, cargo, yarn, etc.) to the Python world. Windows is a first–class citizen, in our world.

were omitted entirely, because those tools have more features.

Issue #2229, opened 2018-05-21
`pipenv clean` fails when `pip freeze` produces output with comments


nelfin:
- [x] Be sure to check the existing issues (both open and closed!).
  * same stack trace but unrelated: #1750 
  * PR to fix a similar issue with `pipenv update`: #1268 

`do_clean` assumes that the output of `pip freeze` produces only requirement lines, but it can in some cases produce `#`-prefixed comments (see https://github.com/pypa/pip/blob/729990c9869148f3f0098a2b2a0c0b92aefb8a69/src/pip/_internal/operations/freeze.py#L188 for one such example). It looks as though this was noticed and fixed for `pipenv update` in 9.1.0 (see f322d93)

<details><summary>$ python -m pipenv.help output</summary>

Pipenv version: `'2018.05.18'`

Pipenv location: `'$HOME/.virtualenvs/pipenv-issue/local/lib/python2.7/site-packages/pipenv'`

Python location: `'$HOME/.virtualenvs/pipenv-issue/bin/python'`

Other Python installations in `PATH`:

  - `2.7`: `$HOME/.virtualenvs/pipenv-issue/bin/python2.7`
  - `2.7`: `$HOME/.virtualenvs/pipenv-issue/bin/python2.7`
  - `2.7`: `/usr/bin/python2.7`
  - `3.5`: `/usr/bin/python3.5m`
  - `3.5`: `/usr/bin/python3.5`

  - `2.7.12`: `$HOME/.virtualenvs/pipenv-issue/bin/python`
  - `2.7.12`: `/usr/bin/python`
  - `2.7.12`: `$HOME/.virtualenvs/pipenv-issue/bin/python2`
  - `2.7.12`: `/usr/bin/python2`
  - `3.5.2`: `/usr/bin/python3`

PEP 508 Information:

```
{'implementation_name': 'cpython',
 'implementation_version': '0',
 'os_name': 'posix',
 'platform_machine': 'x86_64',
 'platform_python_implementation': 'CPython',
 'platform_release': '4.4.0-124-generic',
 'platform_system': 'Linux',
 'platform_version': '#148-Ubuntu SMP Wed May 2 13:00:18 UTC 2018',
 'python_full_version': '2.7.12',
 'python_version': '2.7',
 'sys_platform': 'linux2'}
```

System environment variables:

  - `MANDATORY_PATH`
  - `XDG_GREETER_DATA_DIR`
  - `GNOME_DESKTOP_SESSION_ID`
[...]

PR #2230, opened 2018-05-21
Ignore comments from `pip freeze` when running `pipenv clean`

nelfin:
Addresses #2229

### DIFF
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -46,6 +46,7 @@ from .utils import (
     is_star,
     rmtree,
     split_argument,
+    ignore_comments,
 )
 from ._compat import (
     TemporaryDirectory,
@@ -2573,6 +2574,7 @@ def do_clean(
             '\n'
         ),
     )
+    installed_packages = ignore_comments(installed_packages)
     installed_package_names = []
     for installed in installed_packages:
         r = get_requirement(installed)
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -81,6 +81,12 @@ def _get_requests_session():
     return requests_session
 
 
+def ignore_comments(iterable):
+    for line in iterable:
+        if not line.lstrip().startswith('#'):
+            yield line
+
+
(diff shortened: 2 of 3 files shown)


uranusjr:
While I agree with the logic, it is probably an overkill to create a generator function just for this purpose. It would be much readable to merge this into the previous `filter` call.

nelfin:
HI @uranusjr, thanks for taking the time to review but I disagree that creating a separate function for this is overkill, as it has the advantage that the intent of the action is explicit: the reader can simply choose to trust that the function skips commented lines without having to know the implementation. On the other hand, extending the filter predicate might look something like this:
```py
    installed_packages = filter(
        lambda line: line and not line.lstrip().startswith('#'),
        delegator.run('{0} freeze'.format(which_pip())).out.strip().split(
            '\n'
        ),
    )
```
Here there reader is presented with the "how", and needs to figure out the "what".

This pull request is actually cut down to a minimal change from a larger set of changes where `pip freeze` output is treated in a more consistent manner which would justify having this in one place to prevent re-use. As a first-time contributor to this project however, I wasn't going to push up a pull request that touched many parts of the core (at a glance it looks like lock, purge, clean, and outdated all use the output of `pip freeze` but only purge handles this, see #1268) without first becoming more familiar with the testing and the codebase so that we could all be confident that such a refactor wouldn't break anything existing.

uranusjr:
@nelfin There is a middle ground between filtering + iterator and a lambda:

```python
def is_requirement_line_useful(line):
    line = line.strip()
    if not line:
        return False
    if not line.startswith('#'):
        return False
    return True

installed_packages = filter(
    is_requirement_line_useful,
    delegator.run('{0} freeze'.format(which_pip())).out.split('\n'),
)
```

(Personally I’d prefer a generator to `filter` in this case, but that’s only my preference.)

kennethreitz:
+1 to generator, filter is hard to understand/maintain

PR #2250, opened 2018-05-23
Activate PIPENV_VENV_IN_PROJECT with --local

### LINKED ISSUE #2086: activate PIPENV_VENV_IN_PROJECT with --local

sivakar12:
Fixes #2086 
I think this works. How do I check? Given that the project is already in a virtualenv, can it create a virtualenv inside it?

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -138,6 +138,12 @@ def validate_python_path(ctx, param, value):
     default=False,
     help="Enable site-packages for the virtualenv.",
 )
+@option(
+    '--local',
+    is_flag=True,
+    default=False,
+    help="Use the .venv folder instead of global virtual environment folder"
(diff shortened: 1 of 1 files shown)


techalchemy:
Set PIPENV_IGNORE_VIRTUALENVS=1

Thanks for the contribution! If you get a chance, tests would also be helpful (in tests/integration/test_cli.py) — could be as simple as installing a package and making sure pipenv —venv outputs the right path

sivakar12:
Thanks! I'll check. I'll write the tests too

kennethreitz:
:-1:

kennethreitz:
thanks for the suggestion, we need to be very careful about adding new flags, though. This feature is automatic if you just 'mkdir .venv'.

PR #2262, opened 2018-05-25
Nelfin fix 2229

uranusjr:
Manual merge. Close #2230.


uranusjr:
Urrgh Buildkite is failing and I can’t even see its logs.

Issue #2075, opened 2018-04-27
Allow users to override default PyPI index URL with PyPI mirror URLs (without modifying Pipfile)


JacobHenner:
Hello all,

**The situation**

Currently, there is no easy way to override the default PyPI index URL to use a URL pointed at a mirror. In corporate environments, requiring developers to use a repository mirror is quite common:

1. Corporate firewalls prohibit access to external software repositories.
2. Internal repository mirrors conduct malware and vulnerability analysis, which can be a compliance requirement.
3. Internal mirrors preserve modules that might later be unavailable upstream (due to outage, deletion, etc), which is necessary to ensure the availability and auditability of modules used within the company's environment.

Unfortunately, this doesn't appear to be easily accommodated by pipenv. Although the mirror could be explicitly added to the Pipfile as the source for these packages, this breaks portability.

1. Projects initialized internally will contain unreachable indexes if published externally. Users of the public version will have to modify the Pipfile prior to installing the module's dependencies.
2. Projects initialized externally will not work internally without modification of the Pipfile. These modifications must be maintained locally (but not shared), and reapplied if the Pipfile changes upstream.

There should be a way to override the location of the PyPI index, by specifying a (true) mirror. This would only be applicable to PyPI, and not to other third-party repositories (these would still be specified explicitly in the Pipfile). 

**General proposal**

Docker accommodates this situation by allowing the user to [specify a registry mirror](https://docs.docker.com/registry/recipes/mirror/#configure-the-docker-daemon) in the daemon's configuration file. Likewise, it'd be great if the pipenv user could specify a (true) mirror for PyPI, via an environment variable, configuration file, or command line parameter. If this value is set, pipenv should use the mirror for all PyPI packages, even if a connection to PyPI is available. In some corporate environments, PyPI remains unblocked, but policy dictates that the mirror is used for the other reasons mentioned above.

**Implementation considerations**
1. Pip already allows users to override the default index url through pip's configuration file. Although this would likely be the most obvious source of the internal mirror's url (and would likely be set for these users), this parameter can be used for repositories that aren't true mirrors. Accordingly, it's probably unsuitable for this purpose.
[...]

Issue #2188, opened 2018-05-13
Proposal: Explicit command to create a Pipfile in the current directory


uranusjr:
Pipenv does not currently have an explicit “new project” command, but automatically ensures everything is in order when you run something. While this works well in general, it may leave people in the dark if the environment configuration is slightly off, as is the case of #2185.

#2185 is obviously a user error, and #2186 would help identify the problem, but the user, unless with substantial understanding to Pipenv internals, would still not be able to understand what action should be taken next.

A similar, but non-user error situation is to create nested Pipfile projects. This is suggested in #2148, which introduces the notion of “including” sub-projects. How this inclusion mechanism should be handled, if at all, is up to further discussion, but the idea of a Pipfile sub-project in itself is currently already problematic. You wouldn’t be able to create a sub-project if you create a parent project first, because all Pipenv commands will see the parent project, and refuse to create a new one.

The only reasonable way to work around this is to create the new Pipfile manually (e.g. `touch Pipfile`).  This is however still hackish, and not at all obvious unless you are knowledgable about Pipenv’s internals. With Pipenv encouraging user to let it handle Pipfile edits entirely (with `pipenv install/uninstall [package]` commands), it would also confuse users if it breaks consistency here.

The proposal is to provide an obvious way for user to initialise a virtual environment with a Pipfile created in the current directory, similar to `npm init`. This does not change the current initialisation behaviour, but only provide an additional way to initialise not only the virtual environment, but the Pipfile explicitly, right here. There are several possible ways to implement this:

* An environment variable. This is probably not a good idea, since this needs to be obvious to be useful. Environment variabls are not obvious at all, at least to me. It would also conflict with `PIPENV_PIPFILE`.
* An option to `pipenv`, e.g. `pipenv --python=... --project=.`. This is not as obvious as a top-level subcommand, but is still reasonably visible, and can be combined with other subcommands to perform tasks. It would still conflict with `PIPENV_PIPFILE`. What would `PIPENV_PIPFILE=/path/a/Pipfile pipenv --python=... --project=/path/b` even mean?
* An extra sub-command, *a la* `npm init`. This is the most obvious solution, but given the history of reluctancy to adding new commands, I can understand this may not be the preferred way to go. I do prefer this the most, and it is the only option that does not conflict with `PIPENV_PIPFILE`. It can simply throw an error if a Pipfile already exists at the target location, and `PIPENV_PIPFILE` errors if the specifies location does not exist.

gsemet:
[...]

PR #2281, opened 2018-05-30
Allow users to override default PyPI index URLs with PyPI mirror URLs

JacobHenner:
In response to #2075, allow users to override the default PyPI indices (pypi.org, pypi.python.org) with a mirror specified via CLI parameter or environment variable.

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -24,6 +24,7 @@ from .__version__ import __version__
 
 from . import environments
 from .environments import *
+from .utils import is_valid_url
 
 # Enable shell completion.
 init_completion()
@@ -79,7 +80,10 @@ def validate_python_path(ctx, param, value):
             raise BadParameter('Expected Python at path %s does not exist' % value)
     return value
 
-
+def validate_pypi_mirror(ctx, param, value):
+    if value and not is_valid_url(value):
+        raise BadParameter('Invalid PyPI mirror URL: %s' % value)
+    return value
 @group(
     cls=PipenvGroup,
     invoke_without_command=True,
@@ -303,6 +307,13 @@ def cli(
     help="Specify which version of Python virtualenv should use.",
 )
 @option(
+    '--pypi-mirror',
+    default=PIPENV_PYPI_MIRROR,
+    nargs=1,
+    callback=validate_pypi_mirror,
+    help="Specify a PyPI mirror.",
+)
+@option(
     '--system', is_flag=True, default=False, help="System pip management."
 )
 @option(
@@ -368,6 +379,7 @@ def install(
     dev=False,
     three=False,
     python=False,
+    pypi_mirror=None,
     system=False,
     lock=True,
     ignore_pipfile=False,
@@ -389,6 +401,7 @@ def install(
         dev=dev,
         three=three,
         python=python,
+        pypi_mirror=pypi_mirror,
         system=system,
         lock=lock,
         ignore_pipfile=ignore_pipfile,
@@ -452,6 +465,13 @@ def install(
     default=False,
     help=u"Keep out–dated dependencies from being updated in Pipfile.lock.",
 )
+@option(
+    '--pypi-mirror',
+    default=PIPENV_PYPI_MIRROR,
+    nargs=1,
+    callback=validate_pypi_mirror,
+    help="Specify a PyPI mirror.",
+)
 def uninstall(
     package_name=False,
     more_packages=False,
@@ -463,6 +483,7 @@ def uninstall(
     all=False,
     verbose=False,
     keep_outdated=False,
+    pypi_mirror=None,
 ):
     from .core import do_uninstall
 
@@ -477,6 +498,7 @@ def uninstall(
         all=all,
         verbose=verbose,
         keep_outdated=keep_outdated,
+        pypi_mirror=pypi_mirror,
     )
 
 
@@ -493,6 +515,13 @@ def uninstall(
     nargs=1,
     callback=validate_python_path,
     help="Specify which version of Python virtualenv should use.",
+)
+@option(
+    '--pypi-mirror',
+    default=PIPENV_PYPI_MIRROR,
+    nargs=1,
+    callback=validate_pypi_mirror,
+    help="Specify a PyPI mirror.",
 )
 @option(
     '--verbose',
@@ -531,6 +560,7 @@ def uninstall(
 def lock(
     three=None,
     python=False,
+    pypi_mirror=None,
     verbose=False,
     requirements=False,
     dev=False,
@@ -543,9 +573,9 @@ def lock(
     # Ensure that virtualenv is available.
     ensure_project(three=three, python=python)
     if requirements:
-        do_init(dev=dev, requirements=requirements)
+        do_init(dev=dev, requirements=requirements, pypi_mirror=pypi_mirror)
     do_lock(
-        verbose=verbose, clear=clear, pre=pre, keep_outdated=keep_outdated
+        verbose=verbose, clear=clear, pre=pre, keep_outdated=keep_outdated, pypi_mirror=pypi_mirror
     )
 
 
@@ -696,6 +726,13 @@ def check(
     help="Specify which version of Python virtualenv should use.",
 )
 @option(
+    '--pypi-mirror',
+    default=PIPENV_PYPI_MIRROR,
+    nargs=1,
+    callback=validate_pypi_mirror,
+    help="Specify a PyPI mirror.",
+)
+@option(
     '--verbose',
     '-v',
     is_flag=True,
@@ -747,6 +784,7 @@ def update(
     ctx,
     three=None,
     python=False,
+    pypi_mirror=None,
     system=False,
     verbose=False,
     clear=False,
@@ -774,7 +812,7 @@ def update(
     if not outdated:
         outdated = bool(dry_run)
     if outdated:
-        do_outdated()
+        do_outdated(pypi_mirror=pypi_mirror)
     if not package:
         echo(
             '{0} {1} {2} {3}{4}'.format(
@@ -786,7 +824,7 @@ def update(
             )
         )
         do_lock(
-            verbose=verbose, clear=clear, pre=pre, keep_outdated=keep_outdated
+            verbose=verbose, clear=clear, pre=pre, keep_outdated=keep_outdated, pypi_mirror=pypi_mirror
         )
         do_sync(
             ctx=ctx,
@@ -801,6 +839,7 @@ def update(
             clear=clear,
             unused=False,
             sequential=sequential,
+            pypi_mirror=pypi_mirror,
         )
     else:
         for package in ([package] + list(more_packages) or []):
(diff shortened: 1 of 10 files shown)


JacobHenner:
WIP - Still need to write the test cases and update the docs, but figured I'd start the PR to encourage review.

uranusjr:
The implementation looks reasonable with a brief read (I didn’t actually run it, and yes tests are needed). I’m not sure if the flag is needed for bare `pipenv` calls.

JacobHenner:
I've been working on adding tests for install, but I'm running into an issue testing the --pypi-mirror parameter. It looks like the test suite overrides the default sources to use a locally-run web server, but I've written a test to ensure setting --pypi-mirror doesn't modify the Pipfile's sources. I haven't been able to narrow down where this is occurring, or how to prevent it, however, so the test fails when it reads a source url "http://localhost...".

Also, right now this just covers install. I hope to test and address all functionality currently dependent on the two primary PyPI urls before proceeding.

techalchemy:
Thanks for the PR! The test fixtures and the project.py itself are both configured to use an enviroment variable to configure the test source to always use a local pypi cache for testing. Your test just needs to verify that the real url, when you place it in the sources somewhere, gets overridden during the install process. So you can write a Pipfile with an extra source to actual pypi and use `p.pipenv('install --pypi-mirror')` and use the verbose flag and just verify that the actual url is not present in the output or something, since it should include source urls

JacobHenner:
@techalchemy Thanks, I'll take a look at that.

JacobHenner:
@techalchemy If I go that route, where would you suggest I store that Pipfile?

For the local pypi cache, I can't seem to see where it's set. During each test run, I see the server uses 127.0.0.1 with a random port, but I'm still looking (without much luck) for where this is set, or where PIPENV_TEST_INDEX is set by the test suite.

JacobHenner:
```python
@pytest.mark.install
@flaky
def test_mirror_install(PipenvInstance, pypi):
    with PipenvInstance(pypi=False) as p:
        # This should sufficiently demonstrate the mirror functionality
        # since pypi.org is the default.
        c = p.pipenv('install requests --pypi-mirror https://pypi.python.org/simple')
        assert c.return_code == 0
        # Ensure the --pypi-mirror parameter hasn't altered the Pipfile or Pipfile.lock sources
        assert len(p.pipfile['source']) == 1
        assert len(p.lockfile["_meta"]["sources"]) == 1
        assert 'https://pypi.org/simple' == p.pipfile['source'][0]['url']
        assert 'https://pypi.org/simple' == p.lockfile['_meta']['sources'][0]['url']
        
        assert 'requests' in p.pipfile['packages']
        assert 'requests' in p.lockfile['default']
        assert 'chardet' in p.lockfile['default']
        assert 'idna' in p.lockfile['default']
        assert 'urllib3' in p.lockfile['default']
        assert 'certifi' in p.lockfile['default']
```

still returns 

```
test_mirror_install failed; it passed 0 out of the required 1 times.
	<type 'exceptions.AssertionError'>
	assert 'https://pypi.org/simple' == 'http://127.0.0.1:44423/simple'
  - https://pypi.org/simple
  + http://127.0.0.1:44423/simple
```

techalchemy:
Hey thanks for all your work on this front, so here's what you actually need to do to get your tests to work (almost there!)-- not sure you need to pass `chdir=True` here but it's possible:

```python
@pytest.mark.install
@flaky
def test_mirror_install(PipenvInstance):
            f.write("""
[[source]]
verify_ssl = true
name = "pypi"
            """.strip())
        # This should sufficiently demonstrate the mirror functionality
        # since pypi.org is the default.
        c = p.pipenv('install requests --pypi-mirror https://pypi.python.org/simple')
        assert c.return_code == 0
        # Ensure the --pypi-mirror parameter hasn't altered the Pipfile or Pipfile.lock sources
        assert len(p.pipfile['source']) == 1
        assert len(p.lockfile["_meta"]["sources"]) == 1
        assert 'https://pypi.org/simple' == p.pipfile['source'][0]['url']
        assert 'https://pypi.org/simple' == p.lockfile['_meta']['sources'][0]['url']

        assert 'requests' in p.pipfile['packages']
        assert 'requests' in p.lockfile['default']
        assert 'chardet' in p.lockfile['default']
        assert 'idna' in p.lockfile['default']
        assert 'urllib3' in p.lockfile['default']
        assert 'certifi' in p.lockfile['default']
```

And FYI, I grabbed a copy of your branch and ran this test on it and it passes, so nice work! :)

JacobHenner:
Thanks @techalchemy. I've gone ahead and added that. 

I'm now going to work on the other functions which depend on connectivity to PyPI, such as updates.

JacobHenner:
This is almost finished, I have two things left to address:
1. There's currently an issue with setting an environment variable on Windows running Python 2, according to appveyor.
2. I need to write additional integration tests for update, sync, lock, and uninstall. The challenge will be figuring out which mirror can be used for testing, and how to ensure that mirror is being used (and ensure that PyPI isn't being used silently in the background).

Hopefully I'll have time to work on this tomorrow. If not, I'll try to get to it by Wednesday.

uranusjr:
Maybe you can use the actual PyPI, and override it with the local (mocked) PyPI as mirror? I *guess* you could peek into the logs or peek into it directly to see if it gets used as expected.

JacobHenner:
Looks like all that's left are the additional tests. Thanks for the suggestions @uranusjr and @techalchemy, I'll investigate them soon (hopefully tomorrow).

techalchemy:
That's roughly what i've been attempting to suggest on IRC, but it might not be clear :p

JacobHenner:
Looks like that worked.

```python
def test_mirror_install(PipenvInstance, pypi):
        mirror_url = os.environ['PIPENV_TEST_INDEX']
        del os.environ['PIPENV_TEST_INDEX']
        # This should sufficiently demonstrate the mirror functionality
        # since pypi.org is the default when PIPENV_TEST_INDEX is unset.
        c = p.pipenv('install requests --pypi-mirror {0}'.format(mirror_url))
        assert c.return_code == 0
        # Ensure the --pypi-mirror parameter hasn't altered the Pipfile or Pipfile.lock sources
        assert len(p.pipfile['source']) == 1
        assert len(p.lockfile["_meta"]["sources"]) == 1
        assert 'https://pypi.org/simple' == p.pipfile['source'][0]['url']
        assert 'https://pypi.org/simple' == p.lockfile['_meta']['sources'][0]['url']

        assert 'requests' in p.pipfile['packages']
        assert 'requests' in p.lockfile['default']
        assert 'chardet' in p.lockfile['default']
        assert 'idna' in p.lockfile['default']
        assert 'urllib3' in p.lockfile['default']
        assert 'certifi' in p.lockfile['default']
```

I'll go ahead and use this pattern for the rest of my tests. So far, I intend to add tests for:

1. update
2. uninstall
3. lock
4. sync

This should cover all of the functions that now have the extra pypi_mirror parameter. Not sure if I'll get to this today. If not, I'll have a look tomorrow.

JacobHenner:
I've written tests that should cover installs, uninstalls, locks, and syncs using the --pypi-mirror parameter. Please let me know if you think there are additional tests I should add.

JacobHenner:
@uranusjr @techalchemy I performed additional manual tests in a restrictive environment, with no access to PyPI. All tested operations worked as expected. Looks like this one is ready for review, unless there are additional integration tests you'd like me to include.

pipenv/cli.py
```diff
@@ -302,6 +306,13 @@ def cli(
     callback=validate_python_path,
     help="Specify which version of Python virtualenv should use.",
 )
+@option(
+    '--pypi-mirror',
+    default=PIPENV_PYPI_MIRROR,
+    nargs=1,
+    callback=validate_pypi_mirror,
+    help="Specify a PyPI mirror.",
+)
```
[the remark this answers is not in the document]
JacobHenner: It looks like this one is indeed for the install operation.

pipenv/utils.py
```diff
@@ -277,6 +277,7 @@ class PipCommand(basecommand.Command):
     pypi = PyPIRepository(
         pip_options=pip_options, use_json=True, session=session
     )
+    print(pip_options)
```
uranusjr: stray debug?

pipenv/utils.py
```diff
@@ -442,6 +445,7 @@ def resolve_deps(
             collected_hashes = []
             if any('python.org' in source['url'] or 'pypi.org' in source['url']
                    for source in sources):
+                #FIXME Add support for mirrors.
```
uranusjr: Is this still relevant?

JacobHenner: Nope, this was a note-to-self that was missed in my pre-commit diff grep. Removed.

pipenv/utils.py
JacobHenner: Yep, missed in my pre-commit diff grep. Removed.

pipenv/utils.py
JacobHenner: That being said, it looks like something might have been missed with venv_resolve_deps, I'll take a look now.

JacobHenner: Fixed the missing bit, will push shortly.

JacobHenner:
Removed some accidentally included debugging artifacts, added pypi_mirror support to venv_resolve_deps in get_vcs_deps. Should be good to continue review.

brettdh:
One note: starting from scratch, I had expected the first `pipenv install --pypi-mirror <package>` to set `url = <mirror url>` in the Pipfile's `[[source]]` section, but it didn't. Either I have to manually correct this (even though I've already typed the URL once), or someone else coming along and running `pipenv install` would then unintentionally install the packages from PyPI.

Is there a more ergonomic way to support the use case of initializing a project to use a mirror for all the packages? Or configuring an existing project to use a mirror from now on? I'd expect that this is more common than selectively installing some packages from PyPI and some from a mirror.

uranusjr:
I think the behaviour is intended. The flag is used to override the index during lock and install, not replace it in the declaration. The “right” way to do what you want is to write the mirror into pypirc, or produce your own Pipfile by hand. This really goes back to #2188 IMO—we can’t have an ergonic way to do this unless we have an actual init command.

brettdh:
Ah! Yes, I absolutely agree - a hypothetical `pipenv init` would be the right place for this. Thanks.

brettdh:
Side note: I'm not sure how to do this with `.pypirc` (does pipenv use settings from there?), but editing the Pipfile by hand is fine for now.

pipenv/core.py
```diff
@@ -1131,7 +1138,7 @@ def do_lock(
             lockfile['default'][dep['name']]['markers'] = dep['markers']
     # Add refs for VCS installs.
     # TODO: be smarter about this.
-    _vcs_deps, vcs_lockfiles = get_vcs_deps(project, pip_freeze, which=which, verbose=verbose, clear=clear, pre=pre, allow_global=system, dev=False)
+    _vcs_deps, vcs_lockfiles = get_vcs_deps(project, pip_freeze, which=which, verbose=verbose, clear=clear, pre=pre, allow_global=system, dev=False, pypi_mirror=pypi_mirror)
```
techalchemy: FYI I did realize how ugly and horrible this is, there is a minor refactor PR over in #2315 that will change this a bit

techalchemy:
Merging, thanks for your hard work :cake:

PR #2287, opened 2018-06-01
Cleaner shell history

skurfer:
ZSH has a `HIST_IGNORE_SPACE` option that prevents any command prefixed with a  space from appearing in the history. I’ve added a space to the activation command to take advantage of this.

For most users, it will make no difference, but for ZSH users with that option set, it’ll prevent useless noise in their history.


uranusjr:
Makes sense to me. Bash has something similar too.

uranusjr:
Need admin privilege to merge this one :(

techalchemy:
I'm good with this once we merge the other prs, I'll just update this one and we can verify it passes before merging

Issue #2279, opened 2018-05-30
Passing `-e directory` multiple times attempts to install `-e` as a package


ncoghlan:
`pipenv install -e local/subdir1 -e local/subdir2` fails with an assertion traceback.

Version: 2018.05.18

------------

##### Expected result

That `pipenv install -e local/subdir1 -e local/subdir2` would add both directories to `Pipfile` as editable installs,  similar to the effect of `pip install -e local/subdir1 -e local/subdir2`.

##### Actual result

The second `-e` gets interpreted as a package to try to install, which then fails:

```
[snip successfully setting up the first editable install]
Installing -e…
Traceback (most recent call last):
  File "/home/ncoghlan/.local/bin/pipenv", line 11, in <module>
    sys.exit(cli())
  File "/home/ncoghlan/.local/lib/python3.5/site-packages/pipenv/vendor/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/home/ncoghlan/.local/lib/python3.5/site-packages/pipenv/vendor/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/home/ncoghlan/.local/lib/python3.5/site-packages/pipenv/vendor/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/ncoghlan/.local/lib/python3.5/site-packages/pipenv/vendor/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/ncoghlan/.local/lib/python3.5/site-packages/pipenv/vendor/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/home/ncoghlan/.local/lib/python3.5/site-packages/pipenv/cli.py", line 402, in install
    selective_upgrade=selective_upgrade,
  File "/home/ncoghlan/.local/lib/python3.5/site-packages/pipenv/core.py", line 1978, in do_install
    extra_indexes=extra_indexes,
  File "/home/ncoghlan/.local/lib/python3.5/site-packages/pipenv/core.py", line 1420, in pip_install
    package_name.split('--hash')[0].split('--trusted-host')[0]
  File "/home/ncoghlan/.local/lib/python3.5/site-packages/pipenv/utils.py", line 142, in get_requirement
    req = [r for r in requirements.parse(dep)][0]
[...]

PR #2299, opened 2018-06-03
WIP: Make install editable option an option

mlouielu:
let the -e option show in `pipenv install -h` and to be an option.


uranusjr:
ref #2279

uranusjr:
Something’s wrong :/

techalchemy:
Thanks for the PR!

the `-e ` flag is handled in about 40 different places in the codebase, so simply adding it as a flag is not going to work out here

I wish it were this simple, but unfortunately its' not, if you want to iterate on this I'm happy to point you in the right direction

mlouielu:
@techalchemy @uranusjr yep, I found that it is handled in multiple places in the code... I'll dig it out and iterate this PR.

techalchemy:
No worries and no rush, this one is gonna be a pain :p

techalchemy:
does this belong in a subcommand with the `multiple` flag from click? That way we can have unlimited `editable` things?

uranusjr:
Yes, `multiple=True` should be added. We should probably add a test case for that usage as well.

http://click.pocoo.org/5/options/#multiple-options

techalchemy:
(you shoudl probably link to version 6)

mlouielu:
OK, I'll add the multiple and the test

kennethreitz:
i don't think this is neccessary, but thank you.

kennethreitz:
everything after 'install' is passed to pip. it's a pip feature, not a pipenv feature.

mgeisler:
> everything after 'install' is passed to pip. it's a pip feature, not a pipenv feature.

Oh, that's interesting! I didn't know that `pipenv install` worked like that. Is it documented anywhere?

mgeisler:
I don't think the logic that is supposed to pass command line options to Pip really works as intended. Here I tried to use `--no-index` with `pipenv install`, but it tries to install a package named `--no-index`:
```
[mg]% pipenv install --no-index --find-links . foobar==1.2.3
Installing --no-index...
⠋/nix/store/lw18hsacf1h82yvwg89nn5nsns9hcxij-pipenv-2018.7.1/lib/python3.6/site-packages/pipenv/vendor/requirements/parser.py:44: UserWarning: Private repos not supported. Skipping.
  warnings.warn('Private repos not supported. Skipping.')
  ...
AttributeError: 'NoneType' object has no attribute 'specifier'
```

kennethreitz:
that's not exactly how it works

PR #2300, opened 2018-06-03
Reduce image filesize for documentation

bittner:
I've noticed that the image in the upper left corner of the [Pipenv documentation](https://pipenv.readthedocs.io/) loads very slowly. This is because the file is actually **1.3 MB** of size. Now I've:

1. Scaled down the image to 68% of the original image size
1. Ran the image file through [TinyPNG](https://tinypng.com/) to additionally reduce file size

The result is a file of **420KB** of size and image dimensions still large enough for the near future. (I think it's fair towards users consuming the documentation to make them download a smaller amount of data.)

Thanks for merging!


techalchemy:
don't mind the failures, we have a non-deterministic build in the build process

thanks for the PR, I never noticed this until now.  I'll have to run it by kenneth and make sure he's ok with it...

PR #2315, opened 2018-06-07
Refactor and cleanup some locking code

techalchemy:
- This has been driving me mad.

### DIFF
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -1056,7 +1056,8 @@ def do_lock(
             'dev': False
         }
     }
-    for section_name, settings in sections.items():
+    for section_name in ['dev', 'default']:
+        settings = sections[section_name]
         if write:
             # Alert the user of progress.
             click.echo(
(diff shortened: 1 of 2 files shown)


pipenv/core.py
```diff
@@ -1044,102 +1033,83 @@ def do_lock(
...
+        },
+        'default': {
+            'packages': project.packages,
+            'vcs': project.vcs_packages,
+            'pipfile_key': 'packages',
+            'lockfile_key': 'default',
+            'log_string': 'packages',
+            'dev': False
+        }
+    }
```
uranusjr: This will make the locking order non-deterministic in older Python versions. Probably better to use a list instead. (Or the user could be confused.)

pipenv/core.py
```diff
@@ -1044,102 +1033,83 @@ def do_lock(
...
-        # Add version information to lockfile.
-        pipfile_version = project.packages[dep['name']] if dep['name'] in project.packages else None
-        if pipfile_version and hasattr(pipfile_version, 'keys') and any(k for k in ['file', 'path'] if k in pipfile_version):
-            lockfile['default'].update(
-                {dep['name']: dict(pipfile_version)}
-            )
-            continue
-        lockfile['default'].update(
-            {dep['name']: {'version': '=={0}'.format(dep['version'])}}
+        # Add develop dependencies to lockfile.
```
uranusjr: Stale comment?

pipenv/core.py
techalchemy: fixed

techalchemy:
That resolver failure is so consistent on appveyor now, I wonder what the real cause of it is...

PR #2325, opened 2018-06-07
WIP: Operation pewetomy

uranusjr:
Supercedes #2161. Remove all dependence on Pew, and replace all functionality with our own implementation.

Most are quite straightforward (but still needs testing), only `new` and `shell` requires particular attention. `new` is converted to use virtualenv directly, with a little post-creation patching to make it compatible with Pew (mostly the project association stuff).

`shell` is the coolest part. I rewrote all shell detection as `pipenv.shelltools` (except the NT backend is 99% `_win_utils` from the previous implementation), and re-implemented `pew.workon_cmd`. I’m not that familiar with the compat shell, and largely did copy-pasting job in that part, so more testing are probably required to avoid regression.

We need to manually test the hell out of this before it can land in a stable release.


uranusjr:
Moved part of this to https://github.com/pypa/pipenv/projects/3. Purge can be done later.

Issue #1255, opened 2017-12-31
Goal: ensure lock file consistency after all installation commands


ncoghlan:
As noted in https://github.com/pypa/pipenv/issues/1137, `pipenv` doesn't currently check for requirement conflicts between `[packages]` and `[dev-packages]`, which can cause problems when attempting to generate a flat `requirements.txt` that covers both, as well as causing general weirdness when dependencies conflict.

The way `pipenv install` works (running the installation and then updating the lock file) means it is also relatively easy for the local environment to get out of sync with the lock file: if the installation request covers an already installed package, then it won't be updated locally, but the lock file will be updated to the latest version available from any configured package indices.

This issue doesn't cover an individual feature request. Instead, it establishes the goal of helping to ensure consistency between the lock file and the local environment by structuring installation commands to modify the lock file *first*, and then use the updated lock file to drive the actual installation step. (We don't have any kind of target ETA for how long this work will take - writing it down is just intended to help ensure we're all heading in the same direction, and are all comfortable with that direction)

A key aspect of this is going to be clarifying the division of responsibilities between `pipenv install`, `pipenv uninstall`, `pipenv lock`, and `pipenv update` (and potentially adding one or more new subcommands if deemed necessary).

Proposed substeps:

- [x] add a `pipenv sync` subcommand that ensures that the current environment matches the lock file (akin to `pip-sync` in `pip-tools`). (Implemented by @kennethreitz for 10.0.0, with `pipenv sync` ensuring versions of locked packages match the lock file, while `pipenv clean` removes packages that aren't in the lock file. I like that change relative to `pip-sync`, as it means that including an implicit sync in another command won't unexpectedly delete anything from the virtual environment)
- [x] Keep the current behaviour for `pipenv install` with no arguments, but switch to recommending the use of `pipenv sync` when setting up a fresh environment (that way `pipenv install` is used mainly for *changing* the installed components)
- [x] update `pipenv lock` to always ensure that `[packages]` and `[dev-packages]` are consistent with each other (resolution for #1137)
- [x] change `pipenv update` to be equivalent to `pipenv lock && pipenv sync` (pipenv update is instead being removed entirely - it's old behaviour was actually comparable to what is now `pipenv sync && pipenv clean`)
[...]

PR #2338, opened 2018-06-11
WIP: Investigate missed package upgrade

### LINKED ISSUE #2179: Upgrading package reinstalls old version
> Be sure to check the existing issues (both open and closed!).

Checked #1490 and other issued that mention *update*.

> Describe the issue briefly here.

Upgrading package with `pipenv update PACKAGE` upgrades given PACKAGE but then reinstalls old version.

> Please run `$ python -m pipenv.help`, and paste the results here.

<details><summary>$ python -m pipenv.help output</summary>

Pipenv version: `'11.10.4'`

Pipenv location: `'/usr/local/lib/python3.6/site-packages/pipenv'`

Python location: `'/usr/local/bin/python'`

  - `2.7`: `/usr/bin/python2.7`
  - `2.7`: `/usr/bin/python2.7`
  - `3.6`: `/usr/local/bin/python3.6m`
  - `3.6`: `/usr/local/bin/python3.6`
  - `3.6`: `/usr/local/bin/python3.6`

  - `3.6.5`: `/usr/local/bin/python`
  - `3.6.5`: `/usr/local/bin/python`
  - `2.7.9`: `/usr/bin/python`
  - `2.7.9`: `/usr/bin/python2`
  - `3.6.5`: `/usr/local/bin/python3`
  - `3.6.5`: `/usr/local/bin/python3`

PEP 508 Information:

```
 'implementation_version': '3.6.5',
 'os_name': 'posix',
 'platform_machine': 'x86_64',
 'platform_release': '4.9.87-linuxkit-aufs',
 'platform_system': 'Linux',
 'platform_version': '#1 SMP Wed Mar 14 15:12:16 UTC 2018',
 'python_full_version': '3.6.5',
 'python_version': '3.6',
 'sys_platform': 'linux'}
```

System environment variables:

  - `HOSTNAME`
  - `GPG_KEY`
  - `SHELL`
  - `TERM`
  - `OLDPWD`
  - `PYTHON_VERSION`
  - `PATH`
  - `PWD`
  - `LANG`
  - `PYTHON_PIP_VERSION`
  - `SHLVL`
  - `HOME`
  - `_`
  - `PYTHONDONTWRITEBYTECODE`
  - `PIP_PYTHON_PATH`

  - `PATH`: `/usr/local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin`
  - `SHELL`: `/bin/bash`
  - `LANG`: `C.UTF-8`
  - `PWD`: `/test`

---------------------------

Contents of `Pipfile` ('/test/Pipfile'):

```toml
[[source]]
verify_ssl = true
name = "pypi"

[packages]
requests = "==2.17.0"

[dev-packages]

[requires]
python_version = "3.6"

```

Contents of `Pipfile.lock` ('/test/Pipfile.lock'):

```json
{
    "_meta": {
        "hash": {
            "sha256": "d2e2282e2d97b6e5eb1ea3481d394b64d2598fb4e0c3e7e84f4a477275deacf6"
        },
        "pipfile-spec": 6,
        "requires": {
            "python_version": "3.6"
        },
        "sources": [
            {
                "name": "pypi",
                "verify_ssl": true
            }
        ]
    },
    "default": {
        "certifi": {
            "hashes": [
                "sha256:13e698f54293db9f89122b0581843a782ad0934a4fe0172d2a980ba77fc61bb7",
                "sha256:9fa520c1bacfb634fa7af20a76bcbd3d5fb390481724c597da32c719a7dca4b0"
            ],
            "version": "==2018.4.16"
        },
        "chardet": {
            "hashes": [
                "sha256:84ab92ed1c4d4f16916e05906b6b75a6c0fb5db821cc65e70cbd64a3e2a5eaae",
                "sha256:fc323ffcaeaed0e0a02bf4d117757b98aed530d9ed4531e3e15460124c106691"
            ],
            "version": "==3.0.4"
        },
        "idna": {
            "hashes": [
                "sha256:3cb5ce08046c4e3a560fc02f138d0ac63e00f8ce5901a56b32ec8b7994082aab",
                "sha256:cc19709fd6d0cbfed39ea875d29ba6d4e22c0cebc510a76d6302a28385e8bb70"
            ],
            "version": "==2.5"
        },
        "requests": {
            "hashes": [
                "sha256:73b4088c05f7fb5ca8e68651ed802df3ca40621281acf74bb321b4a8408aab7e",
                "sha256:eff227db5864238d44270cbadc8ac4133e69b69a2e7092b7b316ed1e4761cbd6"
            ],
            "index": "pypi",
            "version": "==2.17.0"
        },
        "urllib3": {
            "hashes": [
                "sha256:8ed6d5c1ff9d6ba84677310060d6a3a78ca3072ce0684cb3c645023009c114b1",
                "sha256:b14486978518ca0901a76ba973d7821047409d7f726f22156b24e83fd71382a5"
            ],
            "version": "==1.21.1"
        }
    },
    "develop": {}
}

```
</details>
------------

##### Expected result

Given I have `requests = "==2.17.0"` in Pipfile and I change it to `requests = "*"`, running `pipenv  update requests` I would expect to have latest version of requests.

##### Actual result

```
(test-BnvVBCnl) root@c4d4399a20d3:/test# pipenv update requests
Installing requests…
Collecting requests
  Using cached https://files.pythonhosted.org/packages/49/df/50aa1999ab9bde74656c2919d9c0c085fd2b3775fd3eca826012bef76d8c/requests-2.18.4-py2.py3-none-any.whl
Requirement not upgraded as not directly required: chardet<3.1.0,>=3.0.2 in /root/.local/share/virtualenvs/test-BnvVBCnl/lib/python3.6/site-packages (from requests) (3.0.4)
Requirement not upgraded as not directly required: certifi>=2017.4.17 in /root/.local/share/virtualenvs/test-BnvVBCnl/lib/python3.6/site-packages (from requests) (2018.4.16)
Requirement not upgraded as not directly required: idna<2.7,>=2.5 in /root/.local/share/virtualenvs/test-BnvVBCnl/lib/python3.6/site-packages (from requests) (2.6)
Requirement not upgraded as not directly required: urllib3<1.23,>=1.21.1 in /root/.local/share/virtualenvs/test-BnvVBCnl/lib/python3.6/site-packages (from requests) (1.22)
Installing collected packages: requests
  Found existing installation: requests 2.17.0
    Uninstalling requests-2.17.0:
      Successfully uninstalled requests-2.17.0
Successfully installed requests-2.18.4

Adding requests to Pipfile's [packages]…
Installing dependencies from Pipfile.lock (b14837)…
  🐍   ▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉ 5/5 — 00:00:01
```

##### Steps to replicate

1. 

```
root@c4d4399a20d3:/test# pipenv install requests==2.17.0
Installing requests==2.17.0…
Collecting requests==2.17.0
  Using cached https://files.pythonhosted.org/packages/5b/b6/9a18db79553524246aa1b081829e6f977667ec558cef684988895c1092d9/requests-2.17.0-py2.py3-none-any.whl
Collecting certifi>=2017.4.17 (from requests==2.17.0)
  Using cached https://files.pythonhosted.org/packages/7c/e6/92ad559b7192d846975fc916b65f667c7b8c3a32bea7372340bfe9a15fa5/certifi-2018.4.16-py2.py3-none-any.whl
Collecting chardet<3.1.0,>=3.0.2 (from requests==2.17.0)
  Using cached https://files.pythonhosted.org/packages/bc/a9/01ffebfb562e4274b6487b4bb1ddec7ca55ec7510b22e4c51f14098443b8/chardet-3.0.4-py2.py3-none-any.whl
Collecting urllib3<1.22,>=1.21.1 (from requests==2.17.0)
  Using cached https://files.pythonhosted.org/packages/24/53/f397db567de0aa0e81b211d81c13c41a779f14893e42189cf5bdb97611b2/urllib3-1.21.1-py2.py3-none-any.whl
Collecting idna<2.6,>=2.5 (from requests==2.17.0)
  Using cached https://files.pythonhosted.org/packages/11/7d/9bbbd7bb35f34b0169542487d2a8859e44306bb2e6a4455d491800a5621f/idna-2.5-py2.py3-none-any.whl
Installing collected packages: certifi, chardet, urllib3, idna, requests
Successfully installed certifi-2018.4.16 chardet-3.0.4 idna-2.5 requests-2.17.0 urllib3-1.21.1

Adding requests==2.17.0 to Pipfile's [packages]…
Updated Pipfile.lock (deacf6)!
Installing dependencies from Pipfile.lock (deacf6)…
  🐍   ▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉ 5/5 — 00:00:01
To activate this project's virtualenv, run the following:
 $ pipenv shell
```

2. Edit Pipfile, replace `requests = "==2.17.0"` with `requests = "*"`
```
(test-BnvVBCnl) root@c4d4399a20d3:/test# more Pipfile
[[source]]
verify_ssl = true
name = "pypi"

[packages]
requests = "*"

[dev-packages]

[requires]
python_version = "3.6"
```

3. Check for outdated packages

```
(test-BnvVBCnl) root@c4d4399a20d3:/test# pipenv update --outdated
Package 'idna' out–of–date: '==2.5' installed, '==2.6' available.
Package 'requests' out–of–date: '==2.17.0' installed, '==2.18.4' available.
Package 'urllib3' out–of–date: '==1.21.1' installed, '==1.22' available.
```

4. Try to update `requests`

```
(test-BnvVBCnl) root@c4d4399a20d3:/test# pipenv update requests
Installing requests…
Collecting requests
  Using cached https://files.pythonhosted.org/packages/49/df/50aa1999ab9bde74656c2919d9c0c085fd2b3775fd3eca826012bef76d8c/requests-2.18.4-py2.py3-none-any.whl
Requirement not upgraded as not directly required: certifi>=2017.4.17 in /root/.local/share/virtualenvs/test-BnvVBCnl/lib/python3.6/site-packages (from requests) (2018.4.16)
Requirement not upgraded as not directly required: chardet<3.1.0,>=3.0.2 in /root/.local/share/virtualenvs/test-BnvVBCnl/lib/python3.6/site-packages (from requests) (3.0.4)
Requirement not upgraded as not directly required: urllib3<1.23,>=1.21.1 in /root/.local/share/virtualenvs/test-BnvVBCnl/lib/python3.6/site-packages (from requests) (1.21.1)
Requirement not upgraded as not directly required: idna<2.7,>=2.5 in /root/.local/share/virtualenvs/test-BnvVBCnl/lib/python3.6/site-packages (from requests) (2.5)
Installing collected packages: requests
  Found existing installation: requests 2.17.0
    Uninstalling requests-2.17.0:
      Successfully uninstalled requests-2.17.0
Successfully installed requests-2.18.4

Adding requests to Pipfile's [packages]…
Installing dependencies from Pipfile.lock (b14837)…
  🐍   ▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉ 5/5 — 
```

5. Check `requests` version

```
(test-BnvVBCnl) root@c4d4399a20d3:/test# pipenv graph
requests==2.17.0
  - certifi [required: >=2017.4.17, installed: 2018.4.16]
  - chardet [required: <3.1.0,>=3.0.2, installed: 3.0.4]
  - idna [required: >=2.5,<2.6, installed: 2.6]
  - urllib3 [required: >=1.21.1,<1.22, installed: 1.22]
```

I found that if I run `pipenv lock` before step 4, `Pipfile.lock` would be updated and everything would work as expected for me. [Upgrade workflow](https://docs.pipenv.org/basics/#example-pipenv-upgrade-workflow) does not mention need to lock before update.

Please note that in my real project I have `"*"` version specified but I wouldn't be able to make reproducable case as it happend when new version of package is released.

uranusjr:
Try to fix #2179.

This currently just adds a failing test to reproduce the problem. The fix will come later? I have no idea where to look actually.


uranusjr:
Some updates on investigation. So `update <package>` in (implementation-wise) the same as `install <package> --selective-upgrade`, which would set `keep_outdated` to `True`. Since the package is already installed in this case, the locking mechanism would elect to keep the old version, instead of fetching the latest. So either the behaviour is correct, or we should tweak the implementation to do different things in `update` from `install --selective-upgrade`.

I’m just in favour more and more to cut the command line interface. So many of the commands are ill-defined, and we don’t even know what is the correct thing to do in each context.

techalchemy:
`selective-upgrade` is meant to resolve and upgrade all packages related to a single dependency, rather than the normal behavior which updates everything. 

`keep-outdated` is meant to never ever touch anything that is in the lockfile no matter what. We use them interchangeably everywhere right now which is a big problem. 

Does this come close in your mind @ncoghlan?

ncoghlan:
Yeah, I think you're right, as the problem I see in #2179 is that if you edit `Pipfile` directly to replace an old constraint with a wildcard constraint, then a plain `pipenv lock --keep-outdated` will see the previously pinned version as still satisfying the wildcard constraint, so it won't update it.

By contrast, the intended behaviour for `pipenv install --selective-upgrade <component>` is that "<component>" and its dependencies will all be exempted from the version pin, such that they *do* get upgraded to the latest version, even if an older version was previously locked.

So I think this may actually be an error in my original `pipenv lock --keep-outdated` design (from https://github.com/pypa/pipenv/issues/1255): without a mechanism to either explicitly ignore the existing lock file entries for particular components, or else to update the lock file pins from the current environment (i.e. running `pipenv sync` backwards), the final `pipenv lock --keep-outdated` in a selective upgrade ends up being *too* conservative, and hence may end up rolling back a change that the installer just made.

uranusjr:
I feel a more sophisticated version resolver would be required to actually solve this proble, With the current piptools-backed solution there is only so much we can do.

One workaround I can think of is to “trick” piptools into thinking it needs to update. After the initial package upgrade (requests to 2.18.4), we can monky-patch the Pipfile entry from `requests = "*"` to `requests = "2.18.4"`, and feed this into piptools for locking. This would satisfy the “only-if-needed” criteria. Would this be resonable, or are there edge cases I didn’t notice?

techalchemy:
I'm not really sure I agree with the premise @ncoghlan -- in looking at the code and thinking about these two commands, I actually think they both *should* exist, and that `keep_outdated` would be *correct* in not locking a newer version of `<package x>` just because you put an unpinned line in your Pipfile.  

The process of locking is confined only to the state of the currently locked environment.  Just because you said somewhere else 'oh and by the way, if you ever get around to it, I don't mind unpinning this', that doesn't mean we should always attempt to upgrade it -- especially not when the user is *explicitly telling us* to keep all the outdated things.  If anything I'd think it's the most correct behavior in that case to completely ignore outdated libraries.

Here is the distinction in my mind:

- `--keep-outdated` -- Used only when installing a new (previously not present -- either via the command line or by adding to the pipfile something that is not yet in the environment) or installing or updating (via the command line only) an existing dependency.  **Never touches any installed dependency besides those explicitly specified**, but can update and install new versions.

- `--selective-upgrade` -- Passes a single dependency to the resolver, gets back the list of dependencies that get installed with it.  Installs/upgrades only that list of dependencies.

` Default behavior -- Install/update re-locks the pipfile with latest versions, confined to the constraints specified in the pipfile

ncoghlan:
@techalchemy I think we're actually agreeing, but I didn't express myself very well.

The bit that's missing currently is a way for the selective upgrade operation to tell the subsequent lock operation that the previous lock should be ignored for the just upgraded components.

techalchemy:
@ncoghlan I have been sneaking in cleanup PRs to help get us there but it’s more that we don’t have functionality to lock just one package. We have access to the temporary state of the prior lock if there was one — in a selective upgrade, my naive approach was to just merge in the resolved state for a single dependency (which would be a sub-lockfile I guess) into the cached lockfile and just have that be the new one. 

This all gets easier to manage with pipfile and lockfile support going directly into @sarugaku/requirementslib which we just did a 1.0 of. After this release of pipenv we are planning a serious code cleanup

techalchemy:
That was what I pieced together when I was looking at it the other day anyway. Does that seem right? Wrong? What am I missing?

ncoghlan:
@techalchemy Calculating the dependency tree for the to-be-upgraded component and then merging it with the existing lock file is the approach that makes sense to me.

So if I'm understanding the status quo correctly, I think that would leave me with two questions:

- when calculating the fresh dependency tree for a selectively upgraded package, does the end result match `pip`'s `only-if-needed` upgrade strategy, or does it match the `eager-upgrade` strategy? While the latter would still be better than eagerly upgrading everything, the former is more what I had in mind when I proposed the `--selective-upgrade` idea (I've changed jobs since then though, and am now in a situation where I'm more likely to do batch updates, which is why I haven't really been tracking the feature's evolution over time)
- regardless of exactly how selective upgrade works, does it make sense to expose a way to request the same "override and/or unpin these versions" logic as part of the `pipenv lock` API? At the moment, `--keep-outdated` isn't flexible enough to let `pipenv install --selective-upgrade ...` be expressed as an equivalent `pipenv lock --keep-outdated --some-other-override-args... && pipenv sync` operation

uranusjr:
I’m likely missing something here, but amending the tree after it is produced sounds a little off to me. What happens if there are conflicting dependencies?

Say we have this (theoratical) depdency tree:

```
A==v1 <- B<v2
C==v1 <- B<v2
```

Initial `pipenv lock` produced a lock file of `A==v1`, `B==v1`, and `C==v1`. Now `A` has a new version `v2`, but it depends on `B>=v2`, that `C==v1` does not support. Under a normal lock sequence, Pipenv emits an error because it cannot resolve.

The above approach for `pipenv update A`, if I understand it correctly, would instead

1. Resolve the versions using the current strategy, getting the same result as before—`A==v1`, `B==v1`, and `C==v1`.
2. Calculate dependency for the updated A (`Av2`), getting `A==v2` and `B==v2`.
3. Overwrite results from 1. with the result of 2., getting `A==v2`, `B==v2`, `C==v1`. C is now broken.

I guess this problem is conceptually the same as Nick’s first question—how would the dependencies be calculated in step 2? To do it correctly, the resolver needs to consider dependencies from the existing tree. This is like `only-if-needed`, but instead of upgrade when needed, it needs to perform extra checks to ensure the upgrade does not break things. But this would mean we need to perform something like a three-way merge in the locking mechanism, and I’m not sure if this is possible with our current resolver’s implementation…

techalchemy:
I’ll answer more completely later but pip added a secret undocumented resolution approach called `to-satisfy-only`

techalchemy:
btw I unskipped this test and also fixed the thing that made it fail AFAIK so it should pass

PR #2342, opened 2018-06-11
Fix VCS Resolution

### LINKED ISSUE #2158: Git dependency with sepcified branch gets overriden by dependency od other package.
This issue is simillar to #1779, but the solution stated there is not working for me.

I need to install specific version of a `PyInstaller` package, but `PyInstaller` is also a dependency of `PyUpdater` which is also required in my project and overrides other my specific `PyInstaller` version.

<details><summary>`$ python -m pipenv.help` output:</summary>

`python -m pipenv.help`

`C:\Program Files (x86)\Python36-32\python.exe: No module named pipenv.help`

I am aware this is not how this output should look like, but I can not make it work. Any help here? I'm on pipenv version 11.0.2.
</details>

------------

##### Expected result

I need latest `develop` branch of PyInstaller installed and I don't want PyUpdater to override this with the `PyInstaller-3.3.1` version. As suggested in #1779 I placed PyUpdater above the PyInstaller in `[packages]` in Pipfile.

##### Actual result

Regardless whether the important part of Pipfile looks like this:
```
PyUpdater = "*"
PyInstaller = {git = "https://github.com/pyinstaller/pyinstaller", ref = "develop"}
```
or like this:
```
PyInstaller = {git = "https://github.com/pyinstaller/pyinstaller", ref = "develop"}
PyUpdater = "*"
```
the PyInstaller version in Pipfile.lock will be `3.3.1` and I end up with this version installed.

What is interesting is that if I uninstall PyInstaller with `pipenv uninstall pyinstaller` and then install it again with `pipenv install https://github.com/pyinstaller/pyinstaller/archive/develop.zip` I get the `Successfully installed PyInstaller-3.4.dev0+b31794cf1` but in fact the `3.3.1` version in installed instead.

##### Steps to replicate
```
# Pipfile
[[source]]

verify_ssl = true
name = "pypi"

[packages]
PyUpdater = "*"
PyInstaller = {git = "https://github.com/pyinstaller/pyinstaller", ref = "develop"}
```
`pipenv install`
Check version: `pipenv run pyinstaller --version`

or

Create new clean environment
`pipenv install pyupdater`
`pipenv uninstall pyinstaller`
`pipenv install https://github.com/pyinstaller/pyinstaller/archive/develop.zip`
Check version: `pipenv run pyinstaller --version`

techalchemy:
- Iterate over VCS dependencies from pipfile instead of iterating over those found in `pip freeze` output
- Only pass editable dependencies to pip-tools for resolution
- Normalize names and ensure that we update lockfile entries accordingly
- Fixes #2158


pipenv/core.py
```diff
@@ -1115,10 +1115,20 @@ def do_lock(
             normalized = pep423_name(dep['name'])
             if not hasattr(dep, 'keys') or not hasattr(dep['name'], 'keys'):
                 continue
-            is_top_level = dep['name'] in vcs_lockfile or normalized in vcs_lockfile
-            lockfile_key = next(k for k in [dep['name'], normalized] if k in vcs_lockfile)
-            lockfile_entry = vcs_lockfile[lockfile_key] if is_top_level else None
-            dep_lockfile = clean_resolved_dep(dep, is_top_level=is_top_level, pipfile_entry=lockfile_entry)
+            is_top_level = (
+                dep['name'] in vcs_lockfile or
```
techalchemy: I should mention this naming is wrong now, `vcs_lockfile` comes from code that doesn't do any resolving anymore.  This is the only place in our codebase that passes deps to the resolver. This is technically `vcs_pipfile` and should be renamed -- that's what makes something a top-level entry... presence in the pipfile.... That's partly the value of this pr

PR #2363, opened 2018-06-15
Upgrade Click-Completion and apply Shellingham for shell detection

uranusjr:
Fix #2352. I want to do this without manually patching Click-Completion if possible.

https://github.com/pypa/pipenv/projects/3


uranusjr:
This introduces an additional environment variable `PIPENV_SHELL`. If set, shell detection is skipped completely. The old `SHELL` introspection is kept as a fallback if detection fails.

A run down of how shell detection works now:

1. If `PIPENV_SHELL` is set, use it.
2. Detect surrounding shell…
    1. (Only on POSIX) The surrounding shell is the login shell. Use the value of `SHELL`.
    2. Use the detected shell executable.
3. Detection failed. Check if `SHELL` is set; if so, use it.
4. Prompt the user to set `PIPENV_SHELL` explicitly.

techalchemy:
looks good to me

frostming:
Does this PR work under CPython only?(see the 'ctypes' imports in Shellingham)
I guess the classifier should be removed from `setup.py`. In fact PyPy is not tested at all.

techalchemy:
We rely heavily on ctypes on window and have done so for some time. We use this as a means of shell detection for spawning subshells. This isn’t new, although I’m guessing the complete lack of issues on the topic is a good indicator that things are ok

uranusjr:
ctypes does exist on PyPy. It is a wrapper of CFFI, and is slow, but works.

PR #2366, opened 2018-06-15
Adds a way to adds new index from the cli

mlhamel:
__What__

In the same way as we we can add package to the `Pipfile` from the command line that would be convenient to be able to add a new source index.

__How__

This PR is proposing to add a new option called `index` to the command line to let people adds a new index from the cli interface.

It would be used like that:
```bash
$ pipenv index <name> <url>
```

By default if ever the `<name>` or the `<url>` already exists we just considered that the index already existed and just do nothing.

__Improvement__

We could possibly improve the clarity of the command by adding a verb to this new option: something like `add-index` would make sense.

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -1061,6 +1061,60 @@ def clean(
     )
 
 
+@command(
+    short_help="Adds a new index of packages in Pipfile."
+)
+@argument('url', default=False)
+@argument('name', default="")
+@option(
+    '--three/--two',
+    is_flag=True,
+    default=None,
+    help="Use Python 3/2 when creating virtualenv.",
+)
+@option(
+    '--python',
+    default=False,
+    nargs=1,
+    callback=validate_python_path,
+    help="Specify which version of Python virtualenv should use.",
+)
+@option(
+    '--dry-run',
+    is_flag=True,
+    default=False,
+    help="Just output unneeded packages.",
+)
+@option(
+    '--pypi-mirror',
+    default=PIPENV_PYPI_MIRROR,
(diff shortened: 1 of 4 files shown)


techalchemy:
Hi there, thanks for taking the time to put this together!  Typically it is a highly recommended route of development to solicit feedback early on in the process, especially if you are looking to add something to the API.  New API options are things that can only really be added, and _never_ really can be removed, so even if they seem intuitive they typically receive a lot of scrutiny.

As it stands right now, I'm not really sure how widely used a command line argument for this would be -- and I do think we would probably want this as an argument rather than its own top level command group.  You probably noticed that we have a small number of top level commands -- that didn't happen by accident,.  They represent the core subset of functionalities that users need to be exposed to on a daily basis... `run, lock, update, install, shell`, none of these is a low level entrypoint for manipulating pipfile level metadata... they provide a direct action to undertake.

Essentially, what is the use case for someone sitting at the command line to simply start typing commands that update index urls in a pipfile? It's a much more likely scenario for them to open their pipfile in an editor or to make that update as part of an action. 

You mentioned that you are looking to do this the same way that users can add packages to their pipfile, but actually there is no command to do that without also installing the package.  See what I'm getting at?  We can talk more about what it is you are looking to accomplish in this PR but I'm not yet convinced that we need something like this. I'd recommend opening an issue to discuss further and if we determine it's a good feature that we want to adopt, we can loop back around and reopen the PR 

In the meantime, I'm going to close this out for the moment -- and again, thanks for putting in the time and we would be more than happy to discuss further in an issue

Issue #876, opened 2017-10-11
up/down arrow history doesn't work on pipenv shell on windows 


crioux:
In the cmd shell, usually you can hit 'up arrow' to get your previous command. Once you run 'pipenv shell', this no longer works. if you run some commands inside the pipenv shell and then hit 'up' it does nothing.

crioux:
FWIW, it works in 'ConEmu', just not the normal 'Windows Console Host' process that you get by running cmd.exe directly, so whatever they're doing in there to pass along the arrow keys is what you'd need. Not sure it's worth it to implement seperately in pipenv, but at least this is some kind of a workaround.

erinxocon:
This is an issue with `pew` since it spawns the subshell on windows.  Can you open an issue over there?  I'm a maintainer of that project too and I'll make sure to address it on that project! 
 https://github.com/berdario/pew/issues

techalchemy:
(FYI in the next release -- which will be today -- we are no longer using pew for managing subshells).

W7PEA:
I tried installing from the repo and after I did pipenv shell, command history still didn't work. This is how I installed:

```
pip install git+https://github.com/pypa/pipenv
Collecting git+https://github.com/pypa/pipenv
  Cloning https://github.com/pypa/pipenv to c:\<snip>\appdata\local\temp\pip-req-build-yv19jpwg
  Installing build dependencies ... done
Requirement already satisfied: pip>=9.0.1 in c:\<snip>\python36\lib\site-packages (from pipenv==2018.7.1.dev0) (10.0.1)
Requirement already satisfied: certifi in c:\<snip>\python36\lib\site-packages (from pipenv==2018.7.1.dev0) (2018.4.16)
Requirement already satisfied: setuptools>=36.2.1 in c:\<snip>\python36\lib\site-packages (from pipenv==2018.7.1.dev0) (39.2.0)
Requirement already satisfied: virtualenv-clone>=0.2.5 in c:\<snip>\python36\lib\site-packages (from pipenv==2018.7.1.dev0) (0.3.0)
Requirement already satisfied: virtualenv in c:\<snip>\python36\lib\site-packages (from pipenv==2018.7.1.dev0) (16.0.0)
Installing collected packages: pipenv
  Running setup.py install for pipenv ... done
Successfully installed pipenv-2018.7.1.dev0
```

I tried it with an existing virtualenv and a newly recreated one. Let me know if I should try another branch. 
Thanks!

[...]

Issue #1762, opened 2018-03-16
Pipenv provides incorrect completion if current shell does not match login shell


QasimK:
It _seems_ as though Pipenv does not correctly detect the fish shell when determining what to output for `pipenv --completion`. This is a regression that appeared within the last couple of weeks for me.

<details><summary>$ python -m pipenv.help output</summary>

Pipenv version: `'11.8.0'`

Pipenv location: `'/home/qasim/.local/lib/python3.6/site-packages/pipenv'`

Python location: `'/usr/bin/python'`

Other Python installations in `PATH`:

  - `2.7`: `/usr/bin/python2.7`
  - `2.7`: `/usr/bin/python2.7`
  - `3.6`: `/usr/bin/python3.6m`
  - `3.6`: `/usr/bin/python3.6`

  - `3.6.4`: `/usr/bin/python`
  - `2.7.14`: `/usr/bin/python2`
  - `3.6.4`: `/usr/bin/python3`

PEP 508 Information:

```
{'implementation_name': 'cpython',
 'implementation_version': '3.6.4',
 'os_name': 'posix',
 'platform_machine': 'x86_64',
 'platform_python_implementation': 'CPython',
 'platform_release': '4.15.9-1-ARCH',
 'platform_system': 'Linux',
 'platform_version': '#1 SMP PREEMPT Sun Mar 11 17:54:33 UTC 2018',
 'python_full_version': '3.6.4',
 'python_version': '3.6',
 'sys_platform': 'linux'}
```

System environment variables:

  - `COLORTERM`
  - `DBUS_SESSION_BUS_ADDRESS`
  - `DESKTOP_AUTOSTART_ID`
  - `DESKTOP_SESSION`
  - `DISPLAY`
  - `GDMSESSION`
  - `GDM_LANG`
  - `GNOME_DESKTOP_SESSION_ID`
  - `HOME`
  - `LANG`
  - `LOGNAME`
  - `MAIL`
  - `MOZ_PLUGIN_PATH`
  - `PATH`
  - `PWD`
  - `SESSION_MANAGER`
  - `SHELL`
  - `SHLVL`
  - `SSH_AUTH_SOCK`
  - `TERM`
  - `USER`
  - `USERNAME`
  - `VTE_VERSION`
  - `WINDOWID`
  - `WINDOWPATH`
  - `XAUTHORITY`
  - `XDG_CURRENT_DESKTOP`
  - `XDG_MENU_PREFIX`
[...]

PR #2371, opened 2018-06-16
Implement our own subshell logic

### LINKED ISSUE #915: pipenv shell starts a CMD shell when run from powershell

uranusjr:
Remove all `pew workon` usages, and implement the subshell functionality from scratch based on Shellingham.

With real shell detection on POSIX implemented, the subshell now defaults to match parent shell, instead of login shell. This would help reduce surprising behaviour such as #1762.

I tested the code on Windows and Ubuntu (WSL) in both fancy and compat mode, and am fairly confident this implementation is reasonably complete. It would probably still be a good idea to rigorously test this manually on various platforms before releasing it to the public.

This also fixes #915 (a bug in Pew).


uranusjr:
I think this is ready. This can only get more manual testing if merged into master.

techalchemy:
merging!

mungojam:
I'm surprised that this doesn't seem to fix #876 since that was a bug with pew. Is there a different bug with Shellingham?

uranusjr:
Unfortunately Shellingham does many things the same as Pew. It fixes a lot of things, but not all of them. But it’d be possible to work on it now, at least.

PR #2384, opened 2018-06-19
Exclude failing markers

techalchemy:
- Dependencies with markers that don't match the current environment will now be skipped during `pipenv lock`


techalchemy:
Apparently I snuck in a script update for invoke and a changelog update

pipenv/core.py
```diff
@@ -103,7 +101,7 @@
     ):
         INSTALL_LABEL = '🎅   '
     else:
-        INSTALL_LABEL = '🐍   '
+        INSTALL_LABEL = '🝝   '
```
uranusjr: wait

tests/integration/test_sync.py
```diff
@@ -22,7 +22,7 @@ def test_sync_error_without_lockfile(PipenvInstance, pypi):
 @pytest.mark.lock
 def test_mirror_lock_sync(PipenvInstance, pypi):
     with temp_environ(), PipenvInstance(chdir=True) as p:
-        mirror_url = os.environ.pop('PIPENV_TEST_INDEX', "https://pypi.python.org/simple")
+        mirror_url = os.environ.pop('PIPENV_TEST_INDEX', "https://pypi.kennethreitz..org/simple")
```
uranusjr: Extra dot?

techalchemy: Oops

PR #2385, opened 2018-06-19
Use shell=False when "run" if possible on Windows

uranusjr:
Do a `where` on the command; if it is found, prevent the intermediate `COMSPEC` and use `CreateProcess` directly. Only use `shell=True` if the command is not an executable.

This prevents some unexpected behaviour caused by the intermediate shell process.


uranusjr:
Need #2386 first for this to work.

techalchemy:
Hm yeah this makes sense to me.

techalchemy:
Tested this out locally on windows and it seems good

PR #2386, opened 2018-06-19
Ensure environs are strings on Python 2 + Windows

uranusjr:
Patch based on theskumar/python-dotenv#101 by @greyli.

### DIFF
--- a/pipenv/vendor/dotenv/main.py
+++ b/pipenv/vendor/dotenv/main.py
@@ -94,6 +94,13 @@ class DotEnv():
         for k, v in self.dict().items():
             if k in os.environ and not override:
                 continue
+            # With Python 2 on Windows, ensuree environment variables are
+            # system strings to avoid "TypeError: environment can only contain
+            # strings" in Python's subprocess module.
+            if sys.version_info.major < 3 and sys.platform == 'win32':
+                from pipenv.utils import fs_str
+                k = fs_str(k)
+                v = fs_str(v)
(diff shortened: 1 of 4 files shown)


pipenv/vendor/dotenv/main.py
```diff
@@ -94,6 +94,14 @@ def set_as_environment_variables(self, override=False):
         for k, v in self.dict().items():
             if k in os.environ and not override:
                 continue
+            # With Python2 on Windows, force environment variables to str to
+            # avoid "TypeError: environment can only contain strings" in
+            # Python's subprocess module.
+            if sys.version_info.major < 3 and sys.platform == 'win32':
```
techalchemy: we can use `pipenv.utils.fs_str()`

pipenv/vendor/dotenv/main.py
```diff
@@ -94,6 +94,13 @@ def set_as_environment_variables(self, override=False):
         for k, v in self.dict().items():
             if k in os.environ and not override:
                 continue
+            # With Python 2 on Windows, ensuree environment variables are
+            # system strings to avoid "TypeError: environment can only contain
+            # strings" in Python's subprocess module.
+            if sys.version_info.major < 3 and sys.platform == 'win32':
+                from pipenv.utils import fs_str
```
techalchemy: Is there any reason not to just do this all the time? Why not just always use `fs_str` here?

Issue #439, opened 2017-07-18
pipenv fails when inside a directory starting with a dash


knkski:
The real issue is filed [here](https://github.com/berdario/pew/issues/143), but would it make sense to add `-` to the list of characters to sanitize [here](https://github.com/kennethreitz/pipenv/blob/master/pipenv/project.py#L66)?

A test case:

```
$ git clone http://github.com/foo/bar.git -- "-deleteme"
Cloning into '-deleteme'...
remote: Counting objects: 1176, done.
remote: Compressing objects: 100% (61/61), done.
remote: Total 1176 (delta 121), reused 131 (delta 104), pack-reused 1001
Receiving objects: 100% (1176/1176), 3.81 MiB | 1.42 MiB/s, done.
Resolving deltas: 100% (473/473), done.
$ cd -deleteme/
$ pipenv --three
Creating a virtualenv for this project...
⠋usage: pew [-h] [-p PYTHON] [-i PACKAGES] [-r REQUIREMENTS] [-d] [-a PROJECT]
           envname
pew: error: argument -d/--dont-activate: ignored explicit argument 'eleteme-srGrfIGO'

Virtualenv location: 
```

nateprewitt:
So I don't think there are any technical reasons we can't add `-` to that list, but that's a breaking change and is likely to irritate a lot of users since `-` is a pretty common character.

Before we start making decisions on how to address this though, I just want to make sure I understand the situation. From what I gathered in your issue on the `pew` repository, Jenkins is stripping path names to meet an 80 character max. Is that correct? Have you attempted to use this [suggested solution](https://issues.jenkins-ci.org/browse/JENKINS-38706?focusedCommentId=274531&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-274531) for Jenkins?

knkski:
You are correct on what the issue is, and I did use that solution to successfully fix it for my situation. It's possible that `pipenv` would be used in a situation where it's not possible to use that solution, such as a hosted Jenkins instance, but that is probably too much of an edge case to consider making a breaking change in `pipenv` for, particularly since the real issue lies in `pew`.

kennethreitz:
closing as 'won't fix'

PR #2399, opened 2018-06-21
Remove leading dashes in venv names

dnoetzel:
This is designed to fix issue #439 without causing a breaking change; it does so by removing only leading slashes, instead of prohibiting slashes anywhere in the virtualenv name, as the original issue proposed.

I slightly changed how `Project._sanitize()` works, but it's not the only potential solution. Another option would be to change how `do_create_virtualenv()` calls `pew`; a double dash could be used to separate options from positional arguments and then even a virtualenv name that started with a dash would work. That change would look something like this:

```diff
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -914,7 +914,6 @@ def do_create_virtualenv(python=None, site_packages=False):
             '-m',
             'pipenv.pew',
             'new',
-            project.virtualenv_name,
             '-d',
             '-a',
             project.project_directory,
@@ -932,6 +931,10 @@ def do_create_virtualenv(python=None, site_packages=False):
         err=True,
     )
     cmd = cmd + ['-p', python]
+
+    if not project.is_venv_in_project():
+        cmd = cmd + ['--', project.virtualenv_name]
+
     # Actually create the virtualenv.
     with spinner():
         try:
```

However, it seems to me that changing the name sanitizing method is a better choice for the following reasons:

1. It's easier and faster to test
Checking a string in a unit test is a fast and easily understood operation. While admittedly the context for the test could be lost if its broader purpose is not made clear, that seems like a decent tradeoff when considering the other option. Changing how `pew` is invoked in `pipenv` necessitates a slow integration test that relies on calling external code. Furthermore, it would require either extensive changes to the `_PipenvInstance` and `TemporaryDirectory` classes to allow specifying a temp dir with a defined name (think `/tmp/-dir-with-leading-dash`) or a lot of messy one-off setup and teardown logic.
2. It potentially covers other edge cases
Should `virtualenv_name` be passed to other external commands, options and positional args would not need to be separated by the double dash. Given that a bug of this nature would be stubborn and hard to find, this PR has the potential save a lot of time by preventing other issues before they are introduced.

### DIFF
--- a/pipenv/project.py
+++ b/pipenv/project.py
@@ -260,11 +260,15 @@ class Project(object):
         #      127 : BINPRM_BUF_SIZE - 1
         #       32 : Maximum length of username
         #
+        # Also remove leading dashes to avoid names being interpreted as
+        # arguments to external commands.
+        #
         # References:
         #   https://www.gnu.org/software/bash/manual/html_node/Double-Quotes.html
         #   http://www.tldp.org/LDP/abs/html/special-chars.html#FIELDREF
         #   https://github.com/torvalds/linux/blob/2bfe01ef/include/uapi/linux/binfmts.h#L18
-        return re.sub(r'[ $`!*@"\\\r\n\t]', '_', name)[0:42]
+        sanitized_name = re.sub(r'[ $`!*@"\\\r\n\t]', '_', name).lstrip('-')
+        return sanitized_name[0:42]
 
     def _get_virtualenv_hash(self, name):
         """Get the name of the virtualenv adjusted for windows if needed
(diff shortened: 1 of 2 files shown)


techalchemy:
That was what I was going to suggest as well. We can just call the pew function I would imagine. I was going to suggest it but I was neck deep in that other PR

Also thanks for putting this together and especially for taking the time to include a failing test. That makes our lives so much easier for exploring options and tossing ideas around and concretely finding out if they work. So just want to let you know, we appreciate the help!

dnoetzel:
Closing in favor of #2415.

PR #2402, opened 2018-06-23
update sphinx-click

frostming:
This simple PR makes readthedocs use the new version of sphinx-click to enable command descriptions in doc.


techalchemy:
Thanks! will merge in the next wave of merges

PR #2403, opened 2018-06-24
Cleanup 16.0

uranusjr:
Follow-up to #2401.

@techalchemy There’s an update to requirementslib. Maybe need to cut another release there?


techalchemy:
Thanks for sorting this out

PR #2408, opened 2018-06-25
Allow check to ignore packages 

erinxocon:
and append it to the call to safety

### DIFF
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -2410,8 +2410,8 @@ def do_check(three=None, python=False, system=False, unused=False, args=None):
     else:
         python = system_which('python')
     try:
-        if '--safety_ignore' in args[0]:
-            ignore = '{0} {1}'.format('--ignore', args[1])
+        if '--safety-ignore' in args[0]:
+            ignore = '-i ' + ' -i'.join(args[1:])
         else:
             ignore = ''
     except IndexError:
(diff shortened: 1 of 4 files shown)


erinxocon:
needs test

pipenv/core.py
```diff
@@ -2409,9 +2409,14 @@ def do_check(three=None, python=False, system=False, unused=False, args=None):
         python = which('python')
     else:
         python = system_which('python')
+
+    try:
+        ignore = ' '.join(args)
```
jarshwah: Using `ignore` here is confusing, because there could be other args. Consider using `safety_args` instead?

erinxocon: @techalchemy?

erinxocon:
@jarshwah I renamed the parameter option to `--safety_ignore [package_name]`.

jarshwah:
@erinxocon it won't allow callers to ignore multiple issues. Can you make safety_ignore args a list, and then explode to multiple `-i` args internally?

Something like this perhaps?

`'-i '.join(vid for vid in safety_ignore_list)`

See: https://github.com/pyupio/safety#--ignore--i

erinxocon:
ahh ok, I wasn't familiar with the api for safety.  Hold please

erinxocon:
Ok @jarshwah, I've updated the syntax to `pipenv check --safety-ignore [package1] [package2] [package3]...`.  I don't know if we need to make a new click option for this.  I'll pin @kennethreitz regarding an api change like that.

jarshwah:
Turn around time on this one was super fast, thanks so much!

techalchemy:
Swapped it over to `--ignore`  and added some tests + news entry, going to merge this before release

erinxocon:
thanks @techalchemy, I should have created the news.  Sorry for leaving print in there, I was debugging.

PR #2415, opened 2018-06-25
Allow virtual env creation in dir with leading dash

### LINKED ISSUE #439: pipenv fails when inside a directory starting with a dash

dnoetzel:
Following feedback from PR #2399, I decided to close that PR and go in a new direction. Instead of sanitizing the virtualenv name, I fixed how `pew` is called. However, I chose to keep the command-line invocation and not use `pew`'s `mkvirtualenv()` function. This was done to keep the codepaths similar, since `virtualenv` needs to continue to be called from the command line. (`virtualenv` does not support supplying a different version of Python in `create_environment()`; it [opens a subprocess](https://github.com/pypa/virtualenv/blob/515503c933f4d71df34923e02cd77b2b14744ebf/virtualenv.py#L670) to achieve that.)

The test is also a much better representation of the problem, albeit more complex.

Fixes #439


techalchemy:
Thanks for the PR!

PR #2417, opened 2018-06-26
Fix bug with `--system --deploy` overwriting pipfile

### LINKED ISSUE #2414: Git submodules in docker
`pipenv, version 11.10.0`

I have a submodule `./dep1` in my git repo and am adding it to pipenv via `pipenv install -e dep1`, which works just fine and adds the line
```
"abc123" = {path = "./dep1", editable = true}
```
to my Pipfile. Then I use a Dockerfile as follows:
```
FROM kennethreitz/pipenv
COPY ./app /app
COPY ./dep1 /app/dep1
CMD python3 start.py
```
where I added the third line, so the submodule is available in the WORKDIR. However, upon `docker build` I am getting the following error 
```
+ pipenv install --deploy --system
Installing -e ./dep1…

Error:  An error occurred while installing -e ./dep1!
./dep1 should either be a path to a local project or a VCS url beginning with svn+, git+, hg+, or bzr+
```
This is unexpected, since ./dep1 *is* a path to a local project, relative to the WORKDIR in the docker image with the Pipfile, which is /app. Am I missing something or is this a bug?

techalchemy:
- Fixes #2414
- Also bump to 2018.6.26.dev0


erinxocon:
I can confirm that this branch fixes the issues raised in #2414

PR #2422, opened 2018-06-26
Install Jinja2 templates for click_completion

wjv:
The new version of click_completion requires its Jinja2 templates to be installed along with the package in order to work. If not, any invocation of `pipenv --completion` results in a `jinja2.exceptions.TemplateNotFound` for the template in question.

I’ve just trivially added the pattern to `setup.py`, and tested that it works.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -116,6 +116,7 @@ setup(
         '': ['LICENSE', 'NOTICES'],
         "pipenv.vendor.requests": ["*.pem"],
         "pipenv.vendor.certifi": ["*.pem"],
+        "pipenv.vendor.click_completion": ["*.j2"],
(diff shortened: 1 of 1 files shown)


uranusjr:
@techalchemy We’ll need another release very soon 🤦‍♂️

techalchemy:
Today is the plan

ilovezfs:
Is this sufficient to make sure they end up in the sdist too?

uranusjr:
@ilovezfs It should, I think.

techalchemy:
We didn’t actually release an sdist yet but I would believe that our manifest doesn’t include these. Feel free to PR a fix and thanks for the heads up!

Issue #1901, opened 2018-04-03
Pipfile python_version not taken into account when generating lockfile


rainyday:
Pipenv lock/install/sync generates a Pipfile.lock with package versions incompatible with the version of Python

output of `python -m pipenv.help`: https://pastebin.com/raw/r4WCG5AR

------------

##### Expected result

Pipenv generates a lockfile with installable packages and then installs them.

##### Actual result

Pipenv generates a Pipfile.lock containing requirements that are incompatible with the version of Python specified in Pipfile using `python_version` or `python_full_version` and installation fails. Example:

Error message:

```
Collecting pylint-quotes==0.1.9 (from -r /var/folders/rj/210lgzrd4q5ccsyby0rdxg0924842d/T/pipenv-_auxxp4x-requirements/pipenv-u4w5lmul-requirement.txt (line 1))
  1 location(s) to search for versions of pylint-quotes:
  * https://pypi.python.org/simple/pylint-quotes/
  Getting page https://pypi.python.org/simple/pylint-quotes/
  Looking up "https://pypi.python.org/simple/pylint-quotes/" in the cache
  Current age based on date: 11
  Freshness lifetime from max-age: 600
  Freshness lifetime from request max-age: 600
  The response is "fresh", returning cached response
  600 > 11
  Analyzing links from page https://pypi.python.org/simple/pylint-quotes/
    Found link https://pypi.python.org/packages/7c/88/7a894a269a58a907fe173282f50d642dcbaa44a0b5d12043f4785347f11b/pylint-quotes-0.1.3.tar.gz#md5=2351ca090248406e0eeafa75d7a1eb10 (from https://pypi.python.org/simple/pylint-quotes/), version: 0.1.3
    Found link https://pypi.python.org/packages/a0/0a/372a8219a6a42a9fdba7bdc00aaffb3a9d88d01138022d76c08f6b83764a/pylint_quotes-0.1.3-py2.py3-none-any.whl#md5=1097eeffc8f0a75cc843cc0c97866be7 (from https://pypi.python.org/simple/pylint-quotes/), version: 0.1.3
    Found link https://pypi.python.org/packages/7d/a7/b1ddde1d38c1141ab2d7c51d307ef7f81d62b970fc035ea8db74911821c4/pylint-quotes-0.1.5.tar.gz#md5=1cd3ebbcfc518653c81593d2486da0d0 (from https://pypi.python.org/simple/pylint-quotes/), version: 0.1.5
[...]

PR #2437, opened 2018-06-27
Add test case for #1901

frostming:
Fix the `HackPythonVersion` function so that it injects env only if the value is valid.


techalchemy:
thanks

Issue #2441, opened 2018-06-27
Relative git ssh module having ssh:// appended


typhoon2099:
I'm trying to install my Pipenv environment on a new machine and have upgraded pipenv from 2018.05.18 to 2018.06.25. A module which I'm pulling from git is now no longer installing with the following error:

```
Command "git clone -q ssh://git@bitbucket.org:redacted/redacted.git /home/redacted/.local/share/virtualenvs/redacted-eQF46Ow3/src/redacted" failed with error code 128 in None
```

Pipenv appears to be appending `ssh://` to the URL where it previously hasn't.

------------

##### Expected result

The module is installed

##### Actual result
The module is not installed

##### Steps to replicate

Try to install the module

EDIT: I've investigated trying to install the module via the command line and get the following output:

Relative Git URL
Command:
```
pipenv install -e git+git@bitbucket.org:company/library.git@1.0.0-rc2#egg=libb
```

Error:
```
Command "git clone -q ssh://git@bitbucket.org:company/library.git /tmp/tmpENxA1wsource/library" failed with error code 128 in None
```

Absolute Git URL
Command:
```
pipenv install -e git+git@bitbucket.org/company/library.git@1.0.0-rc2#egg=libb
```

Error:
```
Command "git clone -q git@bitbucket.org/company/library.git /home/employee/.local/share/virtualenvs/library-joI2jbUq/src/libb" failed with error code 128 in None
```

It looks like the `ssh://` is being appended to the relative URL instead of the absolute URL.

uranusjr:
Fixing one bug, causing another 💥 

We need to be smarter deciding how to fix the SSH URL.

techalchemy:
Closed by #2453

PR #2453, opened 2018-06-28
Update vendored dependencies

### LINKED ISSUE #2440: Absolute paths for -e . in lock -r
Since 2018.6.25, `pipenv install --dev -e .` gets resolved to absolute paths when running `pipenv lock --requirements --dev >dev-requirements.txt`.

That has the downside, that that .txt is only usable by the person who locked the dependencies if you (like us and I’m sure many more) use Pipenv to manage your deps, but use standard .txt files to deploy/tox.

I’m sure you had good reasons to change this behavior but it would be great to have an option to get the old back.

techalchemy:
- Updated requirementslib
- Fix local relative path bug whici caused paths to be output as
  absolute (test added to requirementslib)
- Fix SSH normalization bug (test added to requirementslib)
- Fixes #2440, #2441

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -16,6 +16,7 @@ black = {version="*", markers="python_version >= '3.6'"}
 pytz = "*"
 towncrier = {git = "https://github.com/hawkowl/towncrier.git", editable = true, ref = "master"}
 parver = "*"
+invoke = "*"
 
 [packages]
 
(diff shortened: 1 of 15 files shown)


techalchemy:
@frostming do you mind opening an issue about the patching issue? it's a bit complicated which is why i didn't do it yet (also I'm just bad at keeping up with docs)

PR #2458, opened 2018-06-28
Avoid double `pip` output if `--verbose` is used an output includes "Ignoring"

Tenzer:
If the `pip` output already has been printed due to it containing the string "Ignoring", then avoid printing it again if `--verbose` is enabled.

Ironically, in my simple test of this (`pipenv install --verbose` in a ) the "Ignoring" string is only included in the `pip` output if `--verbose` is supplied.

I've tried to make a screenshot of it here to show the problem, but I had to make it quite small in order to fit it all on screen. You can however see that the yellow and blue output is duplicated:
screen shot 2018-06-28 at 11 57 32

### DIFF
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -730,7 +730,7 @@ def do_install_dependencies(
                 c.block()
             if 'Ignoring' in c.out:
                 click.echo(crayons.yellow(c.out.strip()))
-            if verbose:
+            elif verbose:
                 click.echo(crayons.blue(c.out or c.err))
             # The Installation failed...
             if c.return_code != 0:


Tenzer:
@techalchemy Alright, I'll do that. I did however have a read through both [CONTRIBUTING.md](https://github.com/pypa/pipenv/blob/master/CONTRIBUTING.md) and https://docs.pipenv.org/dev/contributing/ and the latter states:

> If you are contributing, do not feel the need to sit on your contribution until it is perfectly polished and complete. It helps everyone involved for you to seek feedback as early as you possibly can. Submitting an early, unfinished version of your contribution for feedback in no way prejudices your chances of getting that contribution accepted, and can save you from putting a lot of work into a contribution that is not suitable for the project.

So I thought submitting my two character PR would be the right way to go.

techalchemy:
For sure that’s why I left it open, but we are triaging a bunch of stuff trying to cut a bugfix release and if I can’t figure out the problem you’re trying to solve quickly I am probably going to ask for an issue.

PR #2462, opened 2018-06-28
Use --pypi-mirror during virtualenv initialization

### LINKED ISSUE #2455: --pypi-mirror not applied during virtualenv creation
When starting a new project using pipenv with --pypi-mirror specified, virtualenv uses pip without respecting the --pypi-mirror setting. 

If PyPI is inaccessible during this process, virtualenv falls back to using local copies of necessary packages and pipenv silently proceeds. However, if connections to PyPI time out (e.g. if packets are dropped instead of rejected by a firewall), pip (and virtualenv) will hang, and pipenv will timeout.

JacobHenner:
Adds support for the --pypi-mirror parameter for all operations which
may result in a virtualenv initialization.

When a virtualenv is initialized, pip attempts to download several
dependencies from PyPI. If PyPI is unavailable, virtualenv silently uses
local packages instead, which is acceptable in most cases. However, in
some environments connection attempts to PyPI will stall rather than
fail, causing a pipenv timeout. By using the mirror specified by
--pypi-mirror, we can ensure virtualenv will attempt to download
dependencies from an accessible mirror instead of PyPI.

- Fixes #2455


pipenv/core.py
```diff
@@ -933,7 +934,8 @@ def do_create_virtualenv(python=None, site_packages=False):
     # Actually create the virtualenv.
     with spinner():
         try:
-            c = delegator.run(cmd, block=False, timeout=PIPENV_TIMEOUT)
+            pip_config = {'PIP_INDEX_URL': pypi_mirror} if pypi_mirror else {}
```
techalchemy:
```python
from .utils import fs_str
pip_config = {'PIP_INDEX_URL': fs_str(pypi_mirror)} if pypi_mirror else {}
```

JacobHenner: This has been revised

techalchemy:
Thanks!

PR #2474, opened 2018-06-29
Update version map for latest Python releases

greysteil:
Update version map for Wednesday's Python releases.


techalchemy:
Ah yes of course. Thanks!

Issue #2477, opened 2018-06-29
Add pipenv --support instead of python -m pipenv.help


erinxocon:
When homebrew is used to install python, the debug help command doesn't work.  
`pipenv --support` will now be the flag to invoke when getting diagnostic information.

techalchemy:
i'm guessing we want to keep `.help` too and just leave it as a thin wrapper around the new command

erinxocon:
`pipenv --support` and `python -m pipenv.help` are both supported now with `pipenv --support` being preferred.

PR #2478, opened 2018-06-29
Feature/2477

erinxocon:
Change `python -m pipenv.help` to `pipenv --support`.

Note. `python -m pipenv.help` is still supported but should only be used in cases when it's requested.


erinxocon:
Addresses #2477

PR #2488, opened 2018-07-02
Make the docs hierarchy more reasonable

frostming:
- Improve change log docs hierarchy
- Avoid copying file to docs


techalchemy:
:cake: thanks i'm not that good at rst

PR #2491, opened 2018-07-02
Translate readme to spanish

iKenshu:
Hey guys!

I want to contribute to this project helping with the translation of the docs to Spanish. I saw that the website has a Japanese translation that's why I think is a good idea to help with this. 

Let me know if this is not the right way to help. I want to translate all the docs.

### DIFF
--- a/README_es.rst
+++ b/README_es.rst
@@ -1,0 +1,261 @@
+Pipenv: Python Development Workflow for Humans
+==============================================
+
+.. image:: https://img.shields.io/pypi/v/pipenv.svg
+    :target: https://pypi.python.org/pypi/pipenv
+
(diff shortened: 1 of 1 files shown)


uranusjr:
I feel it might be a better idea to translate the documentation instead. There is already an unofficial [Japanese translation](https://pipenv-ja.readthedocs.io/ja/translate-ja/), it would make sense to use a similar approach for Spanish. We can then add links to those documentations in README.

iKenshu:
Thanks @uranusjr I will translate the docs then. After that, should I make another PR to include the links?

uranusjr:
Sure!

(Closing this to keep the tracker tidy.)

PR #2503, opened 2018-07-03
Improve inline virtualenv activation logic

uranusjr:
Set `VIRTUAL_ENV` during activation so py.exe does the right thing.

Before:

```
> pipenv run py -c "import sys; print(sys.executable)"
C:\Users\uranusjr\AppData\Local\Programs\Python\Python36\python.exe
```

After:

```
> pipenv run py -c "import sys; print(sys.executable)"
C:\Users\uranusjr\Documents\programming\pipenv\.venv\Scripts\python.exe
```

Also add venv support—it is actually quite easy, so why not.


uranusjr:
I’m not sure why `VIRTUAL_ENV` was not set before. Maybe there is a reason I am not aware of?

PR #2509, opened 2018-07-04
Fix test case - test_download_file url

mlouielu:
The file was changed by 6c4b973 that rename from .rst to .md


kennethreitz:
fixed already, but thanks so much!

PR #2513, opened 2018-07-04
Remove pew.lsenvs usage from project

uranusjr:


techalchemy:
Nice, seems right to me. One step at a time...

PR #2521, opened 2018-07-05
Un-vendor Pew

uranusjr:
I believe we have removed all Pew usages in the code base.


techalchemy:
Does this cover project creation?

techalchemy:
If this passes and covers project creation I’m good with it.

techalchemy:
I.e `pew new --project`

uranusjr:
Yes it does!

```python
# Associate project directory with the environment.
# This mimics Pew's "setproject".
project_file_name = os.path.join(project.virtualenv_location, '.project')
with open(project_file_name, 'w') as f:
    f.write(fs_str(project.project_directory))
```

`pew new --project` == `pew new` + `pew setproject`.

uranusjr:
p.s. We have test for project creation.

PR #2537, opened 2018-07-07
Allow passing options of commands via environment variables

fridex:
We would like to provide a way to automatically submit options during builds via environment variables. This is not a standard way how to enable click's `auto_envvar_prefix`, but it works with `console_scripts` entrypoint.

### DIFF
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -286,6 +286,9 @@ def cli(
         # Display help to user, if no commands were passed.
         echo(format_help(ctx.get_help()))
 
+    # Allow passing options of commands via environment variables.
+    ctx.auto_envvar_prefix = 'PIPENV'
+
 
 @command(
     short_help="Installs provided packages and adds them to Pipfile, or (if none is given), installs all packages.",


fridex:
Worth to document such change; also the `pipenv --envs` output will probably need to be adjusted to provide some info.

uranusjr:
This would be a very nice addition, but unfortunately it can’t be done this simple :( We have environment variables that are not in line with option names, e.g. `PIPENV_PYTHON` is totally not related to `--python`. We will need to grep-replace-test the whole code base to ensure this won’t happen. Probably even need to find a way to prevent this from accidentally happening in the future.

fridex:
Setting the `auto_envvar_prefix` this way causes availability of environment variables only in commands and not CLI itself (click's context is already instantiated for `cli` when parsing). But I see `PIPENV_PYTHON` being used [explicitly from env](https://github.com/pypa/pipenv/blob/ef6e667971d08af51af094f5732d4c95d97ad63d/pipenv/environments.py#L13) for commands so for example `PIPENV_INSTALL_PYTHON` will clash with `PIPENV_PYTHON`. Is that what you are referring to (priorities for applying options)?

I quickly tried to go through the listing of environment variables and didn't spotted any other collision .

BTW listing in [docs](https://docs.pipenv.org/advanced/#configuration-with-environment-variables) does not state `PIPENV_PYTHON`.

uranusjr:
Yes it is. I didn’t actually check what environment variables we are currently using, but `PIPENV_PYTHON` came to me immediately as a naming conflict when seeing this PR. It is an internal environment variable used to communicate what Python is currently being used between Pipenv processes, so it likely can be safely renamed. A simple find-and-replace inside the repository should be enough.

fridex:
> so it likely can be safely renamed

IMHO I would keep it for backwards compatibility. Based on click's logic

```console
$ export PIPENV_PYTHON=/usr/bin/python2.7
$ pipenv install requests --python /usr/bin/python3.6
```

Is equivalent to:

```console
$ export PIPENV_PYTHON=/usr/bin/python2.7
$ export PIPENV_INSTALL_PYTHON=/usr/bin/python3.6
$ pipenv install requests
```

So this PR should not introduce any new logic for that.

uranusjr:
Oh! I didn’t realise that, good to know! Indeed this likely wouldn’t cause any problems if that is the case. I will find time to review if there are any side effects. Thanks for the clarification 😃

kennethreitz:
Please use the new PEEP process: https://github.com/pypa/pipenv/blob/master/peeps/PEEP-000.md

fridex:
@kennethreitz where can I find steps on how to submit a PEEP?

kennethreitz:
Jest send a PR to the ‘peeps’ directory!

PR #2539, opened 2018-07-08
Revise note format for consistency in basics.rst

gkdoc:
Some notes were preceded by a header while others were not. Added headers to all notes.


erinxocon:
Thanks for the catch @gkdoc

PR #2546, opened 2018-07-10
upgraded shellingham requirement to v1.2.2 

badboybeyer:
upgraded shellingham requirement to v1.2.2 and then invoked vendoring.update. 

Vendoring.update brought in changes from several unexpected libraries: 

- ` pipenv/vendor/requirementslib`

- `pipenv/patched/crayons.py`

- `pipenv/patched/piptools/utils.py`

Two things concern me.
First, the license file ` pipenv/vendor/requirementslib/_vendor/pipfile/LICENSE.APACHE` was removed. There is still a `LICENSE` file which references either `LICENSE.APACHE` or `LICENSE.BSD` as being applicable. 
It may have been an error for the vendoring process to remove the `LICENSE.APAHCE` file, but it also seems that a `LICENSE.BSD` file was missing.

Second, the requirements for shellingham were removed from crayons. 
Is this requirement being removed from the pipenv project?
I still see a requirement in `./pipenv/shells.py`.


techalchemy:
flagging as do not merge until I get the patch in

techalchemy:
I merged the crayons patch into your branch, thanks again

uranusjr:
I just released 1.2.3 for Shellingham, will need to update that as well. So many bugs…

PR #2554, opened 2018-07-10
Delete "Changing Where Pipenv Stores Virtualenvs"

OrangeDog:
It's a partial duplicate of the "Custom Virtual Environment Location" section.


uranusjr:
This really shows how poorly organised the documentation is. Thanks so much!

OrangeDog:
Separating package management (what to install), environment management (where to install) and misc (everything else) may be a good way to structure it. I might have a go at some point, perhaps not any time soon though.

PR #2563, opened 2018-07-11
Only quote arguments during cmdify if needed

uranusjr:
Windows built-in commands don't always work well with quotes.

```cmd
rem The path can be interpreted correctly, but not "/w".
dir "/w" "C:\"

rem Outputs "foo" "bar", with quotes!
echo "foo" "bar"
```

The solution (I think) is to be less aggressive. Only quote things if they need to be quoted (i.e. contains whitespaces).

### DIFF
--- a/news/2563.bugfix
+++ b/news/2563.bugfix
@@ -1,0 +1,2 @@
+Improve quoting logic for ``pipenv run`` so it works better with Windows
+built-in commands.
--- a/pipenv/cmdparse.py
+++ b/pipenv/cmdparse.py
@@ -55,11 +55,18 @@ class Script(object):
 
         The result is then quoted into a pair of double quotes to be grouped.
 
+        An argument is intentionally not quoted if it does not contain
+        whitespaces. This is done to be compatible with Windows built-in
+        commands that don't work well with quotes, e.g. everything with `echo`,
+        and DOS-style (forward slash) switches.
+
         The intended use of this function is to pre-process an argument list
         before passing it into ``subprocess.Popen(..., shell=True)``.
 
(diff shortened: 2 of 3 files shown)


uranusjr:
@techalchemy The use case is a bit different here. That function is good if you already have a string, but here we have a list of arguments, i.e. `["dir", "/w", "C:\Program Files"]`. The problem is to decide when to quote, not how (that is easy, and the two solutions are logically equivalent).

techalchemy:
`[shellquote(arg) for arg in args if " " in arg]`

uranusjr:
That won’t work either because it escapes single quotes, and Windows don’t want you to.

```python
import shlex
import os
import re


def escape_grouped_arguments(s):
    """Prepares a string for the shell (on Windows too!)

    Only for use on grouped arguments (passed as a string to Popen)
    """
    if s is None:
        return None

    # Additional escaping for windows paths
    if os.name == "nt":
        s = "{}".format(s.replace("\\", "\\\\"))
    return '"' + s.replace("'", "'\\''") + '"'


def my_quote(s):
    return '"{0}"'.format(re.sub(r'(\\*)"', r'\1\1\\"', s))

s = "print('hello world')"
print('SO:   ', 'python', '-c', escape_grouped_arguments(s))
print('shlex:', 'python', '-c', shlex.quote(s))
print('mine: ', 'python', '-c', my_quote(s))
```

```
SO:    python -c "print('\''hello world'\'')"
shlex: python -c 'print('"'"'hello world'"'"')'
mine:  python -c "print('hello world')"
```

Only mine works correctly in cmd.exe.

techalchemy:
Damn. You should make a library out of this or something

PR #2572, opened 2018-07-12
New, better pyenv module

uranusjr:
Installation logic refactored into the new module. Also implemented logic to read newest version from `pyenv install --list` automatically, instead of maintaining a custom mapping.

I imagine this would cause a lot of conflicts with the pythonfinder refactor, and will rebase after that is finished. Don’t merge this yet.


techalchemy:
By the way +1 on this (I know i didnt actually review it but hopefully it works)

PR #2577, opened 2018-07-12
Pythonfinder integration, take 3

uranusjr:


uranusjr:
Will need to be rebooted after #2574.

PR #2542, opened 2018-07-09
Add docstrings to show short_help when running --help

sup:
Description
------------
This addresses #2381 by adding docstrings to cli commands which allows Click to display some informative help text when running --help. This change simply maps the existing `short_help` field to a docstring.

Moreover, this doesn't cause the truncation issue mentioned in #2381 because`short_help` is already defined and overrides the 45 character truncation Click uses by default. Indeed, running `pipenv --help` locally results in status quo behavior:

```
Commands:
  check      Checks for security vulnerabilities and against PEP 508 markers
             provided in Pipfile.
  clean      Uninstalls all packages not specified in Pipfile.lock.
  graph      Displays currently-installed dependency graph information.
  install    Installs provided packages and adds them to Pipfile, or (if none
             is given), installs all packages.
  lock       Generates Pipfile.lock.
  open       View a given module in your editor.
  run        Spawns a command installed into the virtualenv.
  shell      Spawns a shell within the virtualenv.
  sync       Installs all packages specified in Pipfile.lock.
  uninstall  Un-installs a provided package and removes it from Pipfile.
  update     Runs lock, then sync.
```

Result
------
```
$ pipenv check --help
Usage: pipenv check [OPTIONS] [ARGS]...

  Checks for security vulnerabilities and against PEP 508 markers provided
  in Pipfile.

Options:
  --three / --two     Use Python 3/2 when creating virtualenv.
  --python TEXT       Specify which version of Python virtualenv should use.
  --system            Use system Python.
  --unused TEXT       Given a code path, show potentially unused dependencies.
  -i, --ignore TEXT   Ignore specified vulnerability during safety checks.
  --pypi-mirror TEXT  Specify a PyPI mirror.
  -h, --help          Show this message and exit.
```

Issue #2381, opened 2018-06-18
Help text doesn't appear in command --help


apnewberry:
```
$ pipenv --version
pipenv, version 2018.05.18

$ pipenv check --help
Usage: pipenv check [OPTIONS] [ARGS]...

Options:
  --three / --two  Use Python 3/2 when creating virtualenv.
  --python TEXT    Specify which version of Python virtualenv should use.
  --system         Use system Python.
  --unused TEXT    Given a code path, show potentially unused dependencies.
  -h, --help       Show this message and exit.

```

There's a field called `short_help` ("Checks for security vulnerabilities and against PEP 508 markers provided in Pipfile.") which appears when running `pipenv` alone, but it doesn't appear when running `pipenv check --help`. It probably should appear there, because that's what people run when they want to know what `pipenv check` does.

See also: https://github.com/pypa/pipenv/pull/1155

uranusjr:
Contributions are welcomed!

toxinu:
In order to have `short_help` printed with `pipenv check` command we must declare help text as `command` function docstring. Is it something we want to do?

Side-effect is that help will be truncated by `click` like that:
```
Commands:
  check      Checks for security vulnerabilities and...
  clean      Uninstalls all packages not specified in Pipfile.lock.
```
Strange behavior that `click` doesn't show `sort_help` when running `pipenv check` as excepted...

techalchemy:
We have a giant blob of text with this info in `core.py` which might be what you’re looking for

erinxocon:
This has been fixed and approved in #2578

PR #2578, opened 2018-07-12
Revert "Revert "Add docstrings to show short_help when running --help""

sup:
This reverts #2571 which mistakingly reverted #2542 which addressed #2381 by adding `short_help` text to the output of `--help`.

Let me know if this works @erinxocon!


erinxocon:
BDFL approved.  Thanks and sorry for the mixup!

PR #2581, opened 2018-07-13
Fix #2527: Add ability to suppress nested venv courtesy notice

### LINKED ISSUE #2527: Is there a way to disable / suppress pipenv's "Courtesy Notice" about virtual environments?
Each time I invoke `pipenv` I see:

> Courtesy Notice: Pipenv found itself running within a virtual environment, so it will automatically use that environment, instead of creating its own for any project. You can set PIPENV_IGNORE_VIRTUALENVS=1 to force pipenv to ignore that environment and create its own instead.

Can this be globally suppressed?

Haplo-Dragon:
This will add an environment variable, PIPENV_SUPPRESS_NESTED_WARNING, to be used to suppress the courtesy notice from warn_in_virtualenv (produced when pipenv finds itself running inside another virtual environment). 
It will also add a sentence to the end of the courtesy notice to tell users how to suppress the notice.

I ended up putting the test in its own file - if there's a more logical place to put it, let me know!

**Fixes #2527.**


tests/unit/test_core.py
```diff
@@ -0,0 +1,20 @@
...
+@mock.patch('pipenv.environments.PIPENV_SUPPRESS_NESTED_WARNING', '1')
+@pytest.mark.core
+def test_suppress_nested_venv_warning():
+    f = io.StringIO()
+    # Capture the stderr of warn_in_virtualenv to test for the presence of the
+    # courtesy notice.
+    with redirect_stderr(f):
+        warn_in_virtualenv()
+    output = f.getvalue()
+    assert 'Courtesy Notice' not in output
```
uranusjr: I’d recommend using [capsys](https://docs.pytest.org/en/2.8.7/capture.html) to redirect stdout/stderr here. I didn’t test myself, and cannot access Buildkite logs either, but suspect this caused the build failure on Python 2.

techalchemy:
Can we just change the name of this to be a more generic ‘be less verbose’ variable? Maybe it should eventually toggle off all of the informational things we write to stderr

Haplo-Dragon:
Thanks for the prompt feedback! I'll switch to capsys - I feel kinda silly for not noticing it earlier. @techalchemy, how do we feel about PIPENV_CONCISE? Or maybe PIPENV_TERSE? I figure PIPENV_SHUT_UP is out of the question?

uranusjr:
Could we make this more generic, like, combining it with the `--verbose` flag in other commands? Something like `PIPENV_VERBOSITY` as an integer value?

Haplo-Dragon:
PIPENV_VERBOSITY as an integer value sounds good to me! I'm assuming 0 will be the default (most verbose) setting, and we can assess higher values as needed later? 
I don't feel qualified to integrate it throughout the rest of the codebase, but I can certainly set it and integrate it with my small changes.

PR #2582, opened 2018-07-14
Dial back Pythonfinder integration effort

uranusjr:
Let’s try this a little at a time…

This only uses Pythonfinder in `find_a_system_python`, and leaves everything else alone. I hope this approach will make rewrites easier to tackle.


pipenv/core.py
```diff
@@ -930,23 +859,17 @@ def do_create_virtualenv(python=None, site_packages=False, pypi_mirror=None):
...
-                    crayons.normal("PATH", bold=True),
-                ),
-                err=True,
-            )
-            sys.exit(1)
-    click.echo(crayons.blue(c.out), err=True)
+        c = delegator.run(
+            cmd, block=False, timeout=PIPENV_TIMEOUT, env=pip_config,
+        )
+    c.block()
```
techalchemy: If we don't block here and we pass `block=False` we can't read the output...

uranusjr: I think it should be indented inside the `with` block?

techalchemy: makes sense to me, I just put it there because of no reason in particular

techalchemy:
just rebased this, I think I'm probably good with it if you are (I can cut 1.0 of pythonfinder if we are satisfied)

Issue #1918, opened 2018-04-05
Provide better error message if the project’s virtual environment is broken


slhck:
## Was: "Your Pipfile requires python_version 3.6, but you are using None" and various errors

-----

I run pipenv, installed via `brew install pipenv` on macOS 10.12. I've installed `python` from Homebrew (3.6.5), as well as `python@2`.

When I open `pipenv shell`, I get this output:

```
➜ pipenv shell
Warning: Your Pipfile requires python_version 3.6, but you are using None (/Users/werner/.local/share/v/b/bin/python).
  $ pipenv check will surely fail.
Spawning environment shell (/bin/zsh). Use 'exit' to leave.
```

When I run `pipenv check`, I get:

```
Checking PEP 508 requirements…
Traceback (most recent call last):
  File "/usr/local/Cellar/pipenv/11.9.0_1/libexec/bin/pipenv", line 11, in <module>
    load_entry_point('pipenv==11.9.0', 'console_scripts', 'pipenv')()
  File "/usr/local/Cellar/pipenv/11.9.0_1/libexec/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/Cellar/pipenv/11.9.0_1/libexec/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/local/Cellar/pipenv/11.9.0_1/libexec/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/Cellar/pipenv/11.9.0_1/libexec/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/Cellar/pipenv/11.9.0_1/libexec/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/Cellar/pipenv/11.9.0_1/libexec/lib/python3.6/site-packages/pipenv/cli.py", line 642, in check
    three=three, python=python, system=system, unused=unused, args=args
  File "/usr/local/Cellar/pipenv/11.9.0_1/libexec/lib/python3.6/site-packages/pipenv/core.py", line 2275, in do_check
[...]

PR #2595, opened 2018-07-17
Show better error if virtual env is broken

sinscary:
This shows the better error to the user if virtual env is broken
also suggests commands that can solve the issue

Closes https://github.com/pypa/pipenv/issues/1918

### DIFF
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -636,12 +636,15 @@ def ensure_project(
                 ):
                     click.echo(
                         "{0}: Your Pipfile requires {1} {2}, "
-                        "but you are using {3} ({4}).".format(
+                        "but you are using {3} ({4}). Running"
+                        "{5} and rebuild the virtual environment"
+                        "may resolve the issue".format(
                             crayons.red("Warning", bold=True),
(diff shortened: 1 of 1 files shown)


uranusjr:
Hi, thanks for the work! I feel the `pipenv install` part could be made more general—people don’t necessarily use it to bootstrap their environments, and generally want to specify a Python version/executable themselves. Something like “Running pipenv --rm and rebuild the virtual environment may resolve the issue” could be better, but I wonder if that is a clear enough instruction for more relatively inexperienced users. What do you think?

uranusjr:
Maybe something like “… rebuild the virtual environment, e.g. `pipenv install`”?

sinscary:
Yeah, that's also good. I will update the error message to what you suggested.

uranusjr:
I’m very bad at figuring out what is more understandable for newcomers, so I’ll leave the decision to you :) I’ll merge either way.

Raffffqq (commented):
- [ ]

Raffffqq (changes requested):
- > []()

> - @__

PR #2600, opened 2018-07-18
Make README badges inline

sup:
This makes the badges in the README inline which pushes some of the more informative content up the page.

Super small, nit change, but a bunch of other projects do the same :P
* https://github.com/pypa/pip
* https://github.com/requests/requests


uranusjr:
They used to be inline (when it was restructuredText), but Kenneth changed it when converting to Markdown. I’m going to assume it was not intentional. Thanks!

PR #2612, opened 2018-07-19
Adding some git+git and git+ssh tests as examples

atzannes:
I was trying to figure out why I was not able to install some `git+git` and some `git+ssh` dependencies with pipenv. Turns out I was falling prey to the slight syntax differences of those two alternatives compared to `git+https`. 

I adapted the `git+https` test for `git+git` and `git+ssh` and thought I'd contribute them. Not sure if it is worth keeping the `gitdb2` check in the two new tests though.

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -28,7 +28,30 @@ def check_internet():
     return True
 
 
+def check_github_ssh():
+    res = False
+    try:
+        # `ssh -T git@github.com` will return successfully with return_code==1
+        # and message 'Hi <username>! You've successfully authenticated, but
+        # GitHub does not provide shell access.' if ssh keys are available and
+        # registered with GitHub. Otherwise, the command will fail with
+        # return_code=255 and say 'Permission denied (publickey).'
+        c = delegator.run('ssh -T git@github.com')
+        res = True if c.return_code == 1 else False
+    except Exception:
+        pass
+    if not res:
+        warnings.warn(
+            'Cannot connect to GitHub via SSH', ResourceWarning
+        )
+        warnings.warn(
+            'Will skip tests requiring SSH access to GitHub', ResourceWarning
+        )
+    return res
+
+
 WE_HAVE_INTERNET = check_internet()
+WE_HAVE_GITHUB_SSH_KEYS = check_github_ssh()
 
 TESTS_ROOT = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 PYPI_VENDOR_DIR = os.path.join(TESTS_ROOT, 'pypi')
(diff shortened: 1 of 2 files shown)


techalchemy:
Thanks for the PR! We probably don't need the gitdb2 test here, but it is probably worth testing for these.  We likely will want to document this somewhere too (contributions welcome!)

The tests failed here mainly due to waiting on input to accept the SSH host ID.  I sorted that out and kicked this off again

atzannes:
I'm not sure where and how to contribute docs. Pointers very welcome.

I am going to remove the `gitdb2` part of the tests. Do you prefer to keep the tests for `git+git` and `git+https` separate (as they are now), or to fold them into the `test_basic_vcs_install`?

atzannes:
I updated the tests to remove the `gitdb2` part and to merge them. Build Kite was red before the update, and it is still failing @techalchemy. I created a Build Kite account (same email as github) but unsurprisingly I cannot look into what the problem is.

techalchemy:
It’s failing due to ssh key pairing issues:

```
Collecting six from git+ssh://git@github.com/benjaminp/six.git@1.11.0#egg=six
  Cloning ssh://git@github.com/benjaminp/six.git (to revision 1.11.0) to /tmp/pip-install-6b1mp6/six
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.
```

atzannes:
I'm guessing this is some environment setting to be configured in Build Kite, right?

techalchemy:
Honestly I don't really know, you can try having it install from `git+ssh://git@github.com/pypa/pipenv.git` and see if you have any luck with that.  I can tweak any settings we need in buildkite / the host machine, but I'm not sure what is causing the issue

atzannes:
So at this point I am skipping the test testing git+ssh and Build Kite is green.
Next, I'm going to push a commit that attempts to detect when SSH keys are not registered with GitHub (in other words, when a `git clone ssh://git@github...` would fail with `git@github.com: Permission denied (publickey)`) and automatically skip those tests that rely on such an integration being set up.

techalchemy:
Sorry, been busy with a work deadline but the issue is that yeah, the ssh keys in question aren't connected to a github account and probably won't be able to pull down the package but this works for me and I think it's good to merge

Much appreciated, and sorry about buildkite, I'll bring the issue back up and see if we can sort it out

atzannes:
Thanks! I'm having some trouble with this automatic check for github ssh-keys (it's skipping locally when it shouldn't) but I'll sort it out and let you know when I'm ready to merge

atzannes:
Ok, I fixed those, this is ready to merge on my side

techalchemy:
that's pretty nifty, thanks for taking the time on this! this is a huge help

PR #2631, opened 2018-07-23
Fix index_indicators condition

### LINKED ISSUE #2494: index_indicators check needs improvement.
##### Issue description

When installing local package, the logical check on this line does not work correctly when package name has `-i` in it.

https://github.com/pypa/pipenv/blob/29eb73aa7f1abc2bef077cd967df21a8719f4b66/pipenv/core.py#L1873

##### Actual result

Pipenv tries installing a package called `-e`

##### Steps to replicate

Suppose you have a local python package named like `marathon-python-internal`

> pipenv install -e ../packages/marathon-python-lnternal/

This results in the code flow falling inside the IF condition vs some other package name that doesn't have the string `-i` in them eg

> pipenv install -e ../packages/funniest/

```
pipenv, version 2018.7.1
```

mlouielu:
This will not match:

    pipenv install -e packages/haha-internal

Fix #2494

### DIFF
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -1844,7 +1844,7 @@ def do_install(
     if any(line.endswith(s) for s in index_indicators):
         # check if cli option is not end of command
         raise click.BadArgumentUsage("Please provide index value")
-    if any(s in line for s in index_indicators):
+    if any(arg == s for arg in line.split() for s in index_indicators):
         line, index = split_argument(line, short="i", long_="index", num=1)
         line, extra_indexes = split_argument(line, long_="extra-index-url")
         package_names = line.split()
--- a/tests/integration/test_install_basic.py
+++ b/tests/integration/test_install_basic.py
@@ -347,6 +347,18 @@ def test_editable_no_args(PipenvInstance):
         assert "Please provide path to editable package" in c.err
 
 
+@pytest.mark.editable
+@pytest.mark.install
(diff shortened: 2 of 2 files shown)


techalchemy:
Hey there, thanks for the PR! I believe the direction we want to go on this is to handle parsing index groups with click if possible, as this logic is kind of messy as is. I’m ok with you taking a crack at it if you like. 

I could be persuaded to merge a temporary fix in the meantime, but I believe it would have to handle parsing everywhere we are currently doing it with this approach.

mlouielu:
@techalchemy yeah, sure. But previous time #2299 to make `--editable` a click option end with closed PR, not sure if it is worth to make it an option or not.

uranusjr:
@mlouielu Just do it (the editable option too, I think they are good if done correctly).

techalchemy:
The editable one didn’t capture the change needed all the way through the codebase as I recall and it would have just broken a lot of things. I’d love to have these as click options, but I want to re-affirm that doing so involves more than just altering the cli. The logic of both of these flows through the entire application and needs to be traced and modified to accommodate those changes.

mlouielu:
@techalchemy @uranusjr OK, I'll open a new PR for this!

techalchemy:
closing for now, I have a fix for this-- thanks for the  work you put in on it!

techalchemy:
https://github.com/pypa/pipenv/pull/2814

PR #2632, opened 2018-07-23
Fix passing multiple ignores to pipenv check

washeck:
Fix missing space in command line arguments to 'safety' when
multiple CVEs are ignored.

### DIFF
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -2326,7 +2326,7 @@ def do_check(
     else:
         python = system_which("python")
     if ignore:
-        ignored = "--ignore {0}".format("--ignore ".join(ignore))
+        ignored = "--ignore {0}".format(" --ignore ".join(ignore))
         click.echo(
             crayons.normal(
                 "Notice: Ignoring CVE(s) {0}".format(crayons.yellow(", ".join(ignore)))


washeck:
When multiple CVEs are ignored in pipenv check it results in error:

```
pipenv check -i 33073 -i 33074
Checking PEP 508 requirements...
Passed!
Checking installed package safety...
Notice: Ignoring CVE(s) 33073, 33074
An error occurred:
Usage: safety check [OPTIONS]

Error: Got unexpected extra argument (33074)
```

It is because the arguments passed to `safety` look like this: `'--ignore 33073--ignore 33074'`

This PR just adds the missing space.

rooterkyberian:
Not sure if we want another testcase for `pipenv check`, but here is one I have used to reproduce this error and test the fix:
```python
@pytest.mark.cli
@flaky
def test_pipenv_check_with_double_ignore(PipenvInstance):
    with PipenvInstance() as p:
        c = p.pipenv('check --ignore 666 --ignore 777')
        assert c.return_code == 0
        assert 'Ignoring CVE(s) 666, 777' in c.err
```

techalchemy:
Would very much appreciate the additional test case, feel free to make a separate PR with that test.  Much appreciated and thanks for your contributions on this!

PR #2644, opened 2018-07-25
Two doc fixes: ghbtns.com->https, removed Konami

javabrett:


uranusjr:
Wouldn’t it be better to make Konami work instead of removed?

javabrett:
> Wouldn’t it be better to make Konami work instead of removed?

I don't know, is that what you are suggesting?

What I removed seemed to be fetching from http://fortunes.herokuapp.com/random/raw , also down/missing, seems unimportant - what is it supposed to do, and would you like to fix it?  Or is Konami doing something else here?

PR #2645, opened 2018-07-25
PIPENV_VERBOSITY

### LINKED ISSUE #2527: Is there a way to disable / suppress pipenv's "Courtesy Notice" about virtual environments?

uranusjr:
Fix #2527, close #2581.

### DIFF
--- a/news/2527.feature
+++ b/news/2527.feature
@@ -1,0 +1,2 @@
+Added environment variable `PIPENV_VERBOSITY` to control output verbosity
+without needing to pass options.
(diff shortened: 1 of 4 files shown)


Abbyboing:
Where should I set this?

> PIPENV_VERBOSITY=-1

A bit new to programming so unable to find the proper solution.

uranusjr:
It depends.

[Windows (cmd.exe or Powershell)](http://www.dowdandassociates.com/blog/content/howto-set-an-environment-variable-in-windows-command-line-and-registry/)

[Bash and ZSH (common shells on macOS and Linux)](https://flaviocopes.com/shell-environment-variables/)

PR #2646, opened 2018-07-25
Unify verbosity specification

uranusjr:
Always use `environments.PIPENV_VERBOSITY` (must not `from ... import PIPENV_VERBOSITY` because this needs to be lazy), instead of passing `verbosity` around. This should make the code base respect the verbosity flag better.


techalchemy:
This is fine, but it would be finer if had a shortcut to it called `PIPENV_QUIET` or something like that :D

techalchemy:
From a UI perspective - I don’t care about globals, but talking about environment vars I can’t imagine `PIPENV_VERBOSITY` Is more intuitive than a simple toggle of `PIPENV_QUIET`

uranusjr:
The difference is if you ever want to add another (or more) verbosity levels, you will probably need to switch to an int-based value.

FWIW there is actually already one place in the code base that has a `debug` logging level (in the resolver) right now, but significantly underused.

techalchemy:
Right I think we should provide both but that `PIPENV_QUIET` should just set an int value for `PIPENV_VERBOSITY`

techalchemy:
my main complaint is that I have literally never seen an env var used the way we are currently using `PIPENV_VERBOSITY`. Not only is this confusing for users, but it's going to be confusing for us as we now have to undertake the overhead of assessing 'is this verbosity level `n` as an extra layer of translation before we determine if it should be suppressed.  

Composer for instance has `quiet` and `no-input` modes.  Yarn has `--silent`, bundler uses `--quiet`, npm has both `quiet` and `silent`, cargo has `quiet` which sets a backend verbosity. apt and apt-get use `-q` for quiet mode. The theme here is that none of these force the user to set a number, they just set a flag.

From a UI/UX standpoint this needs to be a flag

uranusjr:
The environment variable does not preclude the option flag, and the environment variable is only needed when a flag is not viable (i.e. for the “courtesy notice”, and `pipenv run` etc.). From the user’s perspective, verbose mode is still activated with `--verbose` and `-v` as before, only the internal representation of this changed, so we can adapt to a more elaborated verbosity switch mechanism.

Cargo is actually a fine example here—it has a normal mode, `-q`, `-v`, and `-vv`. Internally this is represented a an enum, just like it is represented as an integer here (because Python does not have a language-native enum).

I would agree maybe the user should be able to use boolean-based environment variables to control the internal value though. I’ll add a couple of extra entries for that.

techalchemy:
For sure. All I’ve ever wanted was the extra ability. No need to remove this one but may want to leave it undocumented to avoid too many levers

uranusjr:
Sounds reasonable, I did consider removing the documentation but decided to leave it in.

PR #2649, opened 2018-07-25
Remove odd sentence from index page

benjaoming:
Proposing to remove this sentence. It doesn't make a lot of sense, and it isn't explained. Is it Windows, the OS? Or some other concept of Windows? After removing it, the text is easy to understand.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,6 @@ Pipenv: Python Development Workflow for Humans
 
 **Pipenv** is a tool that aims to bring the best of all packaging worlds
 (bundler, composer, npm, cargo, yarn, etc.) to the Python world.
-*Windows is a first--class citizen, in our world.*
 
 It automatically creates and manages a virtualenv for your projects, as
 well as adds/removes packages from your `Pipfile` as you


techalchemy:
This is about the OS. Sorry if you found that confusing. Please don’t open PRs to randomly delete stuff without asking g about it first

benjaoming:
> Please don’t open PRs to randomly delete stuff without asking g about it first

The PR was the question ;)

techalchemy:
Sure thing, no big deal (I mainly add that in to make sure to discourage high-effort PR's before communicating)

benjaoming:
Okay - but not sure I understand - should I open an issue then to discuss this sentence?

techalchemy:
no we are not getting rid of the sentence :p

techalchemy:
if you want to help out with the rest of the documentation though, it would be most welcome!

uranusjr:
I think Windows, when used as a singular noun (as in here), almost always refers to the Microsoft operating system.

benjaoming:
Friends, I fear that you have only been reading this to each other and in your own minds :wink:

For someone who is new to this paragraph, not familiar with the particular idea of a "first-class citizen", and not familiar with the project, let me share the issues:

>  **Pipenv** is a tool that aims to bring the best of all packaging worlds
> (bundler, composer, npm, cargo, yarn, etc.) to the Python world.
> *Windows is a first--class citizen, in our world.*

1) Is it Windows the OS?
2) Which Windows? 2003? 7, 8, 10?
3) Why is this necessary to reflect upon? Is it not a normal POV?
4) Do you live in your own world? (the sentence might imply that)
5) Do you own the Python world and can make statements on its behalf? (the sentence might also imply that)
6) Why two dashes in [first-class](https://en.wikipedia.org/wiki/First-class_citizen)?
7) Does this exclude other OS'es that aren't mentioned as first-class citizens?
8) Is this statement correcting or contradicting another belief?
9) Are you changing the Python world in this regard?

Delete the sentence and the problem goes away.

With regards to Windows support, perhaps write an elaborated section about Platform support! That will make it easier to document known issues. For instance when you find that Windows XP support is broken :stuck_out_tongue_closed_eyes: 

I can add a PR that adds that section and deletes this sentence if it's more acceptable.

uranusjr:
The sentence has been there from almost day one, way before I joined as a maintainer, and I’ve never been able to read it in another way. Literally every line in the documentation can be picked on like this if a person wants to go into that rabbit hole, but realistically, the line is as clear as every other sentence (if not clearer… the documentation do contain a lot of unclear descriptions) if a reader think inside the context of the current Python open source landscape.

But I’m fine if you are unable, or don’t want to think in that context. Fortunately, I (and every maintainer here) don’t need your approval to put sentences in the documentation.

PR #2651, opened 2018-07-26
Use COMSPEC for shell detection fallback

uranusjr:
The subshell logic now considers:

* PIPENV_SHELL
* Auto detection
* SHELL (should very likely work for POSIX)
* PYENV_SHELL
* COMSPEC (should always work for Windows)

This should be able to close sarugaku/shellingham#9 in the sense of not raising an exception.


uranusjr:
Far Manager can be detected, but I don’t know it well enough and don’t know if it is even a good idea to spawn it (it’s more like a CLI file manager than a shell, complete with menus and things, similar to old DOS applications). This looks like a reasonable thing to do anyway (considering we already detect `SHELL` and even `PYENV_SHELL`), and does kind of resolve the problem.

Issue #2367, opened 2018-06-16
Provide better logging for “run”


uranusjr:
This is more like a TODO item for myself. Ref #2335.

* Add an environment variable to control verbosity. Probably consolidate this with `--verbose`? But `run` cannot have a `--verbose` flag now (unless we improve parsing).
* Show more information when the flag is set, e.g. what command will be run.

sinscary:
@uranusjr Can I work on this?

sinscary:
Thank you :)

uranusjr:
Alright I’ve merged that, you should be able to use `PIPENV_VERBOSITY` to control this.

sinscary:
@uranusjr I wonder what other information we can show other than which command is about to run if `--verbose/-v` flag is set.

frostming:
This request has been implemented in d47604892a915ddf0b2a130dadaf01376442ae6a, so close it now.

PR #2670, opened 2018-07-29
Add verbose flag to run

sinscary:
This adds support for --verbose flag to run method
If flag is set more information about
command will be shown.

Closes https://github.com/pypa/pipenv/issues/2367


uranusjr:
Hi, thanks for the work! Unfortunately I seemed to have broken the patch when merging another PR, could you rebase and fix the problems for me?

I do have some doubts on whether the message shown is clear enough for the user, so it is not mistaken as the actual command output. Maybe some extra description (such as _Running command XXX (YYY)_) would be more explicit. It would also be a good idea to echo this to stderr instead, since we send most verbose messages there (e.g. the .env one).

Finally, is there a reason why this is only done for `do_run_posix`, not `do_run_nt`?

sinscary:
Oh sorry, I forgot about `do_run_nt`. Yeah, I will rebase it and update the message also.

techalchemy:
FYI we have allowed arguments to `run` in the past and decided to not to this anymore so before you move ahead with this api change you will need to make sure Kenneth is ok with it and that it will still allow people to pass`--verbose` as an argument to the thing they are running.

uranusjr:
@techalchemy I thought the same but this seems to the trick, Click seems to become a lot smarter in recent versions. 

```
λ pipenv run flake8 .\tests\unit\test_core.py                                                                                                                                                       
Loading .env environment variables…                                                                                                                                                                 
c:\users\uranusjr\documents\programming\pipenv\.venv\lib\site-packages\pycodestyle.py:113: FutureWarning: Possible nested set at position 1                                                         
  EXTRANEOUS_WHITESPACE_REGEX = re.compile(r'[[({] | []}),;:]')                                                                                                                                     
                                                                                                                                                                                                    
λ pipenv run flake8 --verbose .\tests\unit\test_core.py                                                                                                                                             
Loading .env environment variables…                                                                                                                                                                 
flake8.plugins.manager    MainProcess    234 INFO     Loading entry-points for "flake8.extension".                                                                                                  
[... 100 lines of flake8 verbose stuff]                                                                                               
                                                                                                                                                                                                    
λ pipenv run --verbose flake8 .\tests\unit\test_core.py                                                                                                                                             
Loading .env environment variables…                                                                                                                                                                 
Running command flake8 (C:\Users\uranusjr\Documents\programming\pipenv\.venv\Scripts\flake8.exe)                                                                                                    
c:\users\uranusjr\documents\programming\pipenv\.venv\lib\site-packages\pycodestyle.py:113: FutureWarning: Possible nested set at position 1                                                         
  EXTRANEOUS_WHITESPACE_REGEX = re.compile(r'[[({] | []}),;:]')                                                                                                                                     
```

The `--verbose` flag added in this PR is only triggered if put before the command (i.e. between `pipenv` and `run`, or immediately after `run`), and does not eat flags passed to the command. A `--verbose` flag passed to the command also does not trigger the Click flag.

For the off-chance there is a command called `-v` or `--verbose`, you can do `pipenv run -- --verbose ...`, and Click does the right thing. As far as I can tell all the based are covered quite well…

uranusjr:
I think we can separate this into two parts:

1. Provide extra information in verbose mode
2. Add `--verbose` and `-v` to `run` to activate verbose mode

The first one in uncontrovertial since there would be no impact to the `run` command. The verbose mode can be activated via environment variables `PIPENV_VERBOSE=1`. Personally I think the second one is doable, but I can understand the hesitance over it.

I am +1 if this only contains the first part. We can off-load the second part into a standalone issue and make sure it works more slowly.

techalchemy:
I have no issues with either approach as long as it behaves as you mention. I’d be hesitant to include the api change if I’m cutting a release tomorrow

techalchemy:
So I think we want to drop the click option here for now and then we are good to go for the current release

sinscary:
@techalchemy  I think for now removing the `--verbose` option would be good. That way as @uranusjr suggested if `PIPENV_VERBOSE` env variable is set only than it will print the verbose output.

What I understood from your suggestion is removing the `click.echo` statement, which will not print anything even if verbose argument is passed to command. Please correct me if I got it wrong. :)

sinscary:
@techalchemy @uranusjr should I remove `--verbose` option for now then ?

sinscary:
Sorry my bad, clicked wrong button.

techalchemy:
Thanks for the cleanup on this and sorry for the slow turn around

@uranusjr assuming you're good with this I am ok to merge

sinscary:
@techalchemy @uranusjr Is there anything else pending on this task? I see PR is still open.

sinscary:
@uranusjr @techalchemy Are there any more changes ?

kennethreitz:
Please use the new PEEP process: 

PR #2671, opened 2018-07-29
Update basics.rst with `$SHELL`.

HopeForToday:
State the role of `$SHELL` in `pipenv shell` command.


uranusjr:
Hi, the shell detection went through a couple of changes in recent versions. The current logic works like this:

1. Check if `PIPENV_SHELL` is set—use it if available.
2. Detect the current shell automatically.
3. If the detection fails, use `SHELL` or `PYENV_SHELL`.
4. If either of the above two is set, use `COMSPEC` (added in master, will be in the next version).
5. Give up

So setting `SHELL` is less useful in the current version. We encourage users to try to rely on auto shell detection, and set `PIPENV_SHELL` if that fails. Could you expand the section to reflect this behaviour better, so it can be more useful to more people? A link to the [environment variable section](https://docs.pipenv.org/advanced/#configuration-with-environment-variables) would also be awesome.

Thanks for the work! Documentation contribution is the best  ❤️❤️❤️❤️

HopeForToday:
Ok, I'll try it, as my English is poor. My draft may need your correction.

HopeForToday:
The 4th point, may be `neither` rather than `either`? As I googled `COMSPEC`, it's one of the environment variables used in DOS, OS/2 and Windows.

uranusjr:
Ah yes, sorry, typo :)

HopeForToday:
I just updated it, can you check it?

uranusjr:
Hi, this looks good! Only one thing—can you switch the link to use Sphinx’s [reference syntax](https://stackoverflow.com/questions/4385315) instead?

HopeForToday:
En, I updated it again.

docs/basics.rst
```diff
@@ -326,6 +326,7 @@ You should do this for your shell too, in your ``~/.profile`` or ``~/.bashrc`` o
 
 .. note:: The shell launched in interactive mode. This means that if your shell reads its configuration from a specific file for interactive mode (e.g. bash by default looks for a ``~/.bashrc`` configuration file for interactive mode), then you'll need to modify (or create) this file.
 
+If you get something wrong with ``$ pipenv shell``, just check ``PIPENV_SHELL`` environment variable, ``$ pipenv shell`` will use it if available. For detail, see :ref:`configuration-with-environment-variables`.
```
techalchemy: > If you get something wrong with 

Can you change this to 'If you experience issues with`? Minor change but I think it's a tiny bit more natural.  

> just check ``PIPENV_SHELL`` environment variable

~~just check ``PIPENV_SHELL`` environment variable, ``$ pipenv shell``~~ just check **the** ``PIPENV_SHELL`` environment variable, **which** ``$ pipenv shell`` **will use** if available.

Minor tense changes mainly, thanks for contributing this!

HopeForToday: Updated.

HopeForToday:
Just updated.

PR #2674, opened 2018-07-30
Add link to PEP-440 version specifiers in documentation

mic159:
It's currently hard to find the list of version matching operators supported by Pipenv.

I think there should be a link to it in the basics doc under "Specifying Versions of a Package".


uranusjr:
I think you need one (named) or two (unnamed) underscores after the hyperlink. I would strongly recommend using the out-of-line named syntax as well:

```rst
blah blah `my link`_

.. _`my link`: https://my.link
```

Also, if the text is “PEP 440”, I would simply link to the page; if you want to link to the section specifically, I would alter the text to something like “relevant section in PEP 440“.

Thanks so much fo the contribution! I ❤️ documentation patches.

PR #2690, opened 2018-08-02
Properly handle pre-existing .venv files

### LINKED ISSUE #2680: Pipenv breaks with pre-existing usage of .venv files
#### Issue description
tl;dr `.venv` files already exist in some dev setups but when present, `pipenv` attempts to use that file as a directory and install a virtualenv into it.

Virtualenvwrapper has suggested for awhile to use `.venv` files in your project directory to hold the name of your virtualenv so that you could automatically activate virtualenvs.  Spacemacs (an Emacs editor setup) also does this behavior by default.

Two problems:

1. Pipenv will silently fail to create and activate a virtualenv when a `.venv` file is present.
2. If a `.venv` file is added to an existing pipenv project, pipenv breaks because it can't activate the virtualenv at `<project_dir>/.venv` ignoring the place where it had already created a virtualenv.

Prior Art Links:
 - Virtualenvwrapper Tips & Tricks: http://virtualenvwrapper.readthedocs.io/en/latest/tips.html
 - Spacemacs Python Layer:  https://github.com/syl20bnr/spacemacs/tree/develop/layers/%2Blang/python#automatic-activation-of-local-pyenv-version
 - Virtualenv in git directories: https://hmarr.com/2010/jan/19/making-virtualenv-play-nice-with-git/
 - Sample .bash_profile using `.venv` files: https://github.com/justinabrahms/jlilly-bashy-dotfiles/commit/04899f005397499e89da6d562b062545e70d7975#commitcomment-1526375

#### Expected result

At the least, Pipenv should check if `.venv` is a file or a directory before attempting to automatically use it as a directory and install a virtualenv into it.  If it's a file, then behave as if no `.venv` was found at all.

#### Actual result

If a `.venv` file exists in the project directory, Pipenv will attempt to create and use a virtualenv at `<project_dir>/.venv` and yell at the user a bit because it's not a directory but (a) it won't actually fail and (b) it won't actually save a virtualenv _anywhere_.

Output:
```
Pipfile: <project_path>\Pipfile
Using <python_path> to create virtualenv...
Running virtualenv with interpreter <python_path>
ERROR: File already exists and is not a directory.
Please provide a different path or delete the file.

Virtualenv location: <project_path>\.venv
```

#### Steps to replicate

`touch .venv`
`pipenv install`

Also see output of `pipenv --venv`

*REDACTED*

Pipenv version: `'2018.7.1'`

PEP 508 Information:

```
 'implementation_version': '3.6.5',
 'os_name': 'nt',
 'platform_machine': 'AMD64',
 'platform_release': '10',
 'platform_system': 'Windows',
 'platform_version': '10.0.17134',
 'python_full_version': '3.6.5',
 'python_version': '3.6',
 'sys_platform': 'win32'}
```

System environment variables:

  - `ADSK_CLM_WPAD_PROXY_CHECK`
  - `ALIASES`
  - `ALLUSERSPROFILE`
  - `ANSICON`
  - `ANSICON_DEF`
  - `APPDATA`
  - `ARCHITECTURE`
  - `CHOCOLATEYINSTALL`
  - `CHOCOLATEYLASTPATHUPDATE`
  - `CMDER_ROOT`
  - `COMMONPROGRAMFILES`
  - `COMMONPROGRAMFILES(X86)`
  - `COMMONPROGRAMW6432`
  - `COMPUTERNAME`
  - `COMSPEC`
  - `CONEMUANSI`
  - `CONEMUANSILOG`
  - `CONEMUARGS`
  - `CONEMUARGS2`
  - `CONEMUBACKHWND`
  - `CONEMUBASEDIR`
  - `CONEMUBASEDIRSHORT`
  - `CONEMUBUILD`
  - `CONEMUCFGDIR`
  - `CONEMUCONFIG`
  - `CONEMUDIR`
  - `CONEMUDRAWHWND`
  - `CONEMUDRIVE`
  - `CONEMUHOOKS`
  - `CONEMUHWND`
  - `CONEMUPALETTE`
  - `CONEMUPID`
  - `CONEMUSERVERPID`
  - `CONEMUTASK`
  - `CONEMUWORKDIR`
  - `CONEMUWORKDRIVE`
  - `DRIVERDATA`
  - `FPS_BROWSER_APP_PROFILE_STRING`
  - `FPS_BROWSER_USER_PROFILE_STRING`
  - `FSHARPINSTALLDIR`
  - `GIT_INSTALL_ROOT`
  - `GOPATH`
  - `GOROOT`
  - `GTK_BASEPATH`
  - `HOME`
  - `HOMEDRIVE`
  - `HOMEPATH`
  - `LOCALAPPDATA`
  - `LOGONSERVER`
  - `NUMBER_OF_PROCESSORS`
  - `ONEDRIVE`
  - `OS`
  - `PATH`
  - `PATHEXT`
  - `PLINK_PROTOCOL`
  - `PROCESSOR_ARCHITECTURE`
  - `PROCESSOR_IDENTIFIER`
  - `PROCESSOR_LEVEL`
  - `PROCESSOR_REVISION`
  - `PROGRAMDATA`
  - `PROGRAMFILES`
  - `PROGRAMFILES(X86)`
  - `PROGRAMW6432`
  - `PROMPT`
  - `PSMODULEPATH`
  - `PUBLIC`
  - `RUBYOPT`
  - `SESSIONNAME`
  - `SVN_SSH`
  - `SYSTEMDRIVE`
  - `SYSTEMROOT`
  - `TEMP`
  - `TERM`
  - `TMP`
  - `USER-ALIASES`
  - `USERDOMAIN`
  - `USERDOMAIN_ROAMINGPROFILE`
  - `USERNAME`
  - `USERPROFILE`
  - `VERBOSE-OUTPUT`
  - `WINDIR`
  - `WORKON_HOME`
  - `PYTHONDONTWRITEBYTECODE`
  - `PIP_PYTHON_PATH`

Pipenvûspecific environment variables:

---------------------------

Contents of `Pipfile`:

```toml
[[source]]
verify_ssl = true
name = "pypi"

[scripts]
server = "python manage.py runserver"
migrate = "python manage.py migrate"
test = "python manage.py test"

[packages]
"boto3" = "==1.7"
dj-database-url = "==0.5.0"
json-logging-py = "*"
"psycopg2" = "==2.7"
pytz = "==2018.4"
requests = "==2.18.4"
Django = "==2.0.6"
django_csp = "==3.4"
"Jinja2" = "==2.10"
djangorestframework = "*"

[dev-packages]
ipython = "*"
pylint = "*"
pylint-django = "*"
pylint-plugin-utils = "*"
autoflake = "*"
importmagic = "*"
epc = "*"

[requires]
python_version = "3.6"

```
</details>

MarkKoz:
Fixes #2680 as discussed in the issue with bonus support for custom paths in the .venv file. Not sure what to categorise this under for the news fragment so I will wait for someone's advice on that before writing up the news fragment.

Would like reviewers to especially look at the test cases written. Could the repeated code be moved elsewhere? Are path comparisons properly handled? - concerns are accounting for symlinks and case-insensitivity on Windows.


uranusjr:
Hi, thanks for the work! I would recommend splitting this into two PRs instead—we are very close to feature freeze for a new version, and while the `exists` → `isdir` change can easily get in (as a bugfix), the `.venv` file detection would require additional reviews.

MarkKoz:
Sure. How do you recommend I split the PR? Create two new ones or try to salvage this PR by rewriting history?

uranusjr:
Either would do, whatever is more convenient for you :)

pipenv/project.py
```diff
@@ -267,7 +267,18 @@ def virtualenv_exists(self):
     def get_location_for_virtualenv(self):
         if self.is_venv_in_project():
             return os.path.join(self.project_directory, ".venv")
-        return str(get_workon_home().joinpath(self.virtualenv_name))
+
+        name = self.virtualenv_name
+        if self.project_directory:
+            venv_path = os.path.join(self.project_directory, ".venv")
+            if os.path.isfile(venv_path):
```
techalchemy: If you use `Path(venv_path)`  you can use the check `.exists() and not Path(venv_path).isdir()` for files. You should do the same path object instantiation on the files contents to make sure the virtualenv exists.

MarkKoz: Why `not isdir()` rather than just `isfile()`? Furthermore, how would using pathlib over os.path help? Seems to me like an equivalent check can be written with os.path. I like the pathlib but I've used os.path just to remain consistent with the rest of the code.

> You should do the same path object instantiation on the files contents to make sure the virtualenv exists.

Is this just because the path wont be created if it doesn't exist when the venv is being created?

MarkKoz:
Should a test case still be written for the `exists` → `isdir` bugfix when I split the PRs? If so, that test case will probably need to be removed by this PR since the two test cases I've already written will make the other one redundant.

pipenv/project.py
uranusjr: There are edge cases where `isfile` and `isdir` can both be false, such as [device files](https://en.wikipedia.org/wiki/Device_file). While it doesn’t really matter practically, it is semantically more correct to check for “not directory” because those special files are still readable. It also matches the directory check in `is_venv_in_project`, which makes the code easier to reason with.

I’m wondering though, it may be a better idea to merge the two checks. `is_venv_in_project` already checks for directory, so when you reach this part, an `exists` check is enough. We shouldn’t reach here if it’s a directory; if it exists, we want to try reading it (and fails with a straight-up error if that fails).

uranusjr:
It would be best to have a test for the `isdir` change, but redundant tests are not a problem.

MarkKoz:
Just rewriting history to split the PRs. Will  try prioritising the bugfix PR and then come back to this PR once I've finished the former.

pipenv/project.py
MarkKoz: I agree that it's better to rely on `is_venv_in_project`'s directory check. `exists` should be sufficient here.

What should be done when the directory specified in a .venv file does not exist? That hasn't been addressed yet.

uranusjr: It would need to fail back to use `self.virtualenv_name`. Now you mentioned it, I realise the code is a bit tangled here and the implementation doesn’t work well. I’ll try to find some time to sort the structure out so this can progress.

pipenv/project.py
```diff
@@ -267,7 +267,18 @@ def virtualenv_exists(self):
     def get_location_for_virtualenv(self):
         if self.is_venv_in_project():
             return os.path.join(self.project_directory, ".venv")
-        return str(get_workon_home().joinpath(self.virtualenv_name))
+
+        name = self.virtualenv_name
```
techalchemy: No need to check for the presence of the project directory, the instantiation of a project in the first place depends on the project directory existing

given the current situation this can just become something like

```python
def get_location_for_virtualenv(self):
    if self.is_venv_in_project()
        return os.path.join(self.project_directory, ".venv")
    elif os.path.exists(self.path_to('.venv')):
        venv_contents = None
        try:
            with open(self.path_to('.venv'), 'r') as fh:
                venv_contents = fh.read()
        except OSError:
            pass
        else:
            venv_name = venv_contents.strip()
            if any(sep in venv_name for sep in [os.path.sep, os.path.altsep]):
                return _normalized(venv_name)
            elif venv_name:
                return str(get_workon_home().joinpath(venv_name))
    return str(get_workon_home().joinpath(self.virtualenv_name))
```

techalchemy: thoughts @uranusjr ?

tests/integration/test_dot_venv.py
```diff
@@ -41,15 +41,15 @@ def test_reuse_previous_venv(PipenvInstance, pypi):
...
+
 def test_venv_file_exists(PipenvInstance, pypi):
     """Tests virtualenv creation & package installation when a .venv file exists
     at the project root.
     """
     with PipenvInstance(pypi=pypi, chdir=True) as p:
         file_path = os.path.join(p.path, '.venv')
         with open(file_path, 'w') as f:
-            f.write('test-project')
+            f.write('')
```
MarkKoz: The merge should have kept `test_venv_file_exists' over `test_venv_file_exists` (or kept both). We do want to write something to the file so we can test if the written name is used to find/create a venv of the same name in workon home.

pipenv/project.py
MarkKoz: > No need to check for the presence of the project directory

Wouldn't that mean the same check in `is_venv_in_project` is also redundant?

pipenv/project.py
MarkKoz: `virtualenv` just creates the directory if it doesn't exist. I consider that to be a feature, so I don't think we need to fall back to `self.virtualenv_name` after all.

tests/integration/test_dot_venv.py
techalchemy: hmmmmm I think I prioritized your test over the one he wrote, but that wasn't really intentional, I was just trying to triage and get everything moving so I can get a release out soon(tm). There is a test right below that one though and it does write a name of a directory to the file which is why I undid my inclusion of it here

The merge removed this project name in favor of creating a temporary directory in the test below it and using that directory's location as the path in the `,venv` file.

MarkKoz: The difference between the two tests was one wrote a path to the file, the other just a name. If you think that the former adequately covers the latter case too then all is fine as is.

uranusjr: I think those should be merged into one with pytest fixtures. The fixture would be something like

```python
@pytest.mark.fixtures('content, venv_location', [
    ('just-a-name', Path(WORKON_HOME, 'just-a-name')),
    ('/an/absolute/path', Path('/an/absolute/path')),
    ('a/relative/path', Path(project_root, 'a/relative/path')),
])
```

(I added a third test case. If a relative path is specified, the venv should be created relative to the project’s root.)

MarkKoz: Good call, but does it really need to be a fixture if we'll only have one test that uses it? Could just use `@pytest.mark.parametrize`. 

I'm not sure how `WORKON_HOME` and `project_root` be passed. My first thought was fixtures, but pytest currently doesn't support the use of fixtures in `@pytest.mark.parametrize`. Would it work if it was left a fixture, and a fixture was used for the fixture's params?

uranusjr: Ah yes, excellent call, I was totally thinking about `parametrize`, not fixtures!

techalchemy: I don’t think we need to refactor tests to merge this PR but the point about a path vs a name is valid. You can’t parametrize a temporary directory as an input though so that would still need a separate test

MarkKoz: Yeah, I played around with this idea of parametrisation but couldn't come up with any elegant solution - it'd create a bigger mess than it is intended to prevent.

MarkKoz: Do you mind the messy commit history? I could just commit over everything to make the test write a name again, or I could just rewrite history and re-do the merge.

uranusjr: I usually squash the commits a bit before merging, so make it as messy as you need to :p

MarkKoz:
I don't know what's causing the build to fail on Buildkite - I can't see the results. Everything is fine on my end, other than `test_environment_variable_value_does_not_change_hash` which has failed from the beginning because it tries to access anaconda2 (isn't installed on my machine).

techalchemy:
the build is failing for unclear reasons relating to a tlsv1 error now on buildkite, which is basically unusable I guess.  I'll turn travis back on tomorrow if I can't work it out

MarkKoz:
It's fine; priorities are priorities.  Don't worry I'm not going anywhere. Before continuing, I just wanted to make sure my changes weren't the cause of the build failure.

techalchemy:
I really can't tell what's going on here but it's failing to connect to the local package index due to an SSL issue? this is the only branch / test suite that fails on buildkite and i really don't know why...

MarkKoz:
Is it failing to connect while running one of my tests or at some other point? The major changes to the tests were changing how the venv location is retrieved, doing `install` instead of `install requests`, and of course adding a third test case. Perhaps these changes could be reverted one by one to narrow down the search for the offending code, if any. 

However, I don't see anything obviously wrong with the tests and they do all pass on my machine. Maybe some underlying issue has been unearthed?

What's the course of action? I'm not sure how I can help here.

techalchemy:
I don't see anything wrong either, let me check your branch out locally and see if I have any issues... I'm super confused about this one at the moment and I'm wondering if there is some unrelated ssl issue on the build machine

The tests look fine, and the connection that's failing is to localhost... I'll show you whenever buildkite's frontend is willing to let me copy things

techalchemy:
```
=================================== FAILURES ===================================
_____________________________ test_mirror_install ______________________________
[gw0] linux2 -- Python 2.7.15 /root/.local/share/virtualenvs/pipenv-R0iOVi4H-2.7/bin/python2.7
 
PipenvInstance = <class 'tests.integration.conftest._PipenvInstance'>
pypi = <pytest_pypi.serve.Server object at 0x7fb6dd8198d0>
 
    @pytest.mark.install
    @flaky
    def test_mirror_install(PipenvInstance, pypi):
            mirror_url = os.environ.pop(
                "PIPENV_TEST_INDEX", "https://pypi.python.org/simple"
            )
            assert "pypi.org" not in mirror_url
            # This should sufficiently demonstrate the mirror functionality
            # since pypi.org is the default when PIPENV_TEST_INDEX is unset.
            c = p.pipenv("install requests --pypi-mirror {0}".format(mirror_url))
>           assert c.return_code == 0
E           AssertionError: assert 1 == 0
E            +  where 1 = <Command 'pipenv install requests --pypi-mirror https://127.0.0.1:42425/simple'>.return_code
 
tests/integration/test_install_basic.py:55: AssertionError
----------------------------- Captured stdout call -----------------------------
pytest-httpbin server hit an exception serving request: [SSL: TLSV1_ALERT_UNKNOWN_CA] tlsv1 alert unknown ca (_ssl.c:726)
attempting to ignore so the rest of the tests can run
pytest-httpbin server hit an exception serving request: [SSL: TLSV1_ALERT_UNKNOWN_CA] tlsv1 alert unknown ca (_ssl.c:726)
attempting to ignore so the rest of the tests can run
pytest-httpbin server hit an exception serving request: [SSL: TLSV1_ALERT_UNKNOWN_CA] tlsv1 alert unknown ca (_ssl.c:726)
$ pipenv install requests --pypi-mirror https://127.0.0.1:42425/simple
Installing requests…
Looking in indexes: https://127.0.0.1:42425/simple
Collecting requests
  Could not fetch URL https://127.0.0.1:42425/simple/requests/: There was a problem confirming the ssl certificate: HTTPSConnectionPool(host='127.0.0.1', port=42425): Max retries exceeded with url: /simple/requests/ (Caused by SSLError(SSLError(1, u'[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:726)'),)) - skipping
 
Pipfile: /tmp/pipenv-lkEOTs-project/Pipfile
Using /root/.local/share/virtualenvs/pipenv-R0iOVi4H-2.7/bin/python2.7 (2.7.15) to create virtualenv…
Already using interpreter /root/.local/share/virtualenvs/pipenv-R0iOVi4H-2.7/bin/python2.7
Using real prefix '/usr'
New python executable in /tmp/pipenv-lkEOTs-project/.venv/bin/python2.7
Also creating executable in /tmp/pipenv-lkEOTs-project/.venv/bin/python
 
Virtualenv location: /tmp/pipenv-lkEOTs-project/.venv
Error:  An error occurred while installing requests!
  Retrying (Retry(total=4, connect=None, read=None, redirect=None, status=None)) after connection broken by 'SSLError(SSLError(1, u'[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:726)'),)': /simple/requests/
  Retrying (Retry(total=3, connect=None, read=None, redirect=None, status=None)) after connection broken by 'SSLError(SSLError(1, u'[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:726)'),)': /simple/requests/
  Retrying (Retry(total=2, connect=None, read=None, redirect=None, status=None)) after connection broken by 'SSLError(SSLError(1, u'[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:726)'),)': /simple/requests/
  Retrying (Retry(total=1, connect=None, read=None, redirect=None, status=None)) after connection broken by 'SSLError(SSLError(1, u'[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:726)'),)': /simple/requests/
  Retrying (Retry(total=0, connect=None, read=None, redirect=None, status=None)) after connection broken by 'SSLError(SSLError(1, u'[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:726)'),)': /simple/requests/
  Could not find a version that satisfies the requirement requests (from versions: )
No matching distribution found for requests
```

MarkKoz:
Yeah, I also have no idea. Double checked that test and it indeed passes on my machine. Did you manage to test it locally. If so, how did that go?

Issue #2692, opened 2018-08-02
Support the --system flag in `pipenv run`


orf:
##### Is your feature request related to a problem? Please describe.

When using `pipenv` in a docker container you can install dependencies using the `--system` flag. However if you use `pipenv run` in the docker container it will automatically create a virtualenv and there is no way to stop this.

##### Describe the solution you'd like

Provide a `--system` flag to `pipenv run` that skips the creation of a virtualenv and instead uses the system default.

techalchemy:
I believe we have rejected this in the past because when you use `--system` there is no reason to use `pipenv run` at all — you can simply run the command.

orf:
Having the commands defined in the `Pipfile` has it's advantages. We have a `prod` and `dev` command defined so our base docker image just executes `pipenv run prod`. This is quite nice as the commands are in a single place and anyone can get working, plus the docker image just needs to run a single command for all projects so less duplication.

Also in our case `pipenv run dev` executes `pipenv run prod` with some extra development specific arguments.

orf:
All in all it's just a bit inconsistent. You can take advantage of the `Pipfile` (or specifically the `Pipfile.lock`) to install system dependencies, but you cannot make use of `[scripts]` for no clear reason.

techalchemy:
Yeah I agree with you. If @uranusjr agrees I can bring it up with Kenneth and see if we can find a consensus. 

Out of curiosity is your pipfile usage here inspired by node-style package files?

orf:
It's not that it's inspired by 'node-style package files' more that it's a common use case that node-style package files happen to solve in a similar way.

It's just very convenient to use the `[scripts]` section to have a consistent set of commands across projects, much like I guess a `makefile` might give you. Not being able to run these scripts after using `pipenv install --system` without bootstrapping a whole virtualenv and installing some dependencies from the network is a curious/surprising limitation that should definitely be documented.

uranusjr:
I am -0 on this. On one hand I can understand it is convinient to be able to use Pipfile as a script holder, but on the other hand this really feels out-of-place. On yet another hand (OK I know this is not how this expression works) the whole `--system` thing is out of place anyway, so I guess I don’t really care if this is added. There really needs to be a better bare-bone tool to interact with Pipfile without jumping through virtual environment hoops (wait [I actually made onet](https://github.com/sarugaku/pipfile-cli) but nobody uses i).

techalchemy:
[...]

PR #2693, opened 2018-08-02
Add a --system flag to pipenv run

orf:
##### The issue

Issue: https://github.com/pypa/pipenv/issues/2692

`pipenv run` does not support the `--system` flag, and so it will create a virtualenv when called. This is problematic inside a docker container where the dependencies are typically installed using `pipenv install --system`. It would be nice to be able to use the system interpreter with `pipenv run`.

##### The fix

I added a `--system` flag and a test

##### The checklist

* [X] Associated issue

### DIFF
--- a/news/2692.feature
+++ b/news/2692.feature
@@ -1,0 +1,1 @@
+Added a ``--system`` flag to ``pipenv run``
--- a/pipenv/cli.py
+++ b/pipenv/cli.py
@@ -680,12 +680,17 @@ def shell(
     callback=validate_pypi_mirror,
     help="Specify a PyPI mirror.",
 )
-def run(command, args, three=None, python=False, pypi_mirror=None):
+@option(
+    "--system",
+    is_flag=True,
+    default=False
+)
+def run(command, args, three=None, python=False, pypi_mirror=None, system=False):
     """Spawns a command installed into the virtualenv."""
     from .core import do_run
 
     do_run(
-        command=command, args=args, three=three, python=python, pypi_mirror=pypi_mirror
+        command=command, args=args, three=three, python=python, pypi_mirror=pypi_mirror, system=system
     )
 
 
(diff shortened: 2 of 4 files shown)


orf:
Why are we not able to see the build failures?

techalchemy:
```
assert not project.virtualenv_exists
E           assert not True
E            +  where True = <pipenv.project.Project object at 0x7fd4c438b0d0>.virtualenv_exists
 
tests/integration/test_run.py:71: AssertionError
----------------------------- Captured stdout call -----------------------------
$ pipenv run --system printfoo
foo

```

techalchemy:
See my comment on the issue — the api change here has been decided against in the past so would like to hear an argument for changing that decision

orf:
Thanks! I will look fixing this if this idea is accepted, I left [a comment](https://github.com/pypa/pipenv/issues/2692#issuecomment-409946155) in the issue.

kennethreitz:
run --system makes no sense.

orf:
If that is the case then `pipenv install --system` also makes no sense and should be removed.

orf:
I appreciate that, and I am respectful of your past and current efforts (hence trying to make the PR myself!), but this is definitely a confusing from a usability perspective. @uranusjr commented that [the whole --system thing feels out of place](https://github.com/pypa/pipenv/issues/2692#issuecomment-410144060), and I somewhat agree, but consistency is king. Having it on `install`,  `uninstall` and `check`, but then not `run` for somewhat fuzzy reasons is confusing.

Anyway, your decision is yours and I understand, thanks for your ongoing work with pipenv.

kennethreitz:
Would you mind writing a PEEP for this?

orf:
Sure, I'd love to. Do you mean specifically for `--system` being consistent or `--system` being available in `run`?

kennethreitz:
Consistent!

PR #2696, opened 2018-08-02
Improves message when dependencies could not be resolved.

gzagatti:
##### The issue

`pipenv` might fail to resolve dependencies when the cache is not cleared. The error message is uninformative.

##### The fix

This PR will print a more useful error message when dependencies fail to resolve, instructing the user to first try clearing the dependency cache with `$ pipenv lock --clear`.

This PR has been submitted based on the discussions in #2596.

##### The checklist

* [X] Associated issue

### DIFF
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -310,11 +310,13 @@ def actually_resolve_deps(
         click_echo(
             "{0}: Your dependencies could not be resolved. You likely have a "
             "mismatch in your sub-dependencies.\n  "
-            "You can use {1} to bypass this mechanism, then run {2} to inspect "
-            "the situation.\n  "
-            "Hint: try {3} if it is a pre-release dependency."
+            "First try clearing your dependency cache with {1} and trying again.\n "
+            "Alternatively, you can use {2} to bypass this mechanism, then run "
+            "{3} to inspect the situation.\n  "
+            "Hint: try {4} if it is a pre-release dependency."
             "".format(
                 crayons.red("Warning", bold=True),
+                crayons.red("$ pipenv lock --clear"),
                 crayons.red("$ pipenv install --skip-lock"),
                 crayons.red("$ pipenv graph"),
                 crayons.red("$ pipenv lock --pre"),
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -310,7 +310,7 @@ def actually_resolve_deps(
         click_echo(
             "{0}: Your dependencies could not be resolved. You likely have a "
             "mismatch in your sub-dependencies.\n  "
-            "First try clearing your dependency cache with {1} and trying again.\n "
+            "First try clearing your dependency cache with {1}, then try the original command again.\n "
             "Alternatively, you can use {2} to bypass this mechanism, then run "
             "{3} to inspect the situation.\n  "
             "Hint: try {4} if it is a pre-release dependency."


pipenv/utils.py
```diff
@@ -310,11 +310,13 @@ class PipCommand(basecommand.Command):
         click_echo(
             "{0}: Your dependencies could not be resolved. You likely have a "
             "mismatch in your sub-dependencies.\n  "
-            "You can use {1} to bypass this mechanism, then run {2} to inspect "
-            "the situation.\n  "
-            "Hint: try {3} if it is a pre-release dependency."
+            "First try clearing your dependency cache with {1} and trying again.\n "
```
uranusjr: This sentense need to be rewritten. First try… trying?

pipenv/utils.py
```diff
@@ -310,11 +310,13 @@ class PipCommand(basecommand.Command):
...
+            "Alternatively, you can use {2} to bypass this mechanism, then run "
+            "{3} to inspect the situation.\n  "
+            "Hint: try {4} if it is a pre-release dependency."
             "".format(
                 crayons.red("Warning", bold=True),
+                crayons.red("$ pipenv lock --clear"),
                 crayons.red("$ pipenv install --skip-lock"),
                 crayons.red("$ pipenv graph"),
                 crayons.red("$ pipenv lock --pre"),
```
uranusjr: You don’t really need to shift all the indexes, but you already did, so I guess it’s fine either way.

pipenv/utils.py
gzagatti: True, I have corrected the sentence.

pipenv/utils.py
gzagatti: I didn't know. Thought they had to be in order. At least they are better organised like this.

gzagatti:
@techalchemy, I am not very familiar with the codebase to understand all the nuances. But my understanding is that [we initiate a `Resolver` class with a number of `constraints`](https://github.com/pypa/pipenv/blob/master/pipenv/utils.py#L300). In the iterative process of resolving the constraints, if there are conflicting constraints a [`NoCandidateFound`](https://github.com/pypa/pipenv/blob/master/pipenv/patched/piptools/exceptions.py#L5) error will be raised that lists the conflicting package.

The resolver resolution is run within a `try...except...`, if the `NoCandidateFound` error is found we will print the message which suggests how to solve the issue (the message which I fixed on this PR). Followed by the error message from the `NoCandidateFound` error which lists the package which has an unresolved dependency.

I believe that if we wanted to highlight the packages which are causing trouble, we would need to pass the original packages to `NoCandidateFound`. [At the moment we pass `ireq`, `all_candidates` and `self.finder`](https://github.com/pypa/pipenv/blob/master/pipenv/patched/piptools/repositories/pypi.py#L183). I would need to figure out what those variables are and how to retrieve the packages causing the original conflict from there. 

Without a good grasp of the codebase, I am not sure if I can do this quickly. What do you think?

techalchemy:
I wouldn't spend too much time on it... we just [rewrote the resolver](https://github.com/sarugaku/passa), [broken into multiple pieces](https://github.com/sarugaku/resolvelib), and split apart the pip elements into their own [shim library](https://github.com/sarugaku/pip-shims)

My only real question in order to move forward on this is: what is the result of the change, in terms of look and feel? is there a screenshot or something I can look at

techalchemy:
I'm going to merge this as is -- thanks for the changes!

Issue #2685, opened 2018-08-01
guide: add example to install from git branch or tag


glensc:
##### Is your feature request related to a problem? Please describe.

https://realpython.com/pipenv-guide/#example-usage

please add here example how to add package from git branch "develop" or tag "1.0.0". it is not obvious for persons who started to use pipenv and not familiar with python ecosystem.

techalchemy:
We don't control that website, unfortunately, but if it's not become clear to you yet you can do this from the command line with `pipenv install -e git+https://github.com/user/repo.git@1.0.0#egg=project_name`

glensc:
appending `@1.0.0` is not self-obvious. that's why example needed.

i tried `&ref=develop`, but eventually just used `vim` on `Pipfile` to specify `ref`.

glensc:
such examples could be present in this project `README.md` as well.

akefeli:
Also it took me a while to figure out you needed to escape & in the url for subdirectory.
This doesn't work:
pipenv install -e git@github.com:some/repo.git#egg=foo&subdirectory=python
This works:
pipenv install -e git@github.com:some/repo.git#egg=foo**\**&subdirectory=python

Is this even documented somewhere?

glensc:
you need to escape `&` for shell. so these are equivalent:

```
pipenv install -e "git@github.com:some/repo.git#egg=foo&subdirectory=python"
pipenv install -e 'git@github.com:some/repo.git#egg=foo&subdirectory=python'
pipenv install -e git@github.com:some/repo.git#egg=foo\&subdirectory=python
```

https://www.tldp.org/LDP/abs/html/special-chars.html

akefeli:
@glensc you are right, for some reason I thought same command was working with pip but not with pipenv, I must be using quotes with pip

techalchemy:
I think we are all set on this front now (merged #2702) -- thanks for opening the issue and thanks for the helpful back and forth all!

kushan-gunasekera:
This is how I solved it,

`pipenv install -e git+https://github.com/some_user/some_package.git#egg=some_package`

hivemall:
How to checkout a tag and install from https://github.com/ccxt/ccxt, yet its a js/php/py repo with python subfolder?

coltonbh:
If you want to install via ssh (say, for a private repo), you'll need to modify the GitHub ssh url as follows:
```
    git@github.com:coltonbh/my-package.git (fetch)
[...]

PR #2702, opened 2018-08-03
Added an example to README.md for installing from git

jeremytiki:

##### The issue

There was a request for [additional documentation on installing from git](https://github.com/pypa/pipenv/issues/2685)

##### The fix

 I updated the README.md with an example to install requests, this could also be added to the official docs as well, but I figured this was a good start. I didn't add any additional information to the `news/` directory because I wasn't sure what the naming method was 

##### The checklist

* [ x ] Associated issue
* [ x ] A news fragment in the `news/` directory to describe this fix with the extension `.bugfix`, `.feature`, `.behavior`, `.doc`. `.vendor`. or `.trivial` (this will appear in the release changelog). Use semantic line breaks and name the file after the issue number or the PR #.

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,6 +216,15 @@ Install packages:
     To activate this project's virtualenv, run the following:
     $ pipenv shell
 
+Install from git:
+
+    $ pipenv install -e git+https://github.com/requests/requests.git@v2.19.1#egg=requests
+    Installing -e git+https://github.com/requests/requests.git@v2.19.1#egg=requests…
+    ...
+    Adding -e git+https://github.com/requests/requests.git@v2.19.1#egg=requests to Pipfile's [packages]...
+
+You can also change the tag/ref (or exclude to get the latest version) by adding @version-or-tag before #egg=project_name
+
 Install a dev dependency:
 
     $ pipenv install pytest --dev
--- a/README.md
+++ b/README.md
@@ -220,26 +220,24 @@ Installing from git:
 
 You can install packages with pipenv from git and other version control systems using URLs formatted according to the following rule:
 
-```
     <vcs_type>+<scheme>://<location>/<user_or_organizatoin>/<repository>@<branch_or_tag>#<package_name>
-```
 
 The only optional section is the `@<branch_or_tag>` section.  When using git over SSH, you may use the shorthand vcs and scheme alias `git+git@<location>:<user_or_organization>/<repository>@<branch_or_tag>#<package_name>`. Note that this is translated to `git+ssh://git@<location>` when parsed.
 
 Valid values for `<vcs_type>` include `git`, `bzr`, `svn`, and `hg`.  Valid values for `<scheme>` include `http,`, `https`, `ssh`, and `file`.  In specific cases you also have access to other schemes: `svn` may be combined with `svn` as a scheme, and `bzr` can be combined with `sftp` and `lp`.
 
-Note that it is **strongly recommended** that you install any version-controlled dependencies in editable mode in order to ensure that dependency resolution can be performed with an up to date copy of the repository each time it is performed, and that it includes all known dependencies.
+Note that it is **strongly recommended** that you install any version-controlled dependencies in editable mode, using `pipenv install -e`, in order to ensure that dependency resolution can be performed with an up to date copy of the repository each time it is performed, and that it includes all known dependencies.
 
 Below is an example usage which installs the git repository located at `https://github.com/requests/requests.git` from tag `v2.19.1` as package name `requests`:
 
-```console
     $ pipenv install -e git+https://github.com/requests/requests.git@v2.19#egg=requests
     Creating a Pipfile for this project...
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -216,6 +216,15 @@ Install packages:
...
     $ pipenv shell
 
+Install from git:
+
+    $ pipenv install -e git+https://github.com/requests/requests.git@v2.19.1#egg=requests
+    Installing -e git+https://github.com/requests/requests.git@v2.19.1#egg=requests…
+    ...
+    Adding -e git+https://github.com/requests/requests.git@v2.19.1#egg=requests to Pipfile's [packages]...
+
+You can also change the tag/ref (or exclude to get the latest version) by adding @version-or-tag before #egg=project_name
```
techalchemy: Would you mind doing a touch of cleanup here and adding a section that describes the syntax including the sections of the url? Suggestion below:

Can you add some syntax highlighting here at the bottom? Then you can end the line with a period.

````markdown
Installing from git:

You can install packages with pipenv from git and other version control systems using URLs formatted according to the following rule:

```
    <vcs_type>+<scheme>://<location>/<user_or_organizatoin>/<repository>@<branch_or_tag>#<package_name>
```

The only optional section is the `@<branch_or_tag>` section.  When using git over SSH, you may use the shorthand vcs and scheme alias `git+git@<location>:<user_or_organization>/<repository>@<branch_or_tag>#<package_name>`. Note that this is translated to `git+ssh://git@<location>` when parsed.

Valid values for `<vcs_type>` include `git`, `bzr`, `svn`, and `hg`.  Valid values for `<scheme>` include `http,`, `https`, `ssh`, and `file`.  In specific cases you also have access to other schemes: `svn` may be combined with `svn` as a scheme, and `bzr` can be combined with `sftp` and `lp`.

Note that it is **strongly recommended** that you install any version-controlled dependencies in editable mode in order to ensure that dependency resolution can be performed with an up to date copy of the repository each time it is performed, and that it includes all known dependencies.

Below is an example usage which installs the git repository located at `https://github.com/requests/requests.git` from tag `v2.19.1` as package name `requests`:

```console
    $ pipenv install -e git+https://github.com/requests/requests.git@v2.19#egg=requests
    Installing -e git+https://github.com/requests/requests.git@v2.19.1#egg=requests...
    [...snipped...]
    Adding -e git+https://github.com/requests/requests.git@v2.19.1#egg=requests to Pipfile's [packages]...
    [...]
```
````

jeremytiki:
I went ahead and updated the README with your excellent explaination, although I'm unsure what you meant here:

> Can you add some syntax highlighting here at the bottom? Then you can end the line with a period.

I'm still fairly new to markdown, but it seems that important console section is already syntax highlighted.

slhck:
@jeremytiki Just a minor nitpick: when you add console syntax, it should be formatted with either fenced code (with three backslashes) *or* 4-space indent, not both. If you look at the other code sections in the README, you'll notice that it only uses 4-space indents.

slhck:
Also, coming here from #2714, the proposed README says,

> Note that it is **strongly recommended** that you install any version-controlled dependencies in editable mode in order to ensure that dependency resolution can be performed with an up to date copy of the repository each time it is performed, and that it includes all known dependencies.

It would be best to directly explain to the user how that would work, i.e. write “in editable mode, using `pipenv install -e …`, in order to ensure …”.

jeremytiki:
@uranusjr I'll add a link to the pip documentation as well!

@slhck By all means! Please critique anything that looks out of place! Like I said I'm still trying to learn how a lot of this stuff works so if I am doing something incorrectly or even less than ideal let me know so I can fix it! I went ahead and removed the back-ticks, using only 4 space indent for the code blocks as well as adding in the additional explanation for the `-e` flag.

techalchemy:
Sorry for the delays on this, it looks really good to me :D thanks a ton for putting this together, this is going to be incredibly helpful

coltonbh:
After spending multiple hours not knowing why my installation of a private GitHub repo via ssh was failing I came across [this](https://stackoverflow.com/a/4837571/5728276) unassuming StackOverflow answer. I'd recommend adding something like this to the pipenv docs so that people can avoid this pain--it's a tough issue to track down!

If you want to install via ssh (say, for a private repo), you'll need to modify the GitHub ssh url as follows:

    git@github.com:coltonbh/my-package.git (fetch)
                  ^ change this to a '/' character
Correct:
git@github.com/coltonbh/my-package.git (fetch)

And the install looks like:

pipenv install -e "git+ssh://git@github.com/coltonbh/my-package.git@branch123#egg=mypgk"

Otherwise you'll get a very long install process followed by the cryptic error:

pipenv.patched.notpip._internal.exceptions.InstallationError: Command "git clone -q ssh://git@github.com:coltonbh/my-package.git /var/folders/w2/rkn1bm3d3ndclys0f81clbfw0000gn/T/requirementslib5mkvyj4m/mypgk" failed with error code 128 in None

PR #2709, opened 2018-08-06
Fix typo

svanburen:

##### The issue

##### The fix

##### The checklist

* [ ] Associated issue


techalchemy:
Thanks :D

Issue #2635, opened 2018-07-23
Import Private Github Repo via Environment Variables 


juansc:
You can use `pip` to pull a private Github repo using environment variables. 

```
# requirements.txt
git+https://${AUTH_USER}:${AUTH_PASSWORD}@github.com/user/myproject.git@1.2.3#egg=myproject
```

Then run

```
> pip install -r requirements.txt
```

Is there a way to do this with Pipenv? I tried using the environment variable method described [here](https://docs.pipenv.org/advanced/#injecting-credentials-into-pipfiles-via-environment-variables), but it looks like it's not the right approach.

```
Could not find a version that matches myproject (from -r /var/folders/6r/719kpmn511l0ghtx4v1rbxdm0000gn/T/pipenv-5xvfp5mw-requirements/pipenv-wm4ivcah-constraints.txt (line 11))
No versions found
Were https://pypi.python.org/simple or git+https://${AUTH_USER}:${AUTH_PASSWORD}@github.com/user/myproject.git#egg=myproject reachable?
```

<details><summary>$ pipenv --support</summary>

Pipenv version: `'2018.7.1'`

Pipenv location: `'/usr/local/lib/python2.7/site-packages/pipenv'`

Python location: `'/usr/local/opt/python@2/bin/python2.7'`

Other Python installations in `PATH`:

  - `2.7`: `/usr/local/bin/python2.7`
  - `2.7`: `/usr/local/bin/python2.7`
  - `2.7`: `/usr/bin/python2.7`
  - `3.6`: `/usr/local/bin/python3.6m`
  - `3.6`: `/usr/local/bin/python3.6`

  - `2.7.14`: `/usr/local/bin/python`
  - `2.7.10`: `/usr/bin/python`
  - `2.7.14`: `/usr/local/bin/python2`
  - `3.6.5`: `/usr/local/bin/python3`

PEP 508 Information:

```
{'implementation_name': 'cpython',
 'implementation_version': '0',
 'os_name': 'posix',
 'platform_machine': 'x86_64',
 'platform_python_implementation': 'CPython',
 'platform_release': '17.6.0',
 'platform_system': 'Darwin',
 'platform_version': 'Darwin Kernel Version 17.6.0: Tue May  8 15:22:16 PDT 2018; root:xnu-4570.61.1~1/RELEASE_X86_64',
 'python_full_version': '2.7.14',
 'python_version': '2.7',
 'sys_platform': 'darwin'}
```

System environment variables:

[...]

PR #2710, opened 2018-08-07
Fix a bug with enviroment variable injection to the git repo url

ehengao:

##### The issue

Fix a enviroment injection issue when dealing with git repo url dependencies. The problem issued when pipenv trying to execute command similar to "git clone -q git+https://${USER}:${PASSWORD}/github.com" using python subprocess.Popen, the enviroment variable inside this command is not propely resolved.Which will cause a permission deny error or a repo not found error.

##### The fix

using os.path.expandvars(cmd-parts) before passing to subprocess.

##### The checklist

* [ ] Associated issue

https://github.com/pypa/pipenv/issues/2635 

* [ ] A news fragment:

news/2365.bugfix

### DIFF
--- a/news/2635.bugfix
+++ b/news/2635.bugfix
@@ -1,0 +1,1 @@
+Fix a enviroment injection issue when dealing with git repo url dependencies. The problem issued when pipenv trying to execute command similar to "git clone -q git+https://${USER}:${PASSWORD}/github.com" using python subprocess.Popen, the enviroment variable inside this command is not propely resolved.Which will cause a permission deny error or a repo not found error. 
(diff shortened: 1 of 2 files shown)


ehengao:
By any chance we can merge this?

uranusjr:
I want to sit on this for a while. Generally a patch to `vendor` or `patched` is a last resort we want to avoid, since it may result in unintended subsequences.

techalchemy:
Interesting issue -- I'm pretty sure we're going to un-vendor pip soon though so this might not be the best approach to solving it

pcraciunoiu:
Is there a fix for installing private git repos with `pipenv`? I found this issue and I'm trying to avoid rolling a private pypi just for a repo or two...

uranusjr:
One way is to use SSH, such as [GitHub deploy keys](https://github.blog/2015-06-16-read-only-deploy-keys/).

techalchemy:
Really anything that involves putting a password in a URI directly is a bad practice, and we've had similar versions of this proposed in the past and closed for various security related reasons that I can't completely recall.  If someone can dig through the past issues on the topic it would be helpful

Issue #2718, opened 2018-08-08
pipenv run throws OSERROR on Windows installation


m-stolle:
##### Issue description

pipenv run throws OSERROR on Windows installation. No other python installations present.

It works If I change the following line in pipenv\core.py", line 2201, in do_run_nt

`subprocess.Popen([command] + script.args, **options) `

to

`subprocess.Popen(['python', command] + script.args, **options)`

```
>assoc .py
.py=Python.File

>ftype Python.File
Python.File="C:\WINDOWS\py.exe" "%L" %*
```

##### Expected result

pipenv run without any exceptions

##### Actual result
```
> pipenv run manage.py test
Traceback (most recent call last):
  File "d:\python36-32\lib\runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "d:\python36-32\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "D:\Python36-32\Scripts\pipenv.exe\__main__.py", line 9, in <module>
  File "d:\python36-32\lib\site-packages\pipenv\vendor\click\core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "d:\python36-32\lib\site-packages\pipenv\vendor\click\core.py", line 697, in main
    rv = self.invoke(ctx)
  File "d:\python36-32\lib\site-packages\pipenv\vendor\click\core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "d:\python36-32\lib\site-packages\pipenv\vendor\click\core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "d:\python36-32\lib\site-packages\pipenv\vendor\click\core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "d:\python36-32\lib\site-packages\pipenv\cli.py", line 701, in run
    do_run(command=command, args=args, three=three, python=python, pypi_mirror=pypi_mirror)
  File "d:\python36-32\lib\site-packages\pipenv\core.py", line 2253, in do_run
    do_run_nt(script)
  File "d:\python36-32\lib\site-packages\pipenv\core.py", line 2201, in do_run_nt
    p = subprocess.Popen([command] + script.args, **options)
  File "d:\python36-32\lib\subprocess.py", line 709, in __init__
    restore_signals, start_new_session)
  File "d:\python36-32\lib\subprocess.py", line 997, in _execute_child
    startupinfo)
[...]

PR #2727, opened 2018-08-10
Improve "run" behavior on Windows

uranusjr:
Provide better error-handling for #2718 (so the command never throws OSError).

Some Windows users are used to launch files without specifying a command, e.g.

    > my-script.py

This works in the shell because Windows will automatically choose an command based on file association, and with newer Python versions, the Py Launcher (py.exe) automatically chooses the correct Python based on shebang-parsing.

A similar syntax, unfortunately, does not currently work in Pipenv

    > pipenv run my-script.py

Since my-script.py will be treated as a real application by the subprocess module.

This patch catch Windows error 193 during subprocess initialization, and fall back to use COMSPEC (shell=True) when it happens, to provide better support for this use case.

##### The checklist

* [x] Associated issue

### DIFF
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -2087,15 +2087,30 @@ def inline_activate_virtual_environment():
         os.environ["VIRTUAL_ENV"] = root
 
 
-def do_run_nt(script):
+def _launch_windows_subprocess(script):
     import subprocess
 
     command = system_which(script.command)
     options = {"universal_newlines": True}
-    if command:  # Try to use CreateProcess directly if possible.
-        p = subprocess.Popen([command] + script.args, **options)
-    else:  # Command not found, maybe this is a shell built-in?
-        p = subprocess.Popen(script.cmdify(), shell=True, **options)
+
+    # Command not found, maybe this is a shell built-in?
+    if not command:
+        return subprocess.Popen(script.cmdify(), shell=True, **options)
+
+    # Try to use CreateProcess directly if possible.
+    try:
+        return subprocess.Popen([command] + script.args, **options)
+    except WindowsError as e:
+        if e.winerror != 193:
+            raise
+
+    # Windows error 193 "Command is not a valid Win32 application".
+    # Try shell mode to use Windows's file association for file launch.
+    return subprocess.Popen(script.cmdify(), shell=True, **options)
+
+
+def do_run_nt(script):
+    p = _launch_windows_subprocess(script)
     p.communicate()
     sys.exit(p.returncode)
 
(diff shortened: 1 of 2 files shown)


pipenv/core.py
```diff
@@ -2087,15 +2087,31 @@ def inline_activate_virtual_environment():
...
+
+    # Command not found, maybe this is a shell built-in?
+    if not command:
+        return subprocess.Popen(script.cmdify(), shell=True, **options)
+
+    # Try to use CreateProcess directly if possible. Specifically catch
+    # Windows error 193 "Command is not a valid Win32 application" to handle
+    # a "command" that is non-executable. See pypa/pipenv#2727.
+    try:
+        return subprocess.Popen([command] + script.args, **options)
```
techalchemy: So it's a command, but not an executable?  I think `pythonfinder.which` actually handles this correctly then on windows whenever we get around to figuring that out => it will find things even if the execute bit isn't set as just from testing I found that to be unreliable

uranusjr: Yeah, the cause here is that the DOS CLI does not really have a notion of executable files. As far as it concerns, everything can be launched, and it’s up to the operating system to decide what to do. As a consequence, `where.exe` find all sorts of things, not just applications. It would happily return `django-admin.py` if it’s in `PATH`, and `subprocess.Popen` (actually the `CreateProcess` API it uses under the hood) would be hung to dry.

So the root issue here is that we are mixing two paradigms when implementing Pipenv for Windows because the deep roots of Windows lacking a real user-facing “executable” flag. I’d doubt even a re-implemented `which` function would help, unless you dig deep into Windows API to actually decide if a file is a real application.

techalchemy:
Not sure if we need to complicate it that much. In pythonfinder I think I handle everything with sort order. So prefer by path order and execute bit with priority to known executable extensions. 

At the end of the day we are talking about this: users can run any file on windows from the command line. File get passed to the CreateProcess API which fails if the file can’t be executed. This produces an error message with information. 

We can let pipenv act like a cli application and trust that the user is trying to rub something because they want to, and just try running if if it exists and providing that error if it can’t be executed. Or we can rebuild the API and manually introspect the file to try and check against the api and achieve the exact same result of printing a slightly different error. 

Let ms know Iif I’m missing something here

In case we just match an entry in `$PATH` I will use that instead. No need to introspect. Just treat it like normal input. If the user ran `pipenv shell` and then some command the only difference is it prioritizes builtins which should be known values anyway. Maybe we should just check for builtins and then confirm pythonfinder does what I am describing?

techalchemy:
hm had a chance to look at this properly, i swiped the logic and built it into vistir as well but it looks right to me

PR #2739, opened 2018-08-13
Vendor new libraries: vistir and pip_shims

techalchemy:
- Vendors new libraries for maintainability
- Separates responsibiltiies and adds coverage


uranusjr:
I kind of think we should cut a release before starting to migrate code to use the new libraries (vendoring itself doesn’t matter).

techalchemy:
these libraries are near-exact replication of the code we already have -- I am fine holding off on implementation but the only actual difference is test coverage and edge case handling

PR #2745, opened 2018-08-15
Move misformatted comments

orn688:
##### The issue

Black moved a couple comments to some funky places.

##### The fix

This moves the comments back to the lines to which they refer.

##### The checklist

* [ ] Associated issue


techalchemy:
Seems right to me, thanks for the pr!

PR #2768, opened 2018-08-21
GitHub has a capital H

singingwolfboy:
Just fixing a pet peeve of mine.

### DIFF
--- a/docs/_templates/hacks.html
+++ b/docs/_templates/hacks.html
@@ -60,7 +60,7 @@
 </style>
 
 <!-- GitHub Logo -->
-<a href="https://github.com/kennethreitz/pipenv" class="github-corner" aria-label="View source on Github">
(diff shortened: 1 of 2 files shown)


uranusjr:
The last one you’ll need to submit to pip (notpip is a patched version of it), otherwise I have no objections.

singingwolfboy:
@uranusjr: looks like [that's already been done](https://github.com/pypa/pip/commit/508517f0d7898618d656310ddd797200944b0136).

uranusjr:
Ah that’s great news. Could you remove the change on notpip then? We’ll be able to apply the upstream change once we upgrade, and the change would cause merge conflicts when we do that.

singingwolfboy:
@uranusjr: OK, I've reverted that change.

PR #2797, opened 2018-08-29
Fix typo in basics.rst

gmyers18:
organizatoin => organization


uranusjr:
Strange, I vaguely remember there was a similar PR just a few days ago. We are not very good at typing, apparently :p

PR #2717, opened 2018-08-31
default .env with local overrides

jamesstidard:
Hi,

I'm unsure if this now the correct place for a feature request (the [link](https://kenneth-reitz.uservoice.com/forums/913660-general) in the issue template no longer seems to be working).

I've been using Vue.js for some web development and their build tool have some nice .env support which I think would be nice to have in pipenv.

So with the Vue.js webpack template you can have a `.env` file which you can commit to your source control which is nice to provide some default configurations to all developers. You can then override that `.env` with a `.env.local` which will merge over the top replacing any redeclared environment variable values. This one is obviously not tracked by source control.

Wondering if that's a feature which others would also like to see.

Thanks.

### DIFF
--- a/.env
+++ b/.env
@@ -1,1 +1,2 @@
+# Override with your own in .env.local
 HELLO=WORLD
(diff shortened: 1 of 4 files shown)


kennethreitz:
-1

PR #2819, opened 2018-09-03
PEEP-002: Specify options via environment variables

fridex:


fridex:
The relevant implementation changes originally proposed in https://github.com/pypa/pipenv/pull/2537

peeps/PEEP-002.md
```diff
@@ -0,0 +1,32 @@
+# PEEP-002: Specify options via environment variables
+
+This PEEP describes an addition that would allow configuring pipenv options via environment variables suitable especially for automated systems or CI/CD systems.
+
+☤
+
+Systems running not only on containerized solutions (like Kubernetes or OpenShift) are often parametrized via environment variables. The aim of this PEEP is to provide an extension to the current pipenv implementation that would simplify parametrizing options passed via environment variables.
```
kennethreitz: 'pipenv' should always be capitalized

kennethreitz:
I like this idea, as pip does this, but I think it would be difficult to implement with Click. @techalchemy thoughts?

peeps/PEEP-002.md
fridex: Adjusted.

kennethreitz:
@fridex I meant like title case, `Pipenv` :)

fridex:
> @fridex I meant like title case, Pipenv :)

Fixed :)


> I like this idea, as pip does this, but I think it would be difficult to implement with Click. @techalchemy thoughts?

Please check https://github.com/pypa/pipenv/pull/2537 where this was implemented and discussed with @uranusjr.

kennethreitz:
Agreed! I'll let you take it from here, @uranusjr. +1 from me, assuming all details get worked out.

uranusjr:
(For @fridex: Feel free to update the PEEP and provide extra information regarding the topics I mentioned, including submitting additional PRs)

techalchemy:
+1 from me as well

fridex:
> For @fridex: Feel free to update the PEEP and provide extra information regarding the topics I mentioned, including submitting additional PRs

Sure will do. Unfortunately I will not be available by the end of week and the week after, so please bear with my time. Thanks!

sirosen:
I think we can do a better job of documenting this in `click`, but the `auto_envvar_prefix` code _is_ designed to handle this type of case: https://github.com/pallets/click/blob/123dd717439d8620d8d6be5574d2c9f007952326/click/core.py#L318-L329

As long as you pass a prefix to the top level click context, it should percolate down to any options which have `allow_from_autoenv=True`.

You could therefore set the prefix to `PIPENV` and get reasonable behavior out of `click`, in which options to `pipenv install` would pull from `PIPENV_INSTALL_*` env vars.

Universal options like `--verbose` are probably most consistently handled by setting them to `PIPENV_VERBOSE` if you go down this path.

---

Aside: I think there's a very minor bug in the linked code, and I'll submit a fix which we should hopefully be able to get into `click` 7.

fridex:
> Aside: I think there's a very minor bug in the linked code, and I'll submit a fix which we should hopefully be able to get into click 7.

@sirosen, could you be more specific with the issue (if there needs to be done patching on Pipenv side to get changes in based on click 7 fixes)? Thanks!

sirosen:
No problem.
The fix did make it in for click 7, so if `pipenv` is vendoring click without patching it, an upgrade should suffice.

If an older click is vendored or there's a patch you need to worry about conflicting, here's the exact change:
https://github.com/pallets/click/pull/1105

If you passed an `auto_envvar_prefix` which was lowercase, `click` was not uppercasing it when it was passed explicitly. The uppercasing happened correctly when the envvar prefix was deduced from a parent command name.

uranusjr:
I have a new question. Would it be possible to blacklist certain options, or whole options of commands from being accessible from envvar? There isn’t a use case for this right now, but I feel it would be best if we can have an idea what we can do to amend edge cases when they come up.

I don’t think we are patching our vendored Click. There are a few dependencies that have Click as a secondary dependency (e.g. click-completions), so we’ll need to make sure they can handle the Click 7.0 upgrade.

uranusjr:
Also, before I forget, discussions in this thread will need to be summarised and written up in the PEEP document. It is not enough to add a link pointing to the issue page, since comments on GitHub can be edited, or even deleted.

PR #2833, opened 2018-09-08
Do not use red if it's no an error

yohanboniface:
screenshot-20180908121953-1086x144

Thanks :)

### DIFF
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -849,7 +849,7 @@ def do_create_virtualenv(python=None, site_packages=False, pypi_mirror=None):
         crayons.normal(u"Creating a virtualenv for this project…", bold=True), err=True
     )
     click.echo(
-        u"Pipfile: {0}".format(crayons.red(project.pipfile_location, bold=True)),
+        u"Pipfile: {0}".format(crayons.yellow(project.pipfile_location, bold=True)),
         err=True,
     )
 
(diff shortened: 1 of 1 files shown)


techalchemy:
Hi there, thanks for the PR :) Any UI changes (and any changes at all) are going to require you to start with an Issue and in this case you will need to propose an enhancement by first creating a **PEEP** -- please see https://github.com/pypa/pipenv/blob/master/peeps/PEEP-001.md for more information on how to go about this

I appreciate that red is generally accepted as an error color, but Kenneth does make the call on UI decisions so this one will have to follow the process. Thanks again for your time, looking forward to the follow up proposal

yohanboniface:
Sounds like "process over effectiveness". :(

And, did I just read this [there](https://github.com/pypa/pipenv/blob/master/peeps/PEEP-001.md)? ;)

> Avoid writing PEEPs, unless it seems absolutely neccessary, and in the best interest of the entire Python community.

Honnestly, I'm not going to write any PEEP to explain a BDFL that *red does mean error*. :)

uranusjr:
Unfortunately, you do need to. The mapping of colour code to contextual meaning is highly cultural dependant, and is not as obvious as you would think. It makes sense for you, and likely people around you, but please do remember the world is a lot larger than that, and we need to assume less than we incline to. “Red = error” is not universal.

yohanboniface:
Ah. Can you point me to the *PEEP* defining the colour scheme so I can suggest a modification of it? :)

techalchemy:
No. We didn’t retroactivelt create enhancement proposals for things that already existed, the process is not meant to be complicated. It is meant to be explicit and straightforward and to separate technical decisions from interface design. If you want to introduce changes to the interface it should start as a proposal that can standardize, say, different types of information and what colors they should be. 

At that point we can probably find some accessibility guideline on the topic and take that info into account so that Kenneth can make centralized and informed design decisions without having to get bogged down in issues and PRs. We will have to run changes like this by him anyway, this process just makes it both easier and more uniform, it’s really not just me being a jerk :)

Issue #2825, opened 2018-09-05
Unreachable pyup.io leads to failing `pipenv check`


l0b0:
I'm working on a project which has been running a lot of `pipenv check`s (possibly on the order of hundreds of requests per day). I think we've been throttled at the network level, because this command now results in messages like this:

> Connection to pyup.io timed out. (connect timeout=5)

`curl pyup.io` can't reach the site, but `torify curl pyup.io` (that is, via a proxy) works, and `dig pyup.io` returns the same IP as [SSL Labs detects](https://www.ssllabs.com/ssltest/analyze.html?d=pyup.io), so it's definitely restricted to somewhere between this network and that site.

I've changed our process to run `pipenv check` much less often, but would it be possible to detect or avoid this issue in `pipenv`? Ideas:

- [Configuring our own API key](https://github.com/pypa/pipenv/issues/1651#issuecomment-373320750) so we can reliably stay below any limits.
- Suggesting solutions when the connection times out.
- Configurable caching of the security issues database. I realize this is probably a very risky move since security issues may have to be fixed within hours, but it could be useful for people who for process reasons end up running a lot of checks.

PS: The [recommended forum](https://kenneth-reitz.uservoice.com/forums/913660-general) for these requests returns HTTP 404.

uranusjr:
Allowing user configuration may be the simplest way to go. You can set it to something else (default to our built-in key if not set), or (maybe) a special value to disable pyup checks completely. At least the first part would be fairly straightforward, and contributions would be welcomed :)

techalchemy:
Since we have merged #2835 I believe we can close this one out, thanks for the report / hope the fix works for your use case!

leotrubach:
Currently pyup.io gives 500 error and we to distinguish the following situations:
1) site is unreachable
2) there are outdated packages

This could be done by returning different error codes

leotrubach:
As a workaround this one could be used:
```
pipenv check || if [ $(curl --write-out %{http_code} --silent --output /dev/null pyup.io) -eq 500 ]; then echo "Site unreachable. Skipping check"; else false; fi
```

leotrubach:
Hmm... I couldn't really figure out what's wrong. But when running `pipenv check` it fails but `safety check` succeeds

leotrubach:
Strace output
```
Checking installed package safety…
read(9, "", 50000)                      = 0
read(5, "Unable to load vulnerability database\n", 32768) = 38
[...]

PR #2835, opened 2018-09-09
Add ability to override the pyup.io API key for the check command

ulrikjohansson:
This an attempted fix for part of the problem in #2825, being able to supply your own API key for `pipenv check`
I haven't created any news fragments, if needed I might need some pointers on how to go about that.

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -240,6 +240,9 @@ Example::
    API access throttling based on overall usage rather than individual
    client usage.
 
+   You can also use your own safety API key by setting the
+   environment variable ``PIPENV_CHECK_KEY``.
+
 
 ☤ Community Integrations
 ------------------------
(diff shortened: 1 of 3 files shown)


ulrikjohansson:
Is it the ambiguity in the env name that's troubling you? I can see how the word CHECK might confuse users to think that it might be a boolean flag instead of a place to put your api key.

Perhaps it's better to name it something along the lines of `PIPENV_PYUP_API_KEY` instead?

techalchemy:
Hey thanks for the PR!

Definitely see where you're coming from on this, and I think `PIPENV_PYUP_API_KEY` is totally appropriate as an environment variable.  I'd be happy merging that

If that provides the functionality you need, that is totally workable on our end, so thanks for putting this together!

techalchemy:
Sorry for the delay on merging this, but definitely want to include this in the release.  Thanks again for putting it together!

PR #2844, opened 2018-09-11
Clarify `pipenv install` documentation

g4b1nagy:
`if no packages are given` sounds a bit better than
`if none is given` when referring to `provided packages`.

Also mention that `pipenv install` with no provided packages
installs all packages from Pipfile. It is not obvious for
people new to pipenv whether pipenv is using Pipfile or
Pipfile.lock by default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,8 +181,8 @@ Fish is the best shell. You should use it.
                  provided in Pipfile.
       clean      Uninstalls all packages not specified in Pipfile.lock.
       graph      Displays currently–installed dependency graph information.
-      install    Installs provided packages and adds them to Pipfile, or (if none
-                 is given), installs all packages.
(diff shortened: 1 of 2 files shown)


g4b1nagy:
PS: thank you for all the hard work you've shared with us!
These are just a few minor changes that I think will make documentation a bit more clear. Please let me know if you'd like me to make any changes.

uranusjr:
Looks reasonable to me. Would you mind adding a fragment to the `news` directory to describe this change? It would appear in the changelog.

g4b1nagy:
Sure thing! I've added a commit with the news fragment. Please let me know what you think.

g4b1nagy:
@uranusjr I've found another small issue in the documentation, so I've added a new commit. Please let me know what you think.

g4b1nagy:
Thank you!

Issue #2846, opened 2018-09-12
Pipenv install does not provide a good error message for no Pipfile.


fossum:
##### Issue description

When attempting to run `pipenv install --dev` in the wrong spot I got this error below (not a bug I assume).

```
~ # pipenv install --dev
Creating a Pipfile for this project...
Traceback (most recent call last):
  File "/usr/local/bin/pipenv", line 11, in <module>
    sys.exit(cli())
  File "/usr/local/lib/python3.4/site-packages/pipenv/vendor/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.4/site-packages/pipenv/vendor/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.4/site-packages/pipenv/vendor/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.4/site-packages/pipenv/vendor/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.4/site-packages/pipenv/vendor/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python3.4/site-packages/pipenv/cli.py", line 435, in install
    selective_upgrade=selective_upgrade,
  File "/usr/local/lib/python3.4/site-packages/pipenv/core.py", line 1759, in do_install
    pypi_mirror=pypi_mirror,
  File "/usr/local/lib/python3.4/site-packages/pipenv/core.py", line 639, in ensure_project
    ensure_pipfile(validate=validate, skip_requirements=skip_requirements, system=system)
  File "/usr/local/lib/python3.4/site-packages/pipenv/core.py", line 288, in ensure_pipfile
    project.create_pipfile(python=python)
  File "/usr/local/lib/python3.4/site-packages/pipenv/project.py", line 576, in create_pipfile
    config_parser = ConfigOptionParser(name=self.name)
  File "/usr/local/lib/python3.4/site-packages/pipenv/patched/notpip/_internal/baseparser.py", line 141, in __init__
    assert self.name
AssertionError
~ #
```

##### Expected result

I would expect something about missing a Pipfile or some such. Probaly something along these lines.

```
AssertionError("Could not find Pipfile.")
```

##### Actual result

Instead I just got `AssertionError()`.

##### Steps to replicate

```
[...]

PR #2853, opened 2018-09-14
fix_#2846

snowmanunderwater:
Fix for [#2846](https://github.com/pypa/pipenv/issues/2846).

### DIFF
--- a/pipenv/patched/notpip/_internal/baseparser.py
+++ b/pipenv/patched/notpip/_internal/baseparser.py
@@ -138,7 +138,7 @@ class ConfigOptionParser(CustomOptionParser):
         isolated = kwargs.pop("isolated", False)
         self.config = Configuration(isolated)
 
-        assert self.name
+        assert self.name, 'Could not find Pipfile.'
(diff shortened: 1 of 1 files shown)


pipenv/patched/notpip/_internal/baseparser.py
```diff
@@ -138,7 +138,7 @@ def __init__(self, *args, **kwargs):
         isolated = kwargs.pop("isolated", False)
         self.config = Configuration(isolated)
 
-        assert self.name
+        assert self.name, 'Could not find Pipfile.'
```
techalchemy: So this is probably not the best place to handle this, we really don't want to be making additional patches in this folder if possible -- note that this is patching a vendored copy of pip directly which is definitely not preferred.  Ideally we would handle this in pipenv's main codebase instead

frostming: Agree, we should check the existence somewhere early in `do_install()` instead of patching the vendor.

techalchemy:
closing in favor of #2920

Issue #1015, opened 2017-11-01
Add flag to not load .env file and/or add command to reload .env file when inside a pipenv shell


rodxavier:
Currently, pipenv automatically reads the `.env` file when running either `pipenv run` or `pipenv shell`. It would be nice to have a way to prevent it from doing this. I know I can just do `PIPENV_DOTENV_LOCATION=.fake` but it seems a little hacky.

Moreover, when I edit the `.env` file inside a `pipenv shell`. I have to deactivate and activate the environment again for the changes to take place.

**Environment**
1. OS Type: OSX
2. Python version: 3.6.2
3. Pipenv version: 8.3.1

**Expected Result**
It would be nice to have a flag to make pipenv not automatically load the `.env` file. Also, it would be helpful to have a command to reload the `.env` file.

nateprewitt:
@rodxavier, have you tried using PIPENV_DONT_LOAD_ENV? That should work for your use case.

rodxavier:
@nateprewitt Thanks for replying. Yeah, I have tried using it but it seems a bit hacky.

nateprewitt:
@rodxavier, in what way? By flag are you asking for an option to pass? If we added an option for every possible combination of environment settings we'd have a pretty unmanageable API.

rodxavier:
@nateprewitt Sorry about that. I haven't tried that option and that seem to solve my issue. Is it mentioned somewhere in the docs?

nateprewitt:
@rodxavier it *should* be under "Automatic Loading of .env" in the advanced section, but it's currently not.

This would be a good first PR for anyone looking 😉

rodxavier:
@nateprewitt Thanks. I'm keen to add it in.

PR #2876, opened 2018-09-20
Added inotifier to watch  on a thread. Needs refinement

owenstranathan:
##### The issue

When in an active virtualenv ,changes to `.env` file require manual deactivation/reactivation
It would be great if this was automatic.

related to [Issue #1015](https://github.com/pypa/pipenv/issues/1015)

##### The fix
At first I wanted a command to manually retrigger the load, but since commands are run from a child process it's not possible to export the values to the parent environment.

My solution is to use pyinotify to spawn a thread to watch the dotenv file and manually send the new env to the opened shell in the callback.

I'm not convinced that this is the best solution but it "technically works".
Currently this PR sends a line exporting the values in the subshell. This is not ideal and I'm looking for suggestions on how to improve this in anyway. Does pexpect offer something to set the process's environment?

I'm not sure how important/desirable this behavior is to everyone who uses pipenv. Perhaps this should be configurable, so it only happens when you specifiy it. Please advise.

##### The checklist

- [ x ] Associated issue

* [ ] A news fragment in the `news/` directory to describe this fix with the extension `.bugfix`, `.feature`, `.behavior`, `.doc`. `.vendor`. or `.trivial` (this will appear in the release changelog). Use semantic line breaks and name the file after the issue number or the PR #.  <- Don't really understand this, would welcome guidance

### DIFF
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -2033,8 +2033,9 @@ def do_shell(three=None, python=False, fancy=False, shell_args=None, pypi_mirror
     click.echo("Launching subshell in virtual environment…", err=True)
 
     fork_args = (project.virtualenv_location, project.project_directory, shell_args)
-    dotenv_file = environments.PIPENV_DOTENV_LOCATION or os.sep.join(
-        [project.project_directory, ".env"]
+
+    dotenv_file = environments.PIPENV_DOTENV_LOCATION or os.path.join(
+        project.project_directory, ".env"
     )
    
     if fancy:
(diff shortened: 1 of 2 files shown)


pipenv/core.py
```diff
@@ -2033,13 +2033,16 @@ def do_shell(three=None, python=False, fancy=False, shell_args=None, pypi_mirror
     click.echo("Launching subshell in virtual environment…", err=True)
 
     fork_args = (project.virtualenv_location, project.project_directory, shell_args)
-
+    dotenv_file = environments.PIPENV_DOTENV_LOCATION or os.sep.join(
+        [project.project_directory, ".env"]
+    )
```
uranusjr: better to `os.path.join(project.project_directory, ".env")` instead

owenstranathan: Thanks! I normally would have used `os.path` but saw the sep thing elsewhere in the code and was trying to fit in :)

PR #2877, opened 2018-09-21
Create virtualenvs directory if it does not exist

Enzime:
##### The issue

This PR fixes #2786 as `pipenv` tries to iterate through `WORKON_HOME` (`virtualenvs` folder) without ensuring it exists first.

##### The fix

Creating the `WORKON_HOME` directory from inside `pipenv/utils.py:get_workon_home`.

##### The checklist

* [x] Associated issue
* [x] A news fragment in the `news/` directory to describe this fix with the extension `.bugfix`, `.feature`, `.behavior`, `.doc`. `.vendor`. or `.trivial` (this will appear in the release changelog). Use semantic line breaks and name the file after the issue number or the PR #.

PR #2883, opened 2018-09-23
Fix _get_virtualenv_hash function

Jamim:
##### The issue

Case-sensitive filesystems are handled like case-insensitive and vice versa.
This bug was introduced in #2513 and it affects only `master` branch and does not affect previous release.
I didn't find any related issue.

##### The fix

This PR adds `not ` before the related condition.

##### Additional changes
 - add PyCharm's config directory to `.gitignore`

##### CC
 - @uranusjr
 - @techalchemy

##### The checklist

* [ ] Associated issue


Jamim:
Hi @uranusjr,
I believe you can help me fix `_get_virtualenv_hash` for Windows as well.
Thank you!

Jamim:
@kennethreitz, @uranusjr, @techalchemy 
Does anyone care that there is bug which should be fixed?

techalchemy:
Hello @Jamim -- of course we care about the library we maintain, but I would suggest that the phrasing is probably not the most effective way to have a positive interaction.

I appreciate your contribution, however, the pr template is there to suggest that you should _open and report_ an issue if there is not yet one on the topic, which is how we typically triage problems.  This helps to avoid contributing a fix before it has been determined that it is necessary.  This way, we can assess that the fix indeed is desired before you put effort into attempting it.

Before we merge any PR related to this, I would need to see a failing test and an issue explaining the problem. Thanks for your efforts

Jamim:
Thank you for the feedback, @techalchemy!

I understand that you have some workflow that you prefer to follow, but meanwhile I believe there is too much effort required for a single-line fix.
I'll create an issue a bit later today, but I strongly recommend you to just look at the patch and realize how redundant this discussion around one little `not` is.

techalchemy:
I still need to know exactly what is broken, preferably via a test case (prevents regression) but if you don't feel like it at least I need to see an example of what you're saying doesn't work properly, because it's hard to decipher that just from reading over the text provided

I.e. what do I need to type in to have a problem?

Jamim:
> I.e. what do I need to type in to have a problem?

I found that since I submitted this PR the problem was "fixed" by #2877, but I believe it doesn't work properly. So, you need to revert one commit to see what exactly goes wrong.
```bash
git checkout master
git pull origin master
git checkout -b tmp
git revert 65ccd765b6d0b78e602d4b233ca8e31e1a6b8adb
docker rm pipenv_pipenv-tests_1  # you must delete cached image
docker-compose up
```

Jamim:
@techalchemy By the way, you asked me for some suitable test, but **the main goal** of this PR (and [this one](https://github.com/kennethreitz/pipenv-docker-tests/pull/1) as well) **was make the tests works**.

Jamim:
Hi @techalchemy,

I created a corresponding issue and I need you to confirm that I should not create any `news` fragment, since the bug is in `master` branch only and it's not presented in the release.
And please feel free to ask me anything to clarify the issue.

Thank you!

Jamim:
ping @uranusjr

Jamim:
ping @techalchemy

techalchemy:
will look this over again before i cut the next release, sorry about that

techalchemy:
Based on the commit you referenced it seems like we just need to call `normalize()` on the path we create to avoid the issue, no?

And we should be normalizing it before checking as well?

Jamim:
> Based on the commit you referenced it seems like we just need to call normalize() on the path we create to avoid the issue, no?

> And we should be normalizing it before checking as well?

To be honest, I have no idea. Sorry, I didn't investigate this issue deeper.

By the way, did you read the `Solution` section of the related issue #2919?
> The root cause of the issue is that `os.name != "nt"` condition [was mistakenly converted](https://github.com/pypa/pipenv/pull/2513/files#diff-7c0c15ee4eba8c237e3318a395816a13R283) to `fnmatch.fnmatch('A', 'a')` when it should have been converted to `not fnmatch.fnmatch('A', 'a')`.

That's it. I just know that the condition was inverted by mistake and it causes the problem for me.

techalchemy:
> That's it. I just know that the condition was inverted by mistake and it causes the problem for me.

Look, you opened a PR with code changes and you keep telling me a solution, but honestly it is not possible for me to assess a solution until I know the problem.

You showed me a commit you say I need to revert in order to _cause_ the problem, but you also say this is still a problem.  The commit you showed me just looks for the virtualenv directory but doesn't do any case normalization.  

Ultimately, I really don't know how to ask for this differently, _I need you to show me what is wrong by showing me an example of something that is currently behaving incorrectly_.  Telling me that someone forgot to put a `not` somewhere doesn't help me, this is just increasing my cognitive burden when trying to understand the issue.

Here is an example of what I am looking for:

Hi. I am a user of pipenv, and I ran this command:

```
> some command
...
some output
```

But as you can see, that directory should really have been found because it exists here:

```
> ls some_directory
```

It's going to be hard to move forward on this until I can see something showing me why the current code is wrong.

Jamim:
@techalchemy Should I explain why does it happen?

PR #2889, opened 2018-09-27
Fix "py -m site --user-site" description

quevue:

##### The issue

The command doesn't work with:
"py -m site --user-site"

##### The fix

I think the author's intention was:
"python -m site --user-site"
instead of:
"py -m site --user-site"

so I fixed it.
##### The checklist

* [X] Associated issue

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -119,7 +119,7 @@ To install::
     `modifying ~/.profile`_.
 
     On Windows you can find the user base binary directory by running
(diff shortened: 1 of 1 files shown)


uranusjr:
The `py` command refers to the [Python launcher for Windows](https://www.python.org/dev/peps/pep-0397/). This is a Windows-specific command to automatically point to the “default” Python installation on your machine. This is usually preferred to `python`—most certainly under this context—which we don’t advise adding into your `PATH`.

So, this is intended. Don’t change it :)

Issue #2893, opened 2018-09-28
URL in "Bug report" template is unavailable


ocavue:
##### Issue description

The URL ( https://kenneth-reitz.uservoice.com/forums/913660-general ) in [issue template](
https://github.com/pypa/pipenv/blob/9084792db033ccbe79beead19e655a3561dd8503/.github/ISSUE_TEMPLATE/Bug_report.md
) is unavailable.

##### Expected result

This URL is available.

##### Actual result

Error 404: Account Has Been Deleted

This UserVoice instance is no longer available.

##### Steps to replicate

None

uranusjr:
Thanks for the report. I believe there is already an open issue for this, but I’m not in the mood of digging it out. The resolution was to remove it, but nobody cared enough to do it. Would you mind sending us a pull request?

ocavue:
I'm happy to contribute to this project.

I think such a simple update is not needed for changelog so I didn't add a description in `news/`. It's that ok?

uranusjr:
yeah it’s fine

ocavue:
ok, [here](https://github.com/pypa/pipenv/pulls) is my PR.

PR #2900, opened 2018-09-29
Remove a 404 link in issue templates

ocavue:
##### The issue

https://github.com/pypa/pipenv/issues/2893

##### The fix

> How does this pull request fix your problem? Did you consider any alternatives? Why is this the *best* solution, in your opinion?

I just removed this link.

##### The checklist

* [x] Associated issue


uranusjr:
Don’t know what went wrong the Buildkite, but I don’t think there’s anything here that can make it fail.

Issue #2902, opened 2018-09-30
Editable packages with markers can not be locked


jxltom:
##### Issue description

In Pipfile, if markers are added for an editable package, latest pipenv in master branch can not lock.

In some older version of Pipenv, even editable packages can be locked, the generated ```requirements.txt``` such as  ```-e git+https://github.com/PyCQA/pylint.git@8682a220911a9c22286089e08f7e77b7b15d990b#egg=pylint; python_version >= '3.4'``` can not be installed by pip. And extra space before markers ```;``` for installing by ```pip install -r requirements.txt``` is required such as ```-e git+https://github.com/PyCQA/pylint.git@8682a220911a9c22286089e08f7e77b7b15d990b#egg=pylint ; python_version >= '3.4'```

##### Expected result

```pipenv lock``` should work.

##### Actual result

Following error will be raised.

```
Locking [dev-packages] dependencies…
Locking [packages] dependencies…
pip/_vendor/packaging/requirements.py", line 97, in __init__
    requirement_string[e.loc:e.loc + 8]))
pipenv.patched.notpip._vendor.packaging.requirements.InvalidRequirement: Invalid requirement, parse error at "';'"

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/user/Documents/dev/pipenv/pipenv/resolver.py", line 101, in <module>
    main()
  File "/home/user/Documents/dev/pipenv/pipenv/resolver.py", line 90, in main
    system=system,
  File "/home/user/Documents/dev/pipenv/pipenv/resolver.py", line 73, in resolve
    allow_global=system,
  File "/home/user/Documents/dev/pipenv/pipenv/utils.py", line 424, in resolve_deps
    req_dir=req_dir,
  File "/home/user/Documents/dev/pipenv/pipenv/utils.py", line 295, in actually_resolve_deps
    constraints = [c for c in constraints]
  File "/home/user/Documents/dev/pipenv/pipenv/utils.py", line 295, in <listcomp>
    constraints = [c for c in constraints]
  File "/home/user/Documents/dev/pipenv/pipenv/patched/notpip/_internal/req/req_file.py", line 89, in parse_requirements
    for req in req_iter:
  File "/home/user/Documents/dev/pipenv/pipenv/patched/notpip/_internal/req/req_file.py", line 164, in process_line
    constraint=constraint, isolated=isolated, wheel_cache=wheel_cache
[...]

PR #2903, opened 2018-09-30
Fixed editable package with markers can not be locked

jxltom:
##### The issue

Fixed https://github.com/pypa/pipenv/issues/2902 and also added tests. Without this patch, the new added test will fail.

##### The fix

Fixed in upstream https://github.com/sarugaku/requirementslib/pull/71. If it can be accepted in upstream, then it can be merged here.

##### The checklist

* [x] Associated issue


jxltom:
No problem.

But how about the added tests? It will be nice to have tests which can make sure that later development won't break current working cases

uranusjr:
I’d be fine if there is a test to ensure `pipenv lock -r` works, so I’ll merge if you take out the requirementslib patch (and probably change the news fragment). We’ll need to do it later though, otherwise it’ll break all CI.

jxltom:
Great.

I've udpated the PR which removed the requirementslib patch as well as the news (since the test is too trivial)

Currently the CI is broken but it should be fine after https://github.com/sarugaku/requirementslib/pull/71 is merged from upstream.

techalchemy:
I’ll release requirementslib in the next few days when I am sure we captured everything

jxltom:
Update: it's ready for merge since https://github.com/pypa/pipenv/pull/2935 has been merged

PR #2906, opened 2018-10-01
Use new Azure Pipelines domain in build badges

DavidStaheli:
Azure Pipelines now uses https://dev.azure.com/pypa/pipenv instead of https://pypa.visualstudio.com/pipenv (even though redirects will work for some time).  This updates the build badges to use that new domain.


uranusjr:
LGTM, merging! Interestingly, visualstudio.com does not actually redirects, only shows an identical page. Is this intentional? (Just curious.)

PR #2920, opened 2018-10-05
Provide a good message for error described in issue #2846

jr7square:
##### The issue
Issue #2846 

##### The fix
Implemented fix with suggestions made to previous PR #2853


pipenv/core.py
```diff
@@ -1675,6 +1675,9 @@ def do_install(
     # Don't search for requirements.txt files if the user provides one
     if requirements or package_args or project.pipfile_exists:
         skip_requirements = True
+    # Don't attempt to install develop and default packages if Pipfile is missing
+    if not project.pipfile_exists and not packages and dev:
+        click.echo("Could not find Pipfile.", err=True)
```
uranusjr: Should it abort here? This implementation would emit the error, but still carry on to execute the code afterwards.

jr7square: oh yes, I believe it should. Will fix it.

Issue #2924, opened 2018-10-05
pip 18.1 causes "TypeError: 'module' object is not callable"


rafaelcaricio:
Be sure to check the existing issues (both open and closed!), and make sure you are running the latest version of Pipenv.

Check the [diagnose documentation](https://docs.pipenv.org/diagnose/) for common issues before posting! We may close your issue if it is very similar to one of them. Please be considerate, or be on your way.

Make sure to mention your debugging experience if the documented solution failed.

##### Issue description

New version of pip breaks pipenv, exactly on this line https://github.com/pypa/pipenv/blob/51cbc2e2f76a7c60f879a608fd96e95faca35f2e/pipenv/vendor/pip_shims/shims.py#L101

##### Expected result

The `_strip_extras` not be a module.

##### Actual result

```
...
 File "/opt/myproj/.tox/py36/lib/python3.6/site-packages/pipenv/vendor/requirementslib/models/requirements.py", line 704, in from_line

    line, extras = _strip_extras(line)

TypeError: 'module' object is not callable
```

##### Steps to replicate

Provide the steps to replicate (which usually at least includes the commands and the Pipfile).

-------------------------------------------------------------------------------

Please run `$ pipenv --support`, and paste the results here. Don't put backticks (`` ` ``) around it! The output already contains Markdown formatting.

If you're on macOS, run the following:

    $ pipenv --support | pbcopy

If you're on Windows, run the following:

    > pipenv --support | clip

If you're on Linux, run the following:

    $ pipenv --support | xclip

uranusjr:
Tracked in upstream repo (referenced above).

uranusjr:
On second thought, I’m going to keep this open so people don’t open duplicates.

mbarneyjr:
Same problem here

<details><summary>$ pipenv --support</summary>

Pipenv version: `'2018.7.1'`

Pipenv location: `'/Users/mbarney/.pyenv/versions/3.6.6/lib/python3.6/site-packages/pipenv'`

Python location: `'/Users/mbarney/.pyenv/versions/3.6.6/bin/python3.6'`

Other Python installations in `PATH`:

  - `2.7`: `/usr/bin/python2.7`
  - `2.7`: `/Users/mbarney/.pyenv/shims/python2.7`
  - `2.7`: `/Users/mbarney/.pyenv/shims/python2.7`
  - `2.7`: `/usr/bin/python2.7`
[...]

PR #2926, opened 2018-10-05
A temporary workaround for pip 18.1 issues

fridex:
A quick workaround to fix https://github.com/pypa/pipenv/issues/2924, not tested. If acked, please release a new version of Pipenv not to keep systems failing on every new Pipenv installation.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,7 +22,7 @@ if sys.argv[-1] == "publish":
     sys.exit()
 
 required = [
-    "pip>=9.0.1",
+    "pip>=9.0.1,<18.1",
     "certifi",
     "setuptools>=36.2.1",
     "virtualenv-clone>=0.2.5",


Froskekongen:
Since this has happened before, wouldn't a long-term solution be to always pin `pip` to versions that work, and for a new release check if newer versions of `pip` is acceptable?

What would be better options? I am asking because I really need to have control of a (very) large python environment.

uranusjr:
@Froskekongen The long-term solution is to refactor pip and extract common parts into their own library, so we can have a stable Python API independent to pip releases. This is already being worked on, but pip’s code base is not very difficult to work with.

For short-term (or middle-term), the best way to help would be to constantly help test Pipenv against the master branch of pip, and alert us if anything is failing.

greysteil:
@techalchemy / @fridex I think this PR can be closed now we have `2018.10.9` 🎉

PR #2928, opened 2018-10-06
Level up the template headers

jxltom:
### The issue

When I try to open/view issues or pull requests, I found it a littile bit difficult to distinguish issue/pr headers from their content, since level ```#####``` is too similar with the contents.  Maybe it is better to level up the headers.

Well it is a little bit opinionated, but I still would like to share my thoughts. 😄 

### The fix

Update github templates' header level to ```###```.

### The checklist

* [ ] Associated issue

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -9,19 +9,19 @@ Check the [diagnose documentation](https://docs.pipenv.org/diagnose/) for common
 Make sure to mention your debugging experience if the documented solution failed.
 
 
-##### Issue description
+### Issue description
 
 Describe the issue briefly here.
 
-##### Expected result
+### Expected result
 
 Describe what you expected.
 
-##### Actual result
+### Actual result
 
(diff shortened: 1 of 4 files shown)


jxltom:
By the way, the CI failes because pipenv is using the latest pip version. Maybe it is better to pin specific pip version in CI to have a stable and repeatable tests.

uranusjr:
Can you provide a screenshot of the new level? I’ve been thinking about changing this, but never got myself to compare various settings to decide which is best.

jxltom:
This is the new level.

level up the template headers by jxltom pull request 2928 pypa pipenv

This is the old level.

level up the template headers by jxltom pull request 2928 pypa pipenv 1

jxltom:
Acutally I've already using the new template in this PR 😃 

Personally, I prefer the new level, since it makes much easier to distinguish between different sections, such as ```The issue``, ```The fix``` and etc.

uranusjr:
Yeah this looks good to me. I read GitHub’s CSS and it seems h4 is quite similar to h5, so h3 is probably the best to go.

PR #2929, opened 2018-10-06
Lock pip version in tests

jxltom:
### The issue

To have a stable and repeatable tests, the pip version in tests should be pinned. Otherwise, unexpected bugs such as https://github.com/pypa/pipenv/issues/2924 may fail all CI tests. Passed tests may fail even codebase is not changed as time varies.

The pip version in tests should be the latest but a specific version. After https://github.com/pypa/pipenv/issues/2924 is fixed, it should be pinned back to ```18.1``` again.

How does that sound?

### The fix

Pin pip version in tests to latest but a specific version.

### The checklist

* [ ] Associated issue

### DIFF
--- a/.vsts-ci/steps/install-dependencies.yml
+++ b/.vsts-ci/steps/install-dependencies.yml
@@ -1,3 +1,3 @@
 steps:
-- script: 'python -m pip install --upgrade pip && python -m pip install -e .'
+- script: 'python -m pip install --upgrade pip==18.0 && python -m pip install -e .'
   displayName: Upgrade Pip & Install Pipenv


jxltom:
Woops, it looks like the CI still two failed tests even system pip version is not ```18.1```.

No idea on why ```pip graph``` still fails even system pip is downgrades. It may relate tohttps://github.com/sarugaku/pip-shims/issues/17

techalchemy:
The goal is actually that pipenv can build against any reasonable version of pip, so CI failures are highly desirable for catching this

techalchemy:
(i.e. if you as a user have anything between pip 9.x and 18.x, pipenv should still work)

jxltom:
Yeah that will be perfect.

But sometime unexpected bug may block all commits or PR being merged, since they are not actually passed the tests.

But this PR doesn't fix the test (no idea why). So I think it is not helping anyway.

techalchemy:
(don't spend too much time looking, i'm just looking for issue #s before i push mine)

jxltom:
Great, looking forward to the new release!

jxltom:
Close becasue https://github.com/pypa/pipenv/pull/2935

Issue #2816, opened 2018-09-03
Documentation for major version locking


trondhindenes:
We're trying to figure out a way to lock packages using major/minor versions.

I've tried multiple formats, but the documentation doesn't really go into detail on how to lock versions, and more specifically lock major versions of a package. It would be awesome to see a topic on this in the documentation.

uranusjr:
Pipenv uses the same specifier format as pip; it is outlined in [PEP 440](https://www.python.org/dev/peps/pep-0440/#version-specifiers). Contribution on bridging the gap between Pipenv’s documentation and this standard document is very much welcomed.

trallard:
Hi @uranusjr I just created a PR on this, let me know if you find this suitable

trondhindenes:
Thanks @trallard, sorry I should have specified better:
We're looking to version-lock existing packages, which means that we're trying to figure out the toml format and the available options. Essentially we're looking for the easiest way to change a package from '=*' to a locked major-version.

trallard:
@trondhindenes I think if you use `pipenv update <package>` should allow you to update a package version 
Similarly `--selective-upgrade` should do and then pin the version on both your Pipfile and Profile.lock (I might need to do some digging around to provide more accurate feedback) 

On the other hand the PR tackles the point raised by @uranusjr of bridging the gap in the docs (perhaps I should have opened a new issue for this)

trallard:
hey @trondhindenes I just checked this over,  even if you already have a given package installed you can use `pipenv install "package~=x.y.*`  or (`==` although this will definitely prevent you from getting any updates) 

This will then install the specified version, lock it and modify your Pipfile. Hope this helps

GilbertoCS:
Hi @trallard, @uranusjr and @trondhindenes I made some changes on this issue, but for some reason I can not see the update on GitHub. Well, I take another approach for this issue, It is possible to merge both solutions?

uranusjr:
@GilbertoCS Can you submit a pull request with your changes on this issue so I can compare them with @trallard’s side by side and try to merge them? (Apologies if you already did; please mention the PR #.)

GilbertoCS:
Hi @uranusjr as you request I create an PR #3046.

PR #2930, opened 2018-10-06
Documentation: pip specifiers with Pipenv

trallard:
This PR adds details on the pip specifiers and how to use them in conjunction with pipenv

This is related to #2816


uranusjr:
Thanks for putting time into this, I love it! I do wonder however whether it is a little bit too long. Most people would only need `>=` and `<=` (plus their open-ended counterparts), and maybe `≠`. It might be better to just “give a taste” of what is possible, and encourage people to read PEP 440 if they want more.

Using `==` is strongly discouraged since it prevents you from getting any updates at all, so it might be a good idea to intentionally leave it out (again, people can read PEP 440 if they need it). I also feel `~=X.Y` should be preferred to `==X.*`.

Finally, some of the commands in the write up are not very compatible across various platforms. `pipenv install requests> 2.19`, for example, would result in `pipenv install requests` and outputting stdout to a file named `2.19`, since `>` is the special character for [output redirection](https://robots.thoughtbot.com/input-output-redirection-in-the-shell). It is best practice to always put quotes around the requirement, like this:

```
$ pipenv install "requests>=1.4"
$ pipenv install "requests!=2.*"  # Not strictly required, but best practice to ALWAYS quote.
```

It would be nice if there is a little text explaining why quotes are recommended (since PEP 440 does not cover it).

trallard:
👌🏼 thanks for the feedback @uranusjr, based on this I will apply the following changes:
- Remove the use of '==' 
- Add " " around the specifiers (along with a note saying why this is advised)
- Reword the last text (ref to pep 440) so that people actually check this for more advanced/specific cases

trallard:
I have made the changes now and this should be less verbose while still being able to provide a good idea of how this work, let me know what you think

uranusjr:
Waiting for CI to pass before I hit merge…

Issue #84, opened 2017-01-25
How to deactivate current environment entered by `pipenv shell`?


jeyraof:
```bash
❯ python -V
Python 2.7.10

❯ pipenv shell
Spawning environment shell (/bin/zsh).
source /Users/jeyraof/Work/fashion-keyword/.venv/bin/activate

❯ source /Users/jeyraof/Work/fashion-keyword/.venv/bin/activate

(fashion-keyword)❯ python -V
Python 3.6.0

(fashion-keyword)❯ deactivate

❯ python -V
Python 3.6.0

❯ exit

❯ python -V
Python 2.7.10
```

`exit` command is correct way to deactivate current environment?

kfirbreger:
pipenv starts a shell session with the virtualenv instead of changing the bathing in the current shell session. That is why deactivate does not work. you need to exit the shell session. the `exit` command or CTRL-d will do the trick.

jeyraof:
@kfirbreger thank you!

ebuildy:
What about adding a ``pipenv exit`` to "close" the environment (so it can be documented).

uranusjr:
I do not think it is technically possible though; this is, like the activate script, not something we can hook into.

ebuildy:
Sorry, I was more thinking at a kind of alias: ``pipenv exit``  ==> ``exit``, because pipenv help gives the way to activate an env. but no how to exit (yes developers are ... lazy ^^)

uranusjr:
@ebuildy Yeah, that would be very helpful, but as I said, it is not possible.

clsource:
I think a good idea maybe a message will be enough.
That way is a reminder 👍 

 ```
$ pipenv exit
"if you need to exit the shell session. Use the exit command or press CTRL-d"
```

kfirbreger:
@abekroenem The request was for `pipenv exit`, which is quite different than `exit`. The later being a shell command which works, I believe similar to `Ctrl-D`. The former is where the complexity is at.

kfirbreger:
@Oxfos I am not an expert here, however:

1.  I think `exit` as the mechanism for exiting (and `Ctrl-d` for that matter) are not given much attention because that is the normal way to exit a shell. This is not unique to pipenv but works for any shell and iirc also screen, tux etc. Since you start a new shell, you exit it like any other shell, which is different than activating a virtualenv
[...]

PR #2934, opened 2018-10-07
Added shell deactivation instructions to documentation

IamAdiSri:
Issue #84 

r? @uranusjr 

In the thread, you mentioned that an edit to the documentation regarding this issue would be welcome. Can you review this?

### DIFF
--- a/docs/basics.rst
+++ b/docs/basics.rst
@@ -135,7 +135,6 @@ Example Pipfile.lock
 
 Clone / create project repository::
 
-    …
     $ cd myproject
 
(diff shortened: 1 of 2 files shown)


uranusjr:
I think it is enough to mention only `exit`, which works in all cases. No need to complicate things unnecessarily.

Technical tidbit: CTRL-D sends `exit` to the shell, so it is just a shortcut to `exit` anyway.

IamAdiSri:
Thanks for that; I've modified accordingly.

PR #2939, opened 2018-10-08
Add tests for locking editable packages with ref

jxltom:
### The issue

Currently in master branch, following pipfile is not supported, ```pipenv install``` will throw errors.

```
[packages]
requests = {git = "https://github.com/requests/requests.git@883caaf", editable = true}
```

There are these kind of tests in previous pipenv version such as https://github.com/pypa/pipenv/pull/1128/files, but they are gone now for some reason. I didn't do the digging.

But it will be nice to add these tests to make sure that editable package with ref can be locked and installed correctly, and also to make sure that later development in codebase or in upstream won't break these featurres.

These tests added in this PR can pass only after https://github.com/sarugaku/requirementslib/pull/76 is merged from upstrream.

### The checklist

* [ ] Associated issue


jxltom:
Update: it's ready for merge since https://github.com/pypa/pipenv/pull/2940 has been merged

PR #2942, opened 2018-10-08
Remove unused pased_pure_pipfile property

greysteil:
Looks like this property is both a typo and unused. ✂️


techalchemy:
True and true, I've thought about it every time I see it but I never have enough focus to track it down

PR #2948, opened 2018-10-09
Vendor passa

techalchemy:
- Add new spinner for display during locking while also showing output if desired
- Add passa as fallback resolver when regular resolution fails


tasks/vendoring/vendor_passa.py
```diff
@@ -9,10 +9,11 @@
...
     with TemporaryDirectory(prefix='passa') as passa_dir:
         vendor_dir = _get_vendor_dir(ctx).absolute().as_posix()
+        log("Installing requirements for build...")
+        ctx.run("python -m pip install -e git+https://github.com/sarugaku/passa.git#egg=passa plette[validation] requirementslib distlib pip-shims invoke --exists-action=w")
         ctx.run("git clone https://github.com/sarugaku/passa.git {0}".format(passa_dir.name))
         with ctx.cd("{0}".format(passa_dir.name)):
             # ctx.run("git checkout 0.3.0")
-            ctx.run("pip install plette[validation] requirementslib distlib pip-shims -q --exists-action=i")
             log("Packing Passa")
             ctx.run("invoke pack")
```
frostming: only need to pack passa w/o dependencies, which are already in vendors/patched dirs.

This will reduce the package size.

techalchemy: please don't give feedback on my PR unless you understand why i am doing what i did or tried it yourself

techalchemy: being that it runs in a zipfile and without access potentially to anything in pipenv's path, it doesn't really make any sense to trust the vendoring in pipenv

matteius:
This is the oldest open PR here, and now there are conflicts.  It has a confusing title, doesn't relate back to a particular issue or PEEP as far as I can tell.   What should the outcome be here?

oz123:
@matteius it seems even back then it was not clear what this PR is about.
Passa hasn't seen much development since this PR too.
I suggest leaving this open for 30 more days and then closing it.

PR #2953, opened 2018-10-09
fix virtualenv assertion bug

craynic:
introduced in 999761130e910bb211c75467d66be7db7547e7fe.

issue #2957


craynic:
The failed test passed in my environment.

Jamim:
Thank you, @craynic!
I would like to see this PR merged!

cc @uranusjr, @techalchemy

techalchemy:
seems about right, can you add a news entry for it please? pr template :dagger:

Issue #2960, opened 2018-10-09
TOML parsing fails for outline tables


greysteil:
### Issue description

When using Pipenv with outline tables like the below, parsing the `Pipfile` fails:

```
[[source]]
name = "pypi"
url = "https://pypi.python.org/simple/"
verify_ssl = true

[packages.raven]
version = "== 5.27.1"
```

### Expected result

Successful install

### Actual result

```
Traceback (most recent call last):
  File "/usr/local/bin/pipenv", line 11, in <module>
    sys.exit(cli())
  File "/usr/local/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/pipenv/vendor/click/decorators.py", line 64, in new_func
    return ctx.invoke(f, obj, *args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/pipenv/cli/command.py", line 308, in lock
    ensure_project(three=state.three, python=state.python, pypi_mirror=state.pypi_mirror)
  File "/usr/local/lib/python3.6/site-packages/pipenv/core.py", line 565, in ensure_project
    if project.required_python_version:
  File "/usr/local/lib/python3.6/site-packages/pipenv/project.py", line 229, in required_python_version
    required = self.parsed_pipfile.get("requires", {}).get(
  File "/usr/local/lib/python3.6/site-packages/pipenv/project.py", line 448, in parsed_pipfile
    parsed = self._parse_pipfile(contents)
  File "/usr/local/lib/python3.6/site-packages/pipenv/project.py", line 480, in _parse_pipfile
[...]

PR #2971, opened 2018-10-09
Fix parsing of outline tables

greysteil:
Fixes https://github.com/pypa/pipenv/issues/2960, which was caused by a change in the API for `toml`.


techalchemy:
Awesome, thanks a lot for the fix, don't worry about rebasing all night i'll be merging a few more changes still but i'll grab this

greysteil:
Awesome, thanks @techalchemy, and thanks for releasing `2018.10.9`!

PR #2981, opened 2018-10-10
Fix trusted-host passthru

### LINKED ISSUE #2979: `--trusted-host` not passed to `pip` when installing from a lockfile
### Issue description

On the *latest* pipenv release, running `pipenv install` does not pass `--trusted-host` to `pip`.

Running `pipenv install xyz` *does*.

### Expected result

`--trusted-host` is passed to `pip`.

### Actual result

```
> pipenv install --verbose                                
Installing dependencies from Pipfile…
Installing 'aiohttp'▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉ 0/7 — 00:00:00
$ ['/Users/tom/.local/share/virtualenvs/tmp.YkTM0kln-68P7VPaZ/bin/pip', 'install', '--verbose', '--upgrade', '--no-deps', '-r', '/var/folders/9w/twrv54nd0v19gsbjhjs3mlhc0000gn/T/pipenv-wgs4narf-requirements/pipenv-h6s3thx1-requirement.txt', '-i', 'http://x.x.co.uk:8080', '--require-hashes']
```

And:

```
> pipenv install --verbose pytest
⠋Installing 'pytest'
$ ['/Users/tom/.local/share/virtualenvs/tmp.YkTM0kln-68P7VPaZ/bin/pip', 'install', '--verbose', '--upgrade', 'pytest', '-i', 'http://x.x.co.uk:8080', '--trusted-host', 'x.x.co.uk']
```

### Steps to replicate

Have a `Pipfile` like this:

```
[[source]]
url = "http://INTERNAL_PYPI:8080"
verify_ssl = false
name = "xyz"

[packages]
aiohttp = "*"
```

Notice how the argument is not passed to pip when calling `pipenv install`.

techalchemy:
Fix trusted-host passthru

- Fix marker cleaning
- Fixes #2979

### DIFF
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -780,6 +780,7 @@ def do_install_dependencies(
                 requirements_dir=requirements_dir,
                 extra_indexes=extra_indexes,
                 pypi_mirror=pypi_mirror,
+                trusted_hosts=trusted_hosts
             )
             c.dep = dep
             c.ignore_hash = ignore_hash
@@ -1307,10 +1308,13 @@ def pip_install(
     requirements_dir=None,
     extra_indexes=None,
     pypi_mirror=None,
+    trusted_hosts=None
 ):
     from notpip._internal import logger as piplogger
 
     src = []
+    if not trusted_hosts:
+        trusted_hosts = []
 
     if environments.is_verbose():
         piplogger.setLevel(logging.INFO)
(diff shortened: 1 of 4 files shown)


pipenv/core.py
```diff
@@ -1335,23 +1339,30 @@ def pip_install(
...
     if index:
-        if not is_valid_url(index):
-            index = project.find_source(index).get("url")
-        sources = [{"url": index}]
+        try:
+            index_source = project.find_source(index)
+            index_source = index_source.copy()
+        except SourceNotFound:
+            src_name = project.src_name_from_url(index)
+            verify_ssl = True if index not in trusted_hosts else False
```
orf: You may have considered this already, but while I was testing the issue `index` was set to something like `https://MY_INDEX:8080`, whereas `trusted_hosts` was just a list like`['MY_INDEX']`.

techalchemy: no I absolutely in no way was thinking about that, I should probably parse netlocs :D thanks for the comment!

techalchemy: https://github.com/pypa/pipenv/pull/2993/commits/272e98cb6749089341dd05e509a283d358cf2a59

pipenv/core.py
```diff
@@ -1335,23 +1339,30 @@ def pip_install(
...
                 try:
-                    extra_src = project.find_source(idx).get("url")
+                    extra_src = project.find_source(idx)
                 except SourceNotFound:
-                    extra_src = idx
-                if extra_src != index:
-                    sources.append({"url": extra_src})
+                    src_name = project.src_name_from_url(idx)
+                    verify_ssl = True if idx not in trusted_hosts else False
+                    extra_src = {"url": idx, "verify_ssl": verify_ssl, "name": extra_src}
```
orf: I hate to do this to you, but `extra_src` is undefined here if the exception is triggered

PR #2986, opened 2018-10-10
Keyword arg error in Getting `add_index_to_pipfile`

chiqomar:
Latest release of pipenv gave me this error; seemed like an easy fix

### The issue

Looks like a call to `add_index_to_pipfile` uses a keyword called `trusted_host`, but the keyword is `verify_ssl`

### The fix

Changed `trusted_host` to `verify_ssl`. Doesn't look like it's referenced anywhere else. And it fixed pipenv for me.
Don't believe there is an issue for it though.

### The checklist

* [ ] Associated issue


techalchemy:
Good catch, thanks!

Issue #2996, opened 2018-10-11
Set PIPENV_ACTIVE when running pipenv run


maw:
### Is your feature request related to a problem? Please describe.

I have a script to wrap pyre that ends thus:

```
DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"

if [[ x"$PIPENV_ACTIVE" != x"1" ]] ; then
   echo activate pipenv by running pipenv shell or pipenv run "$DIR"/pyre
   exit 2
fi

exec pyre --search-path "$(pipenv --venv)/lib/python3.6/site-packages" "$@"
```

Unfortunately, the second suggestion doesn't work as things currently are.

### Describe the solution you'd like

`PIPENV_ACTIVE` should be set when you run `pipenv run`, just as it is when you run `pipenv shell`.

### Describe alternatives you've considered

None.

### Additional context

I'm about to submit a pull request. Is running `pipenv --support` really necessary here?

-

techalchemy:
Ah closing this as I think I missed it when I mass merged some prs, thanks for the fix!

PR #2997, opened 2018-10-11
Set PIPENV_ACTIVE when running pipenv run

maw:

### The issue

This is for https://github.com/pypa/pipenv/issues/2996. 

### The fix

It's a very small change, and it makes `pipenv shell` and `pipenv run` more consistent.

### The checklist

* [x] Associated issue

### DIFF
--- a/news/2996.trivial
+++ b/news/2996.trivial
@@ -1,0 +1,1 @@
+Set `PIPENV_ACTIVE=1` when running `pipenv run`.  This is what `pipenv shell` already does.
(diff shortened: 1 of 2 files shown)


uranusjr:
I feel it could be worthwhile to extract this logic to its own function for both `run` and `shell`. I know it is a one-liner, but it is a distinct operation that could be extended further in the future.

maw:
@uranusjr Good point.  There was already some repeated code there; with my change there is a little bit more of it.

uranusjr:
I guess there’s no harm in doing this. It’s probably a good idea in general to minimise behavioural differences between `run` and `shell` anyway, and this would be in the right direction.

Issue #3002, opened 2018-10-11
Unable to lock packages with odd version numbers


orf:
### Issue description

Running `pipenv install pdfminer.six` fails.

### Expected result

It succeeds to install the package

### Actual result

```
> pipenv install pdfminer.six --verbose
Installing pdfminer.six…
⠋Installing 'pdfminer.six'
$ ['/Users/tom/.virtualenvs/tempenv-6fe770333809/bin/pip', 'install', '--verbose', '--upgrade', 'pdfminer.six', '-i', 'https://pypi.org/simple']
Created temporary directory: /private/var/folders/9w/twrv54nd0v19gsbjhjs3mlhc0000gn/T/pip-ephem-wheel-cache-fojnad2a
Created temporary directory: /private/var/folders/9w/twrv54nd0v19gsbjhjs3mlhc0000gn/T/pip-req-tracker-avndeix4
Created requirements tracker '/private/var/folders/9w/twrv54nd0v19gsbjhjs3mlhc0000gn/T/pip-req-tracker-avndeix4'
Created temporary directory: /private/var/folders/9w/twrv54nd0v19gsbjhjs3mlhc0000gn/T/pip-install-putzczfh
1 location(s) to search for versions of pdfminer.six:
* https://pypi.org/simple/pdfminer-six/
Getting page https://pypi.org/simple/pdfminer-six/
Looking up "https://pypi.org/simple/pdfminer-six/" in the cache
Request header has "max_age" as 0, cache bypassed
Starting new HTTPS connection (1): pypi.org:443
https://pypi.org:443 "GET /simple/pdfminer-six/ HTTP/1.1" 304 0
Analyzing links from page https://pypi.org/simple/pdfminer-six/
  Found link https://files.pythonhosted.org/packages/ab/60/8df8a1f2c56e191fef9cb3dc70f951ef7ef5df85523382d45a94180d38de/pdfminer.six-20140915.zip#sha256=b734cbad2d9431a794e2c9cb6edfb1aad36f20461b65d20af8d89c3469453fe6 (from https://pypi.org/simple/pdfminer-six/), version: 20140915
......
Cleaning up...
Removed build tracker '/private/var/folders/9w/twrv54nd0v19gsbjhjs3mlhc0000gn/T/pip-req-tracker-avndeix4'

Adding pdfminer.six to Pipfile's [packages]…
Pipfile.lock not found, creating…
Locking [dev-packages] dependencies…
Locking [packages] dependencies…
[...]

PR #3007, opened 2018-10-11
Fix normalization during install

techalchemy:
Escape command line requirements as necessary

- Fixes #2998

PR #3008, opened 2018-10-11
Correctly normalize package names. Fixes #2956 and #3002

### LINKED ISSUE #2956: Pipenv install parsing wrong package name
### Issue description
After the latest patch for the incompatibility problem with pip 18.1, the pipenv parsing of the package name during installation no longer is working (it's replacing `.` in the package name for a `-` and this leads to pip not finding the available versions).

This happens with pip 18.1 as well as 18.0.

### Expected result
`pipenv install zope.interface==4.5.0` should install from this package https://pypi.org/project/zope.interface/ 

### Actual result
Instead it translates to `zope-interface==4.5.0` and then it don't find the package:
```
Could not find a version that satisfies the requirement zope-interface==4.5.0 (from versions: 4.4.0, 4.4.1, 4.4.2, 4.4.3)
No matching distribution found for zope-interface==4.5.0
```

### Steps to replicate
create new virtualenv with latest pipenv version (2018.10.9) to see it break, and test it with  the previews one (2018.7.1) where it's still ok.

**obs**: I'm not sure if this is a bug in pip or pypi (not finding the correct version when replacing `.` with `-`), but since it started happening with the latest pipenv release then I believe it is actually related to pipenv.

orf:
Fixes #2956 and #3002

### DIFF
--- a/pipenv/patched/notpip/_internal/index.py
+++ b/pipenv/patched/notpip/_internal/index.py
@@ -766,7 +766,7 @@ def egg_info_matches(
         return full_match[full_match.index('-'):]
     name = match.group(0).lower()
     # To match the "safe" name that pkg_resources creates:
-    name = name.replace('_', '-')
+    name = name.replace('_', '-').replace('.', '-')
     # project name and version must be separated by a dash
     look_for = search_name.lower() + "-"
     if name.startswith(look_for):


pipenv/patched/notpip/_internal/index.py
```diff
@@ -766,7 +766,7 @@ def egg_info_matches(
         return full_match[full_match.index('-'):]
     name = match.group(0).lower()
     # To match the "safe" name that pkg_resources creates:
-    name = name.replace('_', '-')
+    name = name.replace('_', '-').replace('.', '-')
```
techalchemy: ^_^ I am actually tempted to do this, but I have a preference for not patching pip whenever possible.

I'm likely going to hold off on accepting this for the next release to see about getting feedback from the pip maintainers when I have a bit more clarity on all of the places we are currently working around this problem, which I think is pretty numerous by now...

orf: It's up to you if you accept it, and I understand if you do not, but this is currently breaking our (am I am sure others) workflow entirely. I can't see a workaround apart from downgrading

techalchemy: @orf, apart from the issues mentioned and the fixes for those specifically (which are in master), can you describe what is continuing to cause you problems? I.e. with the current master branch, how do I see a problem?

orf: I will get back to you here tomorrow morning, after I try installing the specific pipfile I have with the current master

techalchemy: ah, I see the issue in question.  This should be resolved by #3007

cjerdonek:
@techalchemy I don't know much about this aspect of pip. Note that @uranusjr just opened this PR which touches this very code: https://github.com/pypa/pip/pull/5875
Does that PR address the issue being raised here?

pipenv/patched/notpip/_internal/index.py
orf: Awesome! Thanks for your continued work on pipenv, and I'm sorry for the volume of issues!

I will confirm tomorrow that this is fixing it, or though I cannot see any code that would after my brief lookm

techalchemy:
We don't even communicate that well when we are co-maintainers of the same projects

I don't think that one will fix this issue, but it is another case where this code is in disagreement to be sure

cjerdonek:
Yeah, I figured that might have been happening so wanted to be sure you were at least alerted here!

techalchemy:
@uranusjr nope, as far as I understand we don't need that change for this to work

techalchemy:
closing for now since we will catch this upstream in the next release of pip, thanks for the PR and thanks for the input all

Issue #796, opened 2017-10-01
How does pipenv know the virtualenv for current project ?


sachinjain024:
I have just setup a python project using pipenv. One thing I want to know is where does pipenv store the mapping of project directories to virtualenvs. 

To be clear, I am running the following command:

    sacjain-macOS:coursera-classification sacjain$ pipenv --venv
    /Users/sachinjain/.local/share/virtualenvs/coursera-classification-iTBt6WsT

I checked Pipfile and Pipfile.lock, I assumed this information should be present in Pipfile but it was not there.

Q2. How can we install a specific version of package using pipenv ?

uranusjr:
The virtualenv’s name is the project’s root directory name, plus the hash of the full path to the project root. This guarantees the name to be unique and predictable as long as you don’t move the project. (AFAICT this name generation logic is an implementation detail that should not be relied on.)

To install a specific version of a package, use the same syntax as you would with pip and requirements.txt, `package==version` (e.g. `django==1.11.0`).

sachinjain024:
@uranusjr This means that if I rename the directory my project's virtualenv will be lost. That's bad. It can be a common use case to rename a directory and users may fall into this and wonder how their environment stopped working.

I would suggest to fix it and make an entry into Pipfile if that's possible. Or create another .pipenv file to store such metadata in each project directory.

What do you suggest ?

Thanks for answering both the questions!

nateprewitt:
Hey @sachinjain024, we've had a few discussions on the local metadata file but the maintainer consensus has been that pipenv won't be supporting that at this time.

The directory path identifying a project was something that was implemented early on because projects with the same name, or multiple copies of the same project, were causing accidental overwriting of environment state.

In practice, we've found far fewer people have had issues with the location being part of the environment than silently having their work overwritten. Pipenv is an deployment management tool after all, so moving the directory should be a one command fix.

sachinjain024:
Thanks @nateprewitt. It makes sense to quickly jump to easy implementation. Suppose I rename a project at later stage then it means my environment does not exist. 

So what should I do in this situation ?

1. Resetup the environment because I have Pipfile already which has the packages information so it will be easy to resetup the env.
2. Somehow rename the virtualenv directory too
3. ???

What do you suggest in this case ? Just curious in case I get in this situation in future.

uranusjr:
You can use `pew cp` to easily copy the existing virtualenv, but 1. would be the more “canonical” way to it I think.

[...]

PR #3024, opened 2018-10-14
PEEP proposal for venv location

gsemet:
Following #796, my humble proposal

### DIFF
--- a/peeps/PEEP-003.md
+++ b/peeps/PEEP-003.md
@@ -1,0 +1,61 @@
+# Virtual-env folder location
+
+Project maintainer may have different opinion on how their project'v virtualenv should be located.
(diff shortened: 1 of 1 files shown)


yuwash:
I indeed don’t see a necessity for `PIPENV_INSTALL_VENV_DIR=custom/path` or the corresponding `--venv-dir`, but the option `--venv` is basically the same as calling pipenv with `PIPENV_VENV_IN_PROJECT=1`, so I think that’s not a vastly new thing to the project. A short command-line argument is easier to use when it’s only needed on setup.

gsemet:
Now I would recommend a setting inside pipenv.toml that can be overridden by parameter.

It is always messy to let all users know they have to set this variable or create a .venv folder prior to do the pipenv install.

I use poetry now at least we can commit a setting in poetry.toml

PR #3050, opened 2018-10-16
Add brief description and caveats of project - virtualenv mapping

floatingpurr:
### The issue

It is not immediately clear how pipenv knows the virtualenv for a project  (See #796)

### The fix

Docs extension

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -107,16 +107,6 @@ Other Commands
 - ``check`` checks for security vulnerabilities and asserts that PEP 508 requirements are being met by the current environment.
 
 
-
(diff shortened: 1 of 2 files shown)


uranusjr:
Thanks for putting this together! I don’t feel this should be on the front page—it already has too many things (I’ve been thinking if it’s possible to removing things from it, actually). How about putting this in either *Pipenv & Virtual Environments* or *Advanced Usage of Pipenv*?

floatingpurr:
Hey @uranusjr, thank you for your suggestion. I did it.

PR #3060, opened 2018-10-18
Contributing guide - add link to CONTRIBUTING.md

pfctdayelise:
It took me a while to realise the CONTRIBUTING.md had useful info not in the Contributing docs, so I've added a link between them. I imagine this doesn't need a separate issue or changelog callout but please correct me if wrong.

### The checklist

* [ ] Associated issue


uranusjr:
Actually I wonder if it is a good idea to move (or at least copy) the development setup section into documentation, instead of linking to it… (I didn’t even know it exists)

pfctdayelise:
It's not entirely clear to me what the difference between the two documents should be. So I could make the .md file just say "see <url>" and merge the content into the docs page?

uranusjr:
CONTRIBUTING.md is essentially just a customary file to put these kind of information in. It also shows up (as a link) on the right hand side when you submit an issue or PR on GitHub. It would make more sense (IMO) to do this the other way around: add “see <link>” in CONTRIBUTING.md to point to the documentation.

pfctdayelise:
Hmm, pretty sure I mangled this by attempting to rebase. Let me make a clean attempt.

Issue #3041, opened 2018-10-15
Command-line option "--bare" has no effect in "sync" and "clean" actions


cubranic:
### Issue description

The `--bare` flag is supposed to set "minimal output" in `sync` and `clean`, but I don't see any difference with and without it. Looking at the [code](https://github.com/pypa/pipenv/blob/master/pipenv/core.py#L2469) implementing those actions, I see `bare` in the list of parameters for `do_sync` and `do_clean`, but they are not used anywhere in the body of those functions.

### Expected result

No output when `--bare` is used in `pipenv sync`.

techalchemy:
Yep, that's definitely right.  It's been at the back of my mind to do something about this for awhile but there is a pretty sizeable project to tackle around output cleanup in general

frostming:
I think the meaning of 'bare' here is to run the command on itself, instead of called by other commands. So if you run `pipenv sync`, the 'bare' flag should be True by default.

Just to share what is in my head.

cubranic:
@frostming Check the help text for the `--bare` option: that demonstrably is not what it is intended to be.
Then look at any other command where it's actually functional. There are various lines where status is printed wrapped in `if not bare:`.

frostming:
@cubranic ya you are right. I didn't check source code thoroughly.

jcrotts:
Opened a PR for this, I think it addresses the behavior you described @cubranic .

techalchemy:
This was closed in #3126 by way of #3064 and will be in the next release, thanks @jcrotts!

PR #3064, opened 2018-10-19
Add --bare behavior to sync and clean

jcrotts:
Implement some behavior for --bare in sync and clean, and add bare an option to do_clean in cli.py

### The issue

https://github.com/pypa/pipenv/issues/3041

### The fix

Added logic for --bare to sync and clean since they are passed in as arguments. 
Excluding error messages.

Also added --bare as a click option to clean.

### The checklist

* [x] Associated issue

### DIFF
--- a/news/3041.feature
+++ b/news/3041.feature
@@ -1,1 +1,1 @@
---bare now has an effect on clean, and use sync's bare option to reduce output.
+--bare now has an effect on clean, and sync's bare option is now used to reduce output.
--- a/news/3041.feature
+++ b/news/3041.feature
@@ -1,0 +1,1 @@
+--bare now has effect on sync and clean.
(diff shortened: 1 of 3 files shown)


uranusjr:
I only see it be added to `clean`? Also, a news fragment to describe the change would be nice.

jcrotts:
--bare was an option for sync already in command.py, but I don't think it actually did anything.

Sure thing I'll add a fragment, would it go under .trivial ?

uranusjr:
I see. Does the `sync` command’s `--bare` option have the intended behaviour now? Since this PR does not change `sync`, it might be better to only mention adding `--bare` to `clean` in the commit message and news fragment.

jcrotts:
One change was made to sync that removes one output line. I can change the commit message and news fragment if you think it would be clearer.

uranusjr:
It would be best to either 1) remove mention of `sync` from fragment, or 2) use `sync`’s bare option to reduce output. I am fine with either, your call. (And if you opt for 1), another PR on `sync` would be nice as well!)

uranusjr:
You don’t need to, I can squash on merge (but you can squash yourself as well, there’s no difference).

Issue #3068, opened 2018-10-20
Impossible to hide the "Courtesy Notice" about active virtualenv


pior:
### Issue description

When using an existing virtualenv, we get this message:

> Courtesy Notice: Pipenv found itself running within a virtual environment, so it will automatically use that environment, instead of creating its own for any project. You can set PIPENV_IGNORE_VIRTUALENVS=1 to force pipenv to ignore that environment and create its own instead. You can set PIPENV_VERBOSITY=-1 to suppress this warning.

### Expected result

With `export PIPENV_VERBOSITY=-1` the **Courtesy Notice** message should not be shown.

### Actual result

The message still appear with `export PIPENV_VERBOSITY=-1`.

techalchemy:
This was merged so i'll go ahead and close it, thanks for reporting!

PR #3069, opened 2018-10-20
Fix PIPENV_VERBOSITY parsing logic

pior:
### The issue

The issue is describe in #3068 : Impossible to hide the "Courtesy Notice" about active virtualenv.

The origin of the issue is that a value of `-1` for `PIPENV_VERBOSITY` is ignored but the current logic.

### The fix

`isdigit()` is not adapted to test a negative value:
```
In [1]: '-1'.isdigit()
Out[1]: False
```

Use `int()` to cast and catch `ValueError`.

### The checklist

* [x] Associated issue

### DIFF
--- a/news/3069.trivial
+++ b/news/3069.trivial
@@ -1,0 +1,1 @@
+- Fix PIPENV_VERBOSITY parsing logic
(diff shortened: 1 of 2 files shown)


techalchemy:
My only feedback on this is I’d prefer to permanently suppress this specific deprecation warning at runtime irrespective of verbosity settings

pior:
@uranusjr Thanks! I'm catching TypeError now. I feel it's sufficient...

PR #3070, opened 2018-10-21
Fixed unnecessary extras are added in lockfiles

jxltom:
### The issue

Fixed https://github.com/pypa/pipenv/issues/3026

### The fix

This is a annoying critical bug. Fixed by do not append ```extra``` when translating markers since they are unnecessary. Also test is added to avoid regression in further development.

### The checklist

* [x] Associated issue

### DIFF
--- a/news/3026.bugfix
+++ b/news/3026.bugfix
@@ -1,0 +1,1 @@
+Fixed unnecessary extras are added when translating markers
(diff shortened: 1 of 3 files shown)


techalchemy:
Thanks for this @jxltom — I have a few other fixes in some other libraries in the air still, but should cut this tomorrow ideally.

jxltom:
Great 😄

Turbo87:
@techalchemy any news on that release?

PR #3074, opened 2018-10-22
Add specific testing instructions to contributing.rst

jeremycarroll:
### The issue

I wanted to start contributing to the project, the instruction "Run the tests to confirm they all pass on your system" is non-trivial, and took longer than really needed to put into effect.

### The fix

I documented what I did to perform this task.

### Open Concern

The first way I suggest running the tests is `make test`; however, this does not work for me.
I don't know what it would take to have success, but believe it is a legitimate way to achieve the goal.

### The checklist

* [None ] Associated issue
* [Not appropriate] A news fragment in the `news/` directory to describe this fix with the extension `.bugfix`, `.feature`, `.behavior`, `.doc`. `.vendor`. or `.trivial` (this will appear in the release changelog). Use semantic line breaks and name the file after the issue number or the PR #.

### DIFF
--- a/docs/dev/contributing.rst
+++ b/docs/dev/contributing.rst
@@ -131,7 +131,7 @@ Three ways of running the tests are as follows:
 3. Using pipenv::
 
     pipenv install --dev
-    pipenv run py.test
+    pipenv run pytest
 
 For the last two, it is important that your environment is setup correctly, and
 this may take some work, for example, on a specific Mac installation, the following
--- a/docs/dev/contributing.rst
+++ b/docs/dev/contributing.rst
@@ -127,7 +127,7 @@ Run the tests
 Three ways of running the tests are as follows:
 
 1. ``make test`` (which uses ``docker``)
-2. ``./run-tests.sh``
+2. ``./run-tests.sh`` or ``run-tests.bat``
 3. Using pipenv::
 
     pipenv install --dev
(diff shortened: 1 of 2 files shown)


docs/dev/contributing.rst
```diff
@@ -120,3 +120,38 @@ hasn't been reported before. Duplicate bug reports are a huge drain on the time
...
+-------------
+
+Three ways of running the tests are as follows:
+
+1. ``make test`` (which uses ``docker``)
+2. ``./run-tests.sh``
+3. Using pipenv::
+
+    pipenv install --dev
+    pipenv run py.test
```
uranusjr: I believe `pytest` is recommended over `py.test` by Pytest folks now. Also you might want to consider Windows for these instructions :) Most of these won’t work on Windows.

jeremycarroll: I changed to `pytest` as suggested.

OTOH, I am not inclined (or able) to also document what is needed for Windows. I agree that the example config instructions won't work for most people. They are explicitly an example of what might be required, and I think they will be helpful even for people who have to modify them. I also think more examples from more platforms would be good.

techalchemy:
this PR looks good, lets use it as a starting point, although I'm 99% sure that `run-tests.bat` doesn't work on windows. I'm sure we will hear about it to confirm :)

PR #3080, opened 2018-10-22
Add dotenv config

### LINKED ISSUE #3077: dotenv configuration
1. I am using the latest version of Flask which supports .env files when `python-dotenv` package is installed. It conflicts with `dotenv` package vendored within pipenv.

It would be nice to have `dont_load_env` (default `false`) at `pipenv` section within Pipfile so there will be no need to provide `PIPENV_DONT_LOAD_ENV=1` every time running pipenv. It would be useful not only in case of Flask usage but in case I want to use `python-dotenv` on my own.

2. Pipenv loads .env file using `override=True` argument so I cannot overwrite environment from CLI:
`VAR=1 pipenv run ...`.

It would be great to have `dont_override_env` (default `false`) at `pipenv` section within Pipfile (and `PIPENV_DONT_OVERRIDE_ENV` environment variable also).

decaz:
Added ``PIPENV_DONT_OVERRIDE_ENV`` environment variable, ``dont_load_env`` and ``dont_override_env`` settings at Pipfile.

Closes #3077.


decaz:
@techalchemy I have added PEEP.

Issue #3081, opened 2018-10-22
pipenv open MODULE – custom command


metakermit:
I find the `pipenv open` command super useful for doing little module source code inspection, even a quick patch & pull request. Would it be appropriate to add an option for a custom command to open the module with? For example:

```
pipenv open --command=atom MODULE
```

to execute `atom path/to/the/module` and open the project in Atom.app. Or for a quick look in Finder to drag & drop some files:

```
pipenv open --command=open MODULE
```

I don't always work on a computer where `$EDITOR` is something meaningful, so this would save me a few steps of fiddling.

A lighter-weight alternative (or an addition – for quickly `cd`-ing in the terminal too) would be if there was an option to just output the path, e.g.:

```shell
$ pipenv open --path responder
/Users/kermit/.virtualenvs/responder-react-aladPS82/src/responder/responder
```

which would allow me to do:

```shell
atom `pipenv open --path responder`
```

One could easily jump to the folder with a quick:

```shell
cd `pipenv open --path responder`
```

What do you think?

uranusjr:
I believe it respects the customary `EDITOR` environment variable.

metakermit:
True, it's already possible to override the variable temporarily by doing:

```
EDITOR=atom pipenv open responder
```

or

```
EDITOR=open pipenv open responder
```

If we go for that approach, I'd at least mention the variable in the help page for the open command. Something like:

```
$ pipenv open --help

Usage: pipenv open [OPTIONS] MODULE

  View a given module in your editor, as defined by the EDITOR environment variable. You can temporarily override it, for example: EDITOR=atom pipenv open responder
```

uranusjr:
Pull requests are always welcomed!

PR #3083, opened 2018-10-23
pipenv open custom command help #3081

metakermit:
As discussed in https://github.com/pypa/pipenv/issues/3081 clarifies how custom commands can be used with `pipenv open`.

To test:

```
$ pipenv run python -m pipenv open --help
Usage: __main__.py open [OPTIONS] MODULE

  View a given module in your editor, as defined by the EDITOR environment
  variable. You can temporarily override it, for example:

      EDITOR=atom pipenv open requests

Options:
  --clear             Clears caches (pipenv, pip, and pip-tools).
```

### DIFF
--- a/pipenv/cli/command.py
+++ b/pipenv/cli/command.py
@@ -533,7 +533,13 @@ def graph(bare=False, json=False, json_tree=False, reverse=False):
 @argument("module", nargs=1)
 @pass_state
 def run_open(state, module, *args, **kwargs):
-    """View a given module in your editor."""
+    """View a given module in your editor.
+
+    This uses the EDITOR environment variable. You can temporarily override it,
+    for example:
+
+        EDITOR=atom pipenv open requests
+    """
     from ..core import which, ensure_project
 
     # Ensure that virtualenv is available.


pipenv/cli/command.py
```diff
@@ -533,7 +533,11 @@ def graph(bare=False, json=False, json_tree=False, reverse=False):
 @argument("module", nargs=1)
 @pass_state
 def run_open(state, module, *args, **kwargs):
-    """View a given module in your editor."""
+    """View a given module in your editor, as defined by the EDITOR environment variable.
+    You can temporarily override it, for example:
+
+        EDITOR=atom pipenv open requests
+    """
```
uranusjr: I think it’s better to format this like

```
View a given module in your editor.

    This uses the EDITOR environment variable. You can temporarily override it,
    for example:

        EDITOR=atom pipenv open requests
```

It is also a good idea to wrap this to under 72 characters, so the message won’t break on narrow terminals.

metakermit: OK, I've updated the formatting and squashed it into a single commit.

Yeah, I think something in the text parsing treats the paragraph as a whole no matter where we put the line break. Looks OK on my terminal I'd say:

screenshot 2018-10-23 10 42 19

uranusjr: Hmm alright then

PR #3097, opened 2018-10-25
Fix typo in docs/install.rst

greyli:
py --> python


frostming:
IMO `py` stands for Py launcher shipped with Python 3(Windows only). So the original statement looks good to me.

greyli:
@frostming Thanks! I didn't know it before. Does it ship with Python 2?

frostming:
@greyli No. The Py launcher aims at managing multiple Python versions FYI. Glad to help.

Issue #3102, opened 2018-10-25
`TypeError: can only concatenate str (not "type") to str` instead of actual error


aidin36:
### Issue description

In some cases, `pipenv` outputs `TypeError: can only concatenate str (not "type") to str` instead of the actual error (or after printing out the actual error). It faces an error while handling the actual error.

### Steps to replicate

Set WORKON_HOME to somewhere that current user doesn't have permission for writing to. Example:

`WORKON_HOME=/opt/  pipenv install`

-------------------------------------------------------------------------------
I encountered the issue on the latest source code from the Github.

equaliser0:
I get this error by starting `pipenv shell`. There actually two errors I found:

```
ModuleNotFoundError: No module named 'pkgutil'
```
and

```
TypeError: can only concatenate str (not "type") to str
```

directly after running  `pipenv shell`

I am not sure, what I can do

techalchemy:
The issue is a problem with a dependency in pipenv. You can fix this by installing from the working update until the release is out: `pip install git+https://github.com/pypa/pipenv.git@update-vendor#egg=pipenv`

jxltom:
Hi @techalchemy Could you look at this issue again? I still get this issue when I installed latest master branch with commit e3ee806. Thanks.

techalchemy:
Yikes. Any chance you can print or log the type? I’m guessing it’s an EOF still?

jxltom:
@techalchemy Hi, I found it is ```TIMEOUT```. I made a PR https://github.com/pypa/pipenv/pull/3159 for that.

namanipradeep:
Hi ,

When I am trying to build pypika query for casting I am also getting same error, The code snippet will be(TypeError: can only concatenate str (not "type") to str)

col_name = 'fn.Cast(' + aliasName + '.' + columnNamevalue + ','+ SqlTypes +'.'+ castTypevalue+')'

Can any one help on this

PR #3103, opened 2018-10-25
Issue #3102 - Converts output of subprocess to str in `_pexpect_out`

aidin36:
### The issue

Issue #3102

### The fix

Output of `subprocess.after` is not always a `str`, and cannot be concatenate to `str`. This PR converts the variable to `str` before concatenation.

### The checklist

* [x] Associated issue

### DIFF
--- a/pipenv/vendor/delegator.py
+++ b/pipenv/vendor/delegator.py
@@ -111,7 +111,7 @@ class Command(object):
             result += self.subprocess.before
 
         if self.subprocess.after:
-            result += self.subprocess.after
+            result += str(self.subprocess.after)
 
         result += self.subprocess.read()
         return result


aidin36:
The error only happening in the master branch and not released yet, so I'm not sure if I should include a file in `news`.

techalchemy:
Hey there. Thanks for the PR. Things in the `vendor` folder unfortunately need to be modified with a corresponding patch or they will be dropped when we re-vendor the dependencies. For this specific case however the output it is giving in addition to strings is `type` instance which is an EOF object and can’t be stringified. There is an open PR in delegator and a corresponding one here to fix these issues

Thanks for the PR!

Issue #3085, opened 2018-10-23
ValueError when env variable password contains hash


Vayel:
### Issue description

When a password passed through an env variable contains a hash (e.g. `ab#cd`), a `ValueError` is raised when calling `pipenv install`. The error is **not** raised when the username also contains a hash.

### Expected result

Just install the dependencies if the credentials are valid.

### Actual result

```
Pipfile.lock not found, creating…
Locking [dev-packages] dependencies…
Locking [packages] dependencies…
using sources: [{'url': 'https://pypi.org/simple', 'verify_ssl': True, 'name': 'pypi'}, {'url': 'https://user:pass#word@simudyne.jfrog.io/simudyne/api/pypi/simudyne-pypi/simple', 'verify_ssl': True, 'name': 'simudyne_pypi'}]
Using pip: -i https://pypi.org/simple --extra-index-url https://user:pass#word@simudyne.jfrog.io/simudyne/api/pypi/simudyne-pypi/simple

                          ROUND 1                           
Current constraints:
  ipykernel
  simudyne

Finding the best candidates:

Traceback (most recent call last):
  File "/home/simudyne/.local/lib/python3.6/site-packages/pipenv/resolver.py", line 101, in <module>
    main()
  File "/home/simudyne/.local/lib/python3.6/site-packages/pipenv/resolver.py", line 90, in main
    system=system,
  File "/home/simudyne/.local/lib/python3.6/site-packages/pipenv/resolver.py", line 73, in resolve
    allow_global=system,
  File "/home/simudyne/.local/lib/python3.6/site-packages/pipenv/utils.py", line 422, in resolve_deps
    req_dir=req_dir,
  File "/home/simudyne/.local/lib/python3.6/site-packages/pipenv/utils.py", line 302, in actually_resolve_deps
    results = resolver.resolve(max_rounds=environments.PIPENV_MAX_ROUNDS)
  File "/home/simudyne/.local/lib/python3.6/site-packages/pipenv/patched/piptools/resolver.py", line 102, in resolve
    has_changed, best_matches = self._resolve_one_round()
  File "/home/simudyne/.local/lib/python3.6/site-packages/pipenv/patched/piptools/resolver.py", line 198, in _resolve_one_round
    best_matches = {self.get_best_match(ireq) for ireq in constraints}
  File "/home/simudyne/.local/lib/python3.6/site-packages/pipenv/patched/piptools/resolver.py", line 198, in <setcomp>
[...]

PR #3111, opened 2018-10-27
Create a PEEP for dedicated credential fields

Vayel:

### The issue

#3085 

### The fix

Perhaps the fields `username` and `password` should only contain the **name** of an environment variable and not its value. Otherwise, the user can include the credentials in the Pipfile, which is a bad practice:
```
[[source]]
url = "https://mypypi.example.com/simple"
verify_ssl = true
name = "pypi"
username = "my#username"
password = "my/pass"
```

### The checklist

* [x] Associated issue


oz123:
Rotten. Closings.

PR #3108, opened 2018-10-26
Introduced --local option in pipenv install to create VENV within project

GitShaffi:
### The issue

Fixes issue activate PIPENV_VENV_IN_PROJECT with --local #2086
    https://github.com/pypa/pipenv/issues/2086

### The fix

Have introduced --local option to pipenv install which will create the .venv dir within the project

Let me know if there are any feedbacks.

### The checklist

* [x] Associated issue
* [ ] A news fragment in the `news/` directory to describe this fix with the extension `.bugfix`, `.feature`, `.behavior`, `.doc`. `.vendor`. or `.trivial` (this will appear in the release changelog). Use semantic line breaks and name the file after the issue number or the PR #.

PR #3115, opened 2018-10-28
PEEP for local option in pipenv install

GitShaffi:
### The issue

Create VENV directory within project without having to setup dotenv #2086 

### The fix

This is a PEEP to introduce `--local` option to pipenv install which will create the .venv dir within the project

Relevant PR: #3108 

### The checklist

* [x] Associated issue


uranusjr:
Two minor suggestions, partially based on personal preference:

* Only `--local`, not `-l`. I don’t believe Pipenv current has a lot of short options, and `l` is particularly confusing.
* Add a friendly reminder after initialisation to remind the user to add `.venv` to the version control’s ignore file, e.g. `.gitignore`.

GitShaffi:
> Some minor suggestions, partially based on personal preference:
> 
>     * Only `--local`, not `-l`. I don’t believe Pipenv current has a lot of short options, and `l` is particularly confusing.
> 
>     * Add a friendly reminder after initialisation to remind the user to add `.venv` to the version control’s ignore file, e.g. `.gitignore`.
> 
>     * What would happen if the user runs `--local` when there is already a (non-local) virtual environment? Should there be a warning/error/reminder?

@uranusjr 

- I found that few other existing options had short versions, So I had `-l` option. This can even be removed.
- Adding reminder for `.gitignore` isn't specific to `--local`. User can also do the same with dotenv vars. So, adding this warning can be done here if everyone emphasis, else it can go as a separate enhancement.
- Running `--local` on non-local env is a valid concern. I would suggest having a error, if we want to have a hardstop, else we can just throw a warning and ignore the `--local` option.

Everyone, please let me opinions on this.

AlJohri:
will future `pipenv install` or `pipenv shell` automatically use the local `.venv` directory after the first usage?

snow8gaia:
`--local` and ` -l`

brice-gros:
>     * Running `--local` on non-local env is a valid concern. I would suggest having a error, if we want to have a hardstop, else we can just throw a warning and ignore the `--local` option.
I would have said it should have the same behavior than calling:
```bash
pipenv install
# then
PIPENV_VENV_IN_PROJECT=true pipenv install
```
though, this creates a non-local env, and then it creates a local env, but keeps the non-local... so, it actually ignores the non-local env

So, the error seems a better option to me, and the envvar should behave that way IMHO.

PR #3144, opened 2018-11-02
Fix locking of pip.exe on windows during upgrade

### LINKED ISSUE #3142: How do I tell pipenv which version of pip to use?
Pip 18.1 is quite buggy, constantly throwing up `TypeError: 'module' object is not callable` and breaking installs on Windows machines, so I'd like to tell pipenv to only ever use 18.0 until pip gets a fix in some next version. However, while there is a `pipenv --python ...` command, there does not appear to be a `pipenv --pip 18.0` that I can set to ensure pipenv doesn't use the wrong pip version.

I can't find any mention about this in the docs, so: how do I make pipenv use an explicit version of pip when invoking it? Or even better, how do I configure pipenv so that it picks my preferred version of pip by default, unless I tell it not to using a flag?

techalchemy:
- Fixes #3142

### DIFF
--- a/news/3142.bugfix.rst
+++ b/news/3142.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed a bug which caused ``pipenv install`` to sometimes cause an ``EnvironmentError`` while attempting to upgrade ``pip`` inside a managed virtual environment.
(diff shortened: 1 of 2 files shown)


techalchemy:
Hm I dont think this will work

Issue #3154, opened 2018-11-03
Cannot create virtualenv with Python 2 with master


uranusjr:
```
$ pipenv --version
pipenv, version 2018.10.14.dev0

$ pipenv --python=2.7
Creating a virtualenv for this project…
Pipfile: C:\Users\uranusjr\Documents\playground\pipenv-venv-try\Pipfile
Using C:/Users/uranusjr/AppData/Local/Programs/Python/Python27/python.exe (2.7.14) to create virtualenv…
Complete Running... Running virtualenv with interpreter C:/Users/uranusjr/AppData/Local/Programs/Python/Python27/python.exe
New python executable in C:\Users\uranusjr\Documents\playground\pipenv-venv-try\.venv\Scripts\python.exe
Installing setuptools, pip, wheel...done.
Running virtualenv with interpreter C:/Users/uranusjr/AppData/Local/Programs/Python/Python27/python.exe
Traceback (most recent call last):
  File "C:\Users\uranusjr\.local\bin\pipenv-script.py", line 11, in <module>
    load_entry_point('pipenv', 'console_scripts', 'pipenv')()
  File "c:\users\uranusjr\documents\programming\pipenv\pipenv\vendor\click\core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "c:\users\uranusjr\documents\programming\pipenv\pipenv\vendor\click\core.py", line 717, in main
    rv = self.invoke(ctx)
  File "c:\users\uranusjr\documents\programming\pipenv\pipenv\vendor\click\core.py", line 1114, in invoke
    return Command.invoke(self, ctx)
  File "c:\users\uranusjr\documents\programming\pipenv\pipenv\vendor\click\core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "c:\users\uranusjr\documents\programming\pipenv\pipenv\vendor\click\core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "c:\users\uranusjr\documents\programming\pipenv\pipenv\vendor\click\decorators.py", line 64, in new_func
    return ctx.invoke(f, obj, *args, **kwargs)
  File "c:\users\uranusjr\documents\programming\pipenv\pipenv\vendor\click\core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "c:\users\uranusjr\documents\programming\pipenv\pipenv\vendor\click\decorators.py", line 17, in new_func
[...]

PR #3155, opened 2018-11-03
POC venv support

uranusjr:
This is not ready to be merged.

A few problems:

1. There is a “failed” process spinner run if you say run this on Python 2 (because vens is tried first). Ideally we should suppress that.
2. The spinner message is a bit funky.
3. There should be a way to force virtualenv usage (instead of venv), even if the Python is capable.
4. #3154.

Note: venv is used only for Python 3.6 or later. Earlier versions lack some features I want to include. It is theoratically possible to implement partial support in earlier versions (especially 3.5; it is probably not worth it to support 3.4), but I am not particularly interested in it.


techalchemy:
> There is a “failed” process spinner run if you say run this on Python 2 (because vens is tried first). Ideally we should suppress that.

We have a lot of control over the spinner, it's implemented in `vistir.spin` so feel free to tweak it (it's a bit overcomplicated at the moment for a bunch of reasons which enable us to keep the spinner spinning while streaming standard output / provides a dummy spinner with the same API)

> The spinner message is a bit funky.

That's an encoding issue, i have a character map and an output encoder in `._compat` that we can use

frostming:
Since the latest virtualenv has supported a "venv-like" behavior, so i would like to close this PR now, feel free to reopen if necessary

uranusjr:
Agreed, the current virtualenv is a better choice for pipenv.

PR #3166, opened 2018-11-05
 Fix PIP_SHIMs import, pythonfinder, vistir

techalchemy:
Fix PIP_SHIMs import, pythonfinder, vistir

- Add new exceptions
- Improve validation
- Cleanup spinner text
- Fix pyenv finder when `PYENV_ROOT` is not set
- Fix import errors for users installing and using `PIP_SHIMS` in pipenv


techalchemy:
ref: sarugaku/pip-shims#23

Issue #3170, opened 2018-11-06
Pipenv uninstall --all returns error when there is no installed packages in virtuanlenv


jxltom:
### Issue description

In latest master branch, ```pipenv uninstall --all``` will returns error if the virtualenv is already fresh.

```
pipenv uninstall --all
Un-installing all packages from virtualenv...
Found 0 installed package(s), purging...
[pipenv.exceptions.UninstallError]:   File "/home/user/Documents/dev/pipenv/pipenv/vendor/click/core.py", line 555, in invoke
[pipenv.exceptions.UninstallError]:       return callback(*args, **kwargs)
[pipenv.exceptions.UninstallError]:   File "/home/user/Documents/dev/pipenv/pipenv/vendor/click/decorators.py", line 17, in new_func
[pipenv.exceptions.UninstallError]:       return f(get_current_context(), *args, **kwargs)
[pipenv.exceptions.UninstallError]:   File "/home/user/Documents/dev/pipenv/pipenv/cli/command.py", line 294, in uninstall
[pipenv.exceptions.UninstallError]:       ctx=ctx
[pipenv.exceptions.UninstallError]:   File "/home/user/Documents/dev/pipenv/pipenv/core.py", line 2120, in do_uninstall
[pipenv.exceptions.UninstallError]:       do_purge(allow_global=system)
[pipenv.exceptions.UninstallError]:   File "/home/user/Documents/dev/pipenv/pipenv/core.py", line 1170, in do_purge
[pipenv.exceptions.UninstallError]:       raise exceptions.UninstallError(installed, command, c.out + c.err, c.return_code)
[pipenv.exceptions.UninstallError]: Attempted to run command: $ "/home/user/.local/share/virtualenvs/saleor-boilerplate-JGSSEPfX/bin/pip" uninstall  -y
[pipenv.exceptions.UninstallError]: You must give at least one requirement to uninstall (see "pip help uninstall")
ERROR: Failed to uninstall package(s) set([u'wheel', u'pip', u'setuptools'])...
```

It is better to just skip purging and show success information for this case, such as following result.

```
pipenv uninstall --all
Un-installing all packages from virtualenv...
Found 0 installed package(s), purging...
Environment now purged and fresh!
```

PR #3171, opened 2018-11-06
Do not show error for pipenv uninstall --all in a fresh virtuanlenv

jxltom:
### The issue

Closes https://github.com/pypa/pipenv/issues/3170

### The fix

Just return success if there is no packages which needs to be removed.

The final results looks like followings.

```
pipenv uninstall --all
Un-installing all packages from virtualenv...
Found 0 installed package(s), purging...
Environment now purged and fresh!
```

### The checklist

* [x] Associated issue


pipenv/core.py
```diff
@@ -1158,6 +1158,13 @@ def do_purge(bare=False, downloads=False, allow_global=False):
         click.echo(
             fix_utf8("Found {0} installed package(s), purging…".format(len(to_remove)))
         )
+    
+    # Skip purging if there is no packages which needs to be removed
+    if not to_remove:
+        if not bare:
+            click.echo(crayons.green("Environment now purged and fresh!"))
```
techalchemy: I'd prefer to be a bit more explicit about this and say exactly what happened, i.e. that there was nothing to remove.

Then we should return success

jxltom: Yes, that will be better. How about now?

Now the result looks like followings.

```
Un-installing all packages from virtualenv...
Found 0 installed package, skip purging
Environment now purged and fresh!
```

techalchemy: Perfect

PR #3172, opened 2018-11-06
Update skip lock option help info since it is used in several operations

jxltom:
### The issue

Currently ```--skip-lock``` option is used for several commands, such as ```install```, ```uninstall```. The ```Ignore locking mechanisms when **installing**``` is not appropriately anymore.

### The checklist

* [ ] Associated issue

### DIFF
--- a/pipenv/cli/options.py
+++ b/pipenv/cli/options.py
@@ -126,7 +126,7 @@ def skip_lock_option(f):
         state.installstate.skip_lock = value
         return value
     return option("--skip-lock", is_flag=True, default=False, expose_value=False,
-                    help=u"Ignore locking mechanisms when installing—use the Pipfile, instead.",
+                    help=u"Ignore locking mechanisms during operation—use the Pipfile, instead.",
                     callback=callback, type=click.types.BOOL)(f)
 
 
--- a/pipenv/cli/options.py
+++ b/pipenv/cli/options.py
@@ -126,7 +126,7 @@ def skip_lock_option(f):
         state.installstate.skip_lock = value
         return value
     return option("--skip-lock", is_flag=True, default=False, expose_value=False,
-                    help=u"Ignore locking mechanisms during operation—use the Pipfile, instead.",
+                    help=u"Skip locking mechanisms and use the Pipfile instead during operation.",
                     callback=callback, type=click.types.BOOL)(f)
 
 


pipenv/cli/options.py
```diff
@@ -126,7 +126,7 @@ def callback(ctx, param, value):
         state.installstate.skip_lock = value
         return value
     return option("--skip-lock", is_flag=True, default=False, expose_value=False,
-                    help=u"Ignore locking mechanisms when installing—use the Pipfile, instead.",
+                    help=u"Ignore locking mechanisms during operation—use the Pipfile, instead.",
```
techalchemy: looking at this now it seems like "ignoring" might be the wrong way to word this, since we aren't exactly ignoring it at all.  It's not like we are pretending the operation occurred, but in reality it didn't occur -- we are intentionally skipping it. 

Maybe we can say something like "Skip dependency resolution and locking and rely on Pipfile and package installation order instead." Is that too detailed? Idk, maybe this is nitpicky

jxltom: Exactly. I've updated it by emphasizing skip instead of ignore and also keep the info text short. What do you think about it now?

scott-vsi:
Does this mean that the `pipenv install` with the `--skip-lock` flag will install things in the same greedy fashion that `pip` would? If so, I still think the help is confusing. 

Maybe something like, "Skip dependency resolution provided by locking and fall back to a pip-style greedy installation of the requirements specified in the Pipfile"

techalchemy:
This means nothing because nothing is different than it was, and yes, if you skip locking pipenv just passes your dependencies to pip, that isn't greedy, it just doens't use a dependency resolver.

scott-vsi:
As a user who was unfamiliar with the lingo of pipenv, it wasn't clear to me that flag would cause pipenv to fall back to pip-style installation, but I get it now. Thanks

> that isn't greedy, it just doens't use a dependency resolver

Perhaps greedy was the wrong term. I was referring to this, from the [Pip User Guide](https://pip.pypa.io/en/stable/user_guide/#requirements-files), "pip doesn’t have true dependency resolution, but instead simply uses the first specification it finds for a project", which I thought of as a greedy (sub-optimal) algorithm, as opposed to an algorithm which can find the globally optimal solution (a. la dependency resolution).

Issue #3174, opened 2018-11-06
A few issues related with respect --skip-lock option for uninstall


jxltom:
### Issue description

In https://github.com/pypa/pipenv/commit/714ff657ca6d0888b932cdffc6fc768e7c9cb6ca commit, I got confused in several places.

1.  For now, even ```pipenv lock``` have ```--skip-lock``` option? This option does not exist for any released pipenv versions. It is introduced in https://github.com/pypa/pipenv/blob/3ce8d7c6bb8007f93f011bd0a7c32b3e977d379f/pipenv/cli/options.py#L356.

2. According to the doc, ```pipenv uninstall --all``` will purge all files from the virtual environment, but **leave the Pipfile untouched**. But for now, from this code https://github.com/pypa/pipenv/blob/master/pipenv/core.py#L2115-L2133, the pipfile and pipfile.lock will be updated anyway.

3. A ```@option("--skip-lock/--lock", is_flag=True, default=False, help="Lock afterwards.")``` is introduced in https://github.com/pypa/pipenv/blob/3ce8d7c6bb8007f93f011bd0a7c32b3e977d379f/pipenv/cli/command.py#L256 which confused me. A ```skip-lock``` option is already added by ```@uninstall_options```. Besides, it looks like this option is not passed into ```do_uninstall``` which means it does not work at all.

Let me know if I'm wrong. :smile: :handshake: 

-------------------------------------------------------------------------------

Please run `$ pipenv --support`, and paste the results here. Don't put backticks (`` ` ``) around it! The output already contains Markdown formatting.

If you're on macOS, run the following:

    $ pipenv --support | pbcopy

If you're on Windows, run the following:

    > pipenv --support | clip

If you're on Linux, run the following:

    $ pipenv --support | xclip

techalchemy:
> For now, even pipenv lock have --skip-lock option

This was added and not removed as an oversight while I was experimenting with exporting pipfile entries to `requirements.txt` format (currently done via `pipenv lock`, so avoiding the lock operation during export would be quite odd) -- definitely want this removed

> According to the doc, pipenv uninstall --all will purge all files from the virtual environment, but leave the Pipfile untouched.

When was that added to the documentation? That makes no sense, `uninstall` removes things from the pipfile in every other case...

> A skip-lock option is already added by @uninstall_options. 

Right, but it doesn't work, and I wanted an explicit flag to toggle

> Besides, it looks like this option is not passed into do_uninstall which means it does not work at all.

[...]

PR #3175, opened 2018-11-06
Continue respect skip lock in uninstall

jxltom:
### The issue

Try to close https://github.com/pypa/pipenv/issues/3174

### The issue

This part of code is a little bit tricky. This PR only solves the most obvious issues.

### The checklist

* [x] Associated issue


jxltom:
I'm going to close it. I think It is better to let you @techalchemy to handle it. How about that?

Issue #2885, opened 2018-09-24
Pipenv uninstall Fails to Update Pipfile When Package Name Exists in Comment


msopko81:
##### Issue description

`pipenv uninstall` fails to remove package from Pipfile if a comment with the package name exists before the package.

##### Expected result

The package to be remove from the Pipfile

##### Actual result

The comment is moved up a line, but the package remains in both Pipfile and Pipfile.lock

##### Steps to replicate

1. `pipenv install django-phonenumber-field phonenumberslite`

2. Edit and have Pipfile like this:
```TOML
[[source]]
url = "https://pypi.org/simple"
verify_ssl = true
name = "pypi"

[packages]
django-phonenumber-field = "*"
# django-phonenumber-field requires phonenumberslite, but does not properly record it
phonenumberslite = "*"

[dev-packages]

[requires]
python_version = "3.6"
```

3. `pipenv uninstall phonenumberslite`

Output:
```
Un-installing phonenumberslite...
Uninstalling phonenumberslite-8.9.14:
  Successfully uninstalled phonenumberslite-8.9.14

Removing phonenumberslite from Pipfile...
Locking [dev-packages] dependencies...
Locking [packages] dependencies...
Updated Pipfile.lock (ebdcbb)!
```
New Pipfile:
```TOML
[[source]]
url = "https://pypi.org/simple"
verify_ssl = true
name = "pypi"

[packages]
django-phonenumber-field = "*"# django-phonenumber-field requires phonenumberslite, but does not properly record it
phonenumberslite = "*"

[dev-packages]

[requires]
python_version = "3.6"
```

The package is removed if checked within pip:
```
pipenv run python -mpip list
Package                  Version
------------------------ -------
Babel                    2.6.0
Django                   2.1.1
django-phonenumber-field 2.0.1
phonenumbers             8.9.14
pip                      18.0
pytz                     2018.5
setuptools               40.4.3
wheel                    0.31.1
```

Since the package is still listed within the Pipfile, it will get installed for the next pipenv operation.

-------------------------------------------------------------------------------

<details><summary>$ pipenv --support</summary>

Pipenv version: `'2018.7.1'`

Pipenv location: `'C:\\Users\\Michael\\AppData\\Roaming\\Python\\Python36\\site-packages\\pipenv'`

Python location: `'c:\\program files\\python36\\python.exe'`

Other Python installations in `PATH`:

  - `3.6`: `C:\Program Files\Python36\python.exe`

[...]

PR #3177, opened 2018-11-06
Implement tomlkit

techalchemy:
Switch to tomlkit for parsing and writing

- Update tomlkit to preserve inline comments when
  deleting elemeents


pipenv/project.py
```diff
@@ -574,32 +578,55 @@ def clear_pipfile_cache(self):
         _pipfile_cache.clear()
 
     def _parse_pipfile(self, contents):
+        toml_encoder = toml.TomlPreserveInlineDictEncoder()
+        toml_encoder.get_empty_table()
```
frostming: I can't see what the second line is used for.
These lines should be moved to the except block where they are actually used.

pipenv/project.py
```diff
@@ -574,32 +578,55 @@ def clear_pipfile_cache(self):
...
+                        table.update(_data)
+                        data[section][package] = table
             # We lose comments here, but it's for the best.)
             try:
-                return contoml.loads(toml.dumps(data, encoder=toml_encoder))
+                return data
 
             except RuntimeError:
-                return toml.loads(toml.dumps(data, encoder=toml_encoder))
+                return toml.loads(tomlkit.dumps(data, encoder=toml_encoder))
```
frostming: Only `return data` here, it is safe to remove the `try except` block.

techalchemy: awesome, thanks for the review => I also patched tomlkit which I am a lot more unsure about

pipenv/project.py
techalchemy: ah yeah i just left these here by accident

pipenv/vendor/tomlkit/items.py
```diff
@@ -789,6 +789,18 @@ def append(self, key, _item):  # type: (Union[Key, str], Any) -> Table
         return self
 
     def remove(self, key):  # type: (Union[Key, str]) -> Table
```
frostming: I don't know if I am doing anything wrong but the `parsed_pipfile['packages']` is a `tomlkit.container.Container` object and this method is not called when `del parsed_pipfile['packages']['mypackage']`.

pipenv/vendor/tomlkit/container.py
```diff
@@ -189,6 +189,10 @@ def remove(self, key):  # type: (Union[Key, str]) -> Container
         if idx is None:
             raise NonExistentKey(key)
 
+        old_data = self._body[idx][1]
+        trivia = getattr(old_data, "trivia", None)
+        if trivia and idx > 0 and getattr(trivia, "comment", None):
+            self._body[idx - 1][1].comment(trivia.comment)
         self._body[idx] = (None, Null())
```
frostming: It is better to leave the comment at its own line as a line comment:
```python
self._body[idx] = (None, Comment(Trivia(comment_ws="", comment=trivia.comment)))
```

techalchemy: I would do that but we have a regression test that makes sure we move it one line up

Comments in toml spec need to belong to another element as far as I understood, right?

Did you write the test for this? https://github.com/pypa/pipenv/blob/master/tests/integration/test_uninstall.py#L154

frostming: Yes, but the test only checks whether the comment is preserved. In fact it is leaved as a line comment(That is also what old `contoml` does). TOML spec allows [line-level comments](https://github.com/toml-lang/toml#user-content-example), too.

techalchemy: ah perfect, that makes so much more sense. Doesn't this level of context-aware parsing make it much slower to parse though?

pipenv/project.py
```diff
@@ -941,7 +967,10 @@ def remove_package_from_pipfile(self, package_name, dev=False):
         name = self.get_package_name_in_pipfile(package_name, dev)
         key = "dev-packages" if dev else "packages"
         p = self.parsed_pipfile
-        lines = [l for l in p[key].serialized().splitlines()]
+        try:
+            lines = [l for l in p.item(key).as_string().splitlines()]
+        except AttributeError:
+            lines = [l for l in p[key].serialized().splitlines()]
```
frostming: What will happen if toml file contains line comments? I can't find the original bug 83af37a attempts to solve.
As discussed above, TOML do allow line comments.

techalchemy: > # The normal toml parser can't handle deleting packages with preceding newlines

When I wrote this i actually meant it can't handle deleting packages from pipfiles where the preceding line is only made of a comment

I updated the comment on the PR in question to include the rest of the commits which weren't referenced explicitly, but the issue I was looking at was #2885

techalchemy: In any event, this is all the reason why we use 4 different toml parsers (some handle inline comments correctly, some don't; some can't delete lines with preceding comments, some can)

Now I am seeing that contoml/prettytoml can't even handle writing files correctly and I'm pretty much done using them, and tomlkit is the only other one in python that I'm aware of that will retain the original document structure including outline/inline tables and comments. We are already using it in https://github.com/sarugaku/passa, https://github.com/sarugaku/requirementslib, https://github.com/sarugaku/plette, and pretty much everywhere else we have reimplemented pipfile parsing.  So far this is the only remaining issue I have with it, so I decided to go ahead and fix everything else.

pipenv/vendor/tomlkit/container.py
```diff
@@ -189,7 +189,12 @@ def remove(self, key):  # type: (Union[Key, str]) -> Container
         if idx is None:
             raise NonExistentKey(key)
 
-        self._body[idx] = (None, Null())
+        old_data = self._body[idx][1]
+        trivia = getattr(old_data, "trivia", None)
+        if trivia and getattr(trivia, "comment", None):
+            self._body[idx] = (None, trivia)
```
frostming: The trivia contains string, just comment only please

techalchemy: ah ok

pipenv/project.py
frostming: Ah, I didn't notice that bug in `contoml/prettytoml`. `tomlkit` seems to work quite well on this, making these lines a bit protective.

techalchemy: Yeah I deleted this all locally but currently I’m not preserving the comment anymore

Issue #3181, opened 2018-11-06
Index per-line in requirements files are not working


jxltom:
### Issue description

Currently ```pipenv``` with latest requirementslib will output ```pipenv lock -r``` as followings. But actually index in line such as ```flask==1.0.2 -i https://pypi.org/simple``` is not working for ```pip install -r requirements.txt```. Only the top line ```-i https://pypi.org/simple``` works.

So probably it is redundant to add sources per-line?

```
pipenv lock -r
-i https://pypi.org/simple
click==7.0 -i https://pypi.org/simple
flask==1.0.2 -i https://pypi.org/simple
itsdangerous==1.1.0 -i https://pypi.org/simple
jinja2==2.10 -i https://pypi.org/simple
markupsafe==1.1.0 -i https://pypi.org/simple
werkzeug==0.14.1 -i https://pypi.org/simple
```

-------------------------------------------------------------------------------

Please run `$ pipenv --support`, and paste the results here. Don't put backticks (`` ` ``) around it! The output already contains Markdown formatting.

If you're on macOS, run the following:

    $ pipenv --support | pbcopy

If you're on Windows, run the following:

    > pipenv --support | clip

If you're on Linux, run the following:

    $ pipenv --support | xclip

PR #3182, opened 2018-11-06
Do not add index source per-line in requirements files

jxltom:
### The issue

Closes https://github.com/pypa/pipenv/issues/3181

### The fix

Since index source per-line is not working, let's just disable it.

### The checklist

* [x] Associated issue

### DIFF
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -726,7 +726,7 @@ def do_install_dependencies(
         index_args = prepare_pip_source_args(project.sources)
         index_args = " ".join(index_args).replace(" -", "\n-")
         deps = [
-            req.as_line(sources=project.sources, include_hashes=False) for req in deps_list
+            req.as_line(sources=False, include_hashes=False) for req in deps_list
         ]
         # Output only default dependencies
         click.echo(index_args)


jxltom:
I think it is a pip issue. It looks like pip does not support this kind of usage.

For example, I have a ```requirements.txt``` with following content. I install packages by ```pip install -r requirements.txt```, and pip will still install packages from the default index anyway. **The in-line ```-i someindex``` does not work for ```pip install -r```.**

```
click==7.0 -i https://somewrongurl
```

```
Collecting click==7.0 (from -r req (line 1))
  Using cached https://files.pythonhosted.org/packages/fa/37/45185cb5abbc30d7257104c434fe0b07e5a195a6847506c074527aa599ec/Click-7.0-py2.py3-none-any.whl
Installing collected packages: click
Successfully installed click-7.0
```

jxltom:
Only following format works for ```pip install -r```.

```
-i https://somewrongurl
click==7.0
```

And here is the result.

```
Looking in indexes: https://somewrongurl
Collecting click==7.0 (from -r req (line 2))
  Retrying (Retry(total=4, connect=None, read=None, redirect=None, status=None)) after connection broken by 'NewConnectionError('<pip._vendor.urllib3.connection.VerifiedHTTPSConnection object at 0x7f9a5da4dfd0>: Failed to establish a new connection: [Errno -5] No address associated with hostname',)': /click/
  Retrying (Retry(total=3, connect=None, read=None, redirect=None, status=None)) after connection broken by 'NewConnectionError('<pip._vendor.urllib3.connection.VerifiedHTTPSConnection object at 0x7f9a5da4d2d0>: Failed to establish a new connection: [Errno -5] No address associated with hostname',)': /click/
  Retrying (Retry(total=2, connect=None, read=None, redirect=None, status=None)) after connection broken by 'NewConnectionError('<pip._vendor.urllib3.connection.VerifiedHTTPSConnection object at 0x7f9a5da4dcd0>: Failed to establish a new connection: [Errno -5] No address associated with hostname',)': /click/
```

techalchemy:
Ah ok, this actually make sense.  We don't really install from these outputs but instead we pass the file to the resolver (`parse_file` or something like that generates constraints for us).  That's probably why I wrote it to drop the URL at the top in the first place.  So we basically need to use `itertools.groupby` to group files by index first and then resolve...

jxltom:
Ok. I got it . These output are also used to pass to the resolver and then installed by pipenv.

But for ```pipenv lock -r``` command, since pip install does not support in-line index. Let's just take output of ```pipenv lock -r``` as a special case and remove them in the output?

techalchemy:
hm lets just not inline the index arguments for now and handle it with environment variables, I think we set them anyway

PR #3188, opened 2018-11-07
Only output dev packages for pipenv lock -r -d

jxltom:
### The issue

The ```pipenv lock -r``` should only output all default packages. The ```pipenv lock -r -d``` should only output all dev packages. But currently master branch will ouput both default and dev packages for ```pipenv lock -r -d```. This behaviour is different with any released versions.

For example, following Pipfile currently would output all packages by ```pipenv lock -r -d```.

```
[[source]]
verify_ssl = true
name = "pypi"

[dev-packages]
requests = "*"

[packages]
flask = "*"

[requires]
python_version = "3.6"
```

```
D:\test>pipenv lock -r -d
-i https://pypi.org/simple
certifi==2018.10.15
chardet==3.0.4
click==7.0
flask==1.0.2
idna==2.7
itsdangerous==1.1.0
jinja2==2.10
markupsafe==1.1.0
requests==2.20.0
urllib3==1.24.1
werkzeug==0.14.1
```

Actually it should only output dev packages as followings.

```
(D:\test>pipenv lock -r -d
-i https://pypi.org/simple
certifi==2018.10.15
chardet==3.0.4
idna==2.7
requests==2.20.0
urllib3==1.24.1
```

### The fix

I believe ```only``` in ```get_requirements``` means only include dev packages without default packages. But the ```only``` in ```do_install_dependencies``` means only install deps without locking. So ```only``` para in ```get_requirements``` should not inhericated from ```do_install_dependencies``` but pin to ```False``` to only output dev packages or default packages.

### The checklist

* [ ] Associated issue

### DIFF
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -721,7 +721,7 @@ def do_install_dependencies(
     # Allow pip to resolve dependencies when in skip-lock mode.
     no_deps = not skip_lock
     failed_deps_list = []
-    deps_list = list(lockfile.get_requirements(dev=dev, only=only))
+    deps_list = list(lockfile.get_requirements(dev=dev, only=True))
     if requirements:
         index_args = prepare_pip_source_args(project.sources)
         index_args = " ".join(index_args).replace(" -", "\n-")


techalchemy:
How are you keeping up with all of this? I wrote a lot of it and I'm having trouble :|

Granted, I didn't write the lock output section, I find it a bit confusing. You're spot on. Thanks for catching this (again) (why don't we have a test for this)

techalchemy:
I really need to cut a release, but it's probably worth creating a triage permission level at the very least for anyone who wants to help go through and tag out everything we merged but didn't write tests for yet / needs protection against regressions

jxltom:
I use pipenv for my daily development for my own/company projects, so it is easy for me to find any breaking changes in APIs and definitely happy to make a PR for that, only if I can.

Yes, I can see pipenv is still in active development and regressions often happens. Probably the test coverage is still not good enough.

Maybe it worth to consider release beta/pre-releases? So some users can try that and have feedbacks, then developers could fix the regressions, and finally cut the release。

PR #3191, opened 2018-11-08
Switch to tomlkit for parsing and writing

frostming:

### The issue

Incomplete implementation of #3177 

### The fix

1. Remove `dump_dict` method.
2. Reorder pipfile only when outline table is present.
3. Preserve inline table for `toml`.
4. Drop `prettytoml/contoml` from vendors.

### The checklist

* [x] Associated issue


frostming:
Maybe we can drop prettytoml/contoml from the vendors directory at this point?

techalchemy:
That would be very nice, those libraries are so annoying

techalchemy:
Nice failure...

```
2018-11-08T02:40:09.9468075Z E           assert 'requests = { version = "*" }' in 'requests = {version = "*"}\nclick = "*"\n'
```

There are spaces between the `{}` and the contents in LHS, but not in the RHS...

frostming:
@techalchemy :( toml keeps them but tomlkit suppresses them.

frostming:
All CI passed! 🎆

techalchemy:
I cannot believe this is passing. :tada:

techalchemy:
I want to make sure we aren't using this in `Pipfile` or some hidden dependency somewhere... did you check this already?

tests/unit/test_vendor.py
```diff
@@ -71,44 +38,3 @@ def test_inject_environment_variables(self):
...
-])
-def test_token_date(dt, content):
-    token = create_primitive_token(dt)
-    assert token == tokens.Token(tokens.TYPE_DATE, content)
-
-
-def test_dump_nonascii_string():
-    content = 'name = "Stažené"\n'
-    toml_content = contoml.dumps(contoml.loads(content))
-    assert toml_content == content
```
uranusjr: I think these tests in `test_vendor.py` should be kept, but rewrote to use tomlkit.

frostming: Sure, thanks for the review.

pipenv/project.py
```diff
@@ -856,7 +856,7 @@ def write_toml(self, data, path=None):
                 formatted_data = tomlkit.dumps(data).rstrip()
```
uranusjr: What doe this `_body` check mean?

frostming: Same as Line. 578, duck type implies tomlkit parsed result.

uranusjr: Maybe it would be a good idea to extract this into a `is_tomlkit_result` function, and add a docstring to it. This folk knowledge can be easily lost.

uranusjr: Also—is there a reason why tomlkit is only used to dump tomlkit-parsed results? I think it can dump normal dictionaries as well.

frostming: Makes sense, will improve this soon.

techalchemy: note that if you're dumping regular dictionaries you probably will need a way to recursively dump dictionarie ;)

frostming: @techalchemy Yes, `tomlkit` already converts dict recursively, see the newest patch.

frostming:
I found a weird bug handling comments. Please don't merge until I fix this. Thanks

techalchemy:
Marking as WIP for now

techalchemy:
what is the weird handling?

frostming:
Seems a bug of tomlkit, trying to figure out how to update vendors and patch scripts.

frostming:
@techalchemy @uranusjr Can you guys have another look on the latest changes and we are good to go.

tests/integration/test_project.py
```diff
@@ -143,3 +143,21 @@ def test_many_indexes(PipenvInstance, pypi):
...
+@pytest.mark.install
+@pytest.mark.project
+def test_rewrite_outline_table(PipenvInstance, pypi):
+    with PipenvInstance(pypi=pypi, chdir=True) as p:
+        with open(p.pipfile_path, 'w') as f:
+            contents = """
+[packages.requests]
+version = "*"
+            """.strip()
+            f.write(contents)
```
uranusjr: Not entirely related to this PR, but it just occurred to me that we need a test to make sure we are reading Pipfile and Pipfile.lock with the correct encoding. It is likely best to have an abstraction around this in `Project` too (something like `with project.open_pipfile('w') as f:`).

tasks/vendoring/patches/vendor/tomlkit-dump-inline-table.patch
```diff
@@ -0,0 +1,34 @@
...
+@@ -1000,6 +1004,9 @@ class InlineTable(Item, dict):
+         if key is not None:
+             super(InlineTable, self).__setitem__(key, value)
+ 
++        if hasattr(value, "trivia") and value.trivia.comment:
++            value.trivia.comment = ""
++
+         m = re.match("(?s)^[^ ]*([ ]+).*$", self._trivia.indent)
+         if not m:
+             return
```
uranusjr: Is there a bug in TOMLkit, or just a workaround specific to Pipenv? We might want to submit a PR to TOMLkit if it’s the former.

pipenv/project.py
```diff
@@ -578,59 +573,37 @@ def clear_pipfile_cache(self):
...
+            empty_inline_table = toml.TomlDecoder().get_empty_inline_table
+        for section in ("packages", "dev-packages"):
+            table_data = parsed.get(section, {}).copy()
+            for package, value in table_data.items():
+                if hasattr(value, "keys"):
+                    table = empty_inline_table()
+                    table.update(value)
+                    table_data[package] = table
+            parsed[section] = table_data
+        return parsed
```
uranusjr: I would prefer this (and `_is_tomlkit_parsed_result`) to be global functions, instead of under `Project` They are not really related to the project construct, but can be used in any general TOML structures.

pipenv/vendor/tomlkit/container.py
```diff
@@ -196,8 +196,7 @@ def remove(self, key):  # type: (Union[Key, str]) -> Container
             self._body[idx] = (None, Comment(Trivia(comment_ws="", comment=trivia.comment)))
         else:
             self._body[idx] = (None, Null())
-            super(Container, self).__delitem__(key.key)
-
+        super(Container, self).__delitem__(key.key)
```
uranusjr: This diff is probably accidental?

frostming: No, the key should be deleted unconditionally.

tasks/vendoring/patches/vendor/tomlkit-dump-inline-table.patch
frostming: You are right, I'll submit a simple fix but I think the owner might have a better one.

frostming:
Thanks for the quick response from tomlkit and I just updated tomlkit to 0.5.2 and cleaned some patching stuff.

tests/integration/test_project.py
techalchemy: We had this in `Project` -- did it go away?

techalchemy:
Vendoring updates are somewhat complicated, I often take my local copy of pipenv, clone the entire thing to `/tmp`, re-vendor the relevant package, commit that version, then re-generate patches:

```console
export PIPENV_DIR=$(cwd)
cd /tmp
git clone "$PIPENV_DIR" pipenv
cd pipenv/pipenv/vendor
pip install -t . --no-deps --upgrade <pkgname>==<pkgversion>
rm -rf *.dist-info
git checkout -- <pkgname>.LICENSE <pkgdir>/LICENSE # depends on which got deleted
git add <pkgname>...
git commit -m "Temp commit for updating patches"
cp -R "${PIPENV_DIR}/pipenv/vendor/<pkgname>/files-you-want" pipenv/vendor/<pkgname>/
git diff -p pipenv/vendor/<pkgname> > "${PIPENV_DIR}/tasks/vendoring/patches/vendor/pkgname-whatever.patch"
```

I've literally never documented that before so we should probably record that...

frostming:
@techalchemy Yeah, it really took me much time to figure it out. I managed to do it by invoke run.

pipenv/vendor/tomlkit/items.py
```diff
@@ -6,7 +6,10 @@
 from datetime import date
 from datetime import datetime
 from datetime import time
-from enum import Enum
+try:
+    from enum import Enum
+except ImportError:
+    from pipenv.vendor.backports.enum import Enum
```
uranusjr: This uses try-catch, but there’s a `lru_cache` block that uses if-else. I would recommend unifying the style (probably to try-catch; it seems to be recommended).

frostming: `lru_cache` is got from the original code, I will change to `if` to adapt the style.

techalchemy: in _most_ cases, like the `Enum` case, it doesn't matter at all

techalchemy:
if and when @uranusjr is ok with this I can verify the patches and we can merge

techalchemy:
I want to verify that this and #3196 will play nicely together before merging either of them so that’s the next step

PR #3192, opened 2018-11-08
Fix typo in creating virtualenv

jxltom:
### The checklist

* [ ] Associated issue


techalchemy:
Hm yeah I can see that causing some problems... that might be the hidden cause behind some issues with installing pyenv

PR #3227, opened 2018-11-15
fix 'Release 2018.11.14' markup

kashewnuts:
### The issue

Fix 'Release 2018.11.14' markup.

### The fix

From:

To:


frostming:
Thanks for the patch. The whitespace should be mistakenly deleted by some editor auto formatting. I am not sure.

Issue #3236, opened 2018-11-16
2018.11.14 regression: Installing Multiple local editable packages leaves some unusable


christek91:
### Issue description

We use pipenv to install five packages from local directories as 'editable' packages. In the most recent version of pipenv we see that when we do a `pipenv install` of those packages they report success and appear to succeed in the logs. However, we are unable to import them due to `ModuleNotFoundError: No module named XXX`. 

### Expected result

We are used to being able to install all five local editable packages, and then being able to import them in code that uses the virtualenv.

### Actual result

Pipenv installs and reports exit code 0. However, importing modules gives a `ModuleNotFoundError`.

The logs report no errors, and verbose logs show the same actions being performed in the older and newer versions of pipenv. 

### Steps to replicate

Create a Pipfile that has 3 or more entries of the form:
```
"my_module" = {path = "./path/to/module", editable = true}
```

Then run `pipenv install` and try to import all those modules. You will likely get the import error.

-------------------------------------------------------------------------------

Rather than paste the `pipenv --support output`, I've actually already got a fix and a test for it ready to go. 

I spent enough time going down the rabbit hole today that I figured I'd take a pass at fixing it. I'll update the issue w/ the PR in a second.

frostming:
Close in honor of #3237

PR #3237, opened 2018-11-16
Fix race condition when installing 2+ editable non-VCS pkgs at once

christek91:
This regression was recently introduced, and only affects non-vcs
packages. The effect of the race is that installations of multiple
editable non-VCS sourced packages at once may cause some of them to be
un-importable.

The fix is to make editable package installs Blocking just like VCS
installs are.

### The issue

This fixes a race condition that is hit when installing multiple local editable packages at once. There was a regression because `editable` installations used to be blocking, but a change in the recent release changed it so that only VCS installations are blocking. 

While most VCS installations are probably `editable, not all `editable` installations are VCS ones. So that change inadvertently brought back https://github.com/pypa/pipenv/issues/919. 
 
I've also created a new Issue for the new regression as the PR template suggests. It is: https://github.com/pypa/pipenv/issues/3236

### The fix

My fix makes sure that all `editable` installations are blocking, regardless of if they are VCS installs or not. 

The test I wrote for this is a little janky to setup since I couldn't use git to install the test packages since VCS installations don't hit the race. I put a comment in the code explaining the context.

### The checklist

* [x] Associated issue


tests/integration/test_install_twists.py
```diff
@@ -8,6 +8,7 @@
 import pytest
 
 from flaky import flaky
+import delegator
```
frostming: I guess this line can be removed?

tests/integration/test_install_twists.py
```diff
@@ -333,3 +334,55 @@ def test_install_local_uri_special_character(PipenvInstance, testsroot):
         c = p.pipenv("install")
         assert c.return_code == 0
         assert "six" in p.lockfile["default"]
+
+
+@pytest.mark.files
+@pytest.mark.needs_internet
```
frostming: Does it really need internet? You are using the mocked pypi test server in this case.

tests/integration/test_install_twists.py
```diff
@@ -333,3 +334,55 @@ def test_install_local_uri_special_character(PipenvInstance, testsroot):
...
+    once, and which causes some of them to not be importable.
+
+    This issue had been fixed for VCS packages already, but not local 'editable' packages.
+
+    So this test locally installs packages from tarballs that have already been committed in
+    the local `pypi` dir to avoid using VCS packages.
+    """
+    pkgs = {
+        "requests-2.19.1": "requests/requests-2.19.1.tar.gz",
+        "flask-0.12.2": "flask/flask-0.12.2.tar.gz",
```
frostming:
```suggestion
        "flask-0.12.2": "flask/Flask-0.12.2.tar.gz",
```

tests/integration/test_install_twists.py
```diff
@@ -333,3 +334,55 @@ def test_install_local_uri_special_character(PipenvInstance, testsroot):
...
+        "six-1.11.0": "six/six-1.11.0.tar.gz",
+        "jinja2-2.10": "jinja2/jinja2-2.10.tar.gz",
```
frostming:
```suggestion
        "jinja2-2.10": "jinja2/Jinja2-2.10.tar.gz",
```

tests/integration/test_install_twists.py
christek91: Ah, good point. That was left over from an earlier iteration. I'll fix that.

tests/integration/test_install_twists.py
christek91: No, you're right it does not in the final iteration. Will fix that.

tests/integration/test_install_twists.py
christek91: Thanks, updated in new commit!

tests/integration/test_install_twists.py
christek91: Thanks, updated in new commit!

tests/integration/test_install_twists.py
```diff
@@ -333,3 +333,55 @@ def test_install_local_uri_special_character(PipenvInstance, testsroot):
...
+
+        import tarfile
+
+        with tarfile.open(source_path, "r:gz") as tgz:
+            tgz.extractall(path=tmpdir.strpath)
+
+        pipfile_string += "'{0}' = {{path = '{1}', editable = true}}\n".format(pkg_name, unzip_path)
+
+    with PipenvInstance(pypi=pypi, chdir=True) as p:
+        print(pipfile_string)
```
frostming: Please remove the debugging prints.

techalchemy:
once the debug line is gone im good to merge this, i didn't think about adding editable as a conditional there but i'm glad you were able to find that.  Hopefully the code cleanup efforts will make contributing easier

tests/integration/test_install_twists.py
christek91: 👍 will do!

Issue #2778, opened 2018-08-25
Allow per-project pipenv configuration in Pipfile, ".pipenv.cfg", setup.cfg, or some other per-project configuration file


sirosen:
There are quite a few `pipenv` configuration options presented as environment variables. I would like to be able to set these on a per-project basis, committed to version control, loaded and understood by `pipenv` itself and not by a secondary tool.

The [documentation section for them](https://pipenv.readthedocs.io/en/latest/advanced/#configuration-with-environment-variables) recommends using direnv. While I'm sure direnv is a lovely solution, it's another dependency for developer machines and for any infrastructure where we want to use `pipenv`.
I don't think it's unreasonable to want `pipenv` to be able to have its own config file or read from an existing config like `setup.cfg` or even a special section of `Pipfile`.

I'd suggest even using a section of `pyproject.toml`, but that seems downright confusing to mix with `Pipfile`.
One easy way out is to just load a special `.pipenv.cfg` with configparser.

Allowing the use of `--config FILENAME` to use a custom config file location would be a logical added goody.

More than anything else, I want to be able to set `PIPENV_VENV_IN_PROJECT=1` in my projects, especially since `/.venv/` is in my `.gitignore`.

Apologies if this has been requested before, but I don't see it in any of the closed issues.
I'm happy to work on this in a PR if it gets the ol' 👍 from people.

techalchemy:
This issue will need to go through the new enhancement proposal process -- nothing too fancy, just an organized way for triaging stuff with Kenneth 

See [here](https://github.com/pypa/pipenv/blob/master/peeps/PEEP-000.md) for the initial description and then you can simply just create a PR with a markdown description for review

Thanks!

MIKNOTAURO:
Hi guys, I do not know if this is related but, is there a way to set the path of my environment per project basis?
I know if you set PIPENV_VENV_IN_PROJECT=1 you got a .venv in the root directory but this does not work for me... I need to set my env outside of my root directory (one level up, like ../.venev) and it would be awesome if we can set the name of the env.

Any help?

andrewpeterprifer:
[...]

PR #3271, opened 2018-11-21
PEEP: Allow per-project pipenv configuration

cinderblock:
# The issue

https://github.com/pypa/pipenv/issues/2778#issuecomment-417966352 says that a PEEP needs to be created to further the discussion. So, I'm creating one. Please forgive me if this is the wrong methodology. I could not find a clear example of exactly what is expected for PEEPs.

### DIFF
--- a/peeps/PEEP-003.md
+++ b/peeps/PEEP-003.md
@@ -1,0 +1,59 @@
+# PEEP-003: Specify pipenv options from configuration file
+
+This PEEP describes an addition that would allow configuring `pipenv` options via configuration file.
+
+☤
+
+The discussion in [#2778](https://github.com/pypa/pipenv/issues/2778#issuecomment-417966352) suggested that a new PEEP needed to be created to address the feature discussed in the github issue.
+This is an attempt at a PEEP for that feature.
+
+The feature that is specifically being requested is the ability to specify any `pipenv` invocation options in a project (or directory) specific configuration file.
+
+Of particular intrest is an equivalent of the `PIPENV_VENV_IN_PROJECT` option.
+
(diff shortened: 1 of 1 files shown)


jxltom:
Here is a template for PEEP, https://github.com/pypa/pipenv/blob/master/peeps/PEEP-TEMPLATE.md. And this https://github.com/pypa/pipenv/blob/master/peeps/PEEP-002.md is a good example.

IMO, it would be better to re-state your proposal here such as WHY, HOW and etc. Then we can move our discussion under this PEEP without rechecking the original issues.

peeps/PEEP-003.md
```diff
@@ -0,0 +1,59 @@
+# PEEP-003: Specify pipenv options from configuration file
+
+This PEEP describes an addition that would allow configuring `pipenv` options via configuration file.
+
+☤
+
+The discussion in [#2778](https://github.com/pypa/pipenv/issues/2778#issuecomment-417966352) suggested that a new PEEP needed to be created to address the feature discussed in the github issue.
```
techalchemy: As @jxltom mentioned, A PEEP needs to be self-contained, so in this document you need to assume there is no preexisting issue and that nobody cares about reported problems.  This is a document to describe a change you want to see in pipenv, so you need to describe the change in full, what it would do, why you are proposing it, what alternatives you've considered, and you can mention the fact there have been issues reported that led you to propose this. I have a proposal I was working on that I can upload which may come close, it might help you consolidate your ideas here.

Essentially we want as few unnecessary words as possible and we want to answer as many questions as possible ahead of time

techalchemy:
For reference: https://github.com/pypa/pipenv/pull/3304/files#diff-21fbe6b8fbe0011d15f60d4cdb1394e7R1

PR #3317, opened 2018-11-28
Update docs dependencies

frostming:
Update the docs dependencies to match the Pipfile.lock

Fix an issue that the CLI docs can't be displayed.


jxltom:
Looks good to me. Let's see if it solves our issue.

PR #3321, opened 2018-11-28
Eliminate superfulous parse-cmdify pairs

uranusjr:
`delegator.run()` accepts a list, so there’s no need to convert it back to string. I *believe* this should work fine.

### DIFF
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -508,7 +508,7 @@ def resolve(cmd, sp):
     from .vendor.vistir.compat import to_native_string
     EOF.__module__ = "pexpect.exceptions"
     from ._compat import decode_output
-    c = delegator.run(Script.parse(cmd).cmdify(), block=False, env=os.environ.copy())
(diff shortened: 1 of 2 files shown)


pipenv/utils.py
```diff
@@ -508,7 +508,7 @@ def resolve(cmd, sp):
     from .vendor.vistir.compat import to_native_string
     EOF.__module__ = "pexpect.exceptions"
     from ._compat import decode_output
-    c = delegator.run(Script.parse(cmd).cmdify(), block=False, env=os.environ.copy())
+    c = delegator.run(cmd, block=False, env=os.environ.copy())
```
techalchemy: yeah I didn't just randomly cmdify any of these... although to some extent you have to rely on me remembering to have covered the regressions I was keeping protected against... I usually catch other stuff while i'm doing that and you know I never make small changes so that is going to be hard to untangle

The question is do you think that `cmdify` is likely to _cause_ issues?

PR #3322, opened 2018-11-28
Handle alternate names for UTF-8 encoding

### LINKED ISSUE #3313: Incompatibility with pypy3
### Issue description

Recent versions of pipenv no longer work with pypy3.

### Expected result

Expected pipenv to work.

### Actual result

```
❯ PYENV_VERSION=pypy3.5-6.0.0 pyenv exec pipenv --verbose
  File "/Users/ob/.pyenv/versions/pypy3.5-6.0.0/bin/pipenv", line 7, in <module>
    from pipenv import cli
  File "/Users/ob/.pyenv/versions/pypy3.5-6.0.0/site-packages/pipenv/__init__.py", line 47, in <module>
    from .cli import cli
  File "/Users/ob/.pyenv/versions/pypy3.5-6.0.0/site-packages/pipenv/cli/__init__.py", line 3, in <module>
    from .command import cli
  File "/Users/ob/.pyenv/versions/pypy3.5-6.0.0/site-packages/pipenv/cli/command.py", line 19, in <module>
    from .options import (
  File "/Users/ob/.pyenv/versions/pypy3.5-6.0.0/site-packages/pipenv/cli/options.py", line 11, in <module>
    from .. import environments
  File "/Users/ob/.pyenv/versions/pypy3.5-6.0.0/site-packages/pipenv/environments.py", line 7, in <module>
    from ._compat import fix_utf8
  File "/Users/ob/.pyenv/versions/pypy3.5-6.0.0/site-packages/pipenv/_compat.py", line 98, in <module>
    OUT_ENCODING, ERR_ENCODING = force_encoding()
  File "/Users/ob/.pyenv/versions/pypy3.5-6.0.0/site-packages/pipenv/_compat.py", line 68, in force_encoding
    from ctypes import pythonapi, py_object, c_char_p
ImportError: cannot import name 'pythonapi'
```

### Steps to replicate

Simply install pipenv with PyPy3, then try to run any pipenv command.

`pipenv --support` won't work either.

PyPy doesn't support `ctypes.pythonapi`, but the root cause of the issue is that it shouldn't even try to import this module in the first place. I've investigated a bit and it looks with PyPy3, `sys.stdout.encoding` returns `"utf8"` instead of `"utf-8"`. I'm unsure whether this is a PyPy bug or not, but since pipenv used to work fine with PyPy3 prior to (I think) #3096, can we relax the condition on [this line](https://github.com/pypa/pipenv/blob/b03983e61ef284157c725dfedef02fef5829341f/pipenv/_compat.py#L64) to allow for `"utf8"` as well as `"utf-8"`?

ob-stripe:
### The issue

Currently, pipenv does this:
1. In `__init__.py`, `sys.stdout` and `sys.stderr` are replaced by `io.TextIOWrapper`s with `encoding` set to `'utf8'` (no dash): https://github.com/pypa/pipenv/blob/d9ebad335dc4cfff9cb223f7cc3141066044b65b/pipenv/__init__.py#L31-L36
2. In `_compat.py`, the encoding is forced to UTF-8 by checking if `sys.stdout/err.encoding.lower()` is equal to `'utf-8'` (with a dash): https://github.com/pypa/pipenv/blob/d9ebad335dc4cfff9cb223f7cc3141066044b65b/pipenv/_compat.py#L331

Fixes #3313.

### The fix

- Canonicalizes the encoding name using `codecs.lookup()`. This will always return `'utf-8'` for UTF-8 encodings, no matter what alias is used (`'utf8'`, `'UTF-8'`, `'u8'`, etc.)
- Catch `ImportError` when importing `ctypes`, so that this code path doesn't crash with PyPy (which doesn't support ctypes). @frostming mentioned in #3313 that pipenv doesn't officially support PyPy so I can remove this if you'd prefer, but it seems like a net improvement to me.

### The checklist

* [x] Associated issue: #3313
* [x] A news fragment in the `news/` directory: `3313.bugfix.rst`

### DIFF
--- a/news/3313.bugfix.rst
+++ b/news/3313.bugfix.rst
@@ -1,0 +1,1 @@
+Handle alternate names for UTF-8 encoding.
--- a/pipenv/_compat.py
+++ b/pipenv/_compat.py
@@ -26,8 +26,8 @@ warnings.filterwarnings("ignore", category=ResourceWarning)
 
 __all__ = [
     "NamedTemporaryFile", "Path", "ResourceWarning", "TemporaryDirectory",
-    "get_terminal_size", "getpreferredencoding", "DEFAULT_ENCODING", "force_encoding",
-    "UNICODE_TO_ASCII_TRANSLATION_MAP", "decode_output", "fix_utf8"
+    "get_terminal_size", "getpreferredencoding", "DEFAULT_ENCODING", "canonical_encoding_name",
+    "force_encoding", "UNICODE_TO_ASCII_TRANSLATION_MAP", "decode_output", "fix_utf8"
 ]
 
 
@@ -46,6 +46,16 @@ def getpreferredencoding():
 DEFAULT_ENCODING = getpreferredencoding()
 
 
+def canonical_encoding_name(name):
+    import codecs
+    try:
+        codec = codecs.lookup(name)
+    except LookupError:
+        return name
+    else:
+        return codec.name
+
+
 # From https://github.com/CarlFK/veyepar/blob/5c5de47/dj/scripts/fixunicode.py
 # MIT LIcensed, thanks Carl!
 def force_encoding():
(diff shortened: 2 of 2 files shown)


techalchemy:
Thanks! if you run into anything else definitely let us know, I may experiment with adding PyPy to CI if we can

ob-stripe:
Cool, thanks!

ob-stripe:
@techalchemy Sorry to bother you, but do you have any idea when the next pipenv version will be released? The issue I fixed in this PR has been affecting our CI pipeline for a few months now.

techalchemy:
Apologies, I have been forced to split focus on a number of projects. Stewardship here seems to have been excellent so I’m just catching up and getting my own changes together so hopefully soon!

PR #3334, opened 2018-11-30
Point build badges at the current pipelines

DavidStaheli:
### The issue

The Azure Pipelines build badges are pointing at the old pipelines.

### The fix

This PR points the build badges to the new/current pipelines.

Thanks for considering this PR!


jxltom:
Looks good to me. Thanks for taking care of this!

BTW, the new pipeline has no records for the old pipelines, right? If so, we should still better to keep the old pipelines in dashboard though they are not being used anymore.

techalchemy:
Lol... badges.... good thinking. Thanks (feel free to merge PRs from @DavidStaheli I’m pretty sure he is a program manager for azure and is helping keep us out of the Stone Age 🤣

jxltom:
Yeah, I'm just thinking this page https://dev.azure.com/pypa/pipenv/_build?definitionId=12 still shows the paused pipelines and is there anyway to hide it from (I saw a rename/move options) the normal viewers but only to admins, which makes our CI dashboard cleaner :)

techalchemy:
@jxltom if you have an azure account I can just add you to the project, just remind me your email on slack or whatever
 
I’m usually bad at clicking buttons so if you want to work out how to hide these, feel free, but the results from their builds will be gone within 30 days anyway

jxltom:
@techalchemy Thanks, going to have a try :smile:

PR #3337, opened 2018-12-01
Resort imports for pipenv, tasks, tests and docs

jxltom:
### The issue

Currently in codebase some methods are imported but not used and imports are not in good order. Since we are using ```isort``` in https://github.com/pypa/pipenv/blob/d63ed0537e4dc0b0967e7f9e1116cafd98485e62/Pipfile#L21 and in https://github.com/pypa/pipenv/blob/master/setup.cfg#L22_L31, it will be nice to resort the imports to make our code cleaner and easier to read.

### The fix

This PR resorts the imports for ```pipenv```, ```tasks```, ```tests``` and ```docs``` via ```isort```. The ```vendor``` and ```patched``` are excluded.

### The checklist

* [ ] Associated issue


techalchemy:
Thank god

PR #3340, opened 2018-12-01
Fix bug where custom virtualenv can not be activated by pipenv shell

### LINKED ISSUE #3339: pipenv prefers activated virtualenvs when looking for projects
### Issue description

```pipenv shell``` can not activate custom virtualenv created by ```virtualenv```

### Actual result

```
➜  abcd  . /Users/jxltom/.local/share/virtualenvs/abcd-Gj3qkKPD/bin/activate
.: no such file or directory: /Users/liujunxiao/.local/share/virtualenvs/abcd-Gj3qkKPD/bin/activate
```

### Steps to replicate

- ```virtualenv somedir```
- ```cd somedir && source bin/activate```
- ```pipenv shell```

Please run `$ pipenv --support`, and paste the results here. Don't put backticks (`` ` ``) around it! The output already contains Markdown formatting.

If you're on macOS, run the following:

    $ pipenv --support | pbcopy

If you're on Windows, run the following:

    > pipenv --support | clip

If you're on Linux, run the following:

    $ pipenv --support | xclip

### PR DESCRIPTION
**Pull request by jxltom**
### The issue

Closes #3339. This is because in ```PIPENV_ACTIVE``` is checked in project's ```virtualenv_location``` which is introduced in https://github.com/pypa/pipenv/pull/3243/files#diff-7c0c15ee4eba8c237e3318a395816a13R430. So setting up ```PIPENV_ACTIVE``` in ```do_shell``` has to be after finishing reading ```virtualenv_location``` which is a dynamic property.

### The checklist

* [x] Associated issue

### CODE REVIEW

**Comment by techalchemy** [comment]
CI is set to batch run so it sometimes takes awhile to start while it waits to see if you will push more changes

**Comment by techalchemy** [comment]
I think we can make a test for this, I’ll see before we merge

**Comment by jxltom** [comment]
That's what I'm thinking. It is better to have tests for this. But it looks like ```pipenv shell``` is not supported by our tests which I tried but failed as some kind of tty console stuff. I didn't find any tests on ```pipenv shell``` in tests actually. @techalchemy Any suggestions for implementing this test? Thanks

**Comment by techalchemy** [comment]
Hm I didn’t get a chance to try this yet but I’m guessing we need a fake tty wrapper to have it play nice. Possibly this is controlled via `SESSION_IS_INTERACTIVE`. Really we just need to open a subprocess that has `sys.stdout` and `sys.stderr` set to `TextIOWrapper` subclasses (so we can use some context manager to swap these in and back out) as long as the subclass provides `isatty` and `fileno`. Click has some useful implementations that might work, `from click._compat import get_text_stdout_stream, get_text_stderr_stream`

**Comment by jxltom** [comment]
@techalchemy Thanks for sharing! I'm going to have look. Free free to contribute to it if you already have some solutions.

**Comment by frostming** [comment]
I have a different approach about this:

Detect whether we are in a virtualenv with `is_in_virtualenv` and then do not fork subshell or activate if yes, showing a message instead.

**Comment by jxltom** [comment]
I've updated the PR fixing another bug in ```pipen run``` related with this. This PR defnitely requires more tests.

**Comment by jxltom** [comment]
@frostming What do you mean by showing a message? If that means show a warning or error message instead of re-activating it, I think this limit the usage of pipenv.

For example, probably I'm already in a activated virtualenv by ```source bin/activate```, ```pipenv shell``` should still work since pipenv can load env vars from ```.env``` while ```virtualenv``` can not. This is very useful in some cases.

Another case is that even virtualenv is activated, ```pipenv run``` should work anyway.

**Comment by techalchemy** [comment]
To be honest I’m confused. If you’re in a virtualenv you made and activated yourself, pipenv should operate automatically as though you passes `--system` all the time except it will give a courtesy notice that you’re in a virtualenv that pipenv doesn’t own

The only time we should be trying to look for pipenv-owned virtualenvs is if `PIPENV_IGNORE_VIRTUALENVS` is set.

**Comment by jxltom** [comment]
Yes, with this PR, pipenv wil use activated virtualenv automatically when ```pipenv run|shell```. Without this PR, pipenv will create pipenv-owned virtualenv even it is already in a activated virtualenv.

This is for reproduce for ```pipenv run```:

- ```virtualenv abc```
- ```cd abc && source bin/activate```
- ```pipenv run which pip```

This is for reproduce for ```pipenv shell```:

- ```virtualenv abc```
- ```cd abc && source bin/activate```
- ```pipenv shell```

**Comment by jxltom** [comment]
Thanks! @frostming , I'm going to have a look.

**Comment by frostming** [comment]
Maybe you can write a contextmanager to hide the openning/ending bits.

**Comment by frostming** [comment]
Ugh.. it doesn't work for posix..

**Comment by jxltom** [comment]
@frostming We may need emulating a tty for testing this

**Comment by techalchemy** [comment]
At one time I had a full test for this exact situation

**Comment by techalchemy** [comment]
This isn't broken for me locally...

**Comment by techalchemy** [comment]
the correct behavior is to use the activated virtualenv unless `PIPENV_IGNORE_VIRTUALENVS` is set.  If you go to a directory with no `Pipfile`, pipenv currently creates a pipfile first and then tells you info from the activated virtualenv rather than create a new one.

Seems like it works?

**Comment by jxltom** [comment]
Sounds like it is working for you @techalchemy 

Did you try this? It should output ```yourpath/abc/bin/pip``` instead of ```/home/yourusername/.local/share/virtualenvs/abc-HoOUeXpF/bin/pip```

- ```virtualenv abc```
- ```cd abc && source bin/activate```
- ```pipenv run which pip```

**Comment by jxltom** [comment]
Update, for reproducing this, you have to create a pipenv-owned virtualenv first...

- ```virtualenv abc && cd abc```
- ```pipenv --three```
- ```source bin/activate```
- ```pipenv run which pip```

Finally it should output  ```yourpath/abc/bin/pip``` instead of ```/home/yourusername/.local/share/virtualenvs/abc-HoOUeXpF/bin/pip```

**Comment by jxltom** [comment]
Yeah, it still a bug but only happens when ```shell/run``` is excecuted within activated virtualenv (which is not owned by pipenv) while pipenv-owned virtualenv already exists.

**Comment by techalchemy** [comment]
and the bug is that they don't use pipenv's environment? or  what is the bug

**Comment by jxltom** [comment]
The bug is that ```pipenv run|shell``` is not using environment created and activated by ```virtualenv```, but using pipenv-owned environment (only happens when it exist).

Check this https://github.com/pypa/pipenv/pull/3340#issuecomment-444000535 for reproducing. And reproduce it under directory under your ```$HOME```, ```tmp``` directory can not reproduce this, I don't know why.

**Comment by jxltom** [comment]
Update: I've added test for this PR. As for the tests for ```pipenv shell```, I think we can do that in another separate PR.

**Comment by frostming** [comment]
@jxltom For tests, you should create a pipenv project outside of the virtualenv, which may need to rewrite the `virtualenv` fixture to a context manager so that it won't be activated at the beginning of test function.

**Comment by jxltom** [comment]
@frostming Yeah, but this bug can also be reproduced within an activated virtualenv at the beginning only if ```PIPENV_IGNORE_VIRTUALENVS``` is set.

I can confirm this test can not pass without this PR's patch, so it is working anyway...

**Comment by techalchemy** [comment]
Thanks for handling this bizarre edge case, definitely a confusing one but a lot more clear with a failing test

**Comment by techalchemy** [comment]
Ok so reflecting on this (because it is causing a strange test failure for me and I now have to mess with it a bit), this is a substantial behavior change and I don't think it's desirable

I don't know why it was hard to nail down an actual example of what is changing in this pr, but I can provide one:

```console
$ virtualenv /tmp/tmpenv
$ source /tmp/tempenv/bin/activate.fish
$ pipenv run pip install click
... Courtesy notice...
```

Under the old approach, this would install into the virtualenv by default unless `PIPENV_IGNORE_VIRTUALENVS` is set.  This is the correct behavior, because pipenv doesn't care what your project paths would suggest as search paths.  The test merged here is falsely passing because you are only matching against the word `virtualenv`, which sometimes appears in the path to the virtualenv directory and sometimes doesnt.  The actual test requires comparing against `Creating a virtualenv`, which is logged to stderr.

Under the new approach, `pipenv run pip install click` from an activated virtualenv installs click _into the virtualenv_ and ignores the current context.  The _only case_ when this should happen is if `PIPENV_IGNORE_VIRTUALENVS` is set.

PR #3341, opened 2018-12-01
Enable test disabled in travis

jxltom:
### The issue

Since we are moving from travis to azure, let's re-enable the test disabled in travis.

### The checklist

* [ ] Associated issue

### DIFF
--- a/tests/integration/test_install_basic.py
+++ b/tests/integration/test_install_basic.py
@@ -12,7 +12,6 @@ from flaky import flaky
 
 @pytest.mark.install
 @pytest.mark.setup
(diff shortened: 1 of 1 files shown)


frostming:
LGTM, thanks for taking care of this

techalchemy:
Btw you can’t really trust skip reasons per se. if it says it doesn’t work on Travis it really just means it was flaky in Linux test envs. If it says it’s non deterministic, it might work fine now

jxltom:
That's true. Actually this case works even in Travis now.

PR #3112, opened 2018-10-27
add PIPENV_SKIP_LOCK envvar, and suitable doc

devxpy:
Fixes https://github.com/pypa/pipenv/issues/2200.

[🐍🏕️](http://www.pycampers.com/)

PR #3344, opened 2018-12-02
Remove duplicate skip-lock option

jxltom:
### The issue

This duplicate ```--skip-lock``` is introduced in https://github.com/pypa/pipenv/commit/714ff657ca6d0888b932cdffc6fc768e7c9cb6ca to make sure ```--skip-lock``` is respected. Then it is removed in https://github.com/pypa/pipenv/pull/3112/files for fixing a bug.

With this PR, duplicate skip-lock option is removed. Locking behavior is by default for ```pipenv uninstall```. Locking can be skipped by providing ```--skip-lock``` option.

### The checklist

* [ ] Associated issue

### DIFF
--- a/pipenv/cli/command.py
+++ b/pipenv/cli/command.py
@@ -261,7 +261,6 @@ def install(
     short_help="Un-installs a provided package and removes it from Pipfile.",
     context_settings=subcommand_context
 )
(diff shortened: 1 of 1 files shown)


jxltom:
Yes, exactly. It is passed by ```uninstall_options```. :D

PR #3345, opened 2018-12-02
Remove --dev option for uninstall

jxltom:
### The issue

I think ```pipenv uninstall --dev``` is not working currently. We should either support it or remove it from the CLI.

Discussions are welcomed :)

### The fix

A options group ```lock_base_options``` is introduced, which can also be used in ```lock_options``` instead of current ```install_base_options``` since ```--dev``` is actually also not supported in ```pipenv lock```. But it is another PEEP for that, which I'm going to propose later

### The checklist

* [ ] Associated issue

### DIFF
--- a/pipenv/cli/options.py
+++ b/pipenv/cli/options.py
@@ -342,16 +342,21 @@ def common_options(f):
     return f
 
 
-def install_base_options(f):
+def lock_base_options(f):
     f = common_options(f)
-    f = dev_option(f)
     f = pre_option(f)
     f = keep_outdated_option(f)
     return f
 
 
+def install_base_options(f):
+    f = lock_base_options(f)
+    f = dev_option(f)
+    return f
+
+
 def uninstall_options(f):
-    f = install_base_options(f)
+    f = lock_base_options(f)
     f = skip_lock_option(f)
     f = editable_option(f)
     f = package_arg(f)


jxltom:
I checked ```test_uninstall.py``` again, it doesn't have any tests related with ```uninstall --dev```

As for the saying that it's not working, if I understand correctly, ```uninstall``` method below doesn't take ```state.installstate.dev``` as an argument at all.

https://github.com/pypa/pipenv/blob/master/pipenv/cli/command.py#L287-L300

So I don't think ```pipenv uninstall --dev``` is working. User can pass ```--dev``` to ```uninstall```, but it won't take any effect.

techalchemy:
You’re right, but I can’t imagine any justification for this at all...

jxltom:
@techalchemy So that means ```pipenv unisntall --dev``` should be supported?

Just for clafication, as the below Pipfile, ```pipenv uninstall flask``` should uninstall flask and only remove it from ```dev-packages``` part, right?

```
[[source]]
name = "pypi"
verify_ssl = true

[packages]
flask = "*"

[dev-packages]
flask = "*"

[requires]
python_version = "2.7"
```

jxltom:
Closed and track this in https://github.com/pypa/pipenv/issues/3385

PR #3348, opened 2018-12-03
Enable output of script line for 'run' in verbose mode

refack:
### The issue

There is no simple way to see the what is the command line that's used for `pipenv run`

### The fix

Output the command line to the log as `info`, so it's output when run in verbose mode.

### The checklist

* [ ] Associated issue

Seeking early feedback.

### DIFF
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -2303,6 +2303,8 @@ def do_run(command, args, three=None, python=False, pypi_mirror=None):
     inline_activate_virtual_environment()
     try:
         script = project.build_script(command, args)
+        cmd_string = ' '.join([script.command] + script.args)
+        logging.getLogger("pip").info("Run: '{0}'".format(cmd_string))
     except ScriptEmptyError:
         click.echo("Can't run script {0!r}-it's empty?", err=True)
     if os.name == "nt":
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -2305,7 +2305,7 @@ def do_run(command, args, three=None, python=False, pypi_mirror=None):
         script = project.build_script(command, args)
         cmd_string = ' '.join([script.command] + script.args)
         if environments.is_verbose():
-            click.echo(crayons.normal("Run: {0}".format(cmd_string)), err=True)
+            click.echo(crayons.normal("$ {0}".format(cmd_string)), err=True)
     except ScriptEmptyError:
         click.echo("Can't run script {0!r}-it's empty?", err=True)
     if os.name == "nt":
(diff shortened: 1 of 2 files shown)


pipenv/core.py
```diff
@@ -2303,6 +2303,8 @@ def do_run(command, args, three=None, python=False, pypi_mirror=None):
     inline_activate_virtual_environment()
     try:
         script = project.build_script(command, args)
+        cmd_string = ' '.join([script.command] + script.args)
+        logging.getLogger("pip").info("Run: '{0}'".format(cmd_string))
```
frostming: This logger is not used in any other place of this file. To keep consistency, I recommend
```python
if environments.is_verbose():
    click.echo("Run: {!0r}".format(cmd_string), err=True)
```

refack: Sure.

jxltom:
```
➜  abc pipenv --verbose run pip list 
Run: pip list
Package    Version
---------- -------
pip        18.1   
setuptools 40.6.2 
wheel      0.32.3 
```

Here is the example output making easier to review

pipenv/core.py
techalchemy: (as a sidenote, logging using a logger implementation isn't a great practice, especially in our case -- we parse the output that goes to stdout in many cases)

pipenv/core.py
```diff
@@ -2303,6 +2303,9 @@ def do_run(command, args, three=None, python=False, pypi_mirror=None):
     inline_activate_virtual_environment()
     try:
         script = project.build_script(command, args)
+        cmd_string = ' '.join([script.command] + script.args)
+        if environments.is_verbose():
+            click.echo(crayons.normal("Run: {0}".format(cmd_string)), err=True)
```
techalchemy: I have one small request -- in order to be consistent across our output elsewhere, the ideal output format would be something to the effect of:
```suggestion
            click.echo(crayons.normal("$ {0}".format(cmd_string)), err=True)
```

refack:
Another output example
For this `Pipfile`:
```TOML
[scripts]
lint = "flake8 . --exclude=.venv --count --show-source --statistics --select=E901,E999,F821,F822,F823"
```
Console:
```cmd
DEV D:\code\4node\gyp>pipenv -v run lint
Run: flake8 . --exclude=.venv --count --show-source --statistics --select=E901,E999,F821,F822,F823
0

DEV D:\code\4node\gyp>
```

refack:
> can you please include a news fragment in the `news/`

Done, PTAL.

BTW: That's very crufty, could it be extracted from the commit log? Point me to the tools I'll give it a try :)

techalchemy:
It's a pretty common changelog approach, we have a lot of commits and extracting changelogs from commit logs is not really a reasonable approach

PR #3350, opened 2018-12-04
Mention Linuxbrew in installation docs

thedrow:
Installing using Linuxbrew also works just fine.
I figured it's worth mentioning.


jxltom:
Just try that, works for me!

Would you mind also mentioning this in this https://github.com/pypa/pipenv/blob/d191deb744b39eebc88c82c576295e86c061d53d/docs/install.rst#-homebrew-installation-of-pipenv part?

thedrow:
Done.

Issue #3351, opened 2018-12-04
--site-packages is not recognized without --two or --three option


nonylene:
### Issue description

`--site-packages` is not recognized without `--two` or `--three` option.

PR: #3352

### Reason

`site_packages` argument in `cli` function always set to false because `--site-packages` flag was defined in the global state (https://github.com/pypa/pipenv/blob/97657b2797b8a8f045776470d74add6048a95b67/pipenv/cli/options.py#L62). 

However,  with `--two` or `--three` option,  we can run into the block creates virtualenv (https://github.com/pypa/pipenv/blob/97657b2797b8a8f045776470d74add6048a95b67/pipenv/cli/command.py#L201). Since `state.site_packages` is used in this block, pipenv recognizes `--site-packages` flag well and creates virtualenv created with `--system-site-packages`.

### Expected result

`pipenv --site-packgages install` creates virtualenv with `--system-site-packages`

### Actual result

`pipenv --site-packgages install` doesn't create virtualenv that uses system packages.

~~However, `pipenv --two --site-packages install` or `pipenv --three --site-packages` works as expected.~~

*UPDATE*: However `pipenv --two --site-packages install` or `pipenv --three --site-packages` works as expected for the first installation, `pipenv` recreates virtualenv without `--system-site-packages` right after the first install. I think this is a different issue. (`pipenv --two install` also creates twice.)

Alternatively, `pipenv --two --site-packages` works well.

### Steps to replicate

Provide the steps to replicate (which usually at least includes the commands and the Pipfile).

- `$ pipenv --two --site-packages install` (*UPDATE*: or `pipenv --two --site-packages`)

This outputs `Making site-packages available…`.

- `$ pipenv --site-packages install`

This doesn't print `Making site-packages available…`.

PR #3352, opened 2018-12-04
Fix `--site-packages` flag is not recognized

nonylene:
### The issue

#3351

`site_packages` argument in `cli` function always set to false because `--site-packages` flag was defined in the global state (https://github.com/pypa/pipenv/blob/97657b2797b8a8f045776470d74add6048a95b67/pipenv/cli/options.py#L62). 

### The fix

Remove `site_packages` argument form `cli` function and make use `state.site_packages`.

### The checklist

* [x] Associated issue


jxltom:
Thanks for this. It is because ```expose_value=False``` is used for ```--site-packages``` option as in https://github.com/pypa/pipenv/blob/8707fe52571422ff5aa2905a2063fdf5ce14840b/pipenv/cli/options.py#L265-L267.

We should also check options with ```expose_value=False``` but used in https://github.com/pypa/pipenv/blob/8707fe52571422ff5aa2905a2063fdf5ce14840b/pipenv/cli/command.py#L62-L81.

BTW, it will be nice if you also include tests for this. Thanks!

techalchemy:
To be honest we just need a thing that tests every CLI flag and makes sure it makes it through to the corresponding function

jxltom:
Yes, it will be nice to test the behavior as well. :D

nonylene:
Test added in f81e861.

techalchemy:
awesome, thank you so much for handling this! :cake:

Issue #3353, opened 2018-12-05
pipenv --clear is not working


jxltom:
### Issue description

```pipenv --clear``` is not working.

### Actual result

```
➜  Untitled Folder pipenv --clear
Clearing caches...
```

clafoutis42:
Do we have any update on this issue ?

I confronted it today, had to clear caches manually.

jxltom:
https://github.com/pypa/pipenv/pull/3354 should sovles this, I will try to add tests to that PR this weekend. Sorry for the delay.

PR #3354, opened 2018-12-05
Fix pipenv --clear is not working

jxltom:
### The issue

Closes https://github.com/pypa/pipenv/issues/3353

### The fix

- ```pipenv --clear``` is supported by passing ```state.clear```
- The exposed several values from click is removed from ```cli```'s arugment since options such as ```clear```, ```python```, ```three``` may be used in subcommands. It is better to access these values via ```state```.
- Test is added

### The checklist

* [ ] Associated issue


jxltom:
Oh, found a bug in vistir, submitted a PR https://github.com/sarugaku/vistir/pull/47 in upstream

jxltom:
@techalchemy I think this PR is ready to merge since vistir is updated from upstream.

pipenv/cli/command.py
```diff
@@ -62,18 +62,15 @@ def cli(
     state,
     where=False,
     venv=False,
-    rm=False,
-    bare=False,
-    three=False,
-    python=False,
-    help=False,
     py=False,
     envs=False,
```
techalchemy: Given that you are removing `--three` and `--python` as explicit arguments here we should add cli smoke tests to make sure they are still parsed properly just in case

jxltom: Will do

jxltom:
Test for ```--three``` argument is added, test for ```--python``` has already existed in https://github.com/pypa/pipenv/blob/cae8d76c210b9777e90aab76e9c4b0e53bb19cde/tests/integration/test_cli.py#L33-L37.

techalchemy:
for all the stuff that you're removing i'm assuming you have verified it still works?

jxltom:
> for all the stuff that you're removing i'm assuming you have verified it still works?

yep :D

PR #3362, opened 2018-12-06
Preserve inline spaces in toml

frostming:
### The issue
In the previous codebase, https://github.com/pypa/pipenv/blob/master/pipenv/utils.py#L94 is always ignored and the table gets rewritten, losing any spaces between inline table keys. The reason behind this is we can't get the real data structure from the `value` here.

### The fix

Get the real data type from `container._body`

### The checklist

* [ ] Associated issue

### DIFF
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -99,11 +99,12 @@ def convert_toml_outline_tables(parsed):
                 table.update(value)
                 section[package] = table
 
+    is_tomlkit_parsed = isinstance(parsed, tomlkit.container.Container)
     for section in ("packages", "dev-packages"):
         table_data = parsed.get(section, {})
         if not table_data:
             continue
-        if isinstance(parsed, tomlkit.container.Container):
+        if is_tomlkit_parsed:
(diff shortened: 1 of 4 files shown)


pipenv/utils.py
```diff
@@ -83,19 +83,32 @@ def cleanup_toml(tml):
...
+            if not key:
+                continue
+            if hasattr(value, "keys") and not isinstance(value, tomlkit.items.InlineTable):
+                table = tomlkit.inline_table()
+                table.update(value.value)
+                section[key.key] = table
+
+    def convert_toml_table(section):
+        for package, value in section.items():
+            if hasattr(value, "keys") and not isinstance(value, toml.decoder.InlineTableDict):
```
techalchemy: Technically `hasattr` usage is a bad practice, it behaves differently across python versions and should be avoided. https://hynek.me/articles/hasattr/

frostming: Hmmm, reasonable, so `isinstance(value, dict)` will be better.

Issue #3365, opened 2018-12-07
`pipenv sync` doesn't remove packages


konstantinmiller:
If I remove a package from Pipfile.lock and perform `pipenv sync`, I expect that the removed packaged is removed from the virtual environment. This, however, doesn't happen. it's still present in `.venv/lib/Python3.7/site-packages`.

frostming:
This is an API change, you may need to submit a PEEP to the `peeps` dir via a normal PR.

konstantinmiller:
So, you mean `pipenv sync` is not supposed to `sync` the removed packages? :)
In this case, the workflow should always be `pipenv clean` plus `pipenv sync` in case you don't know if some packages have been removed or only added, right?

frostming:
Right

techalchemy:
Pipenv sync installs packages, it is not a destructive operation (if you want, it might be reasonable to propose a flag for this via a `PEEP`)

Closing for now, thanks for the report!

PR #3366, opened 2018-12-08
Filed PEEP 003 to allow `pipenv sync` to delete packages

konstantinmiller:
Associated issue: #3365 

`pipens sync` should be allowed to uninstall packages that have been removed from the `Pipfile.lock` file.

At the moment, if packages have been removed from `Pipfile.lock`, a subsequent `pipenv sync` wouldn't remove them from the environment. Consequently, an update of the environment from `Pipile.lock` is only possible via two commands: `pipenv clean` and `pipenv sync`.

It would make sense to add a flag to `pipenv sync` to allow this to happen in one operation. Also, the name `sync` intuitively suggests that the update works in both ways: adding and removing packages. So that the current behavior might be somewhat unintuitive. Adding the flag would make it more intuitive, at least with the flag.

### DIFF
--- a/peeps/PEEP-003.md
+++ b/peeps/PEEP-003.md
@@ -6,4 +6,10 @@
 
 At the moment, if packages have been removed from `Pipfile.lock`, a subsequent `pipenv sync` wouldn't remove them from the environment. Consequently, an update of the environment from `Pipile.lock` is only possible via two commands: `pipenv clean` and `pipenv sync`.
 
-It would make sense to add a flag to `pipenv sync` to allow this to happen in one operation. Also, the name `sync` intuitively suggests that the update works in both ways: adding and removing packages. So that the current behavior might be somewhat unintuitive. Adding the flag would make it more intuitive, at least with the flag.
+This is somewhat unintuitive, as the name `sync` intuitively suggests that the update works in both ways: adding and removing packages.
+
+It would make sense to make the default behavior of `pipenv sync` to produce an environment that exactly corresponds to the content of the `Pipfile.lock` file. That is, `pipenv sync` should be allowed to both install and uninstall packages.
+
+In addition, a flag such as `--install-only` should be added that would only install new packages but not uninstall those that have been removed from `Pipfile.lock`.
+
+
--- a/peeps/PEEP-003.md
+++ b/peeps/PEEP-003.md
@@ -1,0 +1,9 @@
+# PEEP-042: Allowing `pipenv sync` to remove packages
+
+`pipens sync` should be allowed to uninstall packages that have been removed from the `Pipfile.lock` file.
+
+☤
+
+At the moment, if packages have been removed from `Pipfile.lock`, a subsequent `pipenv sync` wouldn't remove them from the environment. Consequently, an update of the environment from `Pipile.lock` is only possible via two commands: `pipenv clean` and `pipenv sync`.
+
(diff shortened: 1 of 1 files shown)


jxltom:
I'm definitely +1 for this.

Would you mind adding the sync option to the PEEP specifically? Such as ```pipenv sync``` should be equal to ```pipenv clean + pipenv sync``` actually, and ```pipenv sync --keep-stale``` shouldn't clean the stale packages and only install or anything your are going to propose.

Probably you have a better option name for this. :D

konstantinmiller:
@jxltom, I've changed PEEP 003 to have `clean + sync` as the new default behavior of `sync`. Plus proposed the name of the flag to be `--install-only`.

jxltom:
@konstantinmiller Thanks, looks very good to me!

techalchemy:
There is roughly 0% chance we would ever change the default behavior of a command to be destructive without a deprecation cycle which we don’t actually have implemented yet, so first you’d have to propose a mechanism for deprecating that. 

Like I said in the issue, this really need to start as a flag or at least with a warning notice on the command itself before it would be implemented as the default behavior. Behavioral changes to existing commands may seem simple and intuitive, but they alter existing functionality and have a high potential for impacting people’s workflows.

konstantinmiller:
So, how about keeping the default behavior, and introducing the flag `--uninstall-removed` or `--uninstall-removed-packages` or something like that?

jxltom:
Sounds too long and hard to type? I don't know. ```pipenv sync --uninstall```?

frostming:
I prefer the idea of deprecation cycle. Or i will just stick to `pipenv sync && pipenv clean`.

jxltom:
Yeah, deprecation cycle is definitly needed, but we still need to figure out a new flag which is going to deprecate the old behevior.

Or we can only propose/implement a new flag, but keep the default behavior, then deprecation won't be needed.

konstantinmiller:
How can this be decided? If to add a flag for the new behavior, or if to introduce a deprecation cycle and change the default behavior? It seems like opinions diverge here.

jxltom:
I don't have a preference actually.

I'm just thinking, for the deprecation cycle, maybe we also need to provide a new flag for the new behavior? It is weird that only a deprecation warning is shown but no new suggested CLI is provided. If so, these two approaches don't have conflicts.

This is a immature thought but I'd like to share and discuss for the deprecation cycle:
- Stage 1: Show deprecation warning for ```pipenv sync``` that the behavior is going to change., and suggest users to use ```pipenv sync --newflag/--no-newflag``` (which is a boolean option in click) for the new behavior.
- Stage 2: Deprecate ```pipenv sync``` old behavivor and adopts the new behavior. Keep ```pipenv sync --newflag/--no-newflag``` for compatibility and it can also be used as a flag to enable/disable this new behavior manually. As the new flag, it should be a flag to determine whether cleaning staled packages before installing.

jxltom:
@kennethreitz is the final one to decide this, https://github.com/pypa/pipenv/blob/399c8795c77dbf99b48959008881e586be5f8adb/docs/dev/philosophy.rst#management-style, and we are just kindly to help. So this proposal is up to you to some extent.

ctheune:
Interesting discussion. I was surprised to see that sync (which implies a 2-way mechanism) isn't doing deletions by default. So moving forward I understand backwards compatibility and would suggest a similar move to the virtualenv site-packages history: add explicit options for both versions and then change the default in the future.

pipenv sync --clean 
pipenv sync --no-clean

To indicate that --clean does the same as the clean command.

The current structure with separate clean and sync commands seems a bit random to me and means that having a consistent workflow requires running multiple commands. This is both superfluous for humans (i want a reproducible build, which means stuff has to be deleted!) and for automation (I want things to be as atomic as possible.)

PR #3392, opened 2018-12-17
Replaced double hyphens with a single hyphen

RohanTalip:


frostming:
Thanks for taking care of it, I have no clue about the CI failure.

PR #3398, opened 2018-12-19
Propose the use of dotenv to manage Pipenv environments

frostming:
This PR includes a PEEP describing the feature requested.

### DIFF
--- a/peeps/PEEP-007.md
+++ b/peeps/PEEP-007.md
@@ -1,0 +1,24 @@
+# PEEP-007: Use dotenv to manage Pipenv environments
+
+This PEEP proposes to make use of dotenv file to manage Pipenv enviroments.
+
+☤
+
+At present Pipenv supports [multiple environment variables][1] to change its behavior.
+To use one or more of them, users have to specify the variables in their shell. As a workaround,
+people can use the awesome [direnv][2] tool to detect and activate what are present in `.envrc`
+automatically, but it does not work on Windows platform. We need built-in support for environment
+files in Pipenv. Pipenv integrates with support of `dotenv` and `.env` files, however,
+it is only used in `pipenv shell` and `pipenv run`.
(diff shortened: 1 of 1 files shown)


frostming:
Related/similar:
* #3080 
* #3271

techalchemy:
interesting. I don’t use direnv and didn’t know it doesn’t work on windows. The main issue we have raised about this approach has been that pipenv needs to be running already in order to load the dotenv file, which means it already has not loaded that file before running. 

The proposal would require separately loading and parsing `env` files before running any commands and then again during execution, which crosses contexts (things used for the state of the app runtime vs the app runner runtime). Conflating those is both confusing and demands extra context management by pipenv, which we would need to flesh out here. Between myself, Kenneth and @uranusjr we have said no to this at least 8 times, mostly for these reasons

frostming:
> I don’t use direnv and didn’t know it doesn’t work on windows
I didn't check it through, just derived from the fact `cmd` or `batch` is not listed on the official website. This may be not true.

> The proposal would require separately loading and parsing env files before running any commands and then again during execution
Yes, I also understand that.

> Between myself, Kenneth and @uranusjr we have said no to this at least 8 times, mostly for these reasons
I didn't know the histories, lol. This is just opened for discussion and any feedbacks are welcomed.

uranusjr:
I am more -0, i.e. not strongly against it, but does not feel it is a good idea.[*] I do wonder whether it is better to extrapolate `.envrc` instead of `.env`, so there is a clean separation between project and app settings. This would considerably increase the complexity, however.

[*]:  I am more in favour of having a project-level configuration file, e.g. pyproject.toml, but IIRC Kenneth is also against that.

frostming:
I think we can stop here, thanks for everyone. We will just keep on the existing features.

Issue #3410, opened 2018-12-26
Can not read from ssh repos


Corvan:
Hey,

I have some problem with installing from ssh-repos. I first thought this was a problem with gitlab, but I have the same problem on github.

### Issue description

`pipenv install -e "git+git@github.com:Corvan/upyls.git#egg=upyls"` fails. Does not matter if I use the ssh:// way of defining the repo and it fails with any other git repo over ssh for me. Tried this as well in a completely new (empty) pipenv folder.

I have loaded my ssh-key via ssh-add, and I can successfully run `git clone git@github.com:Corvan/upyls.git`

### Expected result

Installing the package works

### Actual result

```
18-12-26 14:31:12 lars@Denephor:/home/lars/Workspace/pipenvtest  $ pipenv install -e "git+git@github.com:Corvan/upyls.git#egg=upyls"
Installing -e git+git@github.com:Corvan/upyls.git#egg=upyls...
Adding upyls to Pipfile's [packages]...
✔ Installation Succeeded
Pipfile.lock (494624) out of date, updating to (a65489)...
Locking [dev-packages] dependencies...
Locking [packages] dependencies...
⠙ Pinning VCS Packages...ssh: connect to host github.com:Corvan port 22: Connection refused
fatal: Konnte nicht vom Remote-Repository lesen. (Could not read from remote repository)

Bitte stellen Sie sicher, dass die korrekten Zugriffsberechtigungen bestehen
und das Repository existiert. (Please make sure you have the proper rights to access the repository and make sure it exists)
Traceback (most recent call last):
  File "/usr/local/bin/pipenv", line 11, in <module>
    sys.exit(cli())
  File "/usr/local/lib/python2.7/dist-packages/pipenv/vendor/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/pipenv/vendor/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python2.7/dist-packages/pipenv/vendor/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python2.7/dist-packages/pipenv/vendor/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python2.7/dist-packages/pipenv/vendor/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
[...]

PR #3415, opened 2018-12-29
Update basics.rst

Corvan:
### The issue

Changed the scheme in VCS 

### The fix
With a colon instead of a slash cloning of the repository fails. This is due to the failing of git clone when ssh:// is used, a colon is considered the delimiter for the host's port and not the username. Please see https://github.com/pypa/pipenv/issues/3410 for this.

### The checklist

* [x] Associated issue

### DIFF
--- a/docs/basics.rst
+++ b/docs/basics.rst
@@ -363,7 +363,7 @@ You can install packages with pipenv from git and other version control systems 
 
     <vcs_type>+<scheme>://<location>/<user_or_organization>/<repository>@<branch_or_tag>#egg=<package_name>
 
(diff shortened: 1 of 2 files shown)


jxltom:
@frostming Hmm yes, thanks for mention it! So it is bug to fix...

Corvan:
So do you want me to open an issue or are you handling this on your own?
I don't care if you do not merge this, my main concern was the problem, and now you acknowledged that you know about it.

Corvan:
@frostming Cool thank you! So I close this Pull request.

PR #3426, opened 2019-01-05
Fix DeprecationWarning: invalid escape sequence in tests

BoboTiG:
### The issue

Fixes `DeprecationWarning: invlid escape sequence` in tests.

### The fix

Use double backslashes (raw strings would have been OK too).

### The checklist

* [n] Associated issue
* [n] A news fragment in the `news/` directory to describe this fix with the extension `.bugfix`, `.feature`, `.behavior`, `.doc`. `.vendor`. or `.trivial` (this will appear in the release changelog). Use semantic line breaks and name the file after the issue number or the PR #.

I do not think adding a NEWs entry is necessary but correct me if I am wrong :)


techalchemy:
Oooh I know exactly what this refers to. I would prefer raw strings for this fix though, if you don’t mind!

tests/unit/test_utils.py
```diff
@@ -218,7 +218,7 @@ def mock_version(path):
     @pytest.mark.windows
     @pytest.mark.skipif(os.name != "nt", reason="Windows test only")
     def test_windows_shellquote(self):
-        test_path = "C:\Program Files\Python36\python.exe"
+        test_path = r"C:\Program Files\Python36\python.exe"
```
techalchemy: Based on the presence of 4 backslashes in the original, I'm guessing you'd need 2 backslashes in a raw string and then this should work

Issue #2553, opened 2018-07-10
Errors if WORKON_HOME is not writable could be better


OrangeDog:
Pipenv version: 2018.7.1
The first error is OK (the `PermissionError`), but then there's a lot more.

```
$ WORKON_HOME=/opt/ pipenv sync
Creating a virtualenv for this project...
Pipfile: /home/me/project/Pipfile
Using /usr/bin/python3.6m (3.6.5) to create virtualenv...
⠋Running virtualenv with interpreter /usr/bin/python3.6m
Using base prefix '/usr'
Traceback (most recent call last):
  File "/home/me/.local/lib/python3.6/site-packages/virtualenv.py", line 2343, in <module>
    main()
  File "/home/me/.local/lib/python3.6/site-packages/virtualenv.py", line 712, in main
    symlink=options.symlink)
  File "/home/me/.local/lib/python3.6/site-packages/virtualenv.py", line 927, in create_environment
    site_packages=site_packages, clear=clear, symlink=symlink))
  File "/home/me/.local/lib/python3.6/site-packages/virtualenv.py", line 1112, in install_python
    mkdir(lib_dir)
  File "/home/me/.local/lib/python3.6/site-packages/virtualenv.py", line 324, in mkdir
    os.makedirs(path)
  File "/usr/lib/python3.6/os.py", line 210, in makedirs
    makedirs(head, mode, exist_ok)
  File "/usr/lib/python3.6/os.py", line 210, in makedirs
    makedirs(head, mode, exist_ok)
  File "/usr/lib/python3.6/os.py", line 220, in makedirs
    mkdir(name, mode)
PermissionError: [Errno 13] Permission denied: '/opt/clump-D6KMH_ut'
Error while trying to remove the /opt/clump-D6KMH_ut env:
No such file or directory

Virtualenv location:                                                                                                                            Traceback (most recent call last):
  File "/home/me/.local/bin/pipenv", line 11, in <module>
    sys.exit(cli())
  File "/home/me/.local/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/home/me/.local/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/home/me/.local/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 1066, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
[...]

PR #3458, opened 2019-01-18
WIP: Raise exception from subprocess.check_output before actually calling …

AdamGold:

**WIP**

### The issue

1. #2553 
2. errors from subprocess not being caught
3. `misc.py` called `sys.stderr` with 3 arguments instead of 1.

### The fix

call `subprocess.check_output` before `Popen` and raise exceptions.

### The checklist

* [x] Associated issue


AdamGold:
What custom exception do you think I should raise when the return code from the command is non zero (==failed)?

frostming:
Changes to vendors should be submitted to upstream as well.

PR #3462, opened 2019-01-19
Correct the options of update command

### LINKED ISSUE #3461: Confusion: --keep-outdated --selective-upgrade
Just to get some clarity. I'm trying to update a specific package. The docs say I should do:

```
pipenv update my_package
```

The help says that `--selective-upgrade` "Updates specified packages."... isn't that what the above command does? And shouldn't we rename it to `selective-update` instead of `selective-upgrade` 😅 

I'm just trying to update a specific package without pipenv trying to update all other outdated dependencies. The docs suggest that this is what `--keep-outdated` does... but shouldn't that be the default behaviour?

frostming:
Fixes #3461 

Correct the options of `update` command to reduce confusion.

### DIFF
--- a/pipenv/cli/command.py
+++ b/pipenv/cli/command.py
@@ -454,7 +454,7 @@ def check(
     "--outdated", is_flag=True, default=False, help=u"List out-of-date dependencies."
 )
 @option("--dry-run", is_flag=True, default=None, help=u"List out-of-date dependencies.")
-@install_options
+@sync_options
 @pass_state
 @pass_context
 def update(


frostming:
I am wrong. Close this PR now.

pipenv/cli/command.py
```diff
@@ -454,7 +454,7 @@ def check(
     "--outdated", is_flag=True, default=False, help=u"List out-of-date dependencies."
 )
 @option("--dry-run", is_flag=True, default=None, help=u"List out-of-date dependencies.")
-@install_options
+@sync_options
```
jxltom: The package is passed to ```state.installstate.packages``` by ```package_arg```, and being used by https://github.com/pypa/pipenv/blob/50dcaca07a60483d83eaf2bca8d3a661fa765472/pipenv/cli/command.py#L482. Why it is replaced by ```sync_options```

frostming: @jxltom, that was a mistake. However, state.installstate.packages are not used after this point.

jxltom: Oh it is closed. Forget about why I said. :p

jxltom: @frostming That's true. We should support it actually. 😄

PR #3463, opened 2019-01-20
conservative check for known exceptions in subprocess stderr.

AdamGold:

### The issue

#2553 
### The fix

Conservative checks of known exceptions when subprocess returns output.

### The checklist

* [x] Associated issue

### DIFF
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -925,9 +925,17 @@ def do_create_virtualenv(python=None, site_packages=False, pypi_mirror=None):
         )
         click.echo(crayons.blue("{0}".format(c.out)), err=True)
         if c.returncode != 0:
-            sp.fail(environments.PIPENV_SPINNER_FAIL_TEXT.format(u"Failed creating virtual environment"))
+            sp.fail(environments.PIPENV_SPINNER_FAIL_TEXT.format("Failed creating virtual environment"))
+            known_exceptions = {
+                "PermissionError": "Permission denied:",
+            }
+            # PermissionError - hide the traceback for better UX
+            for partition in (part
+                              for e, part in known_exceptions.items() if e in c.err):
+                known_exceptions_partition = c.err.rpartition(partition)
+                c.err = "{} {}".format(known_exceptions_partition[1], known_exceptions_partition[2])
             raise exceptions.VirtualenvCreationException(
-                extra=[crayons.blue("{0}".format(c.err)),]
+                extra=[crayons.red("{0}".format(c.err)),]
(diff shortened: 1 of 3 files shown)


AdamGold:
This is a UX improvement PR. Do you guys think there is a better way of doing it?

pipenv/core.py
```diff
@@ -911,8 +911,16 @@ def do_create_virtualenv(python=None, site_packages=False, pypi_mirror=None):
...
         if c.returncode != 0:
             sp.fail(environments.PIPENV_SPINNER_FAIL_TEXT.format("Failed creating virtual environment"))
+            known_exceptions = {
+                "PermissionError": "Permission denied:",
+            }
+            # PermissionError - hide the traceback for better UX
+            for partition in (part
+                              for e, part in known_exceptions.items() if e in c.err):
+                known_exceptions_partition = c.err.rpartition(partition)
+                c.err = "{} {}".format(known_exceptions_partition[1], known_exceptions_partition[2])
```
techalchemy: Won't `known_exceptions_partition[1]` and `known_exceptions_partition[2]` just be `PermissionError: <message>` in  this case? I like where this is headed but I feel like it might belong in `exceptions.py`.  Also note that it really isn't appropriate to manipulate `c.err` -- it is the stored `stderr` stream output from a subprocess that has already or is currently executing.  We really should only ever manipulate how we interpret and display the result.

Possibly we should have some helper method like this:

```python
KNOWN_EXCEPTIONS = {
    "PermissionError": "Permission denied:"
}

def prettify_exc(exc):
    matched_exceptions = [k for k in KNOWN_EXCEPTIONS.keys() if k in exc]
    errors = []
    for match in matched_exceptions:
        exc, exc_name, error = exc.rpartition(KNOWN_EXCEPTIONS[match])
        errors.append("{0} {1}".format(exc_name, crayons.red(decode_for_output(error))))
    else:
        errors.append("{0}".format(decode_for_output(exc)))
    return "\n".join(errors)
```

And then the exceptions themselves can share a dictionary of known exceptions and use `prettify_exc` when `environments.is_verbose() is False`

AdamGold: That's a good idea. I have pushed the ``prettify_exc`` implementation.

techalchemy:
Sorry for being incredibly slow on this, but this is looking pretty good to me... I am excited to get things refactored in this direction.

Thanks for taking care of this!

AdamGold:
@techalchemy, Awesome. Rebased to `master`.

PR #3472, opened 2019-01-23
Added --db, --output, and --squelch flags. Cleaned up duplicate code.

kylecribbs:
--db allows users to specify a local source to the json databse for offline safety check.
--output allows users to specify json, full-report, or bare from safety check
--squelch hides all stdout from pipenv.
Removed some unneeded duplicate code.

# PEEP-044: safety-db integration, squelch, and output.

pipenv check needs offline, ci, and other output capabilities.

☤

Not everyone can utilize pipenv check and access the internet. Safety check knew this and that is why they created safety-db. This repository contains a json database that is updated monthly. Safety check allows you to pass a --db flag that is a local directory containing that database. Safety check also allows you to pass --json, --bare, and --full-report. Pipenv check has their own way of displaying the results that is why I believe there should be a --output flag that allows users to specify json, bare, and full-report from safety check and default for the current pipenv check output. Currently, pipenv check has a lot of stdout messages and makes it harder to pipe the results into something to be checked (especially for continuous integration pipelines). That is why adding a --squelch switch is also important. This will be default False (display all stdout); however, the user has the option to add the --squelch switch to make the output only come from safety check. 

### Example 1
``` bash
pipenv check
Passed!
Checking installed package safety…
25853: insecure-package <0.2.0 resolved (0.1.0 installed)!
This is an insecure package with lots of exploitable security vulnerabilities.
```
### Example 2
``` bash
pipenv check | jq length
parse error: Invalid numeric literal at line 1, column 9
```

### Example 1
``` bash
pipenv check --db /Users/macbookpro/workspace/test/safety-db/data/ --output json --squelch 
[
    [ 
        "insecure-package",
        "<0.2.0",
        "0.1.0",
        "This is an insecure package with lots of exploitable security vulnerabilities.",
        "25853"
    ]
]
```
### Example 2
``` bash
pipenv check --db /Users/macbookpro/workspace/test/safety-db/data/ --output json --squelch | jq length
1
```


kylecribbs:
Any updates?

kylecribbs:
Hey its been over a month and haven't heard anything.

techalchemy:
Sorry for the delay, we have had a lot of work to do catching up the codebase with `PEP517` and `PEP518`. Will get this reviewed shortly

pipenv/cli/command.py
```diff
@@ -416,21 +416,41 @@ def run(state, command, args):
...
 @argument("args", nargs=-1)
 @pass_state
 def check(
     state,
     unused=False,
+    db=False,
     style=False,
     ignore=None,
+    output="default",
+    squelch=False,
```
frostming: Seems this flag works similar to `--verbose`?

kylecribbs: Correct.. But Pipenv always has their STDOUT it seems so i made a "squelch" to hide those outputs.. It makes it easier to use Pipenv in a CI/CD pipeline if you have raw json output.

kylecribbs: Does Pipenv have something for that? if so I can remove.

kylecribbs: @techalchemy

kylecribbs:
bumping..

remarkablerocket:
This feature would be really useful right now since pyup.io appears to be broken today.

Would personally prefer to set an environment variable for DB location, but that's a minor quibble.

kylecribbs:
That’s an easy add. Just not sure if they are going to merge this in.

techalchemy:
hah, yeah this would be useful for sure.  ok let me review this, super sorry for the delay, pipenv itself has been having some rough times but things are finally all functional with many improvements!

peeps/PEEP-044.md
```diff
@@ -0,0 +1,54 @@
...
+        "<0.2.0",
+        "0.1.0",
+        "This is an insecure package with lots of exploitable security vulnerabilities.",
+        "25853"
+    ]
+]
+```
+### Example 2
+``` bash
+pipenv check --db /Users/macbookpro/workspace/test/safety-db/data/ --output json --squelch | jq length
```
techalchemy: typically we use `--quiet` which is already configured as a CLI flag and can be accessed throughout pipenv via `pipenv.environments.is_quiet()`  but I am otherwise in favor of this

pipenv/cli/command.py
techalchemy: I'd favor `quiet` here over `squelch` which is what I believe @frostming is referring to.  I don't mind too much about the output flag, but we do use `--json` for `pipenv graph` (this command has other options and if you intend to respect them I can see this being appropriate)

kylecribbs:
Did you want me to fix any of it or plan on fixing on the merge?

pipenv/cli/command.py
uranusjr: I agree, at least coming from a non-English-native background. I can easily imagine my fellow developers having trouble understanding what `squelch` means (let along spell it correctly).

uranusjr: That said, maybe `quiet` is not the best choice either (I’d expect it to output next to nothing). Maybe something like `--raw` would be both more understandable and still semantically good enough?

techalchemy:
> Did you want me to fix any of it or plan on fixing on the merge?

Sorry for the delay, you can go ahead and make those changes and we can get it merged, thanks!

peeps/PEEP-044.md
kylecribbs: did you want me to from pipenv.environments import is_quiet and then do if quiet or is_quiet(): ?

kylecribbs:
Added changes

kylecribbs:
Bumping

peeps/PEEP-044.md
techalchemy: the function call is a bit more robust as it allows for changes that might happen after pipenv loads.

kylecribbs:
> I think a news file and a rebase makes this pretty much good to go, thanks for all the work you've done on it!

Let me know if you need anything else!

PR #3479, opened 2019-01-24
add test for counting pre option and fix it's duplicate

1estart:

### The issue

Issue:
`pipenv install --help` shows duplicate --pre options #3475

### The fix

I think problem in cli/options.py. install option already inherits --pre option from install_base+options through sync_options


tests/unit/test_help.py
```diff
@@ -9,3 +9,16 @@ def test_help():
         stderr=subprocess.STDOUT, env=os.environ.copy(),
     )
     assert output
+
+
+def test_count_of_description_pre_option():
+    test_command = 'pipenv install --help'
+    test_line = '--pre Allow pre-releases.'
+    out = subprocess.Popen(test_command.split(), stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
```
frostming: This case should be in `integration` directory and use `PipenvInstance` to keep consistency.

techalchemy:
once the changes are made per @frostming we can merge this, thanks for the pr!

techalchemy:
Thanks for the pr

I'm calling this good enough, merging

PR #3498, opened 2019-02-03
Use HTTPS for gif in README

salty-horse:
The current URL uses HTTP.
Since Github and PyPI both mirror the file and serve it through HTTPS, it's less likely to ever encounter "insecure content" warnings, but still nice-to-have.


frostming:
Thanks

PR #3507, opened 2019-02-05
Fix version specifier documentation mistake

DpEpsilon:
The [Specifying Versions of a Package](https://pipenv.readthedocs.io/en/latest/basics/#specifying-versions-of-a-package) section of the basics document has a line which reads:

>The use of ~= is preferred over the == identifier as the former prevents pipenv from updating the packages:

This would imply that ~= locks the entire version number and == does not, but the opposite is true according to [PEP 440](https://www.python.org/dev/peps/pep-0440/#version-specifiers).

This PR replaces "former" with "latter" in the quoted line to correct this mistake.


frostming:
Looks good to me, will merge after CI pass

Issue #3499, opened 2019-02-03
docs: Main page references video that no longer exists


salty-horse:
The file `docs/index.rst`, hosted at https://pipenv.readthedocs.io/en/latest/, embeds a Vimeo video ("Pipenv screencast") that no longer exists:

https://player.vimeo.com/video/233134524

If there is no intention to restore it, perhaps it should be replaced with the gif from the README:

https://s3.amazonaws.com/media.kennethreitz.com/pipenv.gif

nachiketdhamankar:
Unable to get access to the gif.

https://s3.amazonaws.com/media.kennethreitz.com/pipenv.gif

frostming:
Need another store for the image file, any recommendations?

PR #3519, opened 2019-02-08
Replace (non-existant) video on docs index.rst with equivalent gif

frob:
Fix for #3499

### The issue

The video on the home page of the docs no longer exists.

### The fix

Replaces the video with an equivalent gif.


frostming:
Thanks for the PR, it's going to be merged after CI passes

PR #3522, opened 2019-02-11
Docs: clarify quoting in version specs note

davidjb:
### The issue

Previously it was confusing as to what `" "` meant exactly -- I read it
as a whitespace character was needed somewhere when it's referring to
the double quote characters.

### The fix

Adjusts the wording to be clearer and gives a concrete example rather than just characters which could be interpreted otherwise.

### The checklist

* [ ] Associated issue -- there isn't one; hopefully this PR is enough as it is minor


frostming:
Thanks

Issue #3527, opened 2019-02-12
Docs hide navigation in responsive mode


davidjb:
### Issue description

Documentation pages currently hide the "sidebar" in responsive mode (eg small screen sizes), which means the navigation isn't visible on mobiles, making the docs unusable.  The issue is with an over-zealous `display: none` at https://github.com/pypa/pipenv/blob/dfa1dceb6399ac695cb5001d3010214e84303533/docs/_templates/hacks.html#L21 -- PR is incoming to remove this line which fixes the issue.

It's not as obvious on the homepage because there's lots of navigation links _in_ the content, but it's clear on a sub-page like https://pipenv.readthedocs.io/en/latest/dev/philosophy/ (seen on a small screen):

![before-unusable-missing-sidebar](https://user-images.githubusercontent.com/1002811/52607908-f0572000-2e6f-11e9-9126-792657a33e08.png)

### Expected result

The page should have the "sidebar" navigation at the bottom:

![after-showing-sidebar](https://user-images.githubusercontent.com/1002811/52607896-e7664e80-2e6f-11e9-97ff-58e5756255c9.png)

### Actual result

As per the expected result.

### Steps to replicate

1. Load a docs URL https://pipenv.readthedocs.io/en/latest/dev/philosophy/ either on a mobile device or just use responsive design mode in browser.
1. Note that the navigation isn't visible.

PR #3528, opened 2019-02-12
Ensure docs show nav on small-screen devices

davidjb:
### The issue

The sidebar (which is the bottom navigation) on mobile devices is hidden by a line of CSS.  This prevents pages from being navigated on small-screen devices, as mentioned in #3527.

### The fix

Removes the CSS line, which means the navigation gets shown again.  Nav is responsive on small-screen devices care of the Alabaster base theme so there's no issue with showing it.

### The checklist

* [x] Associated issue


techalchemy:
Seems like a good idea to me, thanks for the PR and for your patience while we fix things :)

techalchemy:
Also thanks @dimaqq for the triage

PR #3536, opened 2019-02-13
Removing my testimonial

jezdez:
I'm worried that my testimonial would encourage users of pip to switch to pipenv without fully diving into the subtle differences to the venv+pip development workflow they may be used to. Since I've signed the statement with "former pip maintainer" this can in particular create frustrations to users and lead to wrong expectations.


techalchemy:
@jezdez happy to remove this if you like, I'd encourage you to try out the master branch and suggest other ways to help with that clarification if you think this remains a point of confusion.  We push back pretty strongly against that idea and work closely with the pip maintainers to try and track shared logic so we can work out a path forward for merging code where applicable.

Pip isn't going anywhere.  It's a lot more likely that a bunch of our logic will get incorporated into pip and some shared logic will get refactored out of both places.

jezdez:
@techalchemy Thank you for merging @techalchemy!

FWIW I really appreciate everyone's work on pipenv and want to stress that removing the testimonial is no indicator for my personal interest in pipenv as a project nor anyone's work here. Keep up the great work you're doing!

PR #3561, opened 2019-02-26
removed what is seems as redundant PIPENV_ACTIVE when running pipenv shell

codeinvain:
`pipenv shell` do not respect `VIRTUAL_ENV` env var because `PIPENV_ACTIVE` is set before accessing `project.virtualenv_location` method


frostming:
This is not intended and will surely break things. Thanks for that tho.

Issue #3292, opened 2018-11-23
pytest-pypi has pytest-httpbin README, nonexisitng website


hroncok:
I was looking at pytest-pypi at https://pypi.org/project/pytest-pypi/

The website points to https://github.com/kennethreitz/pytest-pypi and that is 404

The README looks like copy paste from pytest-httpbin.

Is the upstream of this PyPI project at https://github.com/pypa/pipenv/tree/master/tests/pytest-pypi ?

Can the README, DESCRIPTION.rst (= long description on PyPI) and URL by updated?

Thanks

frostming:
Would you like to send a PR for this? I will be glad to review.

hroncok:
Yes but no promises on when. I will try to get it done this week, but there other things piling up.

PR #3562, opened 2019-02-26
Update pytest-pypi documentation not to be pytest-httpbin documentation

hroncok:
Fixes https://github.com/pypa/pipenv/issues/3292

I was not able to write proper docs. So at least remove the non relevant ones.


techalchemy:
thanks for the pr, seems good to me

PR #3578, opened 2019-03-05
add spinner during progress

techalchemy:
add spinner during progress


techalchemy:
The ui change here demands a lot of usability changes that I just don't think are really worth pursuing just to put a spinner next to the progress bar -- printing warnings freezes the bar and rewrites it again for example.

IMO this is just not worthwhile.

PR #3595, opened 2019-03-08
Let windows user configure to show emojis

frostming:
### The issue

Windows can't be pleased by the emojis.

### The fix

Keep the behavior that emojis are hidden by default for windows users, but can be enabled by setting `PIPENV_HIDE_EMOJIS=0`

### The checklist

* [ ] Associated issue


techalchemy:
Seems good. We should write truthy test functions to look for `0`, `False`, `off`, `no` to avoid duplicating this everywhere

frostming:
@techalchemy Sure, I tried but failed to find a way to put that function in `utils` that there is a cyclic import problem.

PR #3606, opened 2019-03-11
Vendor cached property

THEMAXIMUM77:

### The issue

### The fix

### The checklist

* [x] Associated issue


frostming:
@THEMAXIMUM77 Are you playing around with PR or just can't find the correct way to do it?

Issue #2396, opened 2018-06-21
Pin `pipenv install X` to compatible version of X in Pipfile instead of *


dlukes:
**Is your feature request related to a problem? Please describe.**

In my experience, `pipenv` seems to be a bit too eager to update to newer versions of dependencies (see e.g. #1219). Based on other tools like `cargo` and `npm`, I would expect `pipenv install` in a newly checked out project to get me a working environment for running the project. However, since `pipenv install X` pins to `X = "*"` in the Pipfile, it is possible that I'll get newer versions of the deps incompatible with the given project.

**Describe the solution you'd like**

See also <https://github.com/pypa/pipenv/issues/1219#issuecomment-360690738>:

> Maybe Pipenv should do what npm does and make `pipenv install` pin to the major version instead of `'*'`.

I think it would be great if `pipenv install X` pinned to `X = "~=<current_version>"` instead of `X = "*"`, ideally by default (but I admit my perspective may be limited by my particular use cases) or at least through an option that can be specified on the command line and/or in the Pipfile itself, something like `--pin-to-current` / `pin_to_current = true`.

**Describe alternatives you've considered**

What I'm doing currently is trying to remember to run `pipenv install --ignore-pipfile` instead of `pipenv install` when setting up a project for development on a new machine. However, this skips all updates, even compatible ones (bug / security fixes etc.), since the purpose of `--ignore-pipfile` is deterministic builds for production, not setting up a compatible environment for development.

uranusjr:
The general proposition of using `~=` (or even `^=`, which does not exist in pip) is still valid, and I have seen this raised quite a few times now. `*` means every re-lock and update could potentially break user code because a new major version is released and it is not compatible with the application. 

You could argue the user should be aware of this, and use e.g. `django~=2.1.0` in the first place. But still, there is no good way to say “I want the latest ~= possible” instead of manually looking it up. It would be immensely helpful to either a) set this as default, and require `django==*` to explicitly get `*`, or b) have a command flag as mentioned above (an option in Pipfile less so IMO).

While the problem described is probably fixed, the user experience side of this proposition is still worth discussing.

dlukes:
Sorry, I should have given a more concrete user story in the OP to make the issue clearer :) For instance:

1. I start a website project, do `pipenv install flask`. This puts `flask = "*"` in my Pipfile and installs the latest version of flask, say `0.12.3`.
[...]

PR #3631, opened 2019-03-20
PEEP: Use `>=` Specifier in Pipfile by Default

uranusjr:
See #2396 for discussion. Note that this does *not* propose the idea of semvar compatible, only a way to introduce a more useful record of installation, so users can work on the end result more easily.

Rendered result: https://github.com/pypa/pipenv/blob/91056e2f50d0/peeps/PEEP-004.md


ncoghlan (commented):
The proposal makes sense to me, but there's one clarification I would request: more information on what goes into `Pipfile` in the case where `pipenv install` finds that the package is already installed, but not yet mentioned in the `Pipfile`. In that case, I'd expect it to add `>=the.installed.version`, rather than adding `*`, but the current wording suggests the latter ("If the package is already installed, this proposal does not change anything.").

uranusjr:
Thanks for pointing it out, I honestly did not think of this scenario at all. I will update the proposal to include this.

uranusjr:
I have updated the proposal. I also noticed that there’s already a PEEP 004 (but named `peep-0004` instead), so I changed this to 005.

techalchemy:
My main question is this:

As a user (especially as a new user to the python ecosystem) I am primarily interested in being able to re-lock my dependencies in a non-conflicting manner as I introduce additional packages to my pipfile (i.e. I want to minimize my cognitive load).  Does this help with that?

I see the argument is about pinning `>=` with a suggestion that this is the current behavior and we are now making it explicit, but I want to emphasize that this is _not_ the current behavior.  We currently don't pin any version, the desired behavior is to _find a resolution that is compatible with all of the specified dependencies_.

By explicitly enacting a pin to whatever the latest version was at the time of installation, are we not _implicitly_ assuming semver on some level at the very least?

If the user adds an additional dependency which requires an earlier version of another dependency which will now be pinned `>=`, won't this be a substantial problem? And when we do ultimately cut in an actual backtracking resolver (which we are OK to do whenever, based on conversations with Kenneth, so it depends on us being comfortable with it), doesn't this prevent us from leveraging the full functionality of it because we won't allow it to fully backtrack?

I am all for a flag that provides compatibility pinning functionality, btw, I just need some clarification on this last bit.

/cc @ncoghlan in case you aren't subscribed

uranusjr:
> As a user (especially as a new user to the python ecosystem) I am primarily interested in being able to re-lock my dependencies in a non-conflicting manner as I introduce additional packages to my pipfile (i.e. I want to minimize my cognitive load). Does this help with that?

I guess it does not help with that, but generally there should be no consequences at all. The only situation this has a impact is when a package *lowers* its dependency versions when upgrading.

Let’s say you have `A` in Pipfile, which depends on `B`. You  and run `pipenv install B`. Old behaviour is to install B 2.0 and insert `B = "*"`. Now A releases 2.0 that depends on `B<2` (maybe because maintainers realise the previous pin was too lax),  and when you lock, B will be *downgraded* automatically to 1.0. In the new behaviour, however, `B = ">=2.0"` is inserted instead, and locking would fail after A 2.0 is out.

As mentioned in the proposal, I feel this is not a negative, but a neutral change. Failing to relock serves as immediate hint that something is wrong with the assumption of versions, instead of letting the user figure out why B is downgraded after the fact.

> I see the argument is about pinning >= with a suggestion that this is the current behavior and we are now making it explicit, but I want to emphasize that this is not the current behavior. We currently don't pin any version, the desired behavior is to find a resolution that is compatible with all of the specified dependencies.

It is debatable, however, whether the resolution fits the user’s expectation. When users `pipenv install <package>`, they are making an explicit statement of “I want to use this package”, and if pip semantics are inherited, “I want the latest version of this package”. This is different from intrinsic dependencies (i.e. packages included because other packages want them), in which case the user generally does not care about the version. When users directly request a package, it should be expected that their code would use some of its API, and more likely to break when those directly-depended packages are downgraded unexpectedly. A `>=` pin in Pipfile matches the users’ expectation better IMO, and although it would make previously-passing locking fail, I argue it is actually a desirable failure to help users debug package compatibility before it breaks the code.

uranusjr:
(I forgot to include this in the previous comment.)

> By explicitly enacting a pin to whatever the latest version was at the time of installation, are we not implicitly assuming semver on some level at the very least?

No, `>=` does not assume semver, but only requires the versions to be order-able (i.e. a list of versions can be reliably sorted). All packages distributed through a package index should already satisfy this. (Packages distributed directly via e.g. URL specs don’t need to adhere to this, but they are not installable with `pipenv install <name>` anyway, and therefore out of scope of this PEEP).

techalchemy:
I mentioned this on slack, but if I install for example `pipenv 2019.4.1` but I later add a pin that needs `pipenv 2019.3.28` we don’t handle that in resolution.  If I wanted to pin an explicit version I would have done that -- as it stands, and this is pretty common -- the moment I install something that in turn pins a lower version of any package I previously `pipenv installed`, I will have a number of conflicts.

Ultimately, this comes down to a simple point: You think people _mean_ `>=<latest>` when they use `pipenv install <package>`.  I simply don't think that's right.

uranusjr:
The only way to find out what more people expect is to set up an actual situation for voting.

techalchemy:
Not sure voting is great and this isn’t a democracy anyway. I am far more interested in UX and usability than an online poll and that’s probably the only thing that will govern my decision.

Rust and npm enforce semver, so when a user types a package name they can loosely pin to the patch release. In npm at least, if the user then installs something that demands an earlier patch release, it just installs both— no problem.

In pipenv this proposal means that if, as a user, I type 'pipenv install pyxl', there is now an implicit pin in my Pipfile of 'pyxl = ">=2.9"' (for example).  If I try to install anything at all with a dependency on for instance, 'pyxl<2.9', that will now fail no matter what even though I never specified any versions of anything.  That seems pretty jarring and would likely introduce resolution failures all over the place.

I did show this to a few folks and by and large the response was, ‘why would you set versions for things when I didn’t tell you to, especially if it might cause more conflicts I now have to troubleshoot?’  I get that this is a push back against ‘I got a smaller version’ but my answer is just, that’s how dependency resolution works. If you want a higher version, pin it.

Overall I am 100% against changing default behaviors in a breaking way without a clear demonstration of why the change is better. Pipenv is widely deployed and this would be a pretty significant usability change.  If you have this working somewhere i can try it out myself a bit and see how it feels. I’m -1 on this right now but if in practice it doesn’t cause problems I could come around to it.

ncoghlan:
@techalchemy That's a really good point regarding the increased risk of resolution conflicts.

One possible way to thread the design needle would be to add the extra metadata regarding the "first installed version" as an end-of-line comment, rather than changing the actual dependency specification.

With that approach, the current output of:

    django = "*"

Would instead become something like:

    django = "*" # To require first installed version:  django >= "2.1.7"

If folks find the comments messy, they can just delete them (or there could be an option to skip adding them), while if they keep them, then setting a minimum version later should be straightforward.

dlukes:
> One possible way to thread the design needle would be to add the extra metadata regarding the "first installed version" as an end-of-line comment

I like this idea very much! It's conservative, which is a good thing in light of the reservations e.g. @techalchemy has expressed, but at the same time the information about which combination of versions of the dependencies should work is easily accessible. With this feature, I'd probably leave `*` deps by default, and only consider pinning if after setting up a new dev environment, stuff breaks.

matteius:
@uranusjr do you mind if we decline this one?  There have been recent discussions of adding to the `[pipenv]` directives the ability to customize the default behavior of how the Pipfile gets pinned during install.

uranusjr:
Feel free.

PR #3635, opened 2019-03-20
Rewrote sentences to increase clarity

acdoc:
Also reduced hedging

### The issue

### The fix

### The checklist

* [ ] Associated issue


uranusjr:
The changes look good to me. Would you mind adding a file at `news/3635.doc` describing this change?

docs/advanced.rst
```diff
@@ -554,9 +554,9 @@ at all, use the `PIP_IGNORE_INSTALLED` setting::
...
 
-Libraries provide reusable functionality to other libraries and applications (let's use the umbrella term **projects** here). They are required to work alongside other libraries, all with their own set of subdependencies. They define **abstract dependencies**. To avoid version conflicts in subdependencies of different libraries within a project, libraries should never ever pin dependency versions. Although they may specify lower or (less frequently) upper bounds, if they rely on some specific feature/fix/bug. Library dependencies are specified via ``install_requires`` in ``setup.py``.
+Libraries provide reusable functionality to other libraries and applications (let's use the umbrella term **projects** here). They are required to work alongside other libraries, all with their own set of subdependencies. They define **abstract dependencies**. To avoid version conflicts in subdependencies, libraries should never pin dependency versions. Although they may specify lower or (less frequently) upper bounds, if they rely on some specific feature/fix/bug. Library dependencies are specified via ``install_requires`` in ``setup.py``.
```
brainwane: "Although they may specify lower or (less frequently) upper bounds, if they rely on some specific feature/fix/bug."

This is a sentence fragment; maybe use "However" instead of "Although"?

brainwane:
@acdoc Hope my suggestion is helpful. And, as @uranusjr mentioned, please add a `news` fragment.

matteius:
Noting that we should adopt these changes to a new branch based on latest main and add a news fragment.  Then we can close out this PR at that time.

Issue #3644, opened 2019-03-24
test_specific_package_environment_markers fails when requests is de-vendored


jayvdb:
As `test_specific_package_environment_markers` uses requests, the following will always succeed in a test environment where requests is always available because `PYTHONPATH` has been modified to include all de-vendored libraries

> c = p.pipenv('run python -c "import requests;"')

A simple fix should be to use a package name which is not even real, as the environment marker should ensure the package name isnt even resolved.  But maybe it does resolve it anyway, in which case perhaps the test could use `tablib` instead?

Also test_pipenv_graph and test_pipenv_graph_reverse fail when requests is de-vendored, likely for the same reason.

jayvdb:
Note that this test might be working ok only if the requests-*.egg-info is present.

However `test_pipenv_graph` and `test_pipenv_graph_reverse` still fail even then.

jayvdb:
Also failing when requests is de-vendored
- test_update_locks 
- test_uninstall
- test_mirror_uninstall
- test_uninstall_all_local_files
- test_uninstall_all_dev

```console
[ 2195s] ______________________________ test_update_locks _______________________________
[ 2195s] 
[ 2195s] PipenvInstance = <class 'tests.integration.conftest._PipenvInstance'>
[ 2195s] pypi = <pytest_pypi.serve.Server object at 0x7fa35c02eb50>
[ 2195s] 
[ 2195s]     @pytest.mark.update
[ 2195s]     @pytest.mark.lock
[ 2195s]     def test_update_locks(PipenvInstance, pypi):
[ 2195s]     
[ 2195s]         with PipenvInstance(pypi=pypi) as p:
[ 2195s]             c = p.pipenv('install requests==2.14.0')
[ 2195s]             assert c.return_code == 0
[ 2195s]             with open(p.pipfile_path, 'r') as fh:
[ 2195s]                 pipfile_contents = fh.read()
[ 2195s]             pipfile_contents = pipfile_contents.replace('==2.14.0', '*')
[ 2195s]             with open(p.pipfile_path, 'w') as fh:
[ 2195s]                 fh.write(pipfile_contents)
[ 2195s]             c = p.pipenv('update requests')
[ 2195s]             assert c.return_code == 0
[ 2195s]             assert p.lockfile['default']['requests']['version'] == '==2.19.1'
[ 2195s]             c = p.pipenv('run pip freeze')
[ 2195s]             assert c.return_code == 0
[...]

PR #3645, opened 2019-03-24
Avoid using requests in tests

jayvdb:
test_specific_package_environment_markers using requests
fails when requests is pre-installed via PYTHONPATH.

Closes https://github.com/pypa/pipenv/issues/3644


techalchemy:
hm, so is the problem that the already installed version is being imported? if so, why?

jayvdb:
It is being imported because it is in `PYTHONPATH` before the test runner is started.

This is needed because a rpm package contents are installed into a build root which isnt the same as the system image.

So the de-vendored requests exists in `/usr/lib/python3.6/site-packages` and pipenv exists at `/foo/bar/build/root/usr/lib/python3.6/site-packages`.

In order the run the tests, `PYTHONPATH` needs to be set to `/usr/lib/python3.6/site-packages:/foo/bar/build/root/usr/lib/python3.6/site-packages`.

I suspect there is a 'bug' in pipenv where it is not removing items from `PYTHONPATH` when it is running commands supposed to be inside the pipenv venv, so libraries in `PYTHONPATH` in the outer scope are accessible inside `pipenv run`.  I put 'bug' in quotes as de-vendoring is of course not supported, and shouldnt be until there are solid tests which confirm that pipenv works correctly when de-vendored.  At present I think the focus should be on identifying any ways in which pipenv misbehaves when de-vendored, so distros can decide whether the de-vendoring is more important than any resulting strange behaviour.  I'm also seeing some misbehaviour when the de-vendored packages were installed using only distutils instead of setuptools, or if the egg-info isnt present (both are scenarios which happen in rpm managed python packages, although I think openSUSE has egg-info installed with most packages, and I assume Fedora also does, so that may be less of a concern now-a-days.).

uranusjr:
Won’t using tablib have the same problem anyway, just with a different package? If we’re to change that, I’d want it to be changed to an actual environment-free implementation.

jayvdb:
@uranusjr , I provide this patch because it fixes the test in my de-vendored builds.  I also have a fix for `test_pipenv_graph` and `test_pipenv_graph_reverse` mentioned in https://github.com/pypa/pipenv/issues/3644, but need to polish it before I submit it.  The other five errors at https://github.com/pypa/pipenv/issues/3644 are probably the 'real' problems, where it should fail, and I havent started on those yet.

Using `tablib` doesnt have the same problem because `tablib` is not a dependency of `pipenv`, so it is not pre-installed into `PYTHONPATH` in order to test `pipenv`, and thus will not exist unless it is installed into the venv.  The markers in this test case prevent anything from being installed, so neither `tablib` nor `requests` should be accessible in the venv, but as I have explained `requests` does exist because it is in the outer scopes `PYTHONPATH` and somewhere `pipenv` isnt managing `PYTHONPATH` sufficiently to prevent it from existing inside the venv.  The underlying problem may even be impossible to solve, if `pipenv` needs to existing inside the venv, as that means a de-vendored `pipenv` will have all deps also existing inside the venv , and a hell of mess if the Pipfile requires different versions than `pipenv`.  However the underlying problem should have dedicated tests, and documentation about it (because de-vendorers will de-vendor anyway, unless the problems caused are very clear and verifiable), and it shouldnt trigger inexplicable failures all over the test suite.  Using `requests` here is one such trigger, and it is unnecessary as any module could be used in this test case - it doesnt need to be `requests`.

uranusjr:
I see, thanks for the explanation.

techalchemy:
Yeah that all makes sense to me, thanks for the details

Issue #356, opened 2017-05-15
pipenv install is very slow


Diggsey:
Running `pipenv install` after changing *one* dependency takes about ~5 minutes for me, on a windows 10 machine with an SSD.

The vast majority of that time is spent inside `Locking [packages] dependencies...`

It seems like there might be some quadratic-or-worse complexity in this step?

I've included most of our pipfile below, but I had to remove some of our private repo dependencies:
```
[[source]]
url = "https://pypi.python.org/simple"
verify_ssl = true

[packages]
alembic = "==0.8.4"
amqp = "==1.4.7"
analytics-python = "==1.2.5"
anyjson = "==0.3.3"
billiard = "==3.3.0.20"
braintree = "==3.20.0"
celery = "==3.1.18"
coverage = "==4.0.3"
docopt = "==0.4.0"
eventlet = "==0.19.0"
flake8 = "==3.0.4"
Flask-Cors = "==2.1.2"
Flask-Login = "==0.3.2"
Flask = "==0.12.1"
funcsigs = "==0.4"
fuzzywuzzy = "==0.12.0"
gcloud = "==0.14.0"
html2text = "==2016.9.19"
itsdangerous = "==0.24"
Jinja2 = "==2.8"
jsonpatch = "==1.15"
jsonschema = "==2.5.1"
PyJWT = "==1.4.2"
kombu = "==3.0.30"
LayerClient = "==0.1.9"
MarkupSafe = "==0.23"
mixpanel = "==4.3.0"
mock = "==1.3.0"
nose-exclude = "==0.4.1"
nose = "==1.3.7"
numpy = "==1.12.1"
pdfrw = "==0.3"
Pillow = "==4.1.0"
pusher = "==1.6"
pycountry = "==1.20"
pycryptodome = "==3.4.5"
pymongo = "==3.2"
PyMySQL = "==0.7.4"
python-dateutil = "<=2.5.1"
python-Levenshtein = "==0.12.0"
python-magic = "==0.4.6"
python-coveralls = "==2.9.0"
pytz = "==2015.6"
raygun4py = "==3.1.2"
"repoze.retry" = "==1.3"
requests = "==2.8.1"
sendgrid = "==2.2.1"
slacker = "==0.7.3"
SQLAlchemy-Enum34 = "==1.0.1"
SQLAlchemy-Utils = "==0.31.6"
SQLAlchemy = "==1.1.9"
typing = "==3.5.2.2"
twilio = "==5.6.0"
[...]

PR #3667, opened 2019-04-03
Added caching to improve pipenv-resolver performance

WoLpH:
### The issue

The `pipenv-resolver` performance was bugging me a bit so I had a few rounds with a profiler and added caching with a few of the heaviest calls. In my case these were doing over a 100k calls so even with little performance impact it still made a difference.

I suppose it's related to several of these: https://github.com/pypa/pipenv/search?q=performance&type=Issues but specifically this one: https://github.com/pypa/pipenv/issues/356

For what it's worth, this change made quite a lot of difference for me. My initial ~30 seconds `pipenv lock` time was decreased to about ~17 seconds. There might be more low-hanging fruit but this takes a stab at it :)

### The fix

It adds a few minor caches to a few key spots in the code which are called _very_ often according to some `cProfile` runs.

### The checklist

* [x] Associated issue

### If this is a patch to the `vendor` directory…

While these could be submitted to the upstream projects, I think they're mostly very specific to pipenv but feel free to disagree and I'll take care of it.

The case of the `Version()` modification could be modified from everywhere it's being called, but that's still within the `vendor` packages so it wouldn't help. It's a rather minor change here but a clean version upstream would be pretty invasive.


frostming:
Thanks for that patch. It is a great work, but IMO I am not willing to merge such big changes made to vendor libraries(`notpip` is `pip` indeed). This will bring extra workload when updating the vendor dependencies.  Thanks for understanding.

WoLpH:
I understand the reasoning, but I do think a similar solution would be beneficial for pipenv. It nearly halves the time needed for a fairly standard Django project for me.

While pipenv is really great, it's really too slow for comfort. I'll see if I can somehow patch something closer instead but I couldn't find anything quickly accessible that would have a meaningful effect.

WoLpH:
In lieu of changes upstream, would a monkeypatch for these two small sections be an option?

frostming:
@WoLpH Yes, if you have an elegant solution instead of hacking into the vendors I will be glad to review.

WoLpH:
@frostming I believe this is about as elegant as monkey patching can get :)
But it definitely helps a bit depending on the case. I think there's more to be done here but it's a start, if this would be accepted I can do a bit more profiling to see how far we can get.

image

If we want to get any really decent performance we'll have to implement multithreaded and/or async downloading though. Or perhaps even skipping the downloading of the release list and simply check if there's been updates.

WoLpH:
Ok, that last test confirms it. It's not my code, the tests are broken ;)

So please review regardless of the test status :)

WoLpH:
@frostming since a large portion of the time is now spent in downloading, an async multithreaded downloading approach could be very beneficial. That would require a similar edit in notpip and doesn't have to be all that invasive I think.

Simply replace the `_get_pages` method and we're done:
https://github.com/pypa/pipenv/blob/6c62d23cfcd1255b62bf7ef31b97caa5af492f46/pipenv/patched/notpip/_internal/index.py#L758-L774

Would a multithreaded version of that code be acceptable?

mungojam:
> a large portion of the time is now spent in downloading

We use a custom, internal pypi hosted on GitHub pages, so this really hits us as GitHub pages has a fixed cache-length of 10 minutes. It would be help us if the offline cache length for the .whl files could be overridden to be infinite. Otherwise we end up downloading big things like numpy over and over.

WoLpH:
@techalchemy do you have time to review this by any chance?

techalchemy:
I will be able to have a look this week, sorry for the very long delay while all the moving pieces were being pulled together

VonAncken:
Almost a year since the last activity, what is the current status?

WoLpH:
I've personally mostly stopped using pipenv. While I like the idea, it's far too slow to be usable for a decently sized project.

Even with all of these improvements it's only about twice as fast or so, but it needs to be about 10 times as fast to be feasible for me which would require a substantial rewrite.

matteius:
This would have to be revisited for modern python3 and modern pip -- perhaps there is something we can learn from this PR to provide some level of caching in the resolver -- but it would need to be redone.   I have a different performance PR out that greatly reduces the amount of system load to perform a batch install in similar or less wall time than the current version, but it doesn't effect the resolver:  https://github.com/pypa/pipenv/pull/5301

The biggest improvements I can see to the resolver without trying the monkeypatch pip internals, would be to get more support for fetching packages hashes from private pypis without having to download all of the wheels or other artifacts to hash them.

PR #3669, opened 2019-04-03
conftest.py: Ignore KeyboardInterrupt

jayvdb:
During check_internet() and check_github_ssh(), a KeyboardInterrupt
should be interpreted as user desire to escape the check, not escape
the entire test run.  This is especially true during check_github_ssh
which may require a passphrase from the user, which they might feel
uncomfortable giving during a test suite.

After these checks are bypassed, there is user feedback indicating the
tests are running, and so they can trigger KeyboardInterrupt again if
they wish to escape the entire test run.

### The issue

### The fix

### The checklist

* [ ] Associated issue


techalchemy:
Hmm makes sense to me; if it makes your life better it seems fine from my POV

Issue #3630, opened 2019-03-20
Tests using internet but not declared as such


jayvdb:
I am packaging pipenv for openSUSE, which means the tests must run without internet.

For the most part the internet detection and skipping works great, however I believe the following use the internet without having the `internet` marker.

- `test_download_file`
- `test_pipenv_clean_pip_no_warnings`
- `test_environment_variable_value_does_not_change_hash`
- `test_run_in_virtualenv`
- `test_lock_with_incomplete_source`

Only the first is in unit tests, and it is very clearly intentionally needing to use internet.  If the internet marker isnt suitable for unit tests, perhaps a simple requests mock would be sufficient?

The others are integration tests, and it is a bit harder to verify these are supposed to be using the internet, as they should probably be intercepted by pytest_pypi.

Also while on the subject, it would be worth skipping `test_system_and_deploy_work` if the location of system packages isnt writable, otherwise that also fails during rpmbuild as it tries to uninstall build dependencies such as `certifi` and fails with :
```
[pipenv.exceptions.InstallError]: ["Could not install packages due to an EnvironmentError: [Errno 13] Permission denied: '__init__.cpython-37.opt-1.pyc'", 'Consider using the `--user` option or check the permissions.']
```

frostming:
Thanks for reporting this, it would be great to improve our test suites and PRs are welcomed.

jayvdb:
Another five pipenv failures when offline documented at https://github.com/sarugaku/requirementslib/issues/152

jayvdb:
`test_pipenv_clean_pip_no_warnings` and `test_run_in_virtualenv` both using `run pip ..` , and I guess that pip is not instructed to use the fake simple index provided by pytest_pypi.

If so, that feels a bit like a bug actually, as the venv's pipenv has been told to use an alternative pypi index, and so pip inside of the venv "should" be using the same index.  I believe that is possible by setting envvars.  I'll let others chew on that, or toss it out if it is silly.

I guess `test_lock_with_incomplete_source` needs internet because it is using `test.pypi.org`.  If so, then the  test `test_lock_after_update_source_name` (not in the last release) will need the same treatment.

But I do not understand why `test_environment_variable_value_does_not_change_hash` uses the internet.  The test is supposed to be use the pypi fake index.

It fails at this assert
```py
              c = p.pipenv('install')
              assert c.return_code == 0
```

I've tried changing the package being installed, e.g. to tablib, and it doesnt appear to make any difference.

[...]

PR #3670, opened 2019-04-03
Avoid using requests unnecessarily in tests

jayvdb:
This replaces requests with tablib in tests where using requests
in a de-vendored environment causes the tests to fail unnecessarily.

Continuation of https://github.com/pypa/pipenv/issues/3644

### The issue

### The fix

### The checklist

* [ ] Associated issue

### DIFF
--- a/tests/integration/test_cli.py
+++ b/tests/integration/test_cli.py
@@ -84,45 +84,59 @@ def test_pipenv_rm(PipenvInstance):
 @pytest.mark.cli
 def test_pipenv_graph(PipenvInstance):
     with PipenvInstance() as p:
-        c = p.pipenv('install requests')
+        c = p.pipenv('install tablib')
         assert c.ok
         graph = p.pipenv("graph")
         assert graph.ok
-        assert "requests" in graph.out
+        assert "tablib" in graph.out
         graph_json = p.pipenv("graph --json")
         assert graph_json.ok
-        assert "requests" in graph_json.out
+        assert "tablib" in graph_json.out
         graph_json_tree = p.pipenv("graph --json-tree")
         assert graph_json_tree.ok
-        assert "requests" in graph_json_tree.out
+        assert "tablib" in graph_json_tree.out
 
 
 @pytest.mark.cli
 def test_pipenv_graph_reverse(PipenvInstance):
     with PipenvInstance() as p:
-        c = p.pipenv('install requests==2.18.4')
+        c = p.pipenv('install tablib==0.13.0')
         assert c.ok
         c = p.pipenv('graph --reverse')
         assert c.ok
         output = c.out
 
-        requests_dependency = [
-            ('certifi', 'certifi>=2017.4.17'),
-            ('chardet', 'chardet(>=3.0.2,<3.1.0|<3.1.0,>=3.0.2)'),
-            ('idna', 'idna(>=2.5,<2.7|<2.7,>=2.5)'),
-            ('urllib3', 'urllib3(>=1.21.1,<1.23|<1.23,>=1.21.1)')
-        ]
-
-        for dep_name, dep_constraint in requests_dependency:
-            dep_match = re.search(r'^{}==[\d.]+$'.format(dep_name), output, flags=re.MULTILINE)
-            dep_requests_match = re.search(r'^  - requests==2.18.4 \[requires: {}\]$'.format(dep_constraint), output, flags=re.MULTILINE)
-            assert dep_match is not None
-            assert dep_requests_match is not None
-            assert dep_requests_match.start() > dep_match.start()
-
         c = p.pipenv('graph --reverse --json')
         assert c.return_code == 1
         assert 'Warning: Using both --reverse and --json together is not supported.' in c.err
+
+        requests_dependency = [
+            ('backports.csv', 'backports.csv'),
+            ('odfpy', 'odfpy'),
+            ('openpyxl', 'openpyxl>=2.4.0'),
+            ('pyyaml', 'pyyaml'),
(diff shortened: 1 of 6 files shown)


jayvdb:
I have at least another test which should have this type of change, so I'll add it here shortly.

jayvdb:
Note in `test_uninstall.py` I have kept one `requests` uninstall, renamed to `test_uninstall_requests`, so that anyone devendoring will still have a failure to contend with, and a note there to indicate the cause is quite simply the de-vendoring.

At a latter stage, we need to better understand just how much of pipenv fails when de-vendoring.  Probably after the next release, when all distros re-do their de-vendoring, we can focus their attention on those failures which they like to skip and hope for the best.

I can think of a few basic tests which could be added, such as installing specific versions of `requests` and testing whether the requested version is visible inside the venv.  My guess is that also fails.  But that and other tests should be a separate set of analysis and separate PR.

jayvdb:
I believe this is all of the changes needed to resolve this issue.

The only other test failures I have are listed at https://github.com/pypa/pipenv/issues/3630 , and I am fairly confident they are not the same problem as this issue.

Im happy to revise this patch heavily based on feedback.  I probably havent made the best choices of alternative packages, and there are probably changes needed to match the local style, etc.

tests/integration/test_uninstall.py
```diff
@@ -44,90 +65,89 @@ def test_mirror_uninstall(PipenvInstance, pypi):
...
+        assert "django" not in p.lockfile["develop"]
+        assert p.lockfile["develop"] == {}
         # Ensure the --pypi-mirror parameter hasn't altered the Pipfile or Pipfile.lock sources
         assert len(p.pipfile["source"]) == 1
         assert len(p.lockfile["_meta"]["sources"]) == 1
         assert "https://pypi.org/simple" == p.pipfile["source"][0]["url"]
         assert "https://pypi.org/simple" == p.lockfile["_meta"]["sources"][0]["url"]
 
-        c = p.pipenv("run python -m requests.help")
+        c = p.pipenv("run python -m django --version")
```
jayvdb: Note Django was used here to continue using the `run python -m  ..` syntax, and there are not many packages in `tests/pypi/` that have a `__main__`.  We could add another lighter package with a `__main__` to the pypi archive .

Or this could be changed to using `run python -c 'import x'` which is almost as good, with a slightly less understandable error if the test fails.  But I think 99% of the people likely to encounter an error here are going to understand the backtrace caused by `import x` failing.

tests/integration/test_install_basic.py
```diff
@@ -119,16 +119,16 @@ def test_install_without_dev(PipenvInstance, pypi):
 six = "*"
 
 [dev-packages]
-pytz = "*"
+tablib = "*"
```
jayvdb: This replacing of `pytz` with `tablib` is a slightly different from the others, which are replacing `requests`.

pytz is not a dependency of pipenv, however it is a dependency of the test runner, so it is also added in `PYTHONPATH` in order to get the builds working in rpmbuild.

I can workaround this by creating a special venv with only the real dependencies of pipenv, and using that in `PYTHONPATH`.  But that is quite a bit more effort, and I doubt many distro maintainers would accept a much more complicated testing rig just so this one test passes - they would likely prefer to just mark this test as a skip.

Note I do not expect pipenv devs to always be avoiding requests and other dependencies in these tests.  It would be nice if they were avoided, especially requests, but it is not possible to predict in advance all of the dependencies, especially all of the test rig dependencies. 

I think the responsibility of creating fixup patches like this one falls on the de-vendor-ers after each release.

Also we should also be working on creating a job which blocks internet access so that the 'internet' marker becomes 100% reliable here -- so that de-vendor-ers dont feel that some of the failures they see might be due to 'no internet', and hopefully they investigate any failures more before shipping broken package updates to users.

techalchemy:
oops I triggered azure's abuse detection :o 

Sorry this took so long but there were a _lot_ of issues to untangle, once I can run the tests here I will address the specific comments.  Generally speaking I have no issues swapping out one library for another and if you have implementations for moving things offline I am more than happy to merge them.

jayvdb:
Glad to see pipenv has green builds again.  I'll keep a close eye on this PR and happy to change it as needed.

If I have spare time after this, my focus will be to create new versions of these tests which are written so they fail when de-vendoring and directly demonstrate the unsolved problems resulting from de-vendoring, with a pytest mark so they are easy to skip as a group.

frostming:
@jayvdb Please merge and we are good to go.

jayvdb:
ok, on it.

tests/integration/test_cli.py
```diff
@@ -84,46 +84,60 @@ def test_pipenv_rm(PipenvInstance):
...
         assert graph_json_tree.ok
-        assert "requests" in graph_json_tree.out
+        assert "tablib" in graph_json_tree.out
 
 
 @pytest.mark.cli
 def test_pipenv_graph_reverse(PipenvInstance, pypi):
     with PipenvInstance(pypi=pypi) as p:
-        c = p.pipenv('install requests==2.18.4')
+        c = p.pipenv('install tablib==0.13.0')
```
jayvdb: waiting on https://github.com/sarugaku/pipenv-test-artifacts/pull/1

frostming:
Thanks for the work, merging.

techalchemy:
Hey thanks again for all the effort on this, I know this was a lot of work

techalchemy:
@jayvdb if you have any notes on your decisionmaking process it might be helpful to begin updating our contributing guidelines to factor in your process into how we decide what to use on future tests

Issue #3621, opened 2019-03-15
possible rendundant vendor


PatrikKopkan:
Hi, I ' ve noticed that shutilwhich is for importing which function. At description at pypi is said that it is not needed in pythons 3.3 and above. From setup.py I asume you don't support versions below 3.3. https://github.com/pypa/pipenv/blob/d5e446ee7bd0e769c385331714c8616c6b6738b8/setup.py#L130
And it seems you dont import this vendor in your project.
https://github.com/pypa/pipenv/search?q=shutilwhich&unscoped_q=shutilwhich

Hope that the issue is meaningful

frostming:
@PatrikKopkan Thanks for reporting.

> I asume you don't support versions below 3.3.

But we need to support Python 2.7, don't we? I think it is there for Python 2.7.

PatrikKopkan:
Thank you for your response.

From reading your code I still have feeling that it is not actually used. Could you point me to where it is needed?

jayvdb:
No, it isnt needed for Python 2.7.

It is unused.  It was vendored due to pew, but that has been removed/unvendored by @uranusjr  in a5a9583 .  It was used by `pipenv/patched/pew/_utils.py` .

c.f. also https://github.com/pypa/pipenv/issues/3640

uranusjr:
I believe Pipenv still has a few places relying on the system `which` command (or `where` on Windows). We should convert those to `shutil.which` instead. cc @techalchemy

jayvdb:
There are a lot of `which` ... `system_which` should like the one you are referring to.  It falls back to [`pythonfinder`](https://github.com/sarugaku/pythonfinder)`.Finder` , which appears to do the job of `shutil.which`

```
grep -r def.*which pipenv-2018.11.26.1543215385.742ae1b9/pipenv/*.py
pipenv-2018.11.26.1543215385.742ae1b9/pipenv/core.py:def which(command, location=None, allow_global=False):
pipenv-2018.11.26.1543215385.742ae1b9/pipenv/core.py:def which_pip(allow_global=False):
pipenv-2018.11.26.1543215385.742ae1b9/pipenv/core.py:def system_which(command, mult=False):
pipenv-2018.11.26.1543215385.742ae1b9/pipenv/environment.py:    def which(self, search, as_path=True):
pipenv-2018.11.26.1543215385.742ae1b9/pipenv/project.py:    def __init__(self, which=None, python_version=None, chdir=True):
[...]

PR #3675, opened 2019-04-04
Remove unused vendored blindspin

jayvdb:
Closes https://github.com/pypa/pipenv/issues/3640

Thank you for contributing to Pipenv!

### The issue

What is the thing you want to fix? Is it associated with an issue on GitHub? Please mention it.

Always consider opening an issue first to describe your problem, so we can discuss what is the best way to amend it.  Note that if you do not describe the goal of this change or link to a related issue, the maintainers may close the PR without further review.

If your pull request makes a non-insignificant change to Pipenv, such as the user interface or intended functionality, please file a PEEP. 

    https://github.com/pypa/pipenv/blob/master/peeps/PEEP-000.md

### The fix

How does this pull request fix your problem? Did you consider any alternatives? Why is this the *best* solution, in your opinion?

### The checklist

* [ ] Associated issue
* [ ] A news fragment in the `news/` directory to describe this fix with the extension `.bugfix`, `.feature`, `.behavior`, `.doc`. `.vendor`. or `.trivial` (this will appear in the release changelog). Use semantic line breaks and name the file after the issue number or the PR #.

PR #3676, opened 2019-04-04
Remove unused vendored shutilwhich

jayvdb:
This package is no longer used.

Closes https://github.com/pypa/pipenv/issues/3621

### The issue

### The fix

### The checklist

* [ ] Associated issue


jayvdb:
c.f. https://github.com/pypa/pipenv/pull/3675 which magically breaks when `blindspin` is removed.

jayvdb:
I checked the code many times, and this dependency is removed in the de-vendored package now in the main openSUSE package collection.

Moreover, it is unlikely you want this [`shutilwhich`](https://pypi.org/project/shutilwhich/) because it automatically monkey patches the `which` function into the stdlib `shutil`, which is useful to workaround some third party software which isnt ported properly.  You would want to use the less intrusive [`backports.shutil_which`](https://pypi.org/project/backports.shutil_which/).

techalchemy:
ahhhh thanks for the info, that definitely makes sense as to why this seems to break my vendoring scripts.  Thanks for the fixes & context, this will actually fix my vendoring script as well

PR #3679, opened 2019-04-05
* Replace to getdefaultlocale because there is no getdefaultencoding

yong422:
### The issue

According to the existing code, locale_encoding of 302 line is fixed as ascii, and it does not follow system encoding value.
It can be a bug in an area that uses encoding in other languages ​​not English.

### The fix

replace locale.getdefaultlocale with locale.getdefaultencoding because locale.getdefaultencoding is not exist

### DIFF
--- a/pipenv/vendor/vistir/misc.py
+++ b/pipenv/vendor/vistir/misc.py
@@ -514,7 +514,7 @@ def chunked(n, iterable):
 
 
 try:
-    locale_encoding = locale.getdefaultencoding()[1] or "ascii"
+    locale_encoding = locale.getdefaultlocale()[1] or "ascii"
 except Exception:
     locale_encoding = "ascii"
 


frostming:
Oh, thanks for the patch. There is something wrong breaking the tests. Will merge after they are fixed. Don't worry, they are not caused by your change.

jayvdb:
@yong422 , you might like to also fix this upstream where this code is vendored from at https://github.com/sarugaku/vistir/blob/master/src/vistir/misc.py#L517 , so that it isnt lost when a new copy of vistir is released.

yong422:
@jayvdb Thank you for your good info.

techalchemy:
Thanks for the fix, I'm going to close this PR and merge the PR against vistir instead, we will get the fix when I pull in all of the changes to vistir

Thanks!

Issue #3727, opened 2019-05-04
Broken link in bug report template


ppo:
The link "diagnose documentation" to https://docs.pipenv.org/diagnose/ is not working.
Found in [.github/ISSUE_TEMPLATE/Bug_report.md](https://github.com/pypa/pipenv/blob/master/.github/ISSUE_TEMPLATE/Bug_report.md)

The doc seems to exist in [docs/diagnose.rst](https://github.com/pypa/pipenv/blob/master/docs/diagnose.rst)

techalchemy:
thanks for the report and the pr!

frostming:
Closed via #3728

PR #3728, opened 2019-05-05
fixed broken link for diagnose documentation

jadhavmanoj:

### The issue reported here #3727

The link "diagnose documentation" to https://docs.pipenv.org/diagnose/ is not working. 
The doc seems to exist in docs/diagnose.rst.
So changed url here .github/ISSUE_TEMPLATE/Bug_report.md

### The fix

changed URL here .github/ISSUE_TEMPLATE/Bug_report.md

### The checklist

* [x] Associated issue


.github/ISSUE_TEMPLATE/Bug_report.md
```diff
@@ -5,7 +5,7 @@ about: Create a report to help us improve
 
 Be sure to check the existing issues (both open and closed!), and make sure you are running the latest version of Pipenv.
 
-Check the [diagnose documentation](https://docs.pipenv.org/diagnose/) for common issues before posting! We may close your issue if it is very similar to one of them. Please be considerate, or be on your way.
+Check the [diagnose documentation](/docs/diagnose.rst) for common issues before posting! We may close your issue if it is very similar to one of them. Please be considerate, or be on your way.
```
techalchemy: So the documentation is actually available at https://docs.pipenv.org but there is nothing available at https://docs.pipenv.org/docs/diagnose/ because the subfolder `docs` doesn't exist.  The actual diagnose documentation is at https://docs.pipenv.org/en/latest/diagnose/

jadhavmanoj: Thanks for the review. I updated Diagnose URL as you mentioned. (https://docs.pipenv.org/en/latest/diagnose/)

techalchemy:
looks good, we only host english documentation so this should be fine.

Do note that after looking this over I did manage to figure out how to add the redirect for `/diagnose/` to point to the correct place:

image


So the regular link will also work now. We will need to remove the english specific path if we ever acquire a team of translators

PR #3738, opened 2019-05-09
Allows overriding `PIP_EXISTS_ACTION` environment variable (it was forced to `w`)

### LINKED ISSUE #3737: Option to override pip's --exists-action, or PIP_EXISTS_ACTION
Hello,

I'm trying to find a way to specify `--exists-action` for pip install command, but it seems there's no way to override it.

Here:  https://github.com/pypa/pipenv/blob/master/pipenv/core.py#L1506 you force `"PIP_EXISTS_ACTION": vistir.misc.fs_str("w")`

I need a way to set either `PIP_EXISTS_ACTION=i` or `--exists-action=i` (ignore a package if it's already installed)
This would make installation really really (really) fast, if the virtual environment exists.

We need this because we cache and restore `.venv`, and this option would make CI much faster :) 

Thanks

simion:
This should fix #3737


jxltom:
This is nice to speed up the installation process

pipenv/core.py
```diff
@@ -1503,7 +1503,7 @@ def pip_install(
         "PIP_DESTINATION_DIR": vistir.misc.fs_str(
             cache_dir.joinpath("pkgs").as_posix()
         ),
-        "PIP_EXISTS_ACTION": vistir.misc.fs_str("w"),
+        "PIP_EXISTS_ACTION": vistir.misc.fs_str(PIP_EXISTS_ACTION or "w"),
```
uranusjr: Is this a good idea? The line in `environments.py` sets `w` when the variable is missing, so maybe w should respect the user explicitly setting it to `` (empty string).

techalchemy: That would just mean using the pip defaults which seems fine to me, I don't think it matters much -- probably better to be consistent and do less hidden assumptions

simion: @uranusjr may be right, we should leave it to PIP, in case it gets overriden by empty string.
However, explicit is better than implicit.
Your choice.

uranusjr: Honestly I doubt whether anyone would care about this extreme edge case, so in the end it’s only a theoretical preference.

techalchemy: I'll be happy to merge this either way, planning to do so once CI passes

techalchemy:
Thanks for the pr :) would you mind adding a news entry to `news/3738.feature.rst` describing the new functionality?

simion:
> Thanks for the pr :) would you mind adding a news entry to `news/3738.feature.rst` describing the new functionality?

Done.

techalchemy:
thanks!

PR #3745, opened 2019-05-14
make check unused work

frostming:
Normalize the package names to lowercase when comparing used and in-Pipfile packages.


techalchemy:
Can you add a news entry for this?

Issue #3694, opened 2019-04-15
pipenv --py crashes


dlech:
### Issue description

Running `pipenv --py` in a project with no venv setup crashes.

### Expected result

No stack trace, just an error message.

### Actual result

```
$ pipenv --py
Traceback (most recent call last):
  File "/home/david/bin/pipenv", line 11, in <module>
    sys.exit(cli())
  File "/home/david/.local/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/home/david/.local/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/home/david/.local/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 1114, in invoke
    return Command.invoke(self, ctx)
  File "/home/david/.local/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/david/.local/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/home/david/.local/lib/python3.6/site-packages/pipenv/vendor/click/decorators.py", line 64, in new_func
    return ctx.invoke(f, obj, *args, **kwargs)
  File "/home/david/.local/lib/python3.6/site-packages/pipenv/vendor/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/home/david/.local/lib/python3.6/site-packages/pipenv/vendor/click/decorators.py", line 17, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/home/david/.local/lib/python3.6/site-packages/pipenv/cli/command.py", line 140, in cli
    do_py()
  File "/home/david/.local/lib/python3.6/site-packages/pipenv/core.py", line 1635, in do_py
    click.echo(which("python", allow_global=system))
  File "/home/david/.local/lib/python3.6/site-packages/pipenv/core.py", line 108, in which
    raise RuntimeError("location not created nor specified")
RuntimeError: location not created nor specified
```

### Steps to replicate

Using Ubuntu 18.04:
```
sudo apt update
sudo apt install python3-pip
pip3 install pipenv  # --user is implied so pipenv executable gets installed in /home/$USER/.local/bin/
git clone https://github.com/dlech/vscode-python-issue4866.git
[...]

PR #3753, opened 2019-05-20
Update core.py

Cologler:
for #3694


frostming:
IMO the error message is not so clear and needs to be capitalized. The same message as `pipenv --venv` should be just OK.

Cologler:
@frostming "location not created nor specified" was copied from current RuntimeError. This won't make break changes if some tools captured stderr and search the string.

erinxocon:
@Cologler Thanks for contributing to pipenv!  It will be nice to not have the whole stack trace print out.  I hope that I can work on the cause of the underlying problem so we won't have to see the error message 😅

Cologler:
@frostming any news for this?

frostming:
would you mind adding a news entry?

Cologler:
@frostming My English is poor 😂

PR #3758, opened 2019-05-23
Adds info on env variables

docdeveloper:
Adds info about using quotation marks if credentials contain special characters.
I added this piece of information because it took me some time to figure it out.


techalchemy:
Thanks for the PR, sorry for the delayed review -- we will get this reviewed shortly!

techalchemy:
LGTM, I swear somedays azure pipelines are so atrocious.

PR #3759, opened 2019-05-23
Doc: advanced.rst: Use "pytest" as the new recommended entrypoint for…

JordanP:
… Pytest

According to https://github.com/pytest-dev/pytest/issues/1629 the recommended entrypoint for `pytest` is `pytest`. Usage of `py.test` has been deprecated for 2 years now.

### The issue

### The fix

### The checklist

* [ ] Associated issue

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -467,7 +467,7 @@ and the corresponding Makefile::
         pipenv install --dev
 
     test:
-        pipenv run py.test tests
+        pipenv run pytest tests
 
 
 Tox Automation Project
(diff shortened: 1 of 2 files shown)


techalchemy:
Thanks for the PR!

Would you mind adding a news entry to `news/3759.doc.rst` which describes the change, e.g. 

```rst
Updated the documentation with the new ``pytest`` endpoint.
```

Thanks

JordanP:
@techalchemy thanks for your feedback. I've updated the doc just right now.

techalchemy:
Perfect, looks great to me, thanks a bunch for the update!

Issue #2241, opened 2018-05-22
Provide project root environment variable for run command


kristofferpeterhansel:
**Is your feature request related to a problem? Please describe.**
The `pipenv run` is really handy. But it requires a good deal of work to actually make it work for situations where your current directory is not the same as where the Pipenv file is located.

Just for something like using it as a shorthand for running your python application that requires setting the `PYTHONPATH` environment variable. It it is tricky to figure out what to actually set it to. And while it can be made easier by wrapping it in a shell script. Informing Pipenv where said script is located also becomes tricky

**Describe the solution you'd like**
The most simple solution I can think of would be having en environment variable (say, `PIPENV_PROJECT_ROOT`). And could be used in places like the scripts section of Pipfile. And in scripts run from it.

**Describe alternatives you've considered**
For my problem it would also have been nice with a way to have Pipenv update the `PYTHONPATH` environment variable to include a path relative to the project root. But just having the project root in the environment would likely solve more use cases

uranusjr:
There is `PIPENV_PIPFILE` which I think is basically `/path/to/project/root/Pipfile`. I don’t know if it works, but it could worth a try.

kristofferpeterhansel:
Ah right. I actually noticed `PIPENV_PIPFILE`and then forgot about it again when I was looking for a way around this.

 But it is not solving the problem of running in-project scripts defined `Pipfile` if they are referring to a relative path. So maybe it is more a question of always running those commands form the root of the project?

kennethreitz:
I'm +1 to setting this environment variable in `environments.py`.

AlJohri:
@kristofferpeterhansel that's the approach I take- run all scripts and commands from the same directory as the Pipfile (project root). `./scripts/myscript.py` or `pipenv run myscript`

I found everything becomes a _bit_ easier that way

kristofferpeterhansel:
>  that's the approach I take- run all scripts and commands from the same directory as the Pipfile (project root). ./scripts/myscript.py or pipenv run myscript

Well yes. But I am not sure I can think of a reason why pipenv wouldn't just always run the scripts with the project root as working directory to make this consistent.

uranusjr:
My guess is that it would make things like `pipenv run python -c "import os; print(os.listdir())"` very confusing. If there should be consistency, it is probably never changing the working directory, instead of always changing…

kristofferpeterhansel:
No. They do not. 

I have entries like these. That will only run if my CWD is in the project root
[...]

PR #3775, opened 2019-05-30
Populate PIPENV_PIPFILE environment variable if empty

ccope:
This sets the PIPENV_PIPFILE variable in the environment, which allows me to insert the root diretory into my PYTHONPATH, re: #2241. It would be better if the directory was directly available as an env variable, but this seemed like it would be the most minimal change.

### The checklist

* [x] Associated issue


ccope:
Note that while it's possible to set environment variables via `.env`, this does not satisfy my usecase. I need a reliable absolute path regardless of where the user checks out the code, or if their CWD is a child inside the project directory.

PR #3820, opened 2019-06-28
Find the python interpreter for pipenv check --system.

### LINKED ISSUE #3818: pipenv clean missing `--system` parameter
you can do `pipenv install --system` and `pipenv check --system` however there is no `pipenv clean --system` to go before a `pipenv install --system`

### LINKED ISSUE #3819: pipenv check --system finds wrong python ver yielding `JSONDecodeError`
if you run `pipenv check --system` on a system which only has `python3` and not `python` binary, you'll get a `JSONDecodeError` due to an empty response from `which -a python`.  Instead with `--system` it should just use `sys.executable`

frostming:
Fixes #3819  

### The Fix

Iterate over all possible python executables, instead of using `which("python", allow_global=True)`. This is to keep consistent with what `which_pip` does when `system=True`.

Pipenv is usually installed as an isolated application. By this method, `sys.executable`, which is the path of the isolated python interpreter, won't be preferred in the first place.

### DIFF
--- a/news/3819.bugfix.rst
+++ b/news/3819.bugfix.rst
@@ -1,0 +1,1 @@
+``pipenv check --system`` should find the correct Python interpreter when ``python`` does not exist on the system.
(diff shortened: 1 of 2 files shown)


thehesiod:
dang that was quick ! :)

PR #3843, opened 2019-07-13
Make sure symlinks are resolved.

### LINKED ISSUE #3842: Symlinks are not resolved properly to check location.
https://github.com/pypa/pipenv/blob/237034018a8a2053b0b3b744623ef7f87694fb0e/pipenv/project.py#L50-L54

https://github.com/pypa/pipenv/blob/237034018a8a2053b0b3b744623ef7f87694fb0e/pipenv/environment.py#L508-L519

The path is never resolved if it is an absolute path with symlinked parts. It is often the case with Linux, where the virtualenv home is symlinked to `~/.local/share/`, then `dist_is_in_project` will always return false.

Remove the `is_absolute` check in `_normalize` funciton can make it work.

More info FWIW: I have `virtualenvwrapper` installed and `WORKON_HOME` set to `~/.virtualenvs`

frostming:
### The fix

Fixes #3842 

### The checklist

* [x] Associated issue

### DIFF
--- a/news/3842.bugfix.rst
+++ b/news/3842.bugfix.rst
@@ -1,0 +1,1 @@
+Resolve the symlinks when the path is absolute.
--- a/pipenv/project.py
+++ b/pipenv/project.py
@@ -47,11 +47,10 @@ def _normalized(p):
     if p is None:
         return None
     loc = vistir.compat.Path(p)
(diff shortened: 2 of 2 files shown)


techalchemy:
You need to be careful with this code as it has been known to break quite significantly

Issue #2843, opened 2018-09-10
`pipenv run --help` crashes


boustrophedon:
##### Issue description
Pipenv 2018.7.1

In any directory (either containing an existing Pipfile or not), running `pipenv run --help` crashes.

##### Expected result

The help text for the run command.

##### Actual result

https://gist.github.com/boustrophedon/c396bf9e6c287f818f9b2ddf23ecebc7

##### Steps to replicate

Provide the steps to replicate (which usually at least includes the commands and the Pipfile).

run the command `pipenv run --help`

-------------------------------------------------------------------------------
pipenv --support: 
https://gist.github.com/boustrophedon/b68b539241d0ac5fd4aa0974eaf39c89

frostming:
Closed via #3844

boustrophedon:
Thank you!

Issue #2127, opened 2018-05-03
`pipenv run --help` thows excption


halloleo:
In an empty directory (where isn't a `pipfile` in any of the searched parents) issue the command

    pipenv run --help

and you get:

    Creating a virtualenv for this project…
    Using /usr/local/bin/python3.6m (3.6.4) to create virtualenv…
    ⠋Running virtualenv with interpreter /usr/local/bin/python3.6m
    Using base prefix '/usr/local'
    New python executable in /home/me/tmp/tepiru/.venv/bin/python3.6m
    Also creating executable in /home/me/tmp/tepiru/.venv/bin/python
    Installing setuptools, pip, wheel...done.
    
    Virtualenv location: /home/me/tmp/tepiru/.venv
    Creating a Pipfile for this project…
    Traceback (most recent call last):
      File "/usr/local/bin/pipenv", line 11, in <module>
        sys.exit(cli())
      File "/usr/local/lib/python3.5/site-packages/pipenv/vendor/click/core.py", line 722, in __call__
        return self.main(*args, **kwargs)
      File "/usr/local/lib/python3.5/site-packages/pipenv/vendor/click/core.py", line 697, in main
        rv = self.invoke(ctx)
      File "/usr/local/lib/python3.5/site-packages/pipenv/vendor/click/core.py", line 1066, in invoke
        return _process_result(sub_ctx.command.invoke(sub_ctx))
      File "/usr/local/lib/python3.5/site-packages/pipenv/vendor/click/core.py", line 895, in invoke
        return ctx.invoke(self.callback, **ctx.params)
      File "/usr/local/lib/python3.5/site-packages/pipenv/vendor/click/core.py", line 535, in invoke
        return callback(*args, **kwargs)
      File "/usr/local/lib/python3.5/site-packages/pipenv/cli.py", line 637, in run
        do_run(command=command, args=args, three=three, python=python)
      File "/usr/local/lib/python3.5/site-packages/pipenv/core.py", line 2303, in do_run
        do_run_posix(script, command=command)
      File "/usr/local/lib/python3.5/site-packages/pipenv/core.py", line 2283, in do_run_posix
        os.execl(command_path, command_path, *script.args)
      File "/usr/local/lib/python3.5/os.py", line 583, in execl
        execv(file, args)
    FileNotFoundError: [Errno 2] No such file or directory
    
I use the follwoing PIPENV variables:

    PIPENV_DEFAULT_PYTHON_VERSION=3.6
[...]

Issue #1405, opened 2018-02-07
`pipenv run` should support `-h` and `--help`


nodakai:
##### Describe your environment

pipenv, version 9.0.3

##### Expected result

Pipenv should print a help text explaining how to use the `pipenv run` sub-command. This should be done regardless of whether `pipenv run` accepts a complex family of options which is worth a help text or not, in order to make it consistent with other Pipenv sub-commands which all accepts `-h` / `--help`.

##### Actual result

```
$ pipenv run -h
Creating a virtualenv for this project…
⠋Using base prefix '/usr'
New python executable in /home/ec2-user/.local/share/virtualenvs/a-EouiYCWm/bin/python3
Also creating executable in /home/ec2-user/.local/share/virtualenvs/a-EouiYCWm/bin/python
Installing setuptools, pip, wheel...done.

Virtualenv location: /home/ec2-user/.local/share/virtualenvs/a-EouiYCWm
Creating a Pipfile for this project…
Error: the command -h could not be found within PATH.
```

##### Steps to replicate

See above

jrbourbeau:
It looks like `pipenv run --help` was disabled in #940. Although, there wasn't a specific reason given for disabling it. 

I agree with @nodakai that it's a little strange the other `pipenv` commands (e.g. `open`, `lock`, `shell`, etc.) have a help message displayed with `--help` / `-h`, but the `run` command doesn't.

kennethreitz:
run passes everything to what you're trying to run...

PR #3844, opened 2019-07-14
run: re-enable --help option

immerrr:
This PR re-enables `--help` option for `pipenv run` command.

### The issue

There are a few issues related to this problem:
- https://github.com/pypa/pipenv/issues/1405
- https://github.com/pypa/pipenv/issues/2127
- https://github.com/pypa/pipenv/issues/2843

As you see, it is a reoccurring confusion that `pipenv run` command does not support `--help`. It was disabled in #940 and there was no particular motivation referenced. I believe it deserves to be enabled back.

It is brought up by different members the core team that this is working "as intended" because pipenv forwards everything through to the subcommand, but it is not entirely true, e.g. `pipenv run` for example accepts all virtualenv creation parameters:

```
$ pipenv run --python=3 which python
Pipfile: /home/immerrr/Pipfile
Using /home/immerrr/src/pyenv/versions/3.7.0/bin/python3.7m (3.7.0) to create virtualenv...
⠙ Creating virtual environment...Using base prefix '/home/immerrr/src/pyenv/versions/3.7.0'
New python executable in /mnt/extraspace/virtualenvs/immerrr-8YhW2Khy/bin/python3.7m
Not overwriting existing python script /mnt/extraspace/virtualenvs/immerrr-8YhW2Khy/bin/python (you must use /mnt/extraspace/virtualenvs/immerrr-8YhW2Khy/bin/python3.7m)
Installing setuptools, pip, wheel...
done.
Running virtualenv with interpreter /home/immerrr/src/pyenv/versions/3.7.0/bin/python3.7m

Virtualenv location: /mnt/extraspace/virtualenvs/immerrr-8YhW2Khy
/mnt/extraspace/virtualenvs/immerrr-8YhW2Khy/bin/python
```
 
Or even `--verbose`:
```
$ pipenv run --verbose which python
/mnt/extraspace/virtualenvs/immerrr-8YhW2Khy/bin/python
```

The general convention that makes sense to me is that pipenv options end with the first cmdline argument that does not start with a dash. It is also reaffirmed that `pipenv run` is supposed to receive options in the usage generated by click: 
```
$ pipenv run
```

And if you really need to run a command that starts with a dash, you have a `--` separator convention that has you covered:

```
$ pipenv run -- --verbose which python
Error: the command --verbose could not be found within PATH or Pipfile's [scripts].
```

Hence I don't think it makes much sense to keep `--help` disabled maintaining the discrepancy between `run` and the rest of pipenv commands.

### The fix

This PR re-enables `--help` option for `pipenv run` command.

### The checklist

* [x] Associated issue

### DIFF
--- a/news/3844.behavior.rst
+++ b/news/3844.behavior.rst
@@ -1,0 +1,1 @@
+Re-enable ``--help`` option for ``pipenv run`` command.
--- a/pipenv/cli/command.py
+++ b/pipenv/cli/command.py
@@ -392,7 +392,6 @@ def shell(
 
 
 @cli.command(
-    add_help_option=False,
     short_help="Spawns a command installed into the virtualenv.",
     context_settings=subcommand_context_no_interspersion,
 )


techalchemy:
I’m fairly confident this was removed for a reason, you’ll have to check the commit history or the PR that changed it to verify

techalchemy:
Ah I see you referenced a few issues. The topics at hand would have been whether to pass `allow_interspersed_arguments` to subcommands, specifically to `pipenv run` we disable this because (and I am 100% certain we have fielded issues on this directly) users often want to pass arguments to the thing they are invoking with `pipenv run`. 

If we set the parser to parse pipenv arguments, it becomes impossible to pass the `--help` argument to anything invoked via `pipenv run`, for instance. Since there just isn’t much to know about `pipenv run`, the tradeoff is relatively minor. Changes to this behavior will break backwards compatibility— I haven’t reviewed this PR yet but if that’s what it does it most likely will be closed

immerrr:
@techalchemy 

re: `allow_interspersed_arguments`: yes, it makes sense to disable interspersed args for `run`, you want as little magic as possible when forwarding command line arguments.

> If we set the parser to parse pipenv arguments <...>

It is not a matter of "if we enable this", it already happens, the parser parses pipenv parameters, e.g. `--python` or  `--verbose`, as I have shown in the OP.

> it becomes impossible to pass the --help argument to anything invoked via pipenv run, for instance

It will not become impossible. Firstly, because by default pipenv options end with the first non-dash argument, which is usually the name of the command:

```
$ pipenv run echo echo receives --help even without the -- separator
echo receives --help even without the -- separator
``` 

Secondly, because you can force to interpret something that looks like an option as a command with the `--` separator:
```
immerrr@mmrcomp:~/tmp$ pipenv run --help

  Spawns a command installed into the virtualenv.

Options:
  --clear             Clears caches (pipenv, pip, and pip-tools).  [env var:
                      PIPENV_CLEAR]
immerrr@mmrcomp:~/tmp$ pipenv run -- --help
Error: the command --help could not be found within PATH or Pipfile's [scripts].
```

So the only backward incompatibility that would be seen by the users is if they actually had a command called `--help`, which should be very rare and can be easily forced to be interpreted as a command by the `--` separator convention.

techalchemy:
awesome, thanks for the clarification then.  This is probably good to go and will be merged shortly!

boustrophedon:
Thanks for working on this @immerrr !

PR #3845, opened 2019-07-15
Minor Doc improvements

frostming:


frostming:
It's strange that ignore settings in `azure-pipelines.yml` seem not working. 😞

Issue #3846, opened 2019-07-15
Issue Resolving Dependencies with graphql-core


themotu:
Pipenv believes one of my dependencies wants graphql-core<2 but that's incorrect therefore it won't install. `--skip-lock` does work then unable to generate a lock file.

`ERROR: ERROR: Could not find a version that matches graphql-core<2,<3,>=0.5.0,>=2.1,>=2.1.0`

Pipfile

```
[[source]]
name = "pypi"
url = "https://pypi.org/simple"
verify_ssl = true

[dev-packages]

[packages]
graphene-django = ">=2.0"
ipython = "*"
ipdb = "*"
Django = "*"
graphene = "*"
django-graphql-jwt = {git = "https://github.com/themotu/django-graphql-jwt.git", editable = true, ref = "user-by-id"}
psycopg2-binary = "*"
uwsgi = "*"
django-cors-headers = "*"
django-request-logging = "*"
WeasyPrint = "*"
```

`pipenv lock`

```
Locking [dev-packages] dependencies…
Locking [packages] dependencies…
⠙ Pinning VCS Packages...Branch 'user-by-id' set up to track remote branch 'user-by-id' from 'origin'.
Switched to a new branch 'user-by-id'
✘ Locking Failed! 
Switched to a new branch 'user-by-id'
Branch 'user-by-id' set up to track remote branch 'user-by-id' from 'origin'.
[pipenv.exceptions.ResolutionFailure]:       req_dir=requirements_dir
[pipenv.exceptions.ResolutionFailure]:   File "/usr/local/lib/python3.6/dist-packages/pipenv/utils.py", line 726, in resolve_deps
[pipenv.exceptions.ResolutionFailure]:       req_dir=req_dir,
[pipenv.exceptions.ResolutionFailure]:   File "/usr/local/lib/python3.6/dist-packages/pipenv/utils.py", line 480, in actually_resolve_deps
[pipenv.exceptions.ResolutionFailure]:       resolved_tree = resolver.resolve()
[pipenv.exceptions.ResolutionFailure]:   File "/usr/local/lib/python3.6/dist-packages/pipenv/utils.py", line 395, in resolve
[pipenv.exceptions.ResolutionFailure]:       raise ResolutionFailure(message=str(e))
[...]

PR #3848, opened 2019-07-15
The PEEP layout seems to have got confused.

willrogers:
I think this resolves it.

### The issue

There are 6 PEEPs, but one is named PEEP-0004.md unlike the others PEEP-000.md to PEEP-005.md. That means it appears out of order.

Also, inside the file it was called PEEP-003 but there is another PEEP-003.

### The fix

Renamed the file to PEEP-004.md and changed the title inside the file to PEEP-004.

### DIFF
--- a/peeps/PEEP-004.md
+++ b/peeps/PEEP-004.md
@@ -1,4 +1,4 @@
-## PEEP-003: Subcommands
+## PEEP-004: Subcommands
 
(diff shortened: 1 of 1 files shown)


frostming:
It's already addressed in #3846 , thanks anyway

willrogers:
I guess you mean #3845.

Issue #3651, opened 2019-03-27
A way to specify build-dependencies


con-f-use:
Pipenv might need a way to specify built-dependencies and tooling packages that need to be installed in order for pipenv to do it's job correctly.

## Example

In my company we use a wrapper around setuptools to simplify writing `setup.py` scripts for our packaging. Let's call this wrapper `stwrapper`.

A typical `setup.py` would look like this:

```
# setup.py for "package-that-uses-stwrapper-in-its-setup-py"
from stwrapper import setup

setup(
   name='package-that-uses-stwrapper-in-its-setup-py',
   # ... some more arguments
)
```
Where the function `stwrapper.setup` parses arguments, does some extra work and then calls the real `setuptools.setup`. Effectively, the `stwrapper` package is sort of a built dependency for us.

Lets say we have the following `Pipfile`:
```
[dev-packages]
stwrapper = "*"
package-that-uses-stwrapper-in-its-setup-py = { path = ".", editable = true, setup_requires=["stwrapper"] }
```
and do a 
```
pipenv lock --verbose
```
then, with the current version of Pipenv, we get:
```
Traceback:
...
pipenv.patched.notpip._internal.exceptions.InstallationError: Command "python setup.py egg_info" failed with error code 1 in '.'
```
Because `stwarpper` is not installed at the time pipenv tries to parse the `setup.py` of `package-that-uses-stwrapper-in-its-setup-py`. That is even the case, when `stwrapper` **is installed** in the **system's python** installation. Pipenv wants to have it **inside the new virtual environment it creates**. It's a Catch-22.

Our current workaround is to let pipenv create the virutal environment. Then do a `pipenv run pip install stwrapper` inside it and then `pipenv lock`. That way it works.

Is there a solution for this already, that I'm not aware of?

## Possible Solutions

 - Pipenv gets an equivalent to `setup_requires` in setup files or installs explicitly specified `setup_requires`
 - Pipenv uses the system python's installation of `stwrapper` (I'm aware of the `--site-packages` option but that may have unwanted side effects). 
 - We can specify packages in the `requires` section of a pipfile that need to be installed before Pipenv does any work, i.e. before locking. 
    ```
    [requires]
    python_version = "2.7"
    stwrapper = "*"
    ```
## Scope

I guess the problem applies to all projects that use non-standard-libaray packages in their `setup.py` files and/or use setuptools entry points to modify its behavior. You can not put such dependencies in the Pipfile, because they are needed for successfully locking.

[...]

PR #3859, opened 2019-07-18
Create PEEP-006.md

con-f-use:
See
 - https://github.com/pypa/pipenv/issues/3651

### DIFF
--- a/peeps/PEEP-006.md
+++ b/peeps/PEEP-006.md
@@ -1,0 +1,96 @@
+# PEEP-006: Pass extra pip arguments on a per-entry basis
+
+This PEEP proposes the introduction of an argument `extra_pip_options` in Pipfile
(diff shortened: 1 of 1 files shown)


frostming:
IMHO this looks a bit awkward and may break the reproducibility of the Pipfile. For example, the option `--curl-dir=${CURL_CONFIG_PATH}` may not be usable on the production server. For some globally applied options, do we have to duplicate them on each package?

I know the concern and problem you are seeking to solve behind this, but maybe we could have a better solution.

con-f-use:
I'm open to different solutions, that's just the best I could come up with. You are welcome to present one. The way I see it, reproducibility is already broken for these cases.

matteius:
@con-f-use by chance -- we just merged in a PR that does basically this:  https://github.com/pypa/pipenv/pull/5283

PR #3877, opened 2019-07-31
Allow passing through safety args, to work around pyup unreliability

lahwran:
### The issue

https://github.com/pypa/pipenv/issues/2825

This issue has not been fully solved by the ability to provide a key - pyup sometimes goes down, and it usually takes a long time. Since `pipenv check` is the only would-be-fast way to test if a `pipenv install` is needed, this limits its usability.

### The fix

This pull request adds a couple more arguments to pass through to safety, exposing its existing configuration options. It also adds an option to completely skip invoking `safety`, for cases were package installation checks and package security checks need to be run separately.

Because it's a trivial change, I also changed the return code to `2` on `safety` failure.

### DIFF
--- a/news/2825.trivial.rst
+++ b/news/2825.trivial.rst
@@ -1,0 +1,1 @@
+Allow skipping `safety` check, or configuring it to use a local database.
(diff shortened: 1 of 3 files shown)


lahwran:
I was distracted when I wrote this and I both didn't write it very well and didn't notice. My bad! I'll leave it here as inspiration.

Issue #2318, opened 2018-06-07
pipenv basic tutorial should link to valid values of python_version


Julian:
Reading [the basic usage tutorial](https://docs.pipenv.org/basics/) through for the first time mentions `python_version`, but doesn't link to what the acceptable values are for that option.

Concretely, since my project uses PyPy, I put `pypy` in there, which appears to "work" in the sense that the virtualenv was created using PyPy when then running `pipenv install`, but I had to guess whether that was actually a thing.

techalchemy:
PyPy isn't actually a valid python version that can be enforced afaik.  But this is a good point, I probably implemented a bunch of this and I actually have no idea what is valid here.

Julian:
To be clear -- is the `python_version` here just referring to the PEP 508 thing called python_version?

Julian:
Yeah OK, from looking at what module parses that value, which is literally called `pep508parser`, I guess that's a "yes" :)

uranusjr:
That is correct!

PR #3936, opened 2019-09-07
Doc: Add python_version value

tapaswenipathak:
### The issue

Fixes https://github.com/pypa/pipenv/issues/2318.

### The fix

Defined value of `python_version`.

### The checklist

* [X] Associated issue
* [X] A news fragment in the `news/` directory to describe this fix with the extension `.bugfix`, `.feature`, `.behavior`, `.doc`. `.vendor`. or `.trivial` (this will appear in the release changelog).


docs/basics.rst
```diff
@@ -125,7 +125,7 @@ Example Pipfile.lock
 
 - Generally, keep both ``Pipfile`` and ``Pipfile.lock`` in version control.
 - Do not keep ``Pipfile.lock`` in version control if multiple versions of Python are being targeted.
-- Specify your target Python version in your `Pipfile`'s ``[requires]`` section. Ideally, you should only have one target Python version, as this is a deployment tool.
+- Specify your target Python version in your `Pipfile`'s ``[requires]`` section. Ideally, you should only have one target Python version, as this is a deployment tool. ``python_version`` can be ``>`` Python 2.7 and ``<`` Python 3.7.
```
frostming: This line is confusing, people may think he should define `python_version = ">2.7"` but we do not support this.

We shouldn't constrain the value of `python_version`, it should be working for any Python version that virtualenv supports, and no upper bounds for future python versions.

tapaswenipathak: Can you please write what you think you know as proper statement? Did you read the ticket, ticket clearly says the user didn't know the supported Python version numbers, right?

frostming: I think the point is user doesn't know the valid value format of that field.
For `python_version`, `X.Y` is the proper value, without other binary operators, for `python_full_version`, it is `X.Y.Z`

tapaswenipathak: Perfect.

news/3913.doc.rst
```diff
@@ -0,0 +1 @@
+Define python_version value.
```
frostming:
```suggestion
Clarify the proper value of `python_version`.
```

tapaswenipathak: One more variable has been documented, can't accept the edit, ignoring.

tapaswenipathak:
@frostming: anything remaining?

tapaswenipathak:
@frostming: anything remaining?

frostming:
@tapaswenipathak No, thanks for contributing

Issue #3919, opened 2019-08-28
support '&&' in pipfile [scripts]


Madoshakalaka:
`command_a && command_b` is surpported both on windows cmd and unix terminals
```
# Pipfile
[scripts]
release = 'python setup.py sdist && python -m twine upload dist/*'
```
```
$ pipenv run release
invalid command name '&&'
```

Madoshakalaka:
#2160

PR #3942, opened 2019-09-07
Pipfile: Added feature for enabling && character

tapaswenipathak:
### The issue

Fixes https://github.com/pypa/pipenv/issues/3919.

### The fix

Feel free letting me know if you require any behavioral changes or tests for the added feature.

### The checklist

* [x] Associated issue
* [x] A news fragment in the `news/` directory to describe this fix with the extension `.bugfix`, `.feature`, `.behavior`, `.doc`. `.vendor`. or `.trivial` (this will appear in the release changelog).

### DIFF
--- a/news/3916.doc.rst
+++ b/news/3916.doc.rst
@@ -1,0 +1,1 @@
+Added a feature for enabling && for multiple executions.
--- a/tasks/release.py
+++ b/tasks/release.py
@@ -111,6 +111,9 @@ def release(ctx, dry_run=False):
         artifacts = list(ROOT.joinpath("dist").glob(dist_pattern))
         filename_display = "\n".join(f"  {a}" for a in artifacts)
         log(f"Would upload dists: {filename_display}")
+        if dist_pattern.contains('&&')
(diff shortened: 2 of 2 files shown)


frostming:
Your fix doesn't relate to the original issue. In fact `tasks/` folder has nothing to do with the `pipenv` executable.

Thanks for your effort but I would like to reject this now. Feel free to open another PR when you have a true fix.

tapaswenipathak:
@frostming: The time when release command is run having multiple commands both commands should run. What you think is incorrect? Can you please tell me the correct directory?

frostming:
@tapaswenipathak `tasks/release.py` is a dev script for releasing the project, and under dry run mode no actions should be taken.

The original issue is complaining about the `pipenv run [script_name]`, where `script_name` is defined in `Pipfile`, which doesn't support compound commands that contain `&&`. The correct file is `pipenv/core.py:do_run` and further `pipenv/cmdparse.py`

tapaswenipathak:
I see, nice!

PR #3985, opened 2019-10-10
Resolve dependencies only for the artifact.

kleschenko:
### The issue

Pipenv tries to resolve dependencies when installing from the lock-file if there is at least one "artifact" in that list. It is at least one of the issues causing https://github.com/pypa/pipenv/issues/1356

### The fix

Try to resolve dependencies only for the artifact

### The checklist

* [ ] Associated issue

### DIFF
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -718,7 +718,6 @@ def batch_install(
     if sequential_deps is None:
         sequential_deps = []
     failed = not retry
-    install_deps = not no_deps
     if not failed:
         label = INSTALL_LABEL if not environments.PIPENV_HIDE_EMOJIS else ""
     else:
@@ -738,6 +737,7 @@ def batch_install(
     trusted_hosts = []
     # Install these because
     for dep in deps_list_bar:
+        install_deps = not no_deps
         extra_indexes = []
         if dep.req.req:
             dep.req.req = strip_extras_markers_from_requirement(dep.req.req)
@@ -754,7 +754,6 @@ def batch_install(
             is_artifact = True
         if not project.s.PIPENV_RESOLVE_VCS and is_artifact and not dep.editable:
             install_deps = True
-            no_deps = False
 
         with vistir.contextmanagers.temp_environ():
             if not allow_global:


pipenv/core.py
```diff
@@ -713,6 +712,7 @@ def batch_install(deps_list, procs, failed_deps_queue,
     trusted_hosts = []
     # Install these because
     for dep in deps_list_bar:
+        install_deps = not no_deps
```
techalchemy, line 715: so this decision was effectively being overridden? I'm trying to reason about how that would cause the issue in question but I'm not really sure I understand yet

techalchemy:
I'm trying to understand what is happening here -- do you have some ideas about what the underlying problem is? If you run pipenv from the master branch (or from test pypi) do you still have the same issue?

pipenv/core.py
kleschenko: IIRC, the problem occurred when there was at least one dependency fulfilling this condition `not PIPENV_RESOLVE_VCS and is_artifact and not dep.editable` (installed with `git://`?) and all the consecutive deps were failing due to this

kleschenko: @techalchemy and as far as I can see the current `master` is still affected by this issue

techalchemy: `is_artifact` is true for anything that is like a `zip` or `whl` or a source control path that is not in editable mode.

techalchemy: ah sorry I didn't have the code expanded for my first comment, I can see the issue here -- `install_deps` is what we actually pass to `pip_install`, but it gets re-defined under specific conditions (i.e. if `not PIPENV_RESOLVE_VCS and is_artifact and not dep.editable`).

Because it is sometimes altered during a pass of the loop, you chose to reset it at the start of each iteration, which makes sense., and git rid of the now-unused `no_deps` variable. This change makes a lot of sense to me.

matteius:
@kleschenko Are you still working on this one?

kleschenko:
@matteius no, I haven't been able to set up the testing env locally to add a new test when I tried it.

matteius:
I have created a replacement PR for this:  https://github.com/pypa/pipenv/pull/5164

Issue #1760, opened 2018-03-16
new 'pre' syntax in Pipfile


gsemet:
Placeholder ticket for the proposition we talk with Kenneth.

Add the following syntax in the `Pipfile`:
```
...
mypkg = { version = "*", pre="true"}
...
```
This would allow to allow prerelease resolution for one package only.

Change the `pipenv install mypkg --pre` behavior. It will set the setting described above only for this very package. Today it turns on a general 'pre' setting for the whole Pipfile, leading other packages to also resolve to prerelease.

Change the `pipenv update` and `pipenv lock` behavior. They should take into account the presence of the `pre=true` setting for some package.

Here are some impact I think might happen
- if a version of package is describe at `>=1.0` in one dependency, and another package define the `pre=true` while a new version 2.0.0.b1 is available on pypi, `pipenv install` should take it
- I don't see any syntax to turn the general `pre` setting in the Pipfile. Maybe we can document users need to do a `pipenv update --pre` or `pipenv lock --pre` to turn the general 'pre' settings on.

gsemet:
I am assigning myself to this task, but I might take some time to enter into the code

techalchemy:
@frostming how does this proposal interact with the changes you made here recently?

frostming:
The change in my PR is: when installing wildcard versions from CLI, Pipfile won't be modified.

With the proposal here, when we do `pipenv install mypkg --pre`,  we are willing to replace the entry, right? Then we need to add extra arg `pre` to `add_package_to_pipfile`.

I am glad to help when it comes true.

Gr1N:
Hello! Any updates or estimates on this?

remingtonc:
If there's no progress on this I'd like to make it happen, otherwise I will revert to virtualenv/pip.

@frostming With your proposal, how would that end up looking in the Pipfile? I'm not familiar with how requirements are really parsed. Completely ballparking, please correct me. Would it have to follow syntax akin to:
`"0.1.0" = {git = "https://github.com/einarnn/pyang-plugins", pre = true}`
or akin to "extras" specification?
`black[pre] = "*"`
or in another section other than the packages entirely?

jxltom:
I found that pin to specific prerelease version works for even ```allow_prereleases``` or ```--pre``` is provided

ekhaydarov:
@gsemet where are you with making this a reality? Would be happy to help push this along

ajenkins-cargometrics:
[...]

PR #3999, opened 2019-10-22
Add initial support for the "pre" option in Pipfile

kleschenko:
### The issue

This PR adds initial support for the `pre` Pipfile option discussed in https://github.com/pypa/pipenv/issues/1760

It touches a lot of vendored dependencies and I'm not sure that this is the right way to do it but seems like this is the smallest possible change.

### The checklist

* [ ] Associated issue


b-karl:
Would love for this to be fixed. One suggestion: can you change the name of the flag to `allow_prereleases` like the global flag and make it more verbose?

matteius:
@kleschenko are you still interested in working on this?  I am thinking though because this is already supported, the pre option, that we may just be able to close this PR?

matteius:
This won't be possible unless the pip resolver supported the pre option on a per-package level instead of a global level.

PR #4007, opened 2019-10-27
Add Video Tutorials

lifez:
----

I think it would be good to have some pipenv tutorial link in README.


dimaqq:
Sounds like a good idea overall!
(I dunno if these particular tutorials are good or best available)

matteius:
Declining this PR as outdated.

Issue #4059, opened 2019-12-16
Failure during installation TensorFlow on Python 3.7


fridex:
### Issue description

Using the most recent version of Pipenv from master 3e63f078adc fails installation of TensorFlow==2.0.0:

```
$ pipenv install  tensorflow==2.0.0                                                                                                                                                                                                        
Creating a Pipfile for this project…
Installing tensorflow==2.0.0…
Adding tensorflow to Pipfile's [packages]…
✔ Installation Succeeded 
Pipfile.lock not found, creating…
Locking [dev-packages] dependencies…
Locking [packages] dependencies…
Traceback (most recent call last):
  File "/usr/local/bin/pipenv", line 10, in <module>
    sys.exit(cli())
  File "/usr/local/lib/python3.7/site-packages/pipenv/vendor/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/pipenv/vendor/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.7/site-packages/pipenv/vendor/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.7/site-packages/pipenv/vendor/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.7/site-packages/pipenv/vendor/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/pipenv/vendor/click/decorators.py", line 64, in new_func
    return ctx.invoke(f, obj, *args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/pipenv/vendor/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/pipenv/vendor/click/decorators.py", line 17, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/pipenv/cli/command.py", line 251, in install
    site_packages=state.site_packages
  File "/usr/local/lib/python3.7/site-packages/pipenv/core.py", line 2157, in do_install
    skip_lock=skip_lock,
  File "/usr/local/lib/python3.7/site-packages/pipenv/core.py", line 1267, in do_init
    pypi_mirror=pypi_mirror,
[...]

PR #4060, opened 2019-12-16
Fix setting os.environ of Python 3.7+

fridex:
Fixes: #4059

### DIFF
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -1286,7 +1286,9 @@ def venv_resolve_deps(
         os.environ["PIPENV_VERBOSITY"] = str(environments.PIPENV_VERBOSITY)
         os.environ["PIPENV_REQ_DIR"] = fs_str(req_dir)
         os.environ["PIP_NO_INPUT"] = fs_str("1")
-        os.environ["PIPENV_SITE_DIR"] = get_pipenv_sitedir()
+        pipenv_site_dir = get_pipenv_sitedir()
+        if pipenv_site_dir is not None:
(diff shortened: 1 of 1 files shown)


pipenv/utils.py
```diff
@@ -1286,7 +1286,9 @@ def venv_resolve_deps(
         os.environ["PIPENV_VERBOSITY"] = str(environments.PIPENV_VERBOSITY)
         os.environ["PIPENV_REQ_DIR"] = fs_str(req_dir)
         os.environ["PIP_NO_INPUT"] = fs_str("1")
-        os.environ["PIPENV_SITE_DIR"] = get_pipenv_sitedir()
+        pipenv_site_dir = get_pipenv_sitedir()
+        if pipenv_site_dir is not None:
+            os.environ["PIPENV_SITE_DIR"] = pipenv_site_dir
```
uranusjr: This would fail to unset `PIPENV_SITE_DIR` if the result is None. Probably need an `else` block.

fridex: Thanks for the review.

Not sure if I follow  - the behaviour is the same as for `keep_outdated` down below this line. Unsetting the environment is done based on the context manager above that keeps a backup. What would be expected in the `else` block?

uranusjr: I would expect it to unset the value, so something like `os.environ.pop("PIPENV_SITE_DIR", None)`

fridex: Done in 5e767e9c. Thanks!

fridex:
> Now we need to figure out a way to merge this. A rebase would help a lot.

Updated. Thanks.

Issue #3441, opened 2019-01-13
'save-exact' option for saving exact version in Pipfile


SteveByerly:
I've looked through the docs and the cli, but cannot seem to find a way to save the version of the installed package explicitly without editing the Pipfile after installing a package.

Example:
`pipenv install scipy` would add `scipy = "*"` to the Pipfile.

I would have expected a way to have `scipy = "==1.2.0"` to the Pipfile instead.

I was wondering if this sort of functionality already exists and/or if there is any reason not to go this route.

This would be similar to running `npm install react --save-exact`, where the installed version is added to the `package.json` file.

Apologies in advance if this is not the right forum/format for this question. I had tried looking through the docs, Stackoverflow, and search engines without luck.

jxltom:
I'm +1 on this but it requries a PEEP since it is a API change.

For now, you can only edit the Pipfile manually and lock it again by ```pipenv lock``` or ```pipenv install --dev``` if you would like to lock/install.

xoelop:
BTW,  I've just seen that you can do `pipenv install --ignore-pipfile` and you install the versions that are saved in Pipfile.lock, ignoring Pipfile. 

So there's no option so save-exact to Pipfile because it's being done already to Pipfile.lock every time you do `pipenv install`. 

Just my 2 cents :)

Demetri0:
I think that it should be **by default** behavior. I'm totally don't mean `any` version when I install something, but I mean `the latest on the current moment`. And I with bugfixes to be applicable so it should be not `==` but `~=`

unique1o1:
What I do to get the exact version is:

``` bash
pipenv run pip freeze > requirements.txt
pipenv install -r requirements.txt     
```

Just a simple workaround so, that I don't have to dig the lock file.

matteius:
`pipenv sync` installs the versions from the lock file, and `pipenv install` re-lock with the Pipfile specifiers and then installs that updated lock file.

PR #4076, opened 2019-12-19
Create PEEP-006.md

xoelop:
PEEP to add the option to save exact libraries versions to Pipfile

### The issue

As described in https://github.com/pypa/pipenv/issues/3441, it would be interesting if when installing a certain package with pipenv *without specifying the version*  (`pipenv install scipy`, for example), there was a flag that saved the *exact* version that was installed to the Pipfile. Something like the `--save-exact` flag in npm.

### The fix

Right now this can be accomplished, AFAIK, manually editing the version numbers in Pipfile, but it's quite annoying. 

### The checklist

* [ YES] Associated issue: https://github.com/pypa/pipenv/issues/3441

### DIFF
--- a/peeps/PEEP-006.md
+++ b/peeps/PEEP-006.md
@@ -1,0 +1,15 @@
+# PEEP-006: add flag to save exact libraries' versions to Pipfile
+
+PROPOSED
+
+As suggested in [this issue](https://github.com/pypa/pipenv/issues/3441), it would be very convinient if there was a flag like in npm that, when doing `pipenv install`, made that the versions saved to Pipfile were the exact ones used, instead of `"*"`
+
+Copying here some text from that issue that explains the problem a bit better:
+
+> Example:
+> pipenv install scipy would add scipy = "*" to the Pipfile.
+>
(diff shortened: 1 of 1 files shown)


hoshsadiq:
This already exists:

```bash
$ pipenv install scipy==1.2.0
<clipped>
$ cat Pipfile
[[source]]
name = "pypi"
verify_ssl = true

[dev-packages]

[packages]
scipy = "==1.2.0"

[requires]
python_version = "3.6"
```

`pipenv` accepts any version that `pip` can accept when it comes to version constraints.

xoelop:
I've just edited the PR so that it's clearer to see what I meant. 

I know you can specify a version to pipenv like that, but that adds an extra step of looking what is the latest available version for every package you want to install. Plus there may be libraries whose latest versions aren't compatible, a problem you wouldn't run into just doing `pipenv install scipy` only, and letting pipenv figure out the latest versions of all the necessary dependencies. 

Ideally, one would be able to save the exact versions to pipfile, without specifying those versions but saving the ones currently installed, so that other people can install them as well

Those versions are saved in pipfile.lock. Is there any way to install all the libraries from pipfile.lock instead of from pipfile?

PR #4103, opened 2020-01-18
Improve wording in "basics.rst"

offby1:


techalchemy:
If you don't mind, could you resolve the conflict? The only difference I can see is `pipenv install` was properly code-formatted in the documentation, but I don't want to push to your master branch

PR #4113, opened 2020-01-27
Add missing format operator ( % )

### LINKED ISSUE #4112: Missing % in patched ipaddress.py breaks pipenv
In   https://github.com/pypa/pipenv/blob/21e9e242449b4a1dbf112dbec168744528a14ecc/pipenv/patched/notpip/_vendor/ipaddress.py#L1106
the % is forgotten so it breaks pipenv

raise TypeError("%s and %s are not of the same version" (a, b))
must be
raise TypeError("%s and %s are not of the same version" % (a, b))

usta:
The format operator ( % ) was missing.

### The fix

Fixes #4112

### DIFF
--- a/pipenv/patched/notpip/_vendor/ipaddress.py
+++ b/pipenv/patched/notpip/_vendor/ipaddress.py
@@ -1103,7 +1103,7 @@ class _BaseNetwork(_IPAddressBase):
         try:
             # Always false if one is v4 and the other is v6.
             if a._version != b._version:
-                raise TypeError("%s and %s are not of the same version" (a, b))
+                raise TypeError("%s and %s are not of the same version" % (a, b))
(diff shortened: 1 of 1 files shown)


usta:
ping ? @techalchemy

techalchemy:
Thanks for the fix -- only reason I haven't merged it yet is because I have been hoping this would be resolved in an update at some point (but would require upgrading pip I suppose). Perhaps we will debundle pip's vendored ipaddress module

techalchemy:
This is now fixed due to an update to the vendored version of pip -- see

https://github.com/pypa/pipenv/blob/master/pipenv/patched/notpip/_vendor/ipaddress.py#L352

Thanks for the contribution, closing the PR for now to avoid additional patches to maintain!

Issue #4133, opened 2020-02-18
Script arguments with special symbols (& < > |)  are not enclosed by double-quotes when pipenv run is used


Phobosmir:
### Issue description

When the script is run with `pipenv run` command, the actual arguments containing "&" sign go to the script without quotes:

pipenv run somescript.py --argument "with_&_sign".

### Expected result

>>> print(sys.argv)
['somescript.py', '--argument', 'with_&_sign']

### Actual result
>>> print(sys.argv)
['somescript.py', '--argument', 'with_']

The same for other symbols:
< > | 

-------------------------------------------------------------------------------

<details><summary>$ pipenv --support</summary>

Pipenv version: `'2018.11.27.dev0'`

Pipenv location: `'c:\\program files (x86)\\python37-32\\lib\\site-packages\\pipenv'`

Python location: `'c:\\program files (x86)\\python37-32\\python.exe'`

Python installations found:

  - `3.6.3`: `C:\Users\MYMironov.TOPCON\AppData\Local\Programs\Python\Python36\python.exe`
  - `3.6.3`: `C:\ProgramData\Anaconda3\python.exe`
  - `3.6.2`: `C:\Program Files\Python36\python.exe`
  - `3.6.5`: `C:\Users\MYMironov.TOPCON\AppData\Local\Continuum\anaconda3\python.exe`

PEP 508 Information:

```
{'implementation_name': 'cpython',
 'implementation_version': '3.7.3',
 'os_name': 'nt',
 'platform_machine': 'AMD64',
 'platform_python_implementation': 'CPython',
 'platform_release': '10',
 'platform_system': 'Windows',
 'platform_version': '10.0.18362',
 'python_full_version': '3.7.3',
 'python_version': '3.7',
 'sys_platform': 'win32'}
```

System environment variables:

  - `PROCESSOR_ARCHITEW6432`
  - `SYSTEMDRIVE`
  - `VIRTUAL_ENV`
  - `PSMODULEPATH`
  - `ROUTER_PASS`
  - `OS`
  - `BOOST_LIBRARYDIR`
  - `LOCALAPPDATA`
  - `__COMPAT_LAYER`
  - `PROGRAMW6432`
  - `ONEDRIVECOMMERCIAL`
  - `M2_HOME`
  - `USERNAME`
  - `DRIVERDATA`
  - `HOMEDRIVE`
  - `VS110COMNTOOLS`
  - `ROUTER_LOGIN`
[...]

PR #4134, opened 2020-02-18
Fixed enclosing by double-quotes for arguments with symbols & < > |

Phobosmir:
#4133 
Fixed enclosing by double-quotes for script arguments.

### DIFF
--- a/pipenv/cmdparse.py
+++ b/pipenv/cmdparse.py
@@ -94,5 +94,5 @@ class Script(object):
         """
         return " ".join(itertools.chain(
             [_quote_if_contains(self.command, r'[\s^()]')],
-            (_quote_if_contains(arg, r'[\s^]') for arg in self.args),
+            (_quote_if_contains(arg, r'[\s^&<>|]') for arg in self.args),
         ))
--- a/pipenv/cmdparse.py
+++ b/pipenv/cmdparse.py
@@ -72,6 +72,10 @@ class Script(object):
         * Whitespaces.
         * Carets (^). (pypa/pipenv#3307)
         * Parentheses in the command. (pypa/pipenv#3168)
+        * Ampersand (&). (pypa/pipenv#4133)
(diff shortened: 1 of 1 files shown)


techalchemy:
> Is it possible for an executable to contain these characters? If so, they should be added to the regex pattern in the line above.

According to https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file:

> The following reserved characters:
>
> -     < (less than)
> -     > (greater than)
> -     : (colon)
> -     " (double quote)
> -     / (forward slash)
> -     \ (backslash)
> -     | (vertical bar or pipe)
> -     ? (question mark)
> -     * (asterisk)

However, these are allowed on *nix

chris-twosense:
It looks like the changes merged into `pypa:master` never made it into a live release, or were subsequently deleted. Does anyone know what happened here? @uranusjr 

Latest release as of this comment:
https://github.com/pypa/pipenv/blame/660db3ef1a72112b8350b67b1f8bf78c611b37e9/pipenv/cmdparse.py#L76

matteius:
@chris-twosense Do you have record of what the missing changes are?  This seems to be from before I started working on the project, and I'm having trouble finding what you are referring to.  There is a deleted branch on this thread so that doesn't really help.

matteius:
Oh duh, I am looking at the PR, thought this was an issue report.  Ok so this was closed and never merged.  Should it have been merged?

chris-twosense:
It looks like the PR was approved, but then deleted? The fix implemented by this branch is still an issue, special characters `& < > |` are treated as separate arguments in pipenv scripts, so I'm not sure why it shouldn't have been merged. Thoughts?

matteius:
@chris-twosense I assigned the issue to myself and when I find some time I'll look into it more.  Unless you think you can get to it sooner and want to open a new similar PR with a news fragment and unit test?

Issue #1839, opened 2018-03-23
Built-in Sphinx documentation search index broken (also, search link hard to find)


WoLpH:
First of all it's pretty hard to find the search at all in the docs (would be useful in the sidebar) but there is a link at the bottom of the docs homepage: https://docs.pipenv.org/

Which points to this page: https://docs.pipenv.org/search/

Actually searching for something like this: https://docs.pipenv.org/search/?q=test
Results in links like these which only result in 404s: https://docs.pipenv.org/en/latest/advanced?highlight=test

The problem seems to be that the search index uses the standard readthedocs urls containing the versions (`/en/latest/`) instead of using the root. I'm not entirely sure if this bug is restricted to this project's configuration or if it's an upstream bug.

frostming:
The search part seems problematic and is hidden by JS hack.

So it is intentional. Maybe @kennethreitz has some words for it.

techalchemy:
This is fixed

WoLpH:
Are you sure? I've tried a minute ago and it still looks broken to me.

techalchemy:
Searching works fine. Are you complaining about our style sheet?

WoLpH:
Have you actually tried the search and read the original question fully?

All of the search results end up in non-existing (404) pages.

RickDB:
Search result pages all have 404's here as well.

techalchemy:
Ah ok, I didn’t click any of the resulting links. Considering we have it hidden to begin with I’m not that surprised. Not sure what would be required to fix

WoLpH:
I've done some testing by setting up another project on readthedocs with the docs at the root domain and that appears to function fine: http://pipenv-searchable.readthedocs.io/search.html?q=test

So for some reason the search index for this project is broken. Perhaps someone can press the "wipe" button at readthedocs to fix it?

![image](https://user-images.githubusercontent.com/270571/38368879-c652bc08-38e6-11e8-91df-889b8a971650.png)

techalchemy:
@WoLpH I’ll find out who has access and see if I can get it taken care of. Thanks for looking into it!

ashnur:
@techalchemy are you sure? because I just tried, and it didn't work.

ashnur:
[...]

PR #4166, opened 2020-03-27
Remove search link on main page of RTD documentation

brainwane:
Partially and temporarily addresses #1839 while we fix the underlying problem and investigate the search index. Users will temporarily only be able to use the search box in the sidebar.

Please revert this once the index is fixed.

Signed-off-by: Sumana Harihareswara <sh@changeset.nyc>

### The issue

 #1839

* [x] Associated issue
* [x] A news fragment in the `news/` directory to describe this fix with the extension `.trivial`

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -134,4 +134,3 @@ Indices and tables
 
 * :ref:`genindex`
 * :ref:`modindex`
-* :ref:`search`


brainwane:
@techalchemy ping.

brainwane:
Fixing this aspect of the docs may be more important if we expect a bunch of people to be looking at those docs for the first time in a while (which is one thing that a release may cause).

brainwane:
@techalchemy I believe this is ready for review.

PR #4167, opened 2020-03-28
Replace docs.pipenv.org with pipenv.pypa.io

ciphergoth:
`dig pipenv.org ns` currently shows no nameserver for that domain;
pipenv.pypa.io is the new home for documentation, so use that
everywhere (contributors guides, `pipenv --envs` output).


techalchemy:
Thanks for the pr! Will merge shortly

PR #4175, opened 2020-04-03
minor redaction fix in basics doc.

jerojasro:


jerojasro:
@brainwane I don't mind at all, didn't notice that one. Done, and thanks.

PR #4186, opened 2020-04-10
WIP Update local bootstrap instructions

ncoghlan:
### The issue

`pytest` was reporting failures that I suspected were due to missing steps in my local setup, and the existing bootstrapping instructions didn't work for me, so I updated them based on the current CI setup.

Marked as WIP, as I based on the work on the pending vendored library update branch rather than master (this will be a lot smaller and easier to review once that is merged)

### The fix

* git submodule updates are needed to get tests passing locally
* the `--user` global pipenv install is needed for bootstrapping on systems where you don't want to (or can't) do a global install

### The checklist

* [ ] Associated issue


matteius:
Outdated.

Issue #4188, opened 2020-04-11
Your API Key is invalid


tucked:
```
$ pipenv check
Checking PEP 508 requirements…
Passed!
Checking installed package safety…
An error occurred:
Your API Key '1ab8d58f-5122e025-83674263-bc1e79e0' is invalid. See https://goo.gl/O7Y1rS

2020-04-11T10:08:05-0700 exit 1
```

I can provide `pipenv --support` if necessary.

techalchemy:
Thanks for the report -- We had received the ok from the pyup team to include the API key in pipenv, so I'm not sure why it's been revoked, though I suspect this is going to cause some pain for a lot of users. /cc @pyupio @jayfk (related context: #1651) -- was this intentional? How would you like us to move forward?

jayfk:
@Jwomers is now taking care of pyup.io with the help of @harlekeyn and @rafaelpivato. ccd

Ochirgarid:
The same issue is occurring for me. Looking forward for the fix.

alekpikl:
I have the same issue. Looking forward to the fix! :) 
Thanks!

kasioumis:
Is there a way to temporarily ignore this with `pipenv check --ignore [something]`?

joaokrabbe:
I have the same issue. Looking forward to the fix! :)
Thanks!

cemsbr:
A workaround is to install _safety_ and use `safety check` instead of `pipenv check`.

pgagnon:
> A workaround is to install _safety_ and use `safety check` instead of `pipenv check`.

`pipenv lock --requirements | safety check --stdin`

techalchemy:
After a brief conversation with the new pyup.io management,  (and it seems others have observed this already) we won't need the API key anymore to connect with pyup.io. So we will remove the API key.  In the meantime I believe this can be resolved by simply setting the environment variable `PIPENV_PYUP_API_KEY=""`.

kristang:
I don't want to to tell the maintainers how to handle issues and fixes, but maybe keep issues open until a fix is actually released. This closed issue gives the impression that the problem has been fixed, even though it has only been merged into master, but not yet released.

jdobes:
@kristang pipenv check was actually working in last 2 weeks using the old pipenv version, but here we go again...

kristang:
> @kristang pipenv check was actually working in last 2 weeks using the old pipenv version, but here we go again...

[...]

PR #4192, opened 2020-04-14
Fix issue #3901: environment variables parse in commands

### LINKED ISSUE #3901: environment variables in pipfile scripts are not recognized correctly
When running [scripts from pipfile](https://docs.pipenv.org/en/latest/advanced/#custom-script-shortcuts). Unix style environment variables are wrongly recognized as commands.

### Expected result

`hello=world` is recognized as setting an environment variable 

### Actual result

`Error: the command hello=world (from generic-script) could not be found within PATH.`

### Steps to replicate

- have the following part in `pipfile`
    ```
    [scripts]
    generic-script = 'hello=world python --version'
    ```
- `$ pipenv run generic-script`

Hammond95:
### The issue
This fixes #3901  

### The fix
My code will fix the specific use case reported in the issue, however this code still requires some improvements in my opinion. Before proceeding with these improvement, I think that we should define which use cases this feature should cover.

Some examples:
- do we want to solve variables into the command? `${MYVAR} or $MY_VAR`
- do we want to allow the execution of multiple commands if separated by semicolon? `echo "hello"; python --version`
- do we want to allow multiple command execution with `|` or `&&` or `||` operators?

### Some possible improvements
- Also add a specific class to manage windows env variables
- manage reserved keywords
- dynamically instantiate the correct class based on the running OS

### The checklist

* [x] Associated issue


Hammond95:
The tests fails on this particular one:
```python
@pytest.mark.cli
@flaky
def test_pipenv_check(PipenvInstance):
    with PipenvInstance() as p:
        p.pipenv('install requests==1.0.0')
        c = p.pipenv('check')
        assert c.return_code != 0
        assert 'requests' in c.out
        c = p.pipenv('uninstall requests')
        assert c.ok
        c = p.pipenv('install six')
        assert c.ok
        c = p.pipenv('check --ignore 35015')
        assert c.return_code == 0
        assert 'Ignoring' in c.err
```

I get an error also if I try to run the command `pipenv check` manually:
```

Passed!
Checking installed package safety…
[JSONParseError]:   File "/Users/mdeluca/_projects/pipenv/pipenv/vendor/click/core.py", line 1066, in invoke
[JSONParseError]:       return ctx.invoke(self.callback, **ctx.params)
[JSONParseError]:   File "/Users/mdeluca/_projects/pipenv/pipenv/vendor/click/core.py", line 610, in invoke
[JSONParseError]:       return callback(*args, **kwargs)
[JSONParseError]:   File "/Users/mdeluca/_projects/pipenv/pipenv/vendor/click/decorators.py", line 73, in new_func
[JSONParseError]:       return ctx.invoke(f, obj, *args, **kwargs)
[JSONParseError]:   File "/Users/mdeluca/_projects/pipenv/pipenv/vendor/click/core.py", line 610, in invoke
[JSONParseError]:       return callback(*args, **kwargs)
[JSONParseError]:   File "/Users/mdeluca/_projects/pipenv/pipenv/cli/command.py", line 450, in check
[JSONParseError]:       pypi_mirror=state.pypi_mirror,
[JSONParseError]:   File "/Users/mdeluca/_projects/pipenv/pipenv/core.py", line 2657, in do_check
[JSONParseError]:       raise exceptions.JSONParseError(c.out, c.err)
Failed parsing JSON results:
ERROR: Your API Key '1ab8d58f-*************' is invalid. See https://goo.gl/O7Y1rS
```

This doesn't look like an error related to my code, is something I am missing?

chysi:
See #4188 -- pipenv check is broken for the time being while the pyup.io team get around to it.

Hammond95:
@chysi Thank you 😄

uranusjr:
I’m inclined to use structured input instead:

```toml
serve = {
    cmd = "uvicorn mysite:app --reload",
    env = { DEBUG = true },
}
```

This gets rid of all the parsing and platform-specific behaviours.

Multiple commands can be supported by supplying an array to `cmd` (or we can introduce a new `cmds` key instead). A flag can be added to specify how the chain continues after an error occurs (i.e. switch between `;`, `||`, and `&&`).

I don’t think it’s a good idea to re-invent Bash command parsing. Just let the user pass in something structured, it’s easier for everyone (including the user; I don’t think most people in the world use shell syntaxes as a native mental model).

Hammond95:
@uranusjr This makes sense, I felt like I was re-inventing the wheel.
Should we also provide retro-compatibility for the commands defined as:
`command = 'foo --bar=goo'` ?

uranusjr:
@Hammond95 Definitely. It should continue to exist and work identically to `command = { cmd = 'foo --bar=goo' }`

uranusjr:
Oh, I just spotted a mistake in my structured example. Environment variable values must be strings (of course):

```toml
serve = {
    cmd = "uvicorn mysite:app --reload",
    env = { DEBUG = 'true' },
}
```

This brings up an interesting point. We’ll either need to do some introspection before passing the env values to `subprocess` (IIRC it would crash if any value is not string), or `str(value)` on the way in. But we’ll need to handle certain values carefully if we go with the `str()` route: users might not expect `true` to become `True` when they receive it in the subprocess.

Hammond95:
@uranusjr Yeah, that's a good point, I will give a look at this.

Another question:
If we run `pipenv run --help` we get
```

  Spawns a command installed into the virtualenv.

Options:
  --clear             Clears caches (pipenv, pip, and pip-tools).  [env var:
                      PIPENV_CLEAR]

```

Does it make sense to have the options `--python`, `--three/--two`, `--clear`, `--pypi-mirror` associated with this command?

techalchemy:
> Does it make sense to have the options `--python`, `--three/--two`, `--clear`, `--pypi-mirror` associated with this command?

I had a separate discussion recently with @ncoghlan about this question. The answer is 'probably not', but most likely we should address that in a separate PR since we have a lot of extra flags and arguments being passed around

Hammond95:
Not sure if this requires a dedicated issue, but I have also found a possible problem in the file `project.py`:

In the method:
```python3
    def build_script(self, name, extra_args=None):
        try:
            script = Script.parse(self.parsed_pipfile["scripts"][name])
        except KeyError:
            script = Script(name)
        if extra_args:
            script.extend(extra_args)
        return script
```

in case you provide a non-existing key, pipenv will try to treat the string as a command, so the following will work as expected:

`pipenv run python --version`

this won't:

`pipenv run "python --version"`

techalchemy:
> in case you provide a non-existing key, pipenv will try to treat the string as a command, so the following will work as expected:
>
> `pipenv run python --version`

Yes, This is the desired behavior

> this won't: 
>
> `pipenv run "python --version"`

This seems like a bug and it also seems kind of silly that we wouldn't be calling `Script.parse()` here... based on the commit history it doesn't seem like there is any reason beyond the fact that we also accept `extra_args`. I went ahead and found this instance in `core.py` where we are using the code that way (and we also test that functionality):

```python
    try:
        cmd_string = ' '.join([script.command] + script.args)
        if environments.is_verbose():
            click.echo(crayons.normal("$ {0}".format(cmd_string)), err=True)
    except ScriptEmptyError:
        click.echo("Can't run script {0!r}-it's empty?", err=True)
    run_args = [script]
```

I vaguely recall having a conversation with @uranusjr about this and whether there is any reason not to just parse the command all the time. @uranusjr -- can you think of any reason to tackle this a bit differently? As I recall there was a reason...

uranusjr:
I’m confused. Do you mean `pipenv run "python --version"` should call `python` with `--version` as an argument, or call an executable named `"python --version"`? I would definitely expect the latter, and the `build_script` implementation mentioned above looks correct to me.

Hammond95:
@uranusjr Well maybe you're right and the command provided in quotes shouldn't be run as `python` with `--version` argument.

But probably we should avoid inline command execution at all 🤔 and keep that only as a "caller" of the keys defined in the `[scripts]` section, let me know what you think.

Also in this case the command (`pipenv run hello=world python --version`) related to the issue, will fail...

uranusjr:
> But probably we should avoid inline command execution at all 🤔 and keep that only as a "caller" of the keys defined in the `[scripts]` section, let me know what you think.

Honestly, I agree (maybe except I’d keep `python`). But it’s way too late for this, unfortunately. There are simply too many people relying on this feature, removing it is not an option.

Issue #4199, opened 2020-04-17
`pipenv lock -r` should respect `--pypi-mirror` arg


TennyZhuang:
### Is your feature request related to a problem? Please describe.

While using `pipenv lock -r` with `--pypi-mirror`, the generated output (content of requirements.txt) will ignore the `--pypi-mirror` option.

### Describe the solution you'd like

`--pypi-mirror` should be treated as an extension to `project.sources`

### Additional context

-------------------------------------------------------------------------------

Please run `$ pipenv --support`, and paste the results here. Don't put backticks (`` ` ``) around it! The output already contains Markdown formatting.

<details><summary>$ pipenv --support</summary>

Pipenv version: `'2018.11.26'`

Pipenv location: `'/usr/local/Cellar/pipenv/2018.11.26_4/libexec/lib/python3.8/site-packages/pipenv'`

Python location: `'/usr/local/Cellar/pipenv/2018.11.26_4/libexec/bin/python3.8'`

Python installations found:

  - `3.8.2`: `/Users/tianyizhuang/.local/share/virtualenvs/dapeng-ojERrZnV/bin/python3`
  - `3.7.7`: `/usr/local/bin/python3`
  - `3.7.7`: `/usr/local/bin/python3.7m`
  - `3.7.3`: `/usr/bin/python3`
  - `2.7.17`: `/usr/local/bin/python`
  - `2.7.17`: `/usr/local/bin/pythonw`
  - `2.7.16`: `/usr/bin/python`
  - `2.7.16`: `/usr/bin/pythonw`

PEP 508 Information:

```
{'implementation_name': 'cpython',
 'implementation_version': '3.8.2',
 'os_name': 'posix',
 'platform_machine': 'x86_64',
 'platform_python_implementation': 'CPython',
 'platform_release': '19.3.0',
 'platform_system': 'Darwin',
 'platform_version': 'Darwin Kernel Version 19.3.0: Thu Jan  9 20:58:23 PST '
                     '2020; root:xnu-6153.81.5~1/RELEASE_X86_64',
 'python_full_version': '3.8.2',
 'python_version': '3.8',
 'sys_platform': 'darwin'}
```

System environment variables:

  - `PATH`
  - `TERM_PROGRAM`
  - `PIP_PYTHON_PATH`
  - `SHELL`
  - `TERM`
  - `TMPDIR`
  - `TERM_PROGRAM_VERSION`
  - `TERM_SESSION_ID`
  - `ZSH`
  - `USER`
[...]

PR #4200, opened 2020-04-17
`pipenv lock -r` respect pypi_mirror args

TennyZhuang:

### The issue

#4199 

### The fix

extend the `pip_args` in the first line of `requirements.txt`.

### The checklist

* [x] Associated issue


pipenv/core.py
```diff
@@ -791,7 +791,7 @@ def do_install_dependencies(
     skip_lock=False,
     concurrent=True,
     requirements_dir=None,
-    pypi_mirror=False,
+    pypi_mirror=None,
```
techalchemy, line 794: This probably shouldn't need to change

pipenv/core.py
```diff
@@ -826,6 +826,8 @@ def do_install_dependencies(
     deps_list = list(lockfile.get_requirements(dev=dev, only=requirements))
     if requirements:
         index_args = prepare_pip_source_args(project.sources)
+        if pypi_mirror:
+            index_args.extend(["-i", pypi_mirror])
```
techalchemy, line 830: This wouldn't be valid if there were multiple indexes available, since pip expects to receive only one index using `-i`. We have a method for implementing the `pypi_mirror` already by merging it into the project sources, which you can find in the `utils` module and you can use like this:

```python
index_args = prepare_pip_source_args(get_source_list(pypi_mirror=pypi_mirror, project=project))
```

pipenv/core.py
TennyZhuang: Why the `pypi_mirror` becoms False here but None in the rest parts of codes?

pipenv/core.py
TennyZhuang: Fixed it

pipenv/core.py
techalchemy: Most likely there is no good reason for this -- and apologies for not taking the time to look at it more closely! If, as you say, this is `None` everywhere else, and we don't differentiate in the function below (e.g. by checking if it's `False`) then it will be ok to make this change for consistency

PR #4208, opened 2020-04-24
Automate release workflow

techalchemy:
- Automatically release when tags are pushed to master
- Release automation will build wheel and sdist + upload to pypi
- Added test pypi as initial target
- Updated version update scripts to help with automatic version
  management

### DIFF
--- a/.github/workflows/pypi_upload.yml
+++ b/.github/workflows/pypi_upload.yml
@@ -1,0 +1,63 @@
+name: Create Release & Upload To PyPI
+
+on:
+  push:
+    # Sequence of patterns matched against refs/tags
+    tags:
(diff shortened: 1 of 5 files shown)


techalchemy:
tagged reviewers for awareness -- will be adding release documentation

techalchemy:
> I see it is uploading to test-pypi, is it a temporary solution or a standard workflow?

the `test.pypi.org` part will be a test to make sure the action works properly, my intention is to switch over to do real releases this way if everyone thinks it makes sense

frostming:
It makes sense to me. I also tried the release workflow of `pythonfinder`, it works like a charm.

techalchemy:
> It makes sense to me. I also tried the release workflow of `pythonfinder`, it works like a charm.

I didn't even realize you made a new release, that's awesome

PR #4216, opened 2020-04-29
Skip satisfied dependencies during installation

### LINKED ISSUE #3057: Make pipenv sync skip installed dependencies
We are currently using pipenv sync to deploy some of our apps with the versions of packages that we have tested.  The problem we're having is just how long it takes for a sync to run.  We might install just one new package (or update one version), then have to wait several minutes for a sync to complete.  

Using `pipenv sync --verbose` it looks like a pip install is being called for every single dependency.  Wouldn't it be faster to check the existing hashes in the lock file against the dependencies which are already installed to confirm that they match?  I don't actually know what those hashes are calculated over so this is a bit of a naive request.  Ultimately my goal is just to make pipenv sync not spin for several minutes on existing dependencies when they don't need to be updated.

Also of note, we came from using our own virtual environments with a normal requirements file for pip.  Adding a new dependency or updating a version was _much_ quicker then, so it has to be something specific to the way pipenv is running independent installs that takes so long.

Currently using Python 3.6.6 with pip 18.1 and pipenv 2018.10.13

### LINKED ISSUE #3506: Install only deltas when running pipenv sync
When working on a team project in git, it is important to ensure the python interpreter in a branch is deterministic. If I have updated django, say, in branch `update-django-to-21` and I need to go back to master to create a hotfix branch, I'll definitely need to run `pipenv sync` to ensure that the appropriate version of django is being used in the hotfix branch. We have, in fact, used a post-checkout githook to enforce this. The problem is that git sync is so slow...

It'd be excellent if there was a way to compare the current modules to the current Pipfile.lock to determine what, if anything, must be upgraded and then to only install the dependencies that are different.

The alternative is to simply install everything from the lock file every time you checkout a different branch (or any other checkout like `git checkout -- foo.py` to unstage the changes in foo.py).

### Additional context

pipenv 2018.11.26
python 3.6.6

Sorry, we aren't allowed to post the output of `pipenv --support` to the public. I can share it privately if required.

techalchemy:
- Skip satisfied dependencies in the environment during install
  by checking whether the constraint of a specifier is satisfied
- If there is no specifier and a dependency is installed, assume it
  is satisfied
- For editable dependencies, if the dependency in the environment is
  an egg link and points at the same path as the given dep, assume
  it is satisfied
- TODO: Add test
- Fixes #3057 
- Fixes #3506

### DIFF
--- a/pipenv/environment.py
+++ b/pipenv/environment.py
@@ -19,6 +19,7 @@ import six
 import pipenv
 
 from .vendor.cached_property import cached_property
+from .vendor.packaging.utils import canonicalize_name
 from .vendor import vistir
 
 from .utils import normalize_path, make_posix
@@ -711,6 +712,23 @@ class Environment(object):
         """
 
         return any(d for d in self.get_distributions() if d.project_name == pkgname)
+
+    def is_satisfied(self, req):
+        match = next(
+            iter(
+                d for d in self.get_distributions()
+                if canonicalize_name(d.project_name) == req.normalized_name
+            ), None
+        )
+        if match is not None:
+            if req.editable and self.find_egg(match):
+                return req.line_instance.path == match.location
+            elif req.line_instance.specifiers is not None:
+                return req.line_instance.specifiers.contains(
+                    match.version, prereleases=True
+                )
+            return True
+        return False
 
     def run(self, cmd, cwd=os.curdir):
         """Run a command with :class:`~subprocess.Popen` in the context of the environment
--- a/pipenv/environment.py
+++ b/pipenv/environment.py
@@ -722,8 +722,9 @@ class Environment(object):
             ), None
         )
         if match is not None:
-            if req.editable and req.local and req.line_instance.path and self.find_egg(match):
-                return vistir.compat.samefile(req.line_instance.path, match.location)
+            if req.editable and req.line_instance.is_local and self.find_egg(match):
+                requested_path = req.line_instance.path
+                return requested_path and vistir.compat.samefile(requested_path, match.location)
             elif match.has_metadata("direct_url.json"):
                 direct_url_metadata = json.loads(match.get_metadata("direct_url.json"))
                 commit_id = direct_url_metadata.get("vcs_info", {}).get("commit_id", "")
--- a/pipenv/environment.py
+++ b/pipenv/environment.py
@@ -722,7 +722,7 @@ class Environment(object):
             ), None
         )
         if match is not None:
-            if req.editable and self.find_egg(match):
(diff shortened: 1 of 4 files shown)


pipenv/environment.py
```diff
@@ -712,6 +713,23 @@ def is_installed(self, pkgname):
...
+    def is_satisfied(self, req):
+        match = next(
+            iter(
+                d for d in self.get_distributions()
+                if canonicalize_name(d.project_name) == req.normalized_name
+            ), None
+        )
+        if match is not None:
+            if req.editable and self.find_egg(match):
+                return req.line_instance.path == match.location
```
uranusjr, line 725: Use `os.path.samefile` here instead?

pipenv/environment.py
```diff
@@ -712,6 +713,23 @@ def is_installed(self, pkgname):
...
+            elif req.line_instance.specifiers is not None:
+                return req.line_instance.specifiers.contains(
+                    match.version, prereleases=True
+                )
+            return True
```
uranusjr, line 730: Probably need to add one more branch to check direct URL install.

pipenv/environment.py
techalchemy: `samefile` is unix only afaik

pipenv/environment.py
techalchemy: hmmm yes, I assumed they would just surface the version here. thanks!

pipenv/environment.py
uranusjr: More precisely, it’s Unix-only on Python 2. I would use it if possible (and fallback to `==` otherwise)

techalchemy: It looks like `samefile` just compares `os.stat` outputs which should still be valid on python 2, right? Is that still ok on windows?

```python
def samefile(f1, f2):
    """Test whether two pathnames reference the same actual file or directory
    This is determined by the device number and i-node number and
    raises an exception if an os.stat() call on either pathname fails.
    """
    s1 = os.stat(f1)
    s2 = os.stat(f2)
    return samestat(s1, s2)


def samestat(s1, s2):
    """Test whether two stat buffers reference the same file"""
    return (s1.st_ino == s2.st_ino and
            s1.st_dev == s2.st_dev)
```

uranusjr: We won’t know unless we try it out 😉

techalchemy: It works fine on `vistir`'s test suite. `vistir`'s subprocess runner has a threading race condition I am ignoring for now to get this release out, but those are the only failures now.

ncoghlan:
Ooh, I have some janky CI chicanery to try removing now :)

Very nice!

PR #4219, opened 2020-04-29
fix toml table construction

frostming:

### The issue

### The fix

### The checklist

* [ ] Associated issue

### DIFF
--- a/pipenv/project.py
+++ b/pipenv/project.py
@@ -790,7 +790,7 @@ class Project(object):
         except Exception:
             document = tomlkit.document()
             for section in ("packages", "dev-packages"):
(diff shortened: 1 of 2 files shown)


techalchemy:
I can see the problem, but how do I trigger this currently? ideally would catch these

frostming:
See the detailed traceback [here](https://github.com/pypa/pipenv/issues/3369#issuecomment-621122584)

It can be simple reproduced by install from following requirements.txt:
```
-i https://test.pypi.org/simple
requests
```

frostming:
I added a regression case to cover this issue.

PR #4241, opened 2020-05-05
Fix recursive PEP 508 VCS resolution

### LINKED ISSUE #3396: How to install packages with a vcs dependency in their setup.py
I'm working on a project at work for which one of the dependencies cannot be published to pypi, so I have to install it with the git url.  Unfortunately I can't paste the setup.py here.  

What I am trying to accomplish is to use the Pipfile/Pipfile.lock as a development environment for the tool and then deploy the tool with `pip install`.  So I'm trying to specify all my dependencies in the setup.py and then just have the `-e .` line in my Pipfile.  My package installs perfectly fine with `pip install .` and `pipenv install --skip-lock`.  However, `pipenv lock` fails with the error (obviously redacted): 
```
pipenv.exceptions.ResolutionFailure: ERROR: ERROR: Could not find a version that matches packagename@ git+https://GITHUB_TOKEN@github.com/user/repo.git@v0.1.0#egg=packagename from git+https://GITHUB_TOKEN@github.com/user/repo.git@v0.1.0#egg=packagename                          
No versions found    
```

I know it's going to be difficult to give specific advice without more information, but I'm hoping you can give me the format and conditions that must be true (does the repo need to be tagged, etc) for pipenv to resolve this dependency the way pip does.  Thanks

### LINKED ISSUE #4217: pipenv 2020.4.1b1: Invalid fragment string
I installed pipenv 2020.4.1b1 on my local machine by running:

    pip install --upgrade --pre pipenv

Then I updated my dependencies on my local machine by running:

    pipenv update

The Pipfile.lock got updated of course. Here is the relevant diff in Pipfile.lock:

```diff
         "robohash": {
             "git": "https://github.com/e1ven/Robohash.git",
-            "ref": "03ade66998c15735740ec4111b5da8d4383a7577"
+            "ref": "03ade66998c15735740ec4111b5da8d4383a7577",
+            "version": "==1.0"
         },

```

As you can see, this dependency is installed from GitHub, not from PyPI. Using the newer pipenv caused a new key-value to be added, which is `"version": "==1.0"`.

I committed these changes and launched them to a server. The server has Pipenv 2018.5.18 installed. When the server tried to install the dependencies from this Pipfile.lock, I got this error:

```
remote:        WARNING: Invalid fragment string egg=robohash==1.0
remote:        ABORTING INSTALL... You will have to reinstall any packages that failed to install.
remote:        You may have to manually run pipenv lock when you are finished.
```

It looks like the new key-value pair `version` is incompatible with pipenv 2018.5.18, when it applies to packages installed from a Git repo. Is that correct? Is it meant to be that way? If it is, it would be useful to include this compatibility notice in the change log. If not, then maybe the `version` key-value shouldn't be included in packages that are installed from a Git repo.

techalchemy:
- Recursively resolve PEP 508 URLs in subdependencies
  - Fix `piptools` cache storage of VCS dependencies as results
- Avoid saving VCS dependency versions to `Pipfile.lock`
- Fixes #3396
- Fixes #4217

Signed-off-by: Dan Ryan <dan.ryan@canonical.com>

### DIFF
--- a/pipenv/patched/piptools/utils.py
+++ b/pipenv/patched/piptools/utils.py
@@ -181,6 +181,8 @@ def format_requirement(ireq, marker=None, hashes=None):
     """
     if ireq.editable:
         line = "-e {}".format(ireq.link.url)
+    elif ireq.link and ireq.link.is_vcs:
+        line = str(ireq.req)
     elif is_url_requirement(ireq):
         line = ireq.link.url
     else:
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -741,6 +741,10 @@ class Resolver(object):
         if self._pip_options is None:
             pip_options, _ = self.pip_command.parser.parse_args(self.pip_args)
             pip_options.cache_dir = environments.PIPENV_CACHE_DIR
+            pip_options.no_python_version_warning = True
+            pip_options.no_input = True
+            pip_options.progress_bar = "off"
+            pip_options.ignore_requires_python = True
             self._pip_options = pip_options
         return self._pip_options
 
@@ -760,7 +764,7 @@ class Resolver(object):
             from pipenv.patched.piptools.repositories.pypi import PyPIRepository
             self._repository = PyPIRepository(
                 self.pip_args, use_json=False, session=self.session,
-                build_isolation=self.pip_options.build_isolation
+                build_isolation=False
             )
         return self._repository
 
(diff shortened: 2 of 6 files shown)


pipenv/patched/piptools/utils.py
```diff
@@ -181,6 +181,8 @@ def format_requirement(ireq, marker=None, hashes=None):
     """
     if ireq.editable:
         line = "-e {}".format(ireq.link.url)
+    elif ireq.link and ireq.link.is_vcs:
+        line = str(ireq.req)
```
uranusjr, line 185: Should we surface this problem to upstream? or is this issue specific to pipenv.

pipenv/utils.py
```diff
@@ -760,7 +764,7 @@ def repository(self):
             from pipenv.patched.piptools.repositories.pypi import PyPIRepository
             self._repository = PyPIRepository(
                 self.pip_args, use_json=False, session=self.session,
-                build_isolation=self.pip_options.build_isolation
+                build_isolation=False
```
uranusjr, line 767: Any reason behind this?

techalchemy: oh hrm, no... I can't even remember why I did this now

pipenv/patched/piptools/utils.py
techalchemy: I think this is specific to pipenv -- or possibly it is now fixed in piptools, since piptools captures and caches the vcs dependency correctly now

Issue #4245, opened 2020-05-06
Python installers (pyenv and asdf) should not need to be present on PATH to be detected by Pipenv


zbentley:
### Is your feature request related to a problem? Please describe.

We run Linux systems that contain multiple Pipenv projects which span multiple versions of Python. We use `pyenv` to manage those versions. We set `PYENV_ROOT` to a common directory in `/opt` so that they can share Python builds. We set that `PYENV_ROOT=/opt/pyenv in an `.env` file checked into each project. This has two benefits:

- We can refer to that `.env` file as a common "vars that should be set for this project" from other programs that understand such files, e.g. [`EnvironmentFile` directives for Systemd services](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#EnvironmentFile=).
- We can invoke `pipenv` from shells that are not specially configured, e.g. human users, service launcher scripts and makefiles.

However, when Pipenv wants to use Pyenv, it expects it to be on `PATH`. Users of these systems then have to modify their PATH (by whatever means they wish) to point at the `/opt/pyenv` directory. Despite respecting `PYENV_ROOT`, Pipenv does not use that variable to resolve the `pyenv` binary itself.

### Describe the solution you'd like

**First change:** When Pipenv wants to install Python via `pipenv` or `asdf`, it should try additional locations to find an installer executable. At present, it searches `PATH`. I would like it to search the following, in order:

- `PATH`
- The respective "installer root" environment variable + `/bin/{installer_name}`. The variables in question are `ASDF_DIR` and `PYENV_ROOT`.
- The default recommended source-control based install location for the installer + `/bin/{installer_name}`.

If an executable file of the expected name is found in any of those locations it should be used.

**Second change:** whether or not an executable installer can be found according to the above logic should be used to decide whether or not to try the other installer (the `pyenv` -> `asdf` fallback). At present, `pythonfinder`'s methods which only check for whether or not an installer's environment variables are configured will be used. **On users' systems with both `pyenv` and `asdf`, where only `asdf` is activated, this will be a behavior change; `pyenv` will now be used for installation.**

### Describe alternatives you've considered

- Manual workaround: every human user could ensure `pyenv`/`asdf` was on their PATH. Every launcher script could appripriately modify `PATH` before launching. **Pros:** requires no Pipenv changes. **Cons:** all code which wants to do a hands-free `pipenv install` on a system without an installer on the path will have to be aware of this and be modified.
[...]

PR #4246, opened 2020-05-06
Allow pyenv/asdf to be used even if they are not on PATH

zbentley:
### The issue

https://github.com/pypa/pipenv/issues/4245

Because of limitations with `.env` file syntax, it is hard to make a `pipenv` commands work on systems that have not been configured externally (outside of `.env` files) to locate `pyenv` or `asdf` installers. See the linked issue for additional details.

### The fix

This makes the changes described in https://github.com/pypa/pipenv/issues/4245

Specifically, the logic for "find a `pyenv` command to run" is now the following:
- Search `PATH` (the pre-existing logic)
- Search `PYENV_ROOT/bin/pyenv`
- Search `~/.pyenv/bin/pyenv`.

...and equivalent handling for `asdf`.

### The checklist

* [X] Associated issue: https://github.com/pypa/pipenv/issues/4245


pipenv/core.py
```diff
@@ -351,6 +351,42 @@ def find_a_system_python(line):
     return python_entry
 
 
+def find_python_installer(name, env_var):
```
frostming, line 354: Very good documentation, LGTM but I think `installers.py` is a better place for this function. A little refactor may be required for it.

pipenv/core.py
```diff
@@ -351,6 +351,42 @@ def find_a_system_python(line):
...
+    """
+    for candidate in (
+        # Look for the Python installer using the equivalent of 'which'. On
+        # Homebrew-installed systems, the env var may not be set, but this
+        # strategy will work.
+        find_windows_executable('', name),
+        # Check for explicitly set install locations (e.g. PYENV_ROOT, ASDF_DIR).
+        os.path.join(os.path.expanduser(os.getenv(env_var, '/dev/null')), 'bin', name),
+        # Check the pyenv/asdf-recommended from-source install locations
+        os.path.join(os.path.expanduser('~/.{}'.format(name)), 'bin', name),
```
techalchemy, line 384: surfacing the installers in explicit variables is desirable over burying them in a function because it becomes easier to reason about what is happening, or to separate out the location of the install as compared with the process for finding python executables

pipenv/core.py
```diff
@@ -395,18 +431,20 @@ def abort():
...
         # dedicated to python installs so probably the preferred
         # method of the user for new python installs.
-        if PYENV_INSTALLED and not PIPENV_DONT_USE_PYENV:
-            installer = Pyenv("pyenv")
-        elif ASDF_INSTALLED and not PIPENV_DONT_USE_ASDF:
-            installer = Asdf("asdf")
-        else:
-            installer = None
+        installer = None
+        if not PIPENV_DONT_USE_PYENV:
```
techalchemy, line 440: It's not clear to me how this is an improvement / what is changing about the logic here, can you explain what problem this solves and provide an example of how to cause it on the master branch?

frostming:
`Pyenv` and `Asdf` here are used to **install** a non-existing Python version, the code does a different task from `pythonfinder`.

techalchemy:
> `Pyenv` and `Asdf` here are used to **install** a non-existing Python version, the code does a different task from `pythonfinder`.

I see that the goal is to locate `pyenv` or `asdf` to use as an installer -- what I am struggling to see is how that is different from locating `pyenv` or `asdf` to use for some other reason -- fundamentally it's the same, you need to find the root installation -- which is all the installer-finding function is doing here.

I don't understand how the proposal is different from the functionality already in place...per my comments, the proposed code:

```python
        os.path.join(os.path.expanduser(os.getenv(env_var, '/dev/null')), 'bin', name),
        # Check the pyenv/asdf-recommended from-source install locations
        os.path.join(os.path.expanduser('~/.{}'.format(name)), 'bin', name),
```

Seems to boil down to the same functionality that is already implemented in `pythonfinder`, but implements it again in pipenv.  I'm not convinced this is an issue on the master branch of pipenv, as the bug was reported against the 2018.11.26 release

pipenv/core.py
zbentley: I can refactor the code to make the search paths more explicit, would that work?

zbentley: In general, `pythonfinder` is insufficient for this task. It surfaces code that checks whether or not `pyenv`/`asdf` are *in use* (on the invoking shell), but it provides no facilities for figuring out where the `pyenv`/`asdf` binaries *actually are*.

`pythonfinder` does a good job of its namesake (finding Python), but does not contain any facilities for finding python installers.

I suppose I could submit a PR to `pythonfinder`, but that would be a brand-new capability for that package, and I suspect a bit out of scope of its goals. Pipenv, which needs to *invoke* `pyenv`/`asdf` is a better place for that logic to live, I think.

pipenv/core.py
zbentley: Assuming neither of the `DONT_USE` variables are set:


**On master:**
If `PYENV_ROOT` is set, but `pyenv` the binary is not on `PATH`, the stacktrace posted in #4245 occurs.

**In this PR:**
If `PYENV_ROOT` is set, but `pyenv` the binary is not on `PATH`, the following logic is used:
- If a `pyenv` executable can be located in one of three locations, it will be used.
- If not, Pipenv will search for an `asdf` binary using the equivalent search process.

This provides two benefits:
1. `pyenv`/`asdf` is searched for more aggressively and don't need to be on `PATH`.
2. Fallback from `pyenv` to `asdf` will occur more reliably.

zbentley:
I think this is still an issue on `master`.

Specifically, the code which creates the Installers is the same, and passes un-qualified installer names (e.g. `'pyenv'`) to the command runner. **This requires that the installers be present on `PATH`, even if their other environment variables are set; that is what this PR is seeking to improve.**

The code passing the non-pathy names is [here](https://github.com/pypa/pipenv/blob/master/pipenv/core.py#L405).
The dispatch in the installer, which only searches `PATH` and no other locations is [here](https://github.com/pypa/pipenv/blob/9ab617af77cea7d8c5363ba9d10e5c931440e6ff/pipenv/installers.py#L71)

In the vendored version of `pythonfinder`, the code can check *whether pyenv/asdf is in use*, using methods like [this one](https://github.com/pypa/pipenv/blob/9ab617af77cea7d8c5363ba9d10e5c931440e6ff/pipenv/installers.py#L71).

...however, its asking the question "is `pyenv`/`asdf` likely active in the shell?", and Pipenv wants an answer to the question "where is the executable for `pyenv`/`asdf`?". `pythonfinder` does not contain equivalent logic to this PR for the purposes of finding `pyenv`/`asdf`.

zbentley:
@techalchemy you mentioned that this is

> the same functionality that is already implemented in pythonfinder

Could you provide a link to that code? If there is something in `pythonfinder` I can call instead of the env-var existence checks, I agree that using it is a better approach than adding more code to Pipenv core.

pipenv/core.py
zbentley: OK, I can make that move, but would prefer to wait for resolution on the other threads of discussion (re: whether this is needed behavior) here before doing it.

zbentley:
Sorry, I pushed a couple of additional updates because I had a stray `PYENV_YES=1` in my `.bash_profile` that was breaking things.

Also, here's a concrete example of the use case I'd like to support:

### Setup:

Using a POSIX OS (I haven't run this test plan on WSL):

Install Pyenv to a global location.
```
sudo rm -rf /opt/pyenv
sudo mkdir -p /opt/pyenv
sudo chown -R $USER /opt/pyenv
git clone https://github.com/pyenv/pyenv.git /opt/pyenv
```

Make a folder containing these two files:

```
$ cat .env
PYENV_ROOT=/opt/pyenv
PIPENV_YES=1
```

```
$ cat Pipfile
[requires]
python_version = "3.8.2"
```

### Preconditions:

Python 3.8.2 *must not* be the system python version (that invalidates the test).

All of these return nothing; no special shell configuration is present.
```
env | grep -i pyenv
env | grep -i pipenv
which pyenv
which asdf
```

Note that those would also pick up directories on `PATH` containing pyenv/pipenv.

### Test:
Run any of the following

```
pipenv install
pipenv shell
pipenv run echo
```

### Desired result:

Running any of the test commands above should cause the *installation* of `3.8.2` (via Pyenv) if it is not present, and the *use* of `3.8.2` if it is present in Pyenv or on the system.

### Result on `master`:

```
$ pipenv install
Warning: Python 3.8.2 was not found on your system...
You can specify specific versions of Python with:
  $ pipenv --python path/to/python
```

And, manually using the env file (bash):

```
$ set -a
$ source .env
$ env | grep PY
PYENV_ROOT=/opt/pyenv
$ pipenv install
Warning: Python 3.8.2 was not found on your system...
Installing CPython 3.8.2 with pyenv (this may take a few minutes)...
  File "/home/zacbentley/.local/bin/pipenv", line 11, in <module>
  File "/home/zacbentley/pipenv_master/pipenv/vendor/click/core.py", line 829, in __call__
  File "/home/zacbentley/pipenv_master/pipenv/vendor/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/home/zacbentley/pipenv_master/pipenv/vendor/click/core.py", line 1259, in invoke
  File "/home/zacbentley/pipenv_master/pipenv/vendor/click/core.py", line 1066, in invoke
  File "/home/zacbentley/pipenv_master/pipenv/vendor/click/core.py", line 610, in invoke
  File "/home/zacbentley/pipenv_master/pipenv/vendor/click/decorators.py", line 73, in new_func
    return ctx.invoke(f, obj, *args, **kwargs)
  File "/home/zacbentley/pipenv_master/pipenv/vendor/click/core.py", line 610, in invoke
  File "/home/zacbentley/pipenv_master/pipenv/vendor/click/decorators.py", line 21, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/home/zacbentley/pipenv_master/pipenv/cli/command.py", line 251, in install
    site_packages=state.site_packages
  File "/home/zacbentley/pipenv_master/pipenv/core.py", line 1920, in do_install
    site_packages=site_packages,
  File "/home/zacbentley/pipenv_master/pipenv/core.py", line 581, in ensure_project
    pypi_mirror=pypi_mirror,
  File "/home/zacbentley/pipenv_master/pipenv/core.py", line 491, in ensure_virtualenv
    python = ensure_python(three=three, python=python)
  File "/home/zacbentley/pipenv_master/pipenv/core.py", line 444, in ensure_python
    c = installer.install(version)
  File "/home/zacbentley/pipenv_master/pipenv/installers.py", line 136, in install
    timeout=PIPENV_INSTALL_TIMEOUT,
  File "/home/zacbentley/pipenv_master/pipenv/installers.py", line 72, in _run
    c = delegator.run(args, block=False, timeout=timeout)
  File "/home/zacbentley/pipenv_master/pipenv/vendor/delegator.py", line 336, in run
    c.run(block=block, binary=binary, cwd=cwd, env=env)
  File "/home/zacbentley/pipenv_master/pipenv/vendor/delegator.py", line 204, in run
    s = PopenSpawn(self._popen_args, **pexpect_kwargs)
  File "/home/zacbentley/pipenv_master/pipenv/vendor/pexpect/popen_spawn.py", line 53, in __init__
    self.proc = subprocess.Popen(cmd, **kwargs)
  File "/usr/lib/python2.7/subprocess.py", line 711, in __init__
    errread, errwrite)
  File "/usr/lib/python2.7/subprocess.py", line 1343, in _execute_child
    raise child_exception
OSError: [Errno 2] No such file or directory
```

techalchemy:
> This requires that the installers be present on PATH, even if their other environment variables are set; that is what this PR is seeking to improve.

Thank you for highlighting the main point here in bold since I was obviously not understanding what this PR accomplishes... this could arguably be achieved by jumping through `pythonfinder` internals but it is not designed to look for installers -- and would probably overcomplicate the goal here. I don't think I want installer-focused code in there anyhow, so it's best if we keep things as they are.

Thanks as well for the example use case, it's really helpful to have a concrete reference. Once @frostming is satisfied and tests are passing I'm ok to merge as well

pipenv/core.py
techalchemy: I definitely don't want to add the installer capability to pythonfinder, i think it's fine to add here. Overall I don't see any major issues with this as-is and think it's fine to leave

pipenv/core.py
techalchemy: Thanks for the cogent explanation, this clarifies the issue here and what the change accomplishes. I am ok with this -- thanks for putting it together and for explaining!

pipenv/core.py
techalchemy: this seems like a good suggestion that I will agree with and then i would say it's good to go

zbentley: That move has been made, along with some light refactors. Waiting on tests at this moment, if they are green then this is ready for re-review.

zbentley:
Summary of changes I made after @techalchemy and @frostming's review comments:
- Made `install` source `.env` files.
- Moved installer selection into `installers.py`.
- `Installer` is now an ABC.
- Installers are now instantiated without executable names, that's determined internally.
- The `environments` module gets reloaded when an `.env` file is read. *There's a significant general TODO here:* handling of environment variables should probably be reworked in the whole project. Whether or not vars are re-read after `.env` files are applied is done pretty inconsistently at present. It would be nice to have an API instead of `os.environ` for Pipenv to use when accessing env that supports control of "should loaded .env files be honored or should I use the values at Pipenv start time?" behavior rather than relying on ad-hoc re-imports and compile-time environment "freezing" in `environments.py`. I think changing that is out of scope of this PR, though I may put up a PR (or is that big enough to need a PEEP?) for it in the future.

pipenv/cli/command.py
```diff
@@ -226,7 +226,9 @@ def install(
     **kwargs
 ):
     """Installs provided packages and adds them to Pipfile, or (if no packages are given), installs all packages from Pipfile."""
-    from ..core import do_install
+    from ..core import do_install, load_dot_env
+
+    load_dot_env()
```
frostming, line 231: This change is out of your proposed behavior. You should change it and related module reloading part in a separate PR.
There were already many discussions about whether to respect dot env files in **setup** time because it is now used for **run** time only. #3398 #3080 #3271

pipenv/installers.py
```diff
@@ -56,12 +64,52 @@ def __init__(self, desc, c):
 
 
 class Installer(object):
+    __metaclass__ = ABCMeta
```
frostming, line 67: This shows no obvious improvement comparing to the plain parent class and `NotImplementedError`, but I am OK with it.

frostming:
Added my comments for the new changes, the dot-env part should be separated out from this PR. It requires more discussion.

pipenv/installers.py
zbentley: Sorry, force of habit. It's a very slight advantage (some IDEs and linters will automatically warn if you subclass and forget to implement unimplemented methods using ABCmeta), but I'm happy to back it out if you'd prefer.

pipenv/cli/command.py
zbentley: I removed the loading of the env. Prior discussions aside, this feels inconsistent: if we already set up python via `ensure_python` at other times, we should be loading `.env` those times as well.

As it is, my workaround would be something silly like `pipenv shell -- 'pipenv install; exit;'` to achieve the presence of `.env` files, which I can do, but feels lame.

However, the module reload is important, because without it, the check for `PIPENV_YES` fails during install unless that is set in the environment.

In general it feels like the Pipenv code is playing pretty fast and loose with whether or not it uses `.env` modified env vars: it freezes those in `environments.py` at compile time, but sometimes doesn't `import environments` until runtime, deferring environment reading by ... I can't tell if it's an accident or a very oblique way of re-reading the environment. Either way, I think a reload is better--not great, but better; a real solve here would be a better interface to all the environment vars Pipenv cares about, which I mentioned in another comment.

Followup questions:
- Is removing the `.env` load at install time sufficient?
- If environment module reloading must be removed, should I instead fix the `PIPENV_YES` check?
- If not, what remaining course of action achieves the goal (Pipenv can be set up and installed with nothing particular set in the shell)?

zbentley:
@frostming thanks for the help and feedback!

For future work in this area, if I wanted to improve the environment-variable reading API (currently `environments.py`) or make the loading (or not) of `.env` files more consistent, which of the following should I do?
- Submit a PR for review if no documented user-facing behavior is changed.
- Make a new PEEP.
- Adopt/resurrect the PEEP in https://github.com/pypa/pipenv/pull/3271?

Thanks again.

PR #4267, opened 2020-05-25
Make sure to import the vendored attrs

torsava:
### The issue

The `attrs` module is vendored in pipenv, but when the package is installed on the system, the system version is imported first, and fails if the version is different.

We found this issue in Fedora, where we have python-attrs-19.1.0, which causes a failure because version 19.3.0 is expected (and vendored):

    >   @attr.s(slots=True, eq=True, hash=True)
        class SetupInfo(object):
    E   TypeError: attrs() got an unexpected keyword argument 'eq'
    pipenv/vendor/requirementslib/models/setup_info.py:1341: TypeError

### The fix

I've implemented a patch in Fedora that instead of `import attr` uses `from pipenv.vendor import attr`. I think it be useful upstream too, so I opened this PR.

News fragment is probably not necessary, but I can provide one if prompted.

### DIFF
--- a/news/4267.vendor.rst
+++ b/news/4267.vendor.rst
@@ -1,0 +1,1 @@
+Changed attrs import path in vendored dependencies to always import from ``pipenv.vendor``.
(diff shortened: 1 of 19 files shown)


techalchemy:
Thanks for the PR! I went ahead and added a news entry + a patch so this gets updated when we revendor. Will merge after CI completes

Issue #4273, opened 2020-05-28
Pipenv no longer working under the root directory is not listed as a breaking change


gps035:
### Issue description

Our docker builds failed this morning due to a change in behaviour when running `pipenv install` in the root directory.
When running with the latest version (2020.5.28) an error is thrown after dependencies are built and installed `ERROR: Pipenv is not intended to work under the root directory, please choose another path.`
It looks like this is a change introduced in #3386, relating to an issue raised in #3434.
This issue is really just to point out that this behaviour change should be listed as breaking in the changelog, as it doesn't currently appear to be mentioned. 
For the time being we have worked around this by pinning `pipenv` to the last version (2018.11.26). 

### Expected result

N/A

### Actual result

N/A

### Steps to replicate

N/A

mohamedMok:
hi @gps035, 

I'm having the same issue with pipenv. 
Could you please show how you've pipped the `pipenv` version ?

Thanks

jean-malo:
I can confirm that this behaviour does not happen on version `2018.11.26`.

@mohamedMok You can just use `pip install pipenv==2018.11.26` which is the last version before this breaking change.

frostming:
@gps035 Any chance sending a PR to mention it in the CHANGELOG?

eugene-nikolaev:
Hi @gps035,
I've got the same issue but pipenv isn't even being ran in a root.

This is 'pipenv --support' skim:

```
Debug–specific environment variables:
  - `PATH`: `/usr/local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin`
  - `LANG`: `C.UTF-8`
  - `PWD`: `/smartflow`
```

What am I doing wrong?

kbenchaaboun:
Not a funny breaking, all our dockers are using pipenv install during build :/

frostming:
If you look at the original issue https://github.com/pypa/pipenv/issues/2210 you can see `pipenv install` can't work properly under root directory and throws a cryptic error message, on 2018.11.26

killuazhu:
Run into the same issue, using fix in https://github.com/pypa/pipenv/issues/4273#issuecomment-635303079 worked for me.

techalchemy:
> Our docker builds failed this morning due to a change in behaviour when running `pipenv install` in the root directory.

can you explain the workflow here -- are you using `--system`?

techalchemy:
As I just mentioned in #4275:

[...]

PR #3386, opened 2018-12-17
Decouple project from the existence of Pipfile.

frostming:
1. Improve message when working under the root directory
    Related: #2210 
    Fixes #3434 

    Quick example:
    ```bash
    $ cd /
    $ pipenv install
    ...
    ERROR:: Pipenv is not intended to work under root directory, please choose another path.
    ```
2. Show the project path in the output of `pipenv --venv`. This may be useful when the `Pipfile` is not under the current path.
3. Allow `$ pipenv --venv` and `$ pipenv --rm` to work even if Pipfile is not present.

PR #4275, opened 2020-05-28
Don't check against root directory

frostming:
### The issue

#3386 causes some regression issue #4273 

### The fix

This PR reverts some of the changes of #3386, don't check against root directory but fix the `self.name` reference in `create_pipfile()`.

Sorry for the trouble.

### The checklist

* [x] Associated issue

### DIFF
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -126,9 +126,6 @@ def do_clear():
 
     try:
         vistir.path.rmtree(PIPENV_CACHE_DIR, onerror=vistir.path.handle_remove_readonly)
-        vistir.path.rmtree(
-            locations.USER_CACHE_DIR, onerror=vistir.path.handle_remove_readonly
-        )
     except OSError as e:
         # Ignore FileNotFoundError. This is needed for Python 2.7.
         import errno
@@ -136,6 +133,8 @@ def do_clear():
         if e.errno == errno.ENOENT:
             pass
         raise
+    # Other processes may be writing into this directory simultaneously.
+    vistir.path.rmtree(locations.USER_CACHE_DIR, ignore_errors=True)
 
 
 def load_dot_env():
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -126,6 +126,12 @@ def do_clear():
 
     try:
         vistir.path.rmtree(PIPENV_CACHE_DIR, onerror=vistir.path.handle_remove_readonly)
+        # Other processes may be writing into this directory simultaneously.
+        vistir.path.rmtree(
+            locations.USER_CACHE_DIR,
+            ignore_errors=environments.PIPENV_IS_CI,
+            onerror=vistir.path.handle_remove_readonly
+        )
     except OSError as e:
         # Ignore FileNotFoundError. This is needed for Python 2.7.
         import errno
@@ -133,8 +139,6 @@ def do_clear():
         if e.errno == errno.ENOENT:
             pass
         raise
-    # Other processes may be writing into this directory simultaneously.
-    vistir.path.rmtree(locations.USER_CACHE_DIR, ignore_errors=True)
 
 
 def load_dot_env():
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -568,14 +568,6 @@ def ensure_project(
         system = True
     if not project.pipfile_exists and deploy:
         raise exceptions.PipfileNotFound
-    # Fail if working under /
-    if not project.name:
-        click.echo(
-            "{0}: Pipenv is not intended to work under the root directory, "
-            "please choose another path.".format(crayons.red("ERROR")),
-            err=True
-        )
-        sys.exit(1)
     # Skip virtualenv creation when --system was used.
     if not system:
         ensure_virtualenv(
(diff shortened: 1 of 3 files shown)


techalchemy:
The main reason for the change in the first place is due to locating virtual environments & related python paths -- as far as I was aware, this was a substantial cause of bugs and breakages and basically didn't work. The fact that it is breaking workflows is the first I am hearing of it working *at all*.

pipenv/project.py
```diff
@@ -685,8 +685,8 @@ def create_pipfile(self, python=None):
         from .vendor.pip_shims.shims import (
             ConfigOptionParser, make_option_group, index_group
         )
-
-        config_parser = ConfigOptionParser(name=self.name)
+        # Inherit the pip's index configuration of install command.
+        config_parser = ConfigOptionParser(name="install")
```
jayaddison, line 689: Is this change intended, @frostming?  It doesn't seem related to the previous root-path handling modification (#3386)

jayaddison: (to clarify, on taking a second look - this change does appear to affect a similar section of code to the previous modifications, but it's not immediately clear whether it's a direct revert of them; it may be the correct way to go, but I figured it'd be worth checking just in case)

frostming: The original issue is that when under the root directory, `ConfigOptionParser` will complain that `sel.name` is empty. The fix in #3386 prevented it fron happening at early stage, which brought a breakage in many users' docker building setup.

So I revert the early checking and instead make `ConfigOptionaParser` happy by passing a meaningful section name to it.
Because giving an arbitrary name doesn't make any sense, which i didn't realize before. This should be the correct path to resolving the issue.

techalchemy, line 689: Arguably, we should change this to `pip_shims.shims.InstallCommand()` and grab the relevant option groups from there. Trying to construct this ourselves out of option groups is foolish and I don't think it actually works correctly.

```suggestion
        command = InstallCommand()
        extra_indexes = command.cmd_opts.get_option("--extra-index-url").default
```

```python
>>> from pip_shims.shims import InstallCommand
>>> cmd = InstallCommand()
>>> cmd.cmd_opts.get_option("--extra-index-url").default
[]
```

pipenv/project.py
```diff
@@ -685,8 +685,8 @@ def create_pipfile(self, python=None):
         from .vendor.pip_shims.shims import (
             ConfigOptionParser, make_option_group, index_group
```
techalchemy, line 686:
```suggestion
            InstallCommand
```

techalchemy:
I think the changes make sense overall -- my only question is -- wasn't one of our main issues regarding the `which` functionality that is partly derived from project locations? If that is completely resolved (which it should be, I would think) then we shouldn't have any issues allowing pipenv usage from `/`

pipenv/project.py
```diff
@@ -685,8 +685,8 @@ def create_pipfile(self, python=None):
...
         )
-
-        config_parser = ConfigOptionParser(name=self.name)
+        # Inherit the pip's index configuration of install command.
+        config_parser = ConfigOptionParser(name="install")
         config_parser.add_option_group(make_option_group(index_group, config_parser))
```
techalchemy, line 690:
```suggestion
```

pipenv/project.py
```diff
@@ -685,8 +685,8 @@ def create_pipfile(self, python=None):
...
         install = config_parser.option_groups[0]
```
techalchemy, line 691:
```suggestion
        install = config_parser.option_groups[0]
        indexes = command.cmd_opts.get_option("--extra-index-url").default
```

and drop the `indexes` defined in the line below

frostming:
> One small update to how we are determining extra index url defaults -- our syntax here is kind of outdated

Agree, do as suggested

>my only question is -- wasn't one of our main issues regarding the which functionality that is partly derived from project locations? If that is completely resolved (which it should be, I would think) then we shouldn't have any issues allowing pipenv usage from /

Yes, when under the root directory `self.name` is just an empty string, which can work in subsequent location calculations.
https://github.com/pypa/pipenv/blob/3f4abadac255d527f94c8fa60590e1472acd814b/pipenv/project.py#L220-L224

pipenv/core.py
```diff
@@ -126,16 +126,15 @@ def do_clear():
...
-        )
     except OSError as e:
         # Ignore FileNotFoundError. This is needed for Python 2.7.
         import errno
 
         if e.errno == errno.ENOENT:
             pass
         raise
+    # Other processes may be writing into this directory simultaneously.
+    vistir.path.rmtree(locations.USER_CACHE_DIR, ignore_errors=True)
```
techalchemy, line 137: yeah i guess that's fine, or maybe `ignore_errors=environments.PIPENV_IS_CI`

techalchemy: (but I'm not sure that's necessary, and don't delay merging over it)

frostming: I didn't know that env var, that looks better.

pipenv/project.py
jayaddison: Thank you for the explanation!

PR #4294, opened 2020-05-29
Fix #3630: Add 'needs_internet' marker to tests missing it

### LINKED ISSUE #3630: Tests using internet but not declared as such

GPHemsley:
### The issue

#3630

### The fix

Add a `needs_internet` marker to:
* `test_pipenv_clean_pip_no_warnings`
* `test_pipenv_clean_pip_warnings`
* `test_environment_variable_value_does_not_change_hash`
* `test_lock_with_incomplete_source`
* `test_lock_after_update_source_name`
* `test_get_source`
* `test_many_indexes`
* `test_run_in_virtualenv_with_global_context`
* `test_run_in_virtualenv`
* `test_mirror_lock_sync`
* `test_pipenv_clean_windows`

### The checklist

* [x] Associated issue

### DIFF
--- a/tests/integration/test_cli.py
+++ b/tests/integration/test_cli.py
@@ -158,6 +158,7 @@ def test_pipenv_check(PipenvInstance):
 
 
 @pytest.mark.cli
+@pytest.mark.needs_internet
 def test_pipenv_clean_pip_no_warnings(PipenvInstance):
     with PipenvInstance(chdir=True) as p:
         with open('setup.py', 'w') as f:
@@ -172,6 +173,7 @@ def test_pipenv_clean_pip_no_warnings(PipenvInstance):
 
 
 @pytest.mark.cli
+@pytest.mark.needs_internet
 def test_pipenv_clean_pip_warnings(PipenvInstance):
     with PipenvInstance(chdir=True) as p:
         with open('setup.py', 'w') as f:
--- a/tests/integration/test_lock.py
+++ b/tests/integration/test_lock.py
@@ -617,6 +617,7 @@ def test_lockfile_with_empty_dict(PipenvInstance):
 @pytest.mark.lock
 @pytest.mark.install
 @pytest.mark.skip_lock
+@pytest.mark.needs_internet
 def test_lock_with_incomplete_source(PipenvInstance):
     with PipenvInstance(chdir=True) as p:
         with open(p.pipfile_path, 'w') as f:
@@ -693,6 +694,7 @@ def test_vcs_lock_respects_top_level_pins(PipenvInstance):
 
 
 @pytest.mark.lock
+@pytest.mark.needs_internet
 def test_lock_after_update_source_name(PipenvInstance):
     with PipenvInstance(chdir=True) as p:
         contents = """
--- a/tests/integration/test_project.py
+++ b/tests/integration/test_project.py
@@ -37,6 +37,7 @@ pytz = "*"
 
 @pytest.mark.project
 @pytest.mark.sources
+@pytest.mark.needs_internet
 @pytest.mark.parametrize('lock_first', [True, False])
 def test_get_source(PipenvInstance, lock_first):
     with PipenvInstance(chdir=True) as p:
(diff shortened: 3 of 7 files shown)


tests/integration/test_cli.py
```diff
@@ -158,6 +158,7 @@ def test_pipenv_check(PipenvInstance):
 
 
 @pytest.mark.cli
+@pytest.mark.needs_internet
 def test_pipenv_clean_pip_no_warnings(PipenvInstance):
```
frostming, line 162:
```suggestion
def test_pipenv_clean_pip_no_warnings(PipenvInstance, pypi):
```

frostming: This case can be easily changed to using mocked PyPI:
```python
c = p.pipenv('run pip -i {} install pytz'.format(pypi.url))
```

tests/integration/test_cli.py
```diff
@@ -172,6 +173,7 @@ def test_pipenv_clean_pip_no_warnings(PipenvInstance):
 
 
 @pytest.mark.cli
+@pytest.mark.needs_internet
```
frostming, line 176: Same as `test_pipenv_clean_pip_no_warnings`

tests/integration/test_lock.py
```diff
@@ -617,6 +617,7 @@ def test_lockfile_with_empty_dict(PipenvInstance):
 @pytest.mark.lock
 @pytest.mark.install
 @pytest.mark.skip_lock
+@pytest.mark.needs_internet
```
frostming, line 620: Already using mocked PyPI, should not require internet access

tests/integration/test_windows.py
```diff
@@ -71,6 +71,7 @@ def test_local_path_windows_forward_slash(PipenvInstance):
 
 
 @pytest.mark.cli
+@pytest.mark.needs_internet
```
frostming, line 74: Same as `test_pipenv_clean_pip_no_warnings`

tests/integration/test_project.py
```diff
@@ -209,6 +212,7 @@ def test_run_in_virtualenv_with_global_context(PipenvInstance, virtualenv):
 
 @pytest.mark.project
 @pytest.mark.virtualenv
+@pytest.mark.needs_internet
```
frostming, line 215: Same as `test_pipenv_clean_pip_no_warnings`

tests/integration/test_lock.py
```diff
@@ -693,6 +694,7 @@ def test_vcs_lock_respects_top_level_pins(PipenvInstance):
 
 
 @pytest.mark.lock
+@pytest.mark.needs_internet
```
frostming, line 697: Already be mocked, or can you elaborate which part needs internet?

frostming:
I added some comments, for cases calling `pip` directly, it can be modified to use mocked PyPI easily

GPHemsley:
@frostming FWIW, I explained my reasoning here: https://github.com/pypa/pipenv/issues/3630#issuecomment-636200385

Not sure if that changes anything.

frostming:
Close the PR in favor of #4449

Issue #4382, opened 2020-07-07
Prompt to verify in case of no Pipfile at context


kobibarhanin:
I've been using pipenv for a while now, and I've grown accustomed to _deleting unnecessary mistakably created_ Pipfiles.

Whenever running pipenv shell / install / (even) check - in case of no Pipfile at context an empty one will be generated, and in almost every case for me this was a result of simply running the pipenv command in the wrong directory.

I suggest prompting the user in such cases to notify that no Pipfile was found, an empty one will be created, and allowing the user to reject at that stage.

In my opinion it will improve usability.

bendem:
Easy fix, large gain, would love to see this.

kobibarhanin:
@bendem I agree, but when attempting to approach this I've realised that since this is a behavioural type of change it has a wide effect on the entire CI testing schema (which fails due to basing on top of the current convention where a Pipfile is auto generated), so it might require more work then anticipated.

PR #4392, opened 2020-07-13
fast commit

kobibarhanin:
This is a draft PR in the spirit of #early-feedback.
This PR addresses issue #4382.
This draft handles only the case of using `pipenv install` with no Pipfile at context, later this solution should be duplicated for `pipenv shell` and `pipenv check`.

So essentially this is simple - detect an install request with no Pipfile - prompt the user, respond to input.
In case of no input / misformatted input - the command terminates with message.

Of course all the irrelevant commenting will be deleted its only there as placeholders for my navigation.

Feedback is much appreciated.

### DIFF
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -1913,6 +1913,18 @@ def do_install(
     editable_packages = editable_packages if editable_packages else []
     package_args = [p for p in packages if p] + [p for p in editable_packages if p]
     skip_requirements = False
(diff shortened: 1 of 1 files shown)


kobibarhanin:
Of course - most of the testing suite is unadjusted to the new prompt so the tests fail.

frostming:
This breaks the automated non-interactive usages, see how the CI fails

Issue #4379, opened 2020-07-06
Python 3.10 is shown as python_version = "3.1" in Pipfile and Pipfile.lock


chriswayg:
### Issue description

Python 3.10 is shown as `python_version = "3.1"` in Pipfile and Pipfile.lock

```
        "requires": {
            "python_version": "3.1"
        },
```

### Expected result

`python_version = "3.10"`

### Steps to replicate

- On a fresh macOS 10.14.6, installed and activated Python 3.10.0a0 (using pyenv)

```
brew install pyenv

echo -e 'export LC_ALL=en_US.UTF-8\nexport LANG=en_US.UTF-8' >> ~/.bash_profile
echo -e 'export PIPENV_VENV_IN_PROJECT=1\nexport PYENV_ROOT=$HOME/.pyenv\nexport PIPENV_PYTHON=$PYENV_ROOT/shims/python\n' >> ~/.bash_profile
echo -e 'if command -v pyenv 1>/dev/null 2>&1; then\n  eval "$(pyenv init -)"\nfi' >> ~/.bash_profile

# in new shell
pyenv install 3.10-dev
pyenv global 3.10-dev
```

- installed latest pipenv and installed the requests package:

```
pip install git+https://github.com/pypa/pipenv.git
mkdir test1 && cd test1
pipenv install requests
pipenv shell
```

- then checked the Pipfiles

-------------------------------------------------------------------------------

<details><summary>$ pipenv --support</summary>

Pipenv version: `'2020.6.2.dev0'`

Pipenv location: `'/Users/editor/.pyenv/versions/3.10-dev/lib/python3.10/site-packages/pipenv'`

Python location: `'/Users/editor/.pyenv/versions/3.10-dev/bin/python3.10'`

Python installations found:

  - `3.10.0`: `/Users/editor/test1/.venv/bin/python3`
  - `3.10.0`: `/Users/editor/test1/.venv/bin/python3`
  - `3.10.0`: `/Users/editor/.pyenv/versions/3.10-dev/bin/python3`
  - `3.10.0`: `/Users/editor/.pyenv/versions/3.10-dev/bin/python3`
  - `3.6.10`: `/Users/editor/.pyenv/versions/3.6.10/bin/python3`
  - `2.7.10`: `/usr/bin/python2.7`

PEP 508 Information:

```
{'implementation_name': 'cpython',
 'implementation_version': '3.10.0a0',
 'os_name': 'posix',
 'platform_machine': 'x86_64',
[...]

PR #4423, opened 2020-08-13
handle python major and minor version correctly in create_pipfile

yosmoc:
### The issue

#4379 

### The fix

Handle python major and minor version correctly in create_pipfile function.
The previous code only handles one digit of the minor version. This change handles the major minor version correctly.

### The checklist

* [x] Associated issue

### DIFF
--- a/pipenv/project.py
+++ b/pipenv/project.py
@@ -756,8 +756,8 @@ class Project(object):
             else:
                 required_python = self.which("python")
         version = python_version(required_python) or PIPENV_DEFAULT_PYTHON_VERSION
-        if version and len(version) >= 3:
-            data[u"requires"] = {"python_version": version[: len("2.7")]}
+        if version and len(version.split(".")) > 2:
+            data[u"requires"] = {"python_version": ".".join(version.split(".")[:2])}
         self.write_toml(data)
 
     @classmethod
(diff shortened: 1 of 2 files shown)


pipenv/project.py
```diff
@@ -757,7 +757,7 @@ def create_pipfile(self, python=None):
                 required_python = self.which("python")
         version = python_version(required_python) or PIPENV_DEFAULT_PYTHON_VERSION
         if version and len(version) >= 3:
-            data[u"requires"] = {"python_version": version[: len("2.7")]}
+            data[u"requires"] = {"python_version": version.rsplit(".", 1)[0]}
```
frostming, line 760:
```suggestion
            data[u"requires"] = {"python_version": ".".join(version.split(".")[:2])}
```
This should be the desired implementation.

pipenv/project.py
```diff
@@ -757,7 +757,7 @@ def create_pipfile(self, python=None):
                 required_python = self.which("python")
         version = python_version(required_python) or PIPENV_DEFAULT_PYTHON_VERSION
         if version and len(version) >= 3:
```
frostming, line 759:
```suggestion
        if version and len(version.split(".")) > 2:
```

frostming (changes requested):

pipenv/project.py
yosmoc: Thanks for reviewing. I have revised the commit.

frostming (approved):

frostming:
Setuptools 50 is breaking the CI, will track it in another PR.

Issue #3686, opened 2019-04-10
Command to list [scripts] entries


uranusjr:
### Is your feature request related to a problem? Please describe.

I sometimes would start typing `pipenv run` and cannot recall what the command I want is named (`release`? `push`? ehh…), and have to ^U and `cat <path to Pipfile>` and look for the `[scripts]` section. It’d be very helpful if I could quickly do this inside my terminal.

### Describe the solution you'd like

Something like

```console
$ pipenv list-scripts
release     python release
test        pytest
format      black src tests
```

techalchemy:
`list-scripts` is a bit verbose, my first instinct was `pipenv scripts` -- thoughts? Either way it seems odd to add a full command for just this -- would we want other functionality with it?

Invoke uses `--list` as well, we can consider `pipenv run --list` as an option (but I don't particularly like it)

uranusjr:
I don’t really have any preferences, all of the mentioned options would work fine for me.

brendanator:
I think this should be rolled into the `pipenv run` command when no command is provided.
When I run `npm run` with no arguments I get:
```
Lifecycle scripts included in example-project:
  start
    npm run develop
  test
    echo "Write tests! -> https://gatsby.dev/unit-testing"

available via `npm run-script`:
  build
    gatsby build
  clean
    gatsby clean
  develop
    gatsby develop
```
I assume this would no longer need a PEEP

As an additional point I think `pipenv run script_command` should print out the command it is running. npm does this and it is useful, especially when passing additional parameters

GlenRSmith:
A quick hack to accomplish this:

Install pipenv as a dev requirement:

>pipenv install -d pipenv

Add a `script` to your Pipfile:

```
[scripts]
list = "./pipenv_list_scripts.py"
```

Add `pipenv_list_scripts.py`:

```
#!/usr/bin/env python
"""
Utility to allow pipenv command listing `scripts` from Pipfile
"""
try:
    import pipenv.core
except ModuleNotFoundError:
    print("This script is only for use in a development environment")
    exit(1)


def main():
    proj = pipenv.core.project
    if not proj:
        print("project not found")
        exit(1)
    scripts = proj.parsed_pipfile.get('scripts', [])
    print("command\tscript")
    for k, v in scripts.items():
        print(f"{k}\t{v}")
    return


if __name__ == '__main__':
    main()
```

uranusjr:
[...]

PR #4445, opened 2020-08-28
Add `pipenv scripts` command to list scripts in Pipfile

GlenRSmith:
Enhancement discussed in #3686

### DIFF
--- a/pipenv/cli/command.py
+++ b/pipenv/cli/command.py
@@ -664,6 +664,7 @@ def run_open(state, module, *args, **kwargs):
     short_help="Installs all packages specified in Pipfile.lock.",
     context_settings=CONTEXT_SETTINGS
 )
+@system_option
 @option("--bare", is_flag=True, default=False, help="Minimal output.")
 @sync_options
 @pass_state
@@ -691,6 +692,7 @@ def sync(
         unused=unused,
         sequential=state.installstate.sequential,
         pypi_mirror=state.pypi_mirror,
+        system=state.system
     )
     if retcode:
         ctx.abort()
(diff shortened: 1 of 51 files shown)


frostming:
A news fragment is needed to notice what is new. You can place it under the `news/` directory, see the other files about how to write one.

uranusjr:
Maybe we should name this `scripts` instead for consistency. I misspelled `[scripts]` as `[script]` recently and spent almost half an hour wondering why my scripts don’t get picked up by Pipenv.

tests/integration/test_cli.py
```diff
@@ -197,6 +197,19 @@ def test_bare_output(PipenvInstance):
...
+def test_scripts(PipenvInstance):
+    with PipenvInstance() as p:
+        with open(p.pipfile_path, "w") as f:
+            contents = """
+[scripts]
+pyver = "which python"
+            """.strip()
+            f.write(contents)
+        c = p.pipenv('scripts')
+        assert 'pyver' in c.out
```
uranusjr, line 210: I would also test the `which python` part is also in the output.

frostming (changes requested):
One more step before merge: add a news entry file under `news/` directory. @GlenRSmith

pipenv/cli/command.py
```diff
@@ -716,5 +716,26 @@ def clean(ctx, state, dry_run=False, bare=False, user=False):
...
+    from ..core import project
+    if not project:
+        echo(u"project not found", err=True)
+        exit(1)
+    scripts = project.parsed_pipfile.get('scripts', {})
+    rpt = u"command\tscripts\n"
+    for k, v in scripts.items():
+        rpt += u"{0}\t{1}".format(k, v)
+    echo(rpt)
+    return
```
uranusjr, line 737: This return has no function.

GlenRSmith: Fixed.

GlenRSmith:
@frostming Added. Sorry for the omission.

PR #4506, opened 2020-10-29
Remove not accurate claim

ttm56p:
### The issue

Readme contains a misleading claim "*Windows is a first-class citizen, in our world.*".
Since claim itself or use of pipenv overall in Windows is not covered by docs by any means.
Removal of this claim should help save time for newcomers to match alternatives with Windows support.
Claim should be restored once docs properly covering work on Windows and claim can be considered as accurate.

### The fix

Removes a claim for better times.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,6 @@ Pipenv: Python Development Workflow for Humans
 
 **Pipenv** is a tool that aims to bring the best of all packaging worlds
 (bundler, composer, npm, cargo, yarn, etc.) to the Python world.
-*Windows is a first-class citizen, in our world.*
 
 It automatically creates and manages a virtualenv for your projects, as
 well as adds/removes packages from your `Pipfile` as you


frostming:
Sorry, but I can't get what this PR is aiming to do. We are doing a lot of jobs getting Pipenv work on the Windows platform. And I personally also mainly use Windows + Powershell. If you encounter any problem with Pipenv on Windows, please file an issue. I would like to close this PR now.

ttm56p:
@frostming pipenv docs has poor(only found path to cache on Windows is given and nothing else) coverage for Windows.
Here is some links to pages for alternative package system that covers work with Windows, that pipenv is lacking.

https://python-poetry.org/docs/#windows-powershell-install-instructions
https://python-poetry.org/docs/basic-usage/#activating-the-virtual-environment

While pipenv docs provide commands for unix based oses only.

Having this in mind claim "Windows is a first-class citizen, in our world" sounds inaccurate.

Do you understand my point of view?

Overall I appreciate the work you doing!

ttm56p:
@frostming my change should save time for people checking docs in search of instructions for Windows.
For claim "Windows is a first-class citizen, in our world" it is expected to have best user experience + docs coverage across all OS-es.
From what I see - Linux is a first-class citizen for pipenv.

uranusjr:
I use pipenv on Windows and have not had platform specific problems. If you find situations where pipenv works on Linux but not Windows, report them, and we’ll work on fixes. It does not help to throw claims without explanation.

PR #4510, opened 2020-10-30
Fix #4225 - virtualenv name when PIPENV_PYTHON set to filepath

### LINKED ISSUE #4225: Virtualenv name gets botched when PIPENV_PYTHON is a full path and WORKON_HOME is set

Check the [diagnose documentation](https://pipenv.pypa.io/en/latest/diagnose/) for common issues before posting! We may close your issue if it is very similar to one of them. Please be considerate, or be on your way.

Make sure to mention your debugging experience if the documented solution failed.

### Issue description

When trying to `pipenv install` using `PIPENV_PYTHON=$HOME/.pyenv/shims/python` and `WORKON_HOME=$HOME/.pyenv/versions`, the `Project.virtualenv_name` property includes slashes. This causes the virtualenv to get botched since it installs within nested directories.

### Expected result

Generated virtualenv name is something that can be a reasonable folder name.

### Actual result

```
❯ WORKON_HOME=$HOME/.pyenv/versions PIPENV_PYTHON=$HOME/.pyenv/shims/python PYENV_VERSION=3.7.7 pipenv install --dev
Pipfile: /Users/Diego/work/lucerna/test/Pipfile
Using /Users/Diego/.pyenv/shims/python (3.7.7) to create virtualenv…
⠦ Creating virtual environment...created virtual environment CPython3.7.7.final.0-64 in 691ms
  creator CPython3Posix(dest=/Users/Diego/.pyenv/versions/test-VgJQqnWV-/Users/Diego/.pyenv/shims/python, clear=False, global=False)
  seeder FromAppData(download=False, pip=latest, setuptools=latest, wheel=latest, via=copy, app_data_dir=/Users/Diego/Library/Application Support/virtualenv/seed-app-data/v1.0.1)
  activators BashActivator,CShellActivator,FishActivator,PowerShellActivator,PythonActivator,XonshActivator

Virtualenv location: /Users/Diego/.pyenv/versions/test-VgJQqnWV-/Users/Diego/.pyenv/shims/python
```

Honza-m:
### The issue

Fixing #4225

### The fix

If PIPENV_PYTHON is a path, only the last bit of the path gets included in the suffix, thus making sure that the environment name contains no incorrect characters. If it's a python version, the suffix is the version as previously intended. If PIPENV_PYTHON is not set, there is no suffix to the environment name.

I couldn't figure out an appropriate way to write tests for this since there don't seem to be tests created for the original function and new tests would depend on my pyenv setup, but if anyone knows how to test this effectively please let me know.

### DIFF
--- a/news/4225.bugfix.rst
+++ b/news/4225.bugfix.rst
@@ -1,0 +1,1 @@
+Fix a bug where setting PIPENV_PYTHON to file path breaks environment name
--- a/pipenv/project.py
+++ b/pipenv/project.py
@@ -443,7 +443,13 @@ class Project(object):
     def virtualenv_name(self):
         # type: () -> str
         sanitized, encoded_hash = self._get_virtualenv_hash(self.name)
-        suffix = "-{0}".format(PIPENV_PYTHON) if PIPENV_PYTHON else ""
+        suffix = ""
+        if PIPENV_PYTHON:
+            if os.path.isabs(PIPENV_PYTHON):
+                suffix = "-{0}".format(os.path.basename(PIPENV_PYTHON))
+            else:
+                suffix = "-{0}".format(PIPENV_PYTHON)
+
         # If the pipfile was located at '/home/user/MY_PROJECT/Pipfile',
         # the name of its virtualenv will be 'my-project-wyUfYPqE'
         return sanitized + "-" + encoded_hash + suffix


frostming:
thanks for working on this, care to add a news entry?

Honza-m:
First pull request, sorry 😅 Here you go.

frostming (approved):

PR #4568, opened 2020-12-15
Fix description of the compatible-release operator

deepyaman:

### The issue

I didn't create a GitHub issue, because the proposed change is trivial.

### The checklist

* [ ] Associated issue


frostming:
Thanks for the patch

PR #4574, opened 2020-12-22
Remove broken link

mrcalderon:


uranusjr:
I searched for a while, and unfortunately the essay is no longer available anywhere. This is probably the best we can do.

Issue #4602, opened 2021-01-27
Bad version parsing in Python 3.10


tucked:
`pep508checker.py` assumes that Python will always use a single digit for the minor version.
https://github.com/pypa/pipenv/blob/62a6a0e7703bd5f4112dcbe5972f540259a54182/pipenv/pep508checker.py#L34

### Steps to replicate
``` sh
edit Pipfile
```
``` toml
[requires]
python_version = "3.10"
```
``` sh
pipenv check --verbose
```
```
Checking PEP 508 requirements...
Running command: $ /home/tucked/.local/share/virtualenvs/tmp.VtEw2GLJ9V-cilwcAZk/bin/python /home/tucked/.local/pipx/venvs/pipenv/lib/python3.6/site-packages/pipenv/pep508checker.py
Command output: {"os_name": "posix", "sys_platform": "linux", "platform_machine": "x86_64", "platform_python_implementation": "CPython", "platform_release": "4.15.0-134-generic", "platform_system": "Linux", "platform_version": "#138-Ubuntu SMP Fri Jan 15 10:52:18 UTC 2021", "python_version": "3.1", "python_full_version": "3.10.0a2+", "implementation_name": "cpython", "implementation_version": "3.10.0a2"}

Specifier python_version does not match 3.10 (3.1).
Failed!
```

PR #4603, opened 2021-01-27
Correct python_version in pep508checker.py

tucked:
### The issue

Fix https://github.com/pypa/pipenv/issues/4602

### The fix

This includes all characters up to the second `"."` (instead of just the first 3 characters),
which should correct version-handling for Python 3.10.

### The checklist

* [x] Associated issue


frostming:
Thanks for the fix

PR #4605, opened 2021-01-28
drop python2 from CI

dulmandakh:
### The issue

PIP dropped Python 2 support https://github.com/pypa/pip/issues/6148, therefore pipenv needs to drop Python 2 and make performance improvements.

I would like to volunteer sometime to drop Python 2.

### The fix

Let's drop Python 2 from CI, then we can do other cleanups safely.

### The checklist

* [ ] Associated issue

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -39,7 +39,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [2.7, 3.6, 3.7, 3.8]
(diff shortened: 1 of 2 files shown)


frostming:
Thanks for that, to drop Python 2 there are some more to do:

- Remove six and other backport libraries
- Modernize the codebase to Py3.6+
- Remove compatible codes

I will merge this shortly

dulmandakh:
@frostming I would like to work on removing **six** once this merged. Thanks

PR #4607, opened 2021-01-29
PEEP-007: Accepting pre-releases for specific packages

fridex:
Proposing PEEP-007 which would allow specifying pre-releases only for selected packages. This has been discussed in https://github.com/pypa/pipenv/issues/1760, added the original author of the issue (@gsemet). Let me know if more people should be marked as authors and I would be happy for any feedback.

### DIFF
--- a/peeps/PEEP-007.md
+++ b/peeps/PEEP-007.md
@@ -1,0 +1,51 @@
+# PEEP-007: Accepting pre-releases for specific packages
+
+**PROPOSED**
+
+This PEEP describes a change that would allow installing pre-release packages
+selectively.
+
+☤
+
+## Installing pre-releases
+
+The current implementation of Pipenv offers capability to install pre-releases
+based on the ``--pre`` flag supplied. Another option is to enable installing
+pre-releases in the Pipfile itself:
+
+```toml
+[pipenv]
+allow_prereleases = true
+```
+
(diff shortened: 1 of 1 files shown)


fargusplumdoodle (approved):
I would love to see this happen

bryant-finney:
This would be fantastic. 

Currently, [psf/black](https://github.com/psf/black) can only be installed as a pre-release. This requires us to pin the `black` version in [pipenv-setup](https://github.com/Madoshakalaka/pipenv-setup); otherwise, dependent packages must set `allow_prereleases` or encounter package resolution errors.

gitpushdashf:
I personally like to just do `black = "==21.12b0"` and not bother with the `allow_prereleases` setting.

matteius:
@fridex  -- Thanks for doing this effort on the PEEP.  I think you have buy in and we can merge this if you still plan to move forward with it, though I am not sure we are sticking with the PEEP process these days.  FWIW I was mentioning to @oz123 earlier it might be coo to try moving more of the design discussions into the new Discussions section on github instead of requiring PEEPs...   Though your outcome here seems pretty complete so if you want to support adding this feature go ahead and work through it.  I think what is in the PEEP here is a good basis for things to pull into the documentation in the PR that implements the feature.

Issue #4538, opened 2020-11-18
Ampersand (&) in folder name breaks pipenv shell


anton3s:
### Issue description

When a new pipenv is created in a folder that contains an ampersand the new environment can't be activated with pipenv shell.

### Expected result

A successfully activated environment.

### Actual result

Error which indicates the path was split on the &.

```
No such file: B-0WFjT3c-/bin/activate
```
### Steps to replicate

```
mkdir A\&B
pipenv shell
```

### Steps to fix

The regular expression in project.Project._sanitize should also include &.

### Workaround

You can create a pipenv in a subfolder of the folder.

PR #4616, opened 2021-02-07
Fix #4615 - Improve sanitization to allow virtualenv paths with () and [] characters to activate

### LINKED ISSUE #4615: Parentheses escaping bug when running `pipenv shell`
### Summary:

When running `pipenv shell` in POSIX (zsh), running `/bin/activate` will fail if there are parentheses in the virtualenv path due to the regex in `_get_activate_script()` (Line 53, `pipenv/shells.py`) not escaping parentheses properly.

### How to reproduce the issue:

1. Create a new directory `a ( )`
2. Create `.venv` folder inside the directory
3. Run `pipenv install` in `a ( )`
4. Run `pipenv shell`. The issue should now be reproduced.
5. To verify, run `pipenv shell` again to confirm the improper activation.

### What is expected to happen:

Step 4 successfully activates virtual environment without error return code (1).

### What actually happens:

Step 4 output:

``` zsh
$ pipenv shell
Launching subshell in virtual environment...
 . /Users/leohku/a\ (\ )/.venv/bin/activate
zsh: no matches found: /Users/leohku/a ( )/.venv/bin/activate
```

Step 5 output:

``` zsh
$ pipenv shell
Shell for UNKNOWN_VIRTUAL_ENVIRONMENT already activated.
No action taken to avoid nested environments.
```

### Environment:

pipenv, version 2020.11.15
running zsh on macOS Big Sur v11.1 x86

leohku:
### The issue

Fixes #4615.
Related to #4538.

### The fix

Added proper character escaping for `()` and `[]` characters in virtualenv paths, by modifying the regex in `_get_activate_script()` in `pipenv/shells.py`.

Also improved sanitization in `_sanitize()` of `Project` class in `pipenv/project.py` to turn `()` and `[]` characters into `_`.

These changes allows virtualenv paths with such characters to properly activate.

### The checklist

* [x] Associated issue

### DIFF
--- a/news/4615.bugfix.rst
+++ b/news/4615.bugfix.rst
@@ -1,1 +1,1 @@
-Fix bug where environment wouldn't activate in paths containing () and [] symbols - Related to #4538
+Fix bug where environment wouldn't activate in paths containing () and [] symbols
--- a/news/4615.bugfix.rst
+++ b/news/4615.bugfix.rst
@@ -1,0 +1,1 @@
+Fix bug where environment wouldn't activate in paths containing () and [] symbols - Related to #4538
--- a/pipenv/project.py
+++ b/pipenv/project.py
@@ -395,7 +395,7 @@ class Project(object):
         #   https://www.gnu.org/software/bash/manual/html_node/Double-Quotes.html
         #   http://www.tldp.org/LDP/abs/html/special-chars.html#FIELDREF
         #   https://github.com/torvalds/linux/blob/2bfe01ef/include/uapi/linux/binfmts.h#L18
-        return re.sub(r'[ &$`!*@"\\\r\n\t]', "_", name)[0:42]
+        return re.sub(r'[ &$`!*@"()\[\]\\\r\n\t]', "_", name)[0:42]
 
     def _get_virtualenv_hash(self, name):
         # type: (str) -> str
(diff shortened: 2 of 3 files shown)


news/4615.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fix bug where environment wouldn't activate in paths containing () and [] symbols - Related to #4538
```
uranusjr, line 1: Since #4538 is already listed in the changelog, it does not need to be included here.

pipenv/project.py
```diff
@@ -395,7 +395,7 @@ def _sanitize(cls, name):
         #   https://www.gnu.org/software/bash/manual/html_node/Double-Quotes.html
         #   http://www.tldp.org/LDP/abs/html/special-chars.html#FIELDREF
         #   https://github.com/torvalds/linux/blob/2bfe01ef/include/uapi/linux/binfmts.h#L18
-        return re.sub(r'[ &$`!*@"\\\r\n\t]', "_", name)[0:42]
+        return re.sub(r'[ &$`!*@"()\[\]\\\r\n\t]', "_", name)[0:42]
```
uranusjr, line 398: Technically you don’t need to escape `[`, only `]`. Not a big deal though.

news/4615.bugfix.rst
leohku: Got it, thanks. It's updated in the latest push.

uranusjr (approved):

PR #4646, opened 2021-03-08
re-fix broken link

mabuelhagag:
### The issue

A broken link on the index page.
### The fix

Replaced with the correct location in Kenneth's webiste.


frostming:
Thanks for the PR

Issue #4642, opened 2021-03-04
pipenv lock --pre does not create a lockfile with prerelease version


nid5:
### Issue description

When running pipenv lock --pre with a non-specific version in the Pipfile, only release versions are included in the Pipfile.lock

### Expected result

Prerelease version in Pipfile.lock

### Actual result

Release versions in Pipfile.lock

### Steps to replicate

Pipfile containing non-specific version of a package, e.g., mylibrary = ">=1.2.3"

Pipenv lock -pre

Examine Pipfile.lock

In my attempts at debugging this issue I placed breakpoints in a variety of files. It seems that in my case, the issue was occurring around https://github.com/pypa/pipenv/blob/bf7203148e1e0f57ab86b652d76135f17bf66927/pipenv/utils.py#L774-L779
In my case self._repository is none, so a new PypiRepository is created. Examining PypiRepository shows that the self.options field, which is later referenced in the package finder when determining whether or not to use prereleases, is set using the 'pip_args' passed when the PypiRepository is initialized. In my case the pip_args passed contained no reference to prereleases, despite the fact that self.pre (referencing the Resolver object) is set to True. Uncertain if there is a case where self._repository already exists at this point, but I couldn't replicate a situation where it existed.
-------------------------------------------------------------------------------

<details><summary>$ pipenv --support</summary>

Pipenv version: `'2020.11.15'`

Pipenv location: `'/Users/nid5/.pyenv/versions/3.7.8/lib/python3.7/site-packages/pipenv'`

Python location: `'/Users/nid5/.pyenv/versions/3.7.8/bin/python3.7'`

Python installations found:

  - `3.8.2`: `/usr/bin/python3`
  - `3.7.8`: `/Users/nid5/.pyenv/versions/3.7.8/bin/python3`
  - `3.7.8`: `/Users/nid5/.pyenv/versions/3.7.8/bin/python3`
  - `3.7.8`: `/Users/nid5/.pyenv/versions/3.7.8/bin/python3.7m`
  - `3.7.8`: `/Users/nid5/.pyenv/versions/3.7.8/bin/python3.7`
  - `2.7.16`: `/usr/bin/python2`
  - `2.7.16`: `/usr/bin/python2.7`

PEP 508 Information:

```
{'implementation_name': 'cpython',
 'implementation_version': '3.7.8',
 'os_name': 'posix',
 'platform_machine': 'x86_64',
 'platform_python_implementation': 'CPython',
 'platform_release': '19.6.0',
[...]

PR #4652, opened 2021-03-18
4642 fix pipenv lock --pre doesn't include prerelease dependencies

nid5:
### The issue

#4642 
pipenv lock --pre doesn't actually include prerelease dependencies

### The fix

Used the already existing self.pre variable to append --pre to a built command that ultimately results in prerelease dependencies being included in the locking process. This seems to be consistent with conventions at least within that file, and doesn't require additional fields to be defined.

### The checklist

* [x] Associated issue

### DIFF
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -705,7 +705,7 @@ class Resolver(object):
             pip_args.append("--no-use-pep517")
         if build_isolation is False:
             pip_args.append("--no-build-isolation")
-        if pre is True:
+        if self.pre:
             pip_args.append("--pre")
         pip_args.extend(["--cache-dir", environments.PIPENV_CACHE_DIR])
         return pip_args
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -697,7 +697,7 @@ class Resolver(object):
             self._pip_command = self._get_pip_command()
         return self._pip_command
 
-    def prepare_pip_args(self, use_pep517=False, build_isolation=True, pre=False):
+    def prepare_pip_args(self, use_pep517=False, build_isolation=True):
         pip_args = []
         if self.sources:
             pip_args = prepare_pip_source_args(self.sources, pip_args)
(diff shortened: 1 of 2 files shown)


pipenv/utils.py
```diff
@@ -697,24 +697,27 @@ def pip_command(self):
...
-    def prepare_pip_args(self, use_pep517=False, build_isolation=True):
+    def prepare_pip_args(self, use_pep517=False, build_isolation=True, pre=False):
         pip_args = []
         if self.sources:
             pip_args = prepare_pip_source_args(self.sources, pip_args)
         if use_pep517 is False:
             pip_args.append("--no-use-pep517")
         if build_isolation is False:
             pip_args.append("--no-build-isolation")
+        if pre is True:
```
uranusjr, line 708:
```suggestion
        if self.pre:
```

So you don’t need to pass the value around.

uranusjr (changes requested):

pipenv/utils.py
```diff
@@ -697,24 +697,27 @@ def pip_command(self):
...
 
     @property
     def pip_args(self):
+        pre = self.pre
         use_pep517 = environments.get_from_env("USE_PEP517", prefix="PIP")
         build_isolation = environments.get_from_env("BUILD_ISOLATION", prefix="PIP")
         if self._pip_args is None:
             self._pip_args = self.prepare_pip_args(
-                use_pep517=use_pep517, build_isolation=build_isolation
+                use_pep517=use_pep517, build_isolation=build_isolation, pre=pre
```
frostming, line 720: So we can drop the `pre` argument now since it isn't used any more?

nid5: New commit removes unnecessary argument.

frostming (approved):

uranusjr (approved):

frostming:
Sorry to miss the updates here, will merge after the CI finishes

Issue #4689, opened 2021-05-03
Safety vulnerability for pip


InvisibleMan1306:
### Issue description

The latest version of pipenv will install pip version 20.3.4 but all versions under 21.1 are marked as having a vulnerability according to safety.

### Expected result

No vulnerability error.

### Actual result

```
+==============================================================================+
|                                                                              |
|                               /$$$$$$            /$$                         |
|                              /$$__  $$          | $$                         |
|           /$$$$$$$  /$$$$$$ | $$  \__//$$$$$$  /$$$$$$   /$$   /$$           |
|          /$$_____/ |____  $$| $$$$   /$$__  $$|_  $$_/  | $$  | $$           |
|         |  $$$$$$   /$$$$$$$| $$_/  | $$$$$$$$  | $$    | $$  | $$           |
|          \____  $$ /$$__  $$| $$    | $$_____/  | $$ /$$| $$  | $$           |
|          /$$$$$$$/|  $$$$$$$| $$    |  $$$$$$$  |  $$$$/|  $$$$$$$           |
|         |_______/  \_______/|__/     \_______/   \___/   \____  $$           |
|                                                          /$$  | $$           |
|                                                         |  $$$$$$/           |
|  by pyup.io                                              \______/            |
|                                                                              |
+==============================================================================+
| REPORT                                                                       |
| checked 115 packages, using free DB (updated once a month)                   |
+============================+===========+==========================+==========+
| package                    | installed | affected                 | ID       |
+============================+===========+==========================+==========+
| pip                        | 20.3.3    | <21.1                    | 40291    |
+==============================================================================+
| Pip 21.1 stops splitting on unicode separators in git references, which      |
| could be maliciously used to install a different revision on the repository. |
| See: <https://github.com/pypa/pip/issues/9827>. Additionally, pip 21.1       |
| updates urllib3 to 1.26.4 to fix CVE-2021-28363.                             |
+==============================================================================+
``` 

### Steps to replicate

```bash
pipenv install safety
pipenv run safety check
```

<details><summary>$ pipenv --support</summary>

Pipenv version: `'2020.11.15'`

Pipenv location: `'/Users/vkrylov/miniconda3/lib/python3.8/site-packages/pipenv'`

Python location: `'/Users/vkrylov/miniconda3/bin/python'`

Python installations found:

  - `3.8.5`: `/Users/vkrylov/miniconda3/bin/python3`
[...]

PR #4691, opened 2021-05-05
Update __init__.py

carllopez:
Updates pip version to fix security issue reported in pipenv check

### The issue

Using pip <21.1 makes `pipenv check` report a security issue:

```
Pip 21.1 stops splitting on unicode separators in git references, which could be maliciously used to install a different revision on the repository. See: <https://github.com/pypa/pip/issues/9827>. Additionally, pip 21.1 updates urllib3 to 1.26.4 to fix CVE-2021-28363.
```

More info here: https://github.com/pypa/pipenv/issues/4689

### The fix

Pins the latest version of `pip`.

### The checklist

* [ ] Associated issue

### DIFF
--- a/pipenv/patched/notpip/__init__.py
+++ b/pipenv/patched/notpip/__init__.py
@@ -4,7 +4,7 @@ if MYPY_CHECK_RUNNING:
     from typing import List, Optional
 
 
-__version__ = "20.0.2"
+__version__ = "21.1.1"
 
 
 def main(args=None):


frenzymadness:
Would not be better to actually fix the vulnerability rather to making the security scanner happy by changing the version? How does this help to make the bundled pip not vulnerable?

uranusjr:
@frenzymadness is correct, this does not actually fix anything, only “lies” to safety so it fails to correctly detect the vulnerability.

PR #4692, opened 2021-05-07
Windows install command

mikepsinn:
### The issue

The readme doesn't have the command to install on Windows. 

### The fix

Added the command to install on Windows to the readme.


luischapis (commented):
#

PR #4723, opened 2021-06-18
Replace call to add_stylesheet with add_css_file

befeleme:
This way it is possible to build documentation with Sphinx 4.x

### The issue

Package documentation stopped building with Sphinx 4 as a result of removal of `add_stylesheet` function.
Using `add_css_file` instead makes it possible to build on both Sphinx 3.5.x and Sphinx 4.x

### The fix

Replacement of a function call, no other measurements are needed.

### The checklist

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -124,7 +124,7 @@ html_static_path = ['_static']
 
 
 def setup(app):
-    app.add_stylesheet('custom.css')
+    app.add_css_file('custom.css')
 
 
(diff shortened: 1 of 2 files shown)


matteius:
I did this recently with:  https://github.com/pypa/pipenv/pull/5032/files#diff-85933aa74a2d66c3e4dcdf7a9ad8397f5a7971080d34ef1108296a7c6b69e7e3R134

Issue #2183, opened 2018-05-12
The `pipenv install --deploy --system` command is no longer working.


ryan-blunden:
Be sure to check the existing issues (both open and closed!).

The `pipenv install --deploy --system` command is no longer working.

<details><summary>$ python -m pipenv.help output</summary>

Pipenv version: `'11.10.4'`

Pipenv location: `'/usr/local/lib/python3.6/site-packages/pipenv'`

Python location: `'/usr/local/bin/python'`

Other Python installations in `PATH`:

  - `3.6`: `/usr/local/bin/python3.6m`
  - `3.6`: `/usr/local/bin/python3.6`
  - `3.6`: `/usr/local/bin/python3.6`

  - `3.6.5`: `/usr/local/bin/python`
  - `3.6.5`: `/usr/local/bin/python`
  - `3.6.5`: `/usr/local/bin/python3`
  - `3.6.5`: `/usr/local/bin/python3`

PEP 508 Information:

```
{'implementation_name': 'cpython',
 'implementation_version': '3.6.5',
 'os_name': 'posix',
 'platform_machine': 'x86_64',
 'platform_python_implementation': 'CPython',
 'platform_release': '4.9.87-linuxkit-aufs',
 'platform_system': 'Linux',
 'platform_version': '#1 SMP Fri Mar 16 18:16:33 UTC 2018',
 'python_full_version': '3.6.5',
 'python_version': '3.6',
 'sys_platform': 'linux'}
```

System environment variables:

  - `HOSTNAME`
  - `SHLVL`
  - `PYTHON_PIP_VERSION`
  - `HOME`
  - `GPG_KEY`
  - `PATH`
  - `LANG`
  - `PYTHON_VERSION`
  - `PWD`
  - `PYTHONDONTWRITEBYTECODE`
  - `PIP_PYTHON_PATH`

Pipenv–specific environment variables:

Debug–specific environment variables:

  - `PATH`: `/usr/local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin`
  - `LANG`: `C.UTF-8`
  - `PWD`: `/usr/src/app`

---------------------------

Contents of `Pipfile` ('/usr/src/app/Pipfile'):

```toml
[[source]]
url = "https://pypi.python.org/simple"
verify_ssl = true
name = "pypi"

[packages]
records = "*"

[dev-packages]

[requires]
python_version = "3.6"

```

</details>

------------

##### Expected result

[...]

PR #4733, opened 2021-07-10
docs: make --system error message more clear

chapmanjacobd:
There are a few issues like this: https://github.com/pypa/pipenv/issues/2183

I think people generally understand the reason behind it but the message is a little vague


matteius:
I am declining this old PR in favor of work being done to expose the ability to allow system installs for users that accept the risks to do so, work on-going in:  https://github.com/pypa/pipenv/pull/5144

PR #4738, opened 2021-07-16
Improve documentation for PIPENV_YES / Automatic Installation of python versions

TechnicalPirate:
### The issue

When coming to improve our CI infrastructure, It took considerable time to deduce why `pipenv` wasn't giving us a prompt to install python versions specificed in `pipfile` `[requires]` - When reverse shelling into the build agent we would be prompted but not when executing a job remotely. After investigation we found that the `PIPENV_YES` variable must be set for this functionality to work correctly.

### The fix

This PR updates the documentation both for `PIPENV_YES` and also for the `☤ Automatic Python Installation` page to clarify that the environment variable needs setting for `pipenv` and `pyenv` to interact correctly when in a CI environment to make it easier for other Build and Config engineers to get this awesome functionality working :)


matteius:
Superseded by: https://github.com/pypa/pipenv/pull/5793

Issue #4760, opened 2021-08-09
Shell-quote `pip` command when logging


allanlewis:
### Is your feature request related to a problem? Please describe.

In v2021.5.29, `pip` commands are logged as lists. a82bbb7b amended this to join the lists with spaces. However, this will lead to invalid commands if any token contains a space or some other special character that needs to be escaped for standard shells.

### Describe the solution you'd like

Tokens in commands should be shell-quoted for logging.

### Describe alternatives you've considered

I'm not aware of any good alternatives to `shlex.quote`, nor any reason to consider any given the simplicity of the solution.

### Additional context

n/a

-------------------------------------------------------------------------------

Please run `$ pipenv --support`, and paste the results here. Don't put backticks (`` ` ``) around it! The output already contains Markdown formatting.

I don't feel this is relevant here; the description above should be self-explanatory.

I intend to address this in a PR myself.

PR #4761, opened 2021-08-09
Quote tokens when logging `pip` commands

allanlewis:
### The issue

See #4760.

### The fix

In order to improve the logging, this commit shell-quotes each token before joining the results with spaces. This means that users should be able to copy and paste logged commands and run them using any common shell without issue, including when paths contain spaces or other special characters.

### The checklist

* [x] Associated issue

### DIFF
--- a/news/4760.feature.rst
+++ b/news/4760.feature.rst
@@ -1,0 +1,1 @@
+Shell-quote ``pip`` commands when logging.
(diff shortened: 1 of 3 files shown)


allanlewis:
Please advise if I should add a unit test for `cmd_list_to_shell`.

frostming:
That is okay, thanks.

allanlewis:
Thanks for the quick response, @frostming 🙏

Issue #4775, opened 2021-08-27
pipenv ignore environment variable `PIP_TARGET` when calculate dependancy delta.


cm-iwata:
### Issue description

Sometimes I use pipenv with environment variable `PIP_TARGET` for install libraries to specified directory.
But pipenv will skip install libraries if libraries already install to default location.

### Expected result

pipenv should install libraries to specified directory when exec with environment variable `PIP_TARGET`.

### Actual result

pipenv will skip install libraries to specified directory when libraries already installed to default location(`.venv/lib/pythonx.x/site-packages`).

### Steps to replicate

1. install some library and create Pipfile.lock

```
pipenv install six
```

2. Create some directory for specify pip install target.

```mkdir layers
mkdir layer
```

3. Try install library to specified directory.

```
PIP_TARGET=layer  pipenv sync
Installing dependencies from Pipfile.lock (34ad89)…
  🐍   ▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉ 0/0 — 0
```

Install library will skip and output `0/0 — 00:00:00`

4. List specified directory.There are no library.

```
ls layer
```

This command will return empty result.

<details><summary>$ pipenv --support</summary>

Pipenv version: `'2020.6.2'`

Pipenv location: `'/Users/cm-iwata/.anyenv/envs/pyenv/versions/3.8.3/lib/python3.8/site-packages/pipenv'`

Python location: `'/Users/cm-iwata/.anyenv/envs/pyenv/versions/3.8.3/bin/python3.8'`

Python installations found:

  - `3.9.2`: `/usr/local/bin/python3`
  - `3.9.2`: `/usr/local/bin/python3.9`
  - `3.8.3`: `/Users/cm-iwata/.anyenv/envs/pyenv/versions/3.8.3/bin/python3`
  - `3.8.3`: `/Users/cm-iwata/.anyenv/envs/pyenv/versions/3.8.3/bin/python3.8`
  - `3.8.3`: `/Users/cm-iwata/.anyenv/envs/pyenv/versions/3.8.3/bin/python3`
  - `3.7.8`: `/Users/cm-iwata/.anyenv/envs/pyenv/versions/3.7.8/bin/python3`
  - `3.7.3`: `/usr/bin/python3`
  - `3.6.5`: `/Users/cm-iwata/.anyenv/envs/pyenv/versions/3.6.5/bin/python3`
[...]

PR #4776, opened 2021-08-27
Consider environment variable PIP_TARGET when calculate depndancy delta

cm-iwata:
### The issue

#4775

### The fix

When exec subcommand `install` or `sync`,   if `PIP_TARGET` is set to environment variables,  Refer specified directory for calculate delta, instead default directory

### The checklist

* [x] Associated issue


pipenv/environment.py
```diff
@@ -540,7 +540,7 @@ def get_distributions(self):
         :rtype: iterator
         """
 
-        libdirs = self.base_paths["libdirs"].split(os.pathsep)
+        libdirs = os.environ.get('PIP_TARGET', self.base_paths["libdirs"].split(os.pathsep))
```
frostming, line 543: Should be return as a list

cm-iwata: @frostming 
I fixed wrong type

matteius (approved):

oz123 (approved):

Issue #4537, opened 2020-11-18
pipenv always create new virtualenv if PIPENV_SITE_PACKAGES=1 is set


bachdx96:
### Issue description

When the environment variable `PIPENV_SITE_PACKAGES` is set, pipenv always recreate new virtualenv on each invoke of pipenv even if it is not a `pipenv install` command.

### Expected result

- `pipenv run` or `pipenv shell` should use existing virtualenv instead of creating a new one.
- `PIPENV_SITE_PACKAGES` should only affect `pipenv install` command.

### Actual result

```
root@9e08ac3143d2:/tmp# pipenv && pipenv && pipenv
Warning: the environment variable LANG is not set!
We recommend setting this in ~/.profile (or equivalent) for proper expected behavior.
Creating a virtualenv for this project...
Pipfile: /tmp/Pipfile
Using /usr/bin/python3.8 (3.8.5) to create virtualenv...
Making site-packages available...
⠼ Creating virtual environment...created virtual environment CPython3.8.5.final.0-64 in 254ms
  creator CPython3Posix(dest=/root/.local/share/virtualenvs/tmp-XVr6zr33, clear=False, global=True)
  seeder FromAppData(download=False, pip=bundle, setuptools=bundle, wheel=bundle, via=copy, app_data_dir=/root/.local/share/virtualenv)
    added seed packages: pip==20.2.4, setuptools==50.3.2, wheel==0.35.1
  activators BashActivator,CShellActivator,FishActivator,PowerShellActivator,PythonActivator,XonshActivator

✔ Successfully created virtual environment!
Virtualenv location: /root/.local/share/virtualenvs/tmp-XVr6zr33
Creating a Pipfile for this project...
Virtualenv already exists!
Removing existing virtualenv...
Warning: the environment variable LANG is not set!
We recommend setting this in ~/.profile (or equivalent) for proper expected behavior.
Creating a virtualenv for this project...
Pipfile: /tmp/Pipfile
Using /usr/bin/python3.8 (3.8.5) to create virtualenv...
Making site-packages available...
⠼ Creating virtual environment...created virtual environment CPython3.8.5.final.0-64 in 235ms
  creator CPython3Posix(dest=/root/.local/share/virtualenvs/tmp-XVr6zr33, clear=False, global=True)
  seeder FromAppData(download=False, pip=bundle, setuptools=bundle, wheel=bundle, via=copy, app_data_dir=/root/.local/share/virtualenv)
[...]

PR #4780, opened 2021-08-30
Bug solved: creating virtual environment when not present

vibhutimishra:

### The issue

Yes it is associated with and issue #4537 

### The fix

Initially when virtual environment is not present it creates a virtual environment and if it is present then it removes the existing virtual environment and creates a new one if PIPENV_SITE_PACKAGES=1 is set.

I have removed this issue and added a feature which uses the existing virtual environment when present and does not remove the the existing virtual environment when PIPENV_SITE_PACKAGES=1 is set.

### The checklist

* [x]  Associated issue

### DIFF
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -516,7 +516,7 @@ def ensure_project(
     # Automatically use an activated virtualenv.
     # checks if virtual Enviroment exists if it exists then it makes the system variable true
     if project.virtualenv_exists:
-       system = False
+       system = True
     if not project.pipfile_exists and deploy:
         raise exceptions.PipfileNotFound
     # Skip virtualenv creation when --system was used.
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -434,6 +434,7 @@ def ensure_virtualenv(project, three=None, python=None, site_packages=None, pypi
         sys.exit(1)
 
     if not project.virtualenv_exists:
+        print("hello1")
         try:
             # Ensure environment variables are set properly.
             ensure_environment()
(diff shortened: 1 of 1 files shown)


pipenv/core.py
```diff
@@ -472,22 +472,12 @@ def abort():
         # about, so confirm first.
         if "VIRTUAL_ENV" in os.environ:
             if not (
-                project.s.PIPENV_YES or click.confirm("Remove existing virtualenv?", default=True)
+                project.s.PIPENV_YES or click.confirm("Using existing virtualenv?", default=True)
```
melvyn-oviavo, line 475: (1) Using -> Use

pipenv/core.py
```diff
@@ -507,12 +497,15 @@ def ensure_project(
 
     # Clear the caches, if appropriate.
     if clear:
+        print("here")
```
melvyn-oviavo, line 500: (2) Debugging left over?

pipenv/core.py
```diff
@@ -507,12 +497,15 @@ def ensure_project(
...
         print("clearing")
         sys.exit(1)
 
     # Automatically use an activated virtualenv.
-    if project.s.PIPENV_USE_SYSTEM:
-        system = True
+    # checks if virtual Enviroment exists if it exists then it makes the system variable true
+    if project.virtualenv_exists:
```
melvyn-oviavo, line 506: (2) This should probably be an or now:
`if project.s.PIPENV_USE_SYSTEM or project.virtualenv_exists:`

melvyn-oviavo (changes requested):

oz123:
Obsoleted by #4936

Issue #3332, opened 2018-11-30
Environment variables PIPENV_SKIP_LOCK does not work.


mimizhang:
I don't know if it is bug or my misunderstanding of document.

### Issue description

Environment variables PIPENV_SKIP_LOCK does not work.

### Expected result

As the document described, if PIPENV_SKIP_LOCK is setted  to true, it will skip locking mechanisms.

### Actual result

when my OS Environment variables PIPENV_SKIP_LOCK is  true , pipenv still excute locking.

### Steps to replicate
```shell
$ echo $PIPENV_SKIP_LOCK
true
$ pipenv uninstall autopep8
Un-installing autopep8…
Uninstalling autopep8-1.4.3:
  Successfully uninstalled autopep8-1.4.3

Removing autopep8 from Pipfile…
Locking [dev-packages] dependencies…
Locking [packages] dependencies…
Updated Pipfile.lock (1402f9)!
```

frostming:
What version of pipenv are you running?

It works well on `2018.11.26`

mimizhang:
> What version of pipenv are you running?
> 
> It works well on `2018.11.26`
it is on 2018.10.13. So Which version starts to support this feature?

mimizhang:
@frostming  Thanks for your help!

BeyondEvil:
This (still?) doesn't work with the latest version of pipenv. @frostming 

```
$ echo $PIPENV_SKIP_LOCK
true

$ pipenv install pytest
Installing pytest…
Adding pytest to Pipfile's [packages]…
✔ Installation Succeeded 
Pipfile.lock not found, creating…
Locking [dev-packages] dependencies…
Locking [packages] dependencies…
✔ Success! 
Updated Pipfile.lock (1c4d3d)!
Installing dependencies from Pipfile.lock (1c4d3d)…
  🐍   ▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉ 12/12 — 00:00:02
To activate this project's virtualenv, run pipenv shell.
Alternatively, run a command inside the virtualenv with pipenv run.

$ pipenv --version
pipenv, version 2018.11.26

$ python3 --version
Python 3.7.3

$ uname -a
Darwin Kernel Version 18.5.0: Mon Mar 11 20:40:32 PDT 2019; root:xnu-4903.251.3~3/RELEASE_X86_64 x86_64

$ sw_vers
[...]

Issue #2200, opened 2018-05-16
Provide persistent skip-lock option


hmaarrfk:
I'm always frustrated when `pipenv` wants to lock things. This happens when I install and uninstall things. The problem is that it takes FOREVER to lock big scientific packages (that probably include sample data).

I would like to be able to specify completely skipping the locking process within the Pipfile itself.

I've tried to specify `--skip-lock` in many commands that support it, but some commands simply don't support it. Additionally, if I forget to specify `--skip-lock` then I have to force quit the program and hope that it maintains a usable state.

uranusjr:
If this is done it should be a user preference (like the one discussed in #2197), not a project one. Maybe we can implement a `PIPENV_SKIP_LOCK` or `PIPENV_NO_LOCK` environ. With that said, I’m not sure why you’d even bother using Pipenv if you don’t lock at all—Pipfile essentially become requirements.txt in this situation.

hmaarrfk:
> Pipfile essentially become requirements.txt in this situation.
Pipenv gives you the virtual environment management. Quite nice IMO.

hmaarrfk:
Would it be appropriate to place the option within the lockfile itself?
That way it wouldn't be in the Pipfile, but we could still have it on a per-project basis?

techalchemy:
I think if anything we would want an environment variable for this.  We've said no in the past, but I think we've heard it raised enough that we could revisit.  I wonder about the security concerns, any thoughts there @ncoghlan? If not I'll run it by Kenneth and we can make a judgment call

hmaarrfk:
If it makes any difference, I would vote against an environment variable. Having new users to linux/windows set environment variables is rather difficult.

I really think the choice should be on a per project basis for the two reasons:
1. Advanced users: work on one project that requires locked dependencies and an other that is in heavy development where the exact version doesn't matter so much until later in the development cycle.
2. New users: Already have a hard time understanding what python environments are. Since the locking mechanism takes so long on large packages (this doesn't seem like it will change even with caching), it can take 10s of minutes to install something new you find on github. I would prefer asking them to run a command such as: `pipenv --three --unlock install` which would create a lock file that simply has a "unlocked" flag.

techalchemy:
Which packages are taking 10s of minutes? If you look at npm, composer, etc, this type of feature is never specified at the project level and there are a lot of good reasons for this — locking is an important element of the behavior we are providing, so disabling it _should_ be an advanced feature, if we choose to offer that at all. 

[...]

PR #4797, opened 2021-09-26
Use PIPENV_SKIP_LOCK environment variable

ELHoussineT:
### The issue

The value of `PIPENV_SKIP_LOCK` is always False. This has been discussed in previous issues (#2200 and #3332) with no resolution. 

### The fix

`PIPENV_SKIP_LOCK` to consider the value of the environment variable `PIPENV_SKIP_LOCK`. 

This is useful to large project as it enables the environment variable `PIPENV_SKIP_LOCK` to control the locking behavior. 

Default behavior is not changed, still False.

### The checklist

* [x] Associated issue

### DIFF
--- a/news/4797.feature.rst
+++ b/news/4797.feature.rst
@@ -1,0 +1,1 @@
+Use environment variable `PIPENV_SKIP_LOCK` to control the behaviour of lock skipping.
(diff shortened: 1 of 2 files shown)


oz123 (changes requested):
@ELHoussineT Can you please add a NEWS fragment?

ELHoussineT:
> EWS fragment?

@oz123  News fragment added.

oz123 (approved):

PR #4798, opened 2021-09-26
[Installer/Shells] Fixes for Windows/pyenv-win

awkimball:

### The issue

- `subprocess.run()` needs the argument `shell=True` on Windows when spawning new shell
- call to `choose_shell()` was missing non-options argument, `project`
- `detect_info(project)` no longer member of project

### The fix

- Added missing args to `choose_shell(project)` 
- changed `out()` to `stdout()` in `_iter_installable_versions()`
- change `project.detect_info()` to `detect_info(project)`

### DIFF
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -2305,7 +2305,7 @@ def do_shell(project, three=None, python=False, fancy=False, shell_args=None, py
 
     from .shells import choose_shell
 
-    shell = choose_shell()
+    shell = choose_shell(project)
     click.echo(fix_utf8("Launching subshell in virtual environment..."), err=True)
 
     fork_args = (
--- a/pipenv/installers.py
+++ b/pipenv/installers.py
@@ -169,7 +169,7 @@ class Pyenv(Installer):
     def iter_installable_versions(self):
         """Iterate through CPython versions available for Pipenv to install.
         """
-        for name in self._run('install', '--list').out.splitlines():
+        for name in self._run('install', '--list').stdout.splitlines():
             try:
                 version = Version.parse(name.strip())
             except ValueError:
--- a/pipenv/shells.py
+++ b/pipenv/shells.py
@@ -228,12 +228,19 @@ def _detect_emulator():
         keys.append("cmder")
     if os.environ.get("MSYSTEM"):
         keys.append("msys")
+    # VS Code internal terminal support
+    if os.environ.get("TERM_PROGRAM") == "vscode":
+        keys.append("vscode") 
     return ",".join(keys)
 
 
 def choose_shell(project):
-    emulator = project.PIPENV_EMULATOR.lower() or _detect_emulator()
-    type_, command = project.detect_info()
+    emulator = "none"
+    try:
+        emulator = _detect_emulator() or project.PIPENV_EMULATOR.lower()
+    except AttributeError as error:
+        print("Uh oh! Error: ", error)
+    type_, command = detect_info(project)
     shell_types = SHELL_LOOKUP[type_]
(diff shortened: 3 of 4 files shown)


awkimball:
Unsure why the Windows 3.7 & 3.8 builds hang but I manually verified that all tests pass on both 3.9.6, 3.8.10, and 3.7.9

pipenv/shells.py
```diff
@@ -228,12 +228,19 @@ def _detect_emulator():
...
+        keys.append("vscode") 
     return ",".join(keys)
 
 
 def choose_shell(project):
-    emulator = project.PIPENV_EMULATOR.lower() or _detect_emulator()
-    type_, command = project.detect_info()
+    emulator = "none"
+    try:
+        emulator = _detect_emulator() or project.PIPENV_EMULATOR.lower()
```
frostming, line 240: Why change the order of the conditions

matteius: Not sure since `PIPENV_EMULATOR` appears to be the override for auto-detection, but this also has conflict with base now since master is `project.s.PIPENV_EMULATOR.lower()`

frostming: `PIPENV_EMULATOR` is for override of course

mungojam:
can this one be closed now that other pyenv fix PRs have been merged in?

matteius:
> can this one be closed now that other pyenv fix PRs have been merged in?

I think so, I was reviewing to see if there are any changes we should take from this, and my thought was maybe a couple things, but most of it seems to raise more questions, plus with the current merge conflicts.  I'd prefer we shut this one down and address any additional windows bugs with new issue reports/new PRs.

pipenv/core.py
```diff
@@ -2305,7 +2305,7 @@ def do_shell(project, three=None, python=False, fancy=False, shell_args=None, py
 
     from .shells import choose_shell
 
-    shell = choose_shell()
+    shell = choose_shell(project)
```
matteius, line 2308: This is already fixed in master.

pipenv/installers.py
```diff
@@ -117,7 +117,10 @@ def _run(self, *args, **kwargs):
             k = list(kwargs.keys())[0]
             raise TypeError(f'unexpected keyword argument {k!r}')
         args = (self.cmd,) + tuple(args)
-        c = subprocess_run(args, timeout=timeout)
+        if os.name == 'nt':
```
matteius, line 120: I don't think this is necessary.   I've tested the current behavior of `c = subprocess_run(args, timeout=timeout)` on classic windows cmd, power shell and git shell and it does not seem to error out.

pipenv/shells.py
```diff
@@ -228,12 +228,19 @@ def _detect_emulator():
         keys.append("cmder")
     if os.environ.get("MSYSTEM"):
         keys.append("msys")
+    # VS Code internal terminal support
```
matteius, line 231: I don't have a way to test this.   If its still required lets do it in a separate PR for VS code terminal support.

tests/integration/test_install_basic.py
```diff
@@ -289,7 +289,8 @@ def test_requirements_to_pipfile(PipenvInstance, pypi):
         assert c.returncode == 0
         print(c.stdout)
         print(c.stderr)
-        print(subprocess_run(["ls", "-l"]).stdout)
+        if os.name != 'nt':
```
matteius, line 292: the tests all pass in windows these days, so probably not this.

matteius:
Closing down, please see discussion points; the changes were reviewed for their merits and some are already in master.

PR #4800, opened 2021-10-03
Fixes #4790: behaving as no .venv if .venv is in project root and empty

### LINKED ISSUE #4790: Empty .venv file causes entire virtualenvs directory to be deleted
### Summary

if `.venv` is an empty file and I run `pipenv --rm`, the entire `~/.local/share/virtualenvs` directory is deleted. This happens even when there's no `Pipfile` present.

This is on macOS, pipenv version 2021.5.29.

### Issue description

I was messing around with my project environment, and tried to tell my editor where to find the environment by putting it in `.venv`. Turns out that running `pipenv --venv > .venv` fails because it first touches `.venv`, and then the actual command looks at the empty `.venv` and doesn't find an environment. This causes the file to remain empty. (The command did print an error message in STDERR, but somehow I didn't notice it immediately.)

So now with an empty `.venv` in my project folder, I ran `pipenv install <some package>` and wondered why it said it had to create a new virtualenv. I looked at the output more closely and realized that the new env was created in `/Users/myname/.local/share/virtualenvs` (not in a subdirectory).

I got curious about what would happen if I now tried to scrap the thing, and as I feared, running `pipenv --rm` deleted that entire directory, which contained all the other virtualenvs that I had made with Pipenv for other projects earlier. It wasn't a big loss or I wouldn't have tried, but still not something I like to see happening.

### Steps to replicate

In an empty directory:

```
touch .venv
pipenv --rm
```

Now the entire `~/.local/share/virtualenvs` directory is deleted.

The way I ended up in this situation was actually essentially the following:

```
pipenv --python 3.9
pipenv --venv > .venv
pipenv install requests
pipenv --rm
```

kobayashi:
### The issue

fix issue #4790 

### The fix

This is to prevent an user from choosing own workon home as the virtual environment or deleting all virtual environments accidentally under the workon dir. pipenv ignores the content with this change, so that it uses project name base virtual environment If an empty .venv is set in root.

### The checklist

* [ ] Associated issue

### DIFF
--- a/news/4790.bugfix.rst
+++ b/news/4790.bugfix.rst
@@ -1,0 +1,1 @@
+Ignore empty .venv in rood dir and create project name base virtual environment
--- a/pipenv/project.py
+++ b/pipenv/project.py
@@ -270,6 +270,10 @@ class Project:
         with io.open(dot_venv) as f:
             name = f.read().strip()
 
+        # If .venv file is empty, set location based on config.
+        if not name:
+            return str(get_workon_home().joinpath(self.virtualenv_name))
+
         # If content looks like a path, use it as a relative path.
         # Otherwise use directory named after content in WORKON_HOME.
         if looks_like_dir(name):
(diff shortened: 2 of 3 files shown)


frostming:
Thanks, can you add a news fragment?

kobayashi:
Sure thing! Added a file for this fix.

frostming:
Thank you

Issue #4893, opened 2021-12-10
Feature: Command to verify hash in Pipfile.lock against packages in Pipfile (out-of-date check)


den4uk:
When running `pipenv install --deploy`, it checks the hash in the **Pipfile.lock** file matches with the packages in the **Pipfile**. Upon a mismatch, an exception is raised, else it proceeds to install.

**NEW FEATURE**: It would be very used to have the hash verification step to be a standalone command, which could return an exit code if the hash is verified or not.

**User case.** 
This can come in useful in the CI/CD pipelines, where locking could be performed and committed when the hash is not verified. This could save a significant amount of time to avoid re-locking dependencies, especially when there are many of them.

matteius:
I agree that this would be a useful feature.   Perhaps open a PR and we can push for it?

den4uk:
Thanks for a suggestion, I opened a PR for an initial review. https://github.com/pypa/pipenv/pull/4896

PR #4896, opened 2021-12-18
New command: verify Pipfile.lock hash is up to date

den4uk:
This a proposal to have a new command added: `pipenv verify`

The aim is to be able to verify the hash in the Pipfile is up to date, or whether it needs updating.
The feature can be very useful within the CI/CD pipelines where the Pipfile.lock is a part of the repository and it needs to be up to date. The command can make it quick to run a check this is the case.

The PR is related to an open issue https://github.com/pypa/pipenv/issues/4893

### DIFF
--- a/pipenv/cli/command.py
+++ b/pipenv/cli/command.py
@@ -718,6 +718,7 @@ def verify(state):
             err=True
         )
         sys.exit(1)
+    echo(crayons.green('Pipfile.lock is up-to-date.'))
     sys.exit(0)
 
 
--- a/pipenv/cli/command.py
+++ b/pipenv/cli/command.py
@@ -700,5 +700,26 @@ def scripts(state):
     echo("\n".join(fix_utf8(line) for line in lines))
 
 
+@cli.command(
+    short_help="Verify the hash in Pipfile.lock is up-to-date.",
+    context_settings=CONTEXT_SETTINGS,
+)
+@pass_state
+def verify(state):
+    """Verify the hash in Pipfile.lock is up-to-date."""
+    if not state.project.pipfile_exists:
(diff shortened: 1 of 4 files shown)


pipenv/cli/command.py
```diff
@@ -700,5 +700,26 @@ def scripts(state):
...
+        sys.exit(1)
+    if state.project.get_lockfile_hash() != state.project.calculate_pipfile_hash():
+        echo(
+            'Pipfile.lock is out-of-date. Run {} to update.'.format(
+                crayons.yellow("$ pipenv lock", bold=True)
+            ),
+            err=True
+        )
+        sys.exit(1)
+    sys.exit(0)
```
fridex, line 721: Would it be reasonable to report an informative message that the lock file is up to date?

den4uk: I don't have a strong feeling about it, however as pipenv is fairly verbose throughout - it would make sense having a success message.

den4uk: A success message added.

matteius (approved):
Thanks @den4uk -- this change looks correct to me.

matteius:
@den4uk Thanks for working on this, it currently requires more reviewers to chime in, but also I'm just realizing this should require some documentation and a news bit about its addition.

den4uk:
> @den4uk Thanks for working on this, it currently requires more reviewers to chime in, but also I'm just realizing this should require some documentation and a news bit about its addition.

Hi @matteius , I thought this may need some documentation being added 😄 
I will do this in due course, and once pushed you may be able to flag this with more reviewers, perhaps?

PR #4908, opened 2022-01-06
Use a PackageFinder with `ignore_compatibility` when collecting hashes

jfly:
This fixes https://github.com/pypa/pipenv/issues/4885

### The checklist

* [x] Associated issue

### DIFF
--- a/news/4885.bugfix.rst
+++ b/news/4885.bugfix.rst
@@ -1,0 +1,1 @@
+Fix regression where lockfiles would only include the hashes for releases for the platform generating the lockfile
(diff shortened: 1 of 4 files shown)


jfly:
@frostming does this seem like the right direction? If so, I'll gladly polish this up!

frostming:
Setting `ignore_compatibility` is correct but we don't need to modify the `vendors` directory. I think it is good enough to set  the attribute on the created package finder.

Changing `vendors` code needs special care about the patches.

jfly:
@frostming I've updated this PR with a test, and the build still seems to be green! Would love your eyes on this again.

frostming:
LGTM!

jfly:
Thanks for the quick response, @frostming!

PR #4915, opened 2022-01-09
Remove vendored first

oz123:
While first is nice to have, it adds a lot of code in vendor.
This patch achieves the same with less code in vendor (~80 lines less).


frostming:
Thank you.

Issue #3130, opened 2018-10-30
Document that pipenv is not designed to run as privileged user (root, administrator)


tiran:
From https://github.com/pypa/pipenv/issues/3128#issuecomment-434286020

> Pipenv is not supposed to run as root either. If people are doing that, they're using it wrong.

The official documentation of pipenv does not warn that pipenv is not designed to be executed as privileged user. As bare minimum, please advise consumers of pipenv, that pipenv must not be executed as root. Please also inform the user, that most container runtimes (e.g. Docker) execute all commands as root user by default. Users must switch to an unprivileged user with e.g. ``USER`` command https://docs.docker.com/engine/reference/builder/#user.

techalchemy:
> pipenv must not be executed as root

I'm not super knowledgeable about docker deployments and I agree that it is a good idea to warn against it, but I'm not sure that it _must_ not be used as root.  Many people use pip in docker as root, and many people use `pipenv install --system --deploy` to deploy their dependency state to their docker system python installation.

I have no idea why this is a common practice, and it definitely continues to confuse me whenever I try to use docker, but my impression was that since containers themselves are throwaways, the tradeoff is between managing additional tools/tooling for user access management inside the container vs the infrastructure to just scrap the whole thing and start over (which is how I think people are using pipenv atm?).

On the other hand, I know security is your area of expertise, and while the commenter in question isn't a maintainer of pipenv unfortunately, I can see that they are obviously knowledgeable in this space as well. My primary concern is that this would be confusing to many users who already have established workflows around the whole python ecosystem (pipenv included) which relies on root permissions to deploy system level python packages. I think we need to be careful with wording something like this to provide a clear path forward for those users

tiran:
I was puzzled by the fact, too. I was under the impression that pipenv is designed a full replacement for and wrapper around pip + venv. Both tools are designed to work as root. (Although it's discouraged to run pip install as root, but that's for different reasons.)

I mentioned containers, because I assumed that lots of people are using pipenv in containers. It's a common misconception that containers are isolated, secure enclaves. Container are not a security mechanism, **containers don't contain**, e.g. https://www.projectatomic.io/blog/2014/09/yet-another-reason-containers-don-t-contain-kernel-keyrings/ . Non-priviliged container processes in combination with additional safety layers like SELinux or AppArmor are pretty solid. But if you run as root, then all bets are off.

techalchemy:
[...]

Issue #4189, opened 2020-04-12
Provide `--generate-hashes` option to include hashes in requirements.txt output


ncoghlan:
As described in https://github.com/pypa/pipenv/issues/3057#issuecomment-589495763, emitting `requirements.txt` and then running `pip-sync` is much faster than running `pipenv sync` directly.

This speed increase comes from the fact that `pip-sync` is correctly checking for already installed packages, while `pipenv sync` doesn't yet do that.

However, the workaround has a problem, which is that if the `pip-sync` operation *does* need to install anything, then that installation won't check the artifact hashes, as those hashes aren't included in the `requirements.txt` output.

Supporting a `pip-compile` style `--generate-hashes` option to `pipenv lock --requirements` would allow the export to `pip-sync` (and other tools) to keep the expected hash information.

whardier:
The documentation seems to suggest that by default `pipenv lock --requirements` would output hashes.  I would also like to see all the related hashes for all ranged and explicit version specifiers.

oz123:
That's probably easy to resolve:

```

diff --git a/pipenv/core.py b/pipenv/core.py
index 361e21ee..ba265bc8 100644
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -798,8 +798,10 @@ def do_install_dependencies(
             get_source_list(project, pypi_mirror=pypi_mirror)
         )
         index_args = " ".join(index_args).replace(" -", "\n-")
+        include_hashes=True
+        import re
         deps = [
-            req.as_line(sources=False, include_hashes=False) for req in deps_list
+            re.sub(" --hash", " \\\n --hash", req.as_line(sources=False, include_hashes=True)) for req in deps_list
         ]
         click.echo(index_args)
         click.echo("\n".join(sorted(deps)))
```

matteius:
This is resolved when using the fairly new `pipenv requirements` command.

PR #4917, opened 2022-01-10
Document running as root and usage in containers

oz123:
See discussion in #3130.

### DIFF
--- a/docs/basics.rst
+++ b/docs/basics.rst
@@ -411,3 +411,24 @@ production environments for reproducible builds.
     If you'd like a ``requirements.txt`` output of the lockfile, run ``$ pipenv lock -r``.
     This will not include hashes, however. To get a ``requirements.txt``
     you can also use ``$ pipenv run pip freeze``.
+
+
+☤ Pipenv and Docker Conatiners
+------------------------------
+
+In general, you should not need to install pipenv inside docker container to manage
+the dependencies of your software. Instead you can generate a ``requirements.txt`` as described
+above and use it inside your container, to install the dependencies, e.g.:
+
+  $ pipenv lock -r > requirements.txt
+
+The in your ``Dockerfile``::
+
+   FROM docker.io/python3
+
+   ...
+
+   ADD requirements.txt /usr/src/
+   RUN pip3 install -r /usr/src/requirements.txt
+
+   ...
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -190,30 +190,31 @@ dependencies your project needs in case you need to re-install them, such as
 when you share your project with others. You should get output similar to this
 (although the exact paths shown will vary)::
 
-    Creating a Pipfile for this project...
-    Creating a virtualenv for this project...
-    Using base prefix '/usr/local/Cellar/python3/3.6.2/Frameworks/Python.framework/Versions/3.6'
-    New python executable in ~/.local/share/virtualenvs/tmp-agwWamBd/bin/python3.6
-    Also creating executable in ~/.local/share/virtualenvs/tmp-agwWamBd/bin/python
-    Installing setuptools, pip, wheel...done.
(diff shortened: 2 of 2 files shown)


frostming (approved):

docs/basics.rst
```diff
@@ -411,3 +411,24 @@ production environments for reproducible builds.
...
+
+
+☤ Pipenv and Docker Conatiners
+------------------------------
+
+In general, you should not need to install pipenv inside docker container to manage
+the dependencies of your software. Instead you can generate a ``requirements.txt`` as described
+above and use it inside your container, to install the dependencies, e.g.:
+
+  $ pipenv lock -r > requirements.txt
```
jfly, line 423: @oz123 these instructions assume that you've got `pipenv` installed outside of docker? That doesn't feel desirable to me.

oz123: is meant for managing virtual environment*s*. Usually, a container contains just one. That is the reason why you probably don't need it inside the container.
If you really desire you can use multi stage build with one stage building your vent with pipenv and then copying it to the end container. But that's way out of scope of the manual and I assumed that power users of docker aren't the audience of these instructions.
That said, what is your suggestion? Have I overlooked something? Please let me know if I can improve the documentation.

matteius: What about having `pipenv` installed in the docker container and running `pipenv sync`?   Seems about half of the time docker containers are using virtualenvs and half the time they aren't.  At my last job we buil ECR images in docker, and so having to run a command outside of docker in the CI to generate the latest requirements.txt from outside of the container, well we wouldn't have done it that way since the whole thing ran the build in docker.

oz123: I still don't get why one would want to have ~80mb of pipenv inside the container image.
Even if you run your pipeline in container one step can generate requirements.txt and the next one can pickup the file and install the dependencies.
And these steps can be in Jenkins/giltab/GitHub/your-ci or as a I mentioned multi stage docker file. 
As side note: there is some confusion always regarding the terminology.
The docker file is used to build the image .
The container is spawned from it. 
You install pipenv during the image build. And then you have it in all the containers built from it.
Pipenv is a build tool, like npm or make or GCC. IMHO it does not belong in the final container.
Whether you exclude it with my first suggestion or later is eventually the same. I chose to document the most easy one. 
I will add a note about pipenv exclusion from the final container.

matteius: Maybe I am missing something @oz123 but pipenv wheel is < 4mb, so having it in the final container or not seems like a matter of preference.  I was just stating that in my prior case, we were using a local user and virtualenvs within the final container, and while we weren't using `pipenv` at all, if we were, we likely would have converted from `requirements.txt` to `Pipfile` and `Pipfile.lock` and generate the virtualenv using `pipenv sync` to build the dependencies in the lock file during CI build (the Docker files building the image).  There is no need to generate a requirements.txt at all the way I see it, if you are relying on your lock file.  

Also and **most importantly** I just tried the example of `pipenv lock -r > requirements.txt` and it doesn't include any hash values in the requirements.txt so doing it this way basically negates any protection of having the lock file and the benefits of using `pipenv sync` during the build.

oz123: The size I mentioned is wrong (based on the package in gentoo with compiled __pycache__ include). Sorry about that.
As for the hashing, yes, you are right, requirements.txt does not include the hashes.
Thanks for pointing this out. I will update the docs accordingly, and suggest to use multistage build which exclude pipenv from the final image.

jfly: @oz123, I agree that multistage builds feel like The Right Way to do this. Hopefully it's not too much work to write up. Thanks for improving the documentation!

oz123: I would also modify `pipenv lock -r` to add hashes:
see #4189.

PR #4918, opened 2022-01-11
Document  usage of Pipenv in multistage docker build

oz123:
Following the discussion on #4917.
This PR add documentation which demonstrates how to use multistage build, and an explicit not on usage as root.
@jfly,  @matteius, your feed back is appreciated!


jfly (approved):
I love it!

matteius (approved):
Looks good and now leverages the benefit of locked dependency hashes by using `pipenv sync`; thanks for doing this @oz123 !

PR #4921, opened 2022-01-12
Update contributing.rst and run tests

oz123:
Major change to contributing.md and developer docs.
`run-tests.sh` is working again.


run-tests.sh
```diff
@@ -1,53 +1,56 @@
...
+
 # Let's use a temporary cache directory
 export PIPENV_CACHE_DIR=`mktemp -d 2>/dev/null || mktemp -d -t 'pipenv_cache'`
 
-prefix() {
-	sed "s/^/   $1:    /"
-}
+# on some Linux OS python is python3
+PYTHON=${PYTHON:-"python"}
+PIPENV_PYTHON="${PIPENV_PYTHON:-3.7}"
```
matteius, line 16: Not sure I understand the value `PIPENV_PYTHON` here, because it defaults only to `-3.7` but it appears to be used in commands below.  Also, are we only running tests against python 3.7?

run-tests.sh
```diff
@@ -1,53 +1,56 @@
...
 export LANG=C.UTF-8
-export PIP_PROCESS_DEPENDENCY_LINKS="1"
+
 # Let's use a temporary cache directory
 export PIPENV_CACHE_DIR=`mktemp -d 2>/dev/null || mktemp -d -t 'pipenv_cache'`
 
-prefix() {
-	sed "s/^/   $1:    /"
-}
+# on some Linux OS python is python3
```
matteius, line 14: Do we need to account for the fact that its python3 on some platforms?

run-tests.sh
```diff
@@ -1,53 +1,56 @@
...
+# This installs the dependencies for pipenv
+${PIP_CALL} --upgrade pip setuptools wheel virtualenv --upgrade-strategy=eager
+
+VENV_CMD="${PYTHON} -m pipenv --venv"
 RM_CMD="pipenv --rm"
-echo "$ PIPENV_PYTHON=2.7 $VENV_CMD && PIPENV_PYTHON=2.7 $RM_CMD"
-echo "$ PIPENV_PYTHON=3.7 $VENV_CMD && PIPENV_PYTHON=3.7 $RM_CMD"
-{ PIPENV_PYTHON=2.7 $VENV_CMD && PIPENV_PYTHON=2.7 $RM_CMD ; PIPENV_PYTHON=3.7 $VENV_CMD && PIPENV_PYTHON=3.7 $RM_CMD ; }
+
+echo "$ PIPENV_PYTHON=${PIPENV_PYTHON} $VENV_CMD && PIPENV_PYTHON=${PIPENV_PYTHON} $RM_CMD"
```
matteius, line 38: Maybe this could be made a variable so that it isn't repeated between the `echo` and the actual execution of it on line 40 below?

run-tests.sh
```diff
@@ -1,53 +1,56 @@
...
+PIPENV_PYTHON=${PIPENV_PYTHON} $INSTALL_CMD
+
 echo "$ git submodule sync && git submodule update --init --recursive"
 
 git submodule sync && git submodule update --init --recursive
 
-echo "$ pipenv run time pytest"
-PIPENV_PYTHON=2.7 python -m pipenv run time pytest
-PIPENV_PYTHON=3.7 python -m pipenv run time pytest
+echo "$pipenv run pytest -ra -n auto -v --fulltrace tests"
```
matteius, line 54: How did you arrive at these arguments for pytest vs what was there prior?

docs/dev/contributing.rst
```diff
@@ -31,14 +25,11 @@ Be Cordial
 
 Pipenv has one very important rule governing all forms of contribution,
 including reporting bugs or requesting features. This golden rule is
-"`be cordial or be on your way`_".
+"be cordial or be on your way".
 
 **All contributions are welcome**, as long as
 everyone involved is treated with respect.
 
-.. _be cordial or be on your way: https://www.kennethreitz.org/essays/be-cordial-or-be-on-your-way
```
frostming, line 39: Why do we remove the link?
The valid link would be https://kennethreitz.org/essays/2013/01/27/be-cordial-or-be-on-your-way

docs/dev/contributing.rst
```diff
@@ -104,22 +93,14 @@ The following sub-sections go into more detail on some of the points above.
...
 Development Setup
 ~~~~~~~~~~~~~~~~~
 
-To get your development environment setup, run:
-
-.. code-block:: sh
-
-  pip install -e .
-  pipenv install --dev
+See `CONTRIBUTING.md`_ in the code repository.
```
frostming, line 101: It is circular referencing

run-tests.sh
oz123: The tests are running against all Python versions with every commit. This is done in:
https://github.com/pypa/pipenv/blob/638c84628df503e685c3ca80a9e72be95af13048/.github/workflows/ci.yaml#L42

The script is intended to be used by users locally. When you run the script you can override every variable with:
```
PIPENV_PYTHON=3.9 ./run-tests.sh
... results in:
Successfully installed wheel-0.37.1
$ PIPENV_PYTHON=3.9 python -m pipenv --venv && PIPENV_PYTHON=3.9 pipenv --rm
No virtualenv has been created for this project(/home/oznt/Software/pipenv) yet!
Aborted!
Installing dependencies...
$ PIPENV_PYTHON=3.9 python -m pipenv install --deploy --dev
```

run-tests.sh
oz123: It is already accounted. You can set PYTHON:
```
PYTHON=${PYTHON:-"python"}
```
Usage example:
```
PYTHON=python3 PIPENV_PYTHON=3.9 ./run-tests.sh
```

run-tests.sh
oz123: Can you clarify here? I don't understand your comment.

docs/dev/contributing.rst
oz123: Great, thank you for finding the correct link.

docs/dev/contributing.rst
oz123: I thought in the beginning only the `md` version is worth maintaining. Eventually I decided to keep the rst, and only maintain it.

docs/dev/contributing.rst
oz123: I restored the link.

run-tests.sh
oz123: Also this piece of code wasn't used anywhere in the previous version before my change.

run-tests.sh
oz123: I think as long as python3.7 is supported the script should default to it. Also the CI pipeline is testing against this version.

oz123: @matteius I added another commit with explicit explanation of the advanced usage of `run-tests.sh`

frostming (approved):

run-tests.sh
matteius: Just was noticing that echo statement is a repeated construction of the command below it, maybe it could be made a variable?   Not a blocker and I am not a bash scripting expert, so might not be worth it.

matteius (approved):
Looks good!

run-tests.sh
oz123: TBH, These `echo` statements are quite obsolete. If one wants to see what the script does,
there is always `bash -x run-tests.sh`.
I'd leave it like these. In the near future I am going to refactor the makefile, so there is no point
investing too much time here.

matteius:
@oz123 or @frostming I just noticed that on the root of the Issues page it links to the old CONTRIBUTING.md which just references the docs/dev/contributing.rst -- not sure how to, but would be good to update this link to just point right to the correct document:  

image

oz123:
Yes, I agree.

PR #4936, opened 2022-01-29
fixed review from #4780, closes #4537

### LINKED ISSUE #4537: pipenv always create new virtualenv if PIPENV_SITE_PACKAGES=1 is set

willhoh:
### The issue #4537

This pull request should fixes #4537 . 

### The fix

The behavior where `PIPENV_SITE_PACKAGES=1` always deletes and recreates an existing environment when typing `pipenv` is eliminated by this pull request.

### The checklist

* [x] Associated issue

### Comment
Unfortunately I don't know what to insert the news/ folder and how it should look like. Please give me an example or explanation.

### DIFF
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -483,11 +483,11 @@ def ensure_virtualenv(project, three=None, python=None, site_packages=None, pypi
         # about, so confirm first.
         if "VIRTUAL_ENV" in os.environ:
             if not (
-                project.s.PIPENV_YES or click.confirm("Remove existing virtualenv?", default=True)
+                project.s.PIPENV_YES or click.confirm("Use existing virtualenv?", default=True)
             ):
                 abort()
         click.echo(
-            crayons.normal(fix_utf8("Removing existing virtualenv..."), bold=True), err=True
+            crayons.normal(fix_utf8("Using existing virtualenv..."), bold=True), err=True
         )
         # Remove the virtualenv.
         cleanup_virtualenv(project, bare=True)
@@ -517,7 +517,7 @@ def ensure_project(
     """Ensures both Pipfile and virtualenv exist for the project."""
 
     # Automatically use an activated virtualenv.
-    if project.s.PIPENV_USE_SYSTEM:
+    if project.s.PIPENV_USE_SYSTEM or project.virtualenv_exists:
         system = True
     if not project.pipfile_exists and deploy:
         raise exceptions.PipfileNotFound


pipenv/core.py
```diff
@@ -483,11 +483,11 @@ def abort():
         # about, so confirm first.
         if "VIRTUAL_ENV" in os.environ:
             if not (
-                project.s.PIPENV_YES or click.confirm("Remove existing virtualenv?", default=True)
+                project.s.PIPENV_YES or click.confirm("Use existing virtualenv?", default=True)
             ):
                 abort()
         click.echo(
-            crayons.normal(fix_utf8("Removing existing virtualenv..."), bold=True), err=True
+            crayons.normal(fix_utf8("Using existing virtualenv..."), bold=True), err=True
```
oz123, line 490: There is no longer need for `fix_utf-8` since pipenv does not support python2, this can be removed.
Also, `click.echo` does not need crayons.normal. IMHO  both can be removed.

oz123 (changes requested):

pipenv/core.py
```diff
@@ -517,7 +517,7 @@ def ensure_project(
     """Ensures both Pipfile and virtualenv exist for the project."""
 
     # Automatically use an activated virtualenv.
-    if project.s.PIPENV_USE_SYSTEM:
+    if project.s.PIPENV_USE_SYSTEM or project.virtualenv_exists:
         system = True
     if not project.pipfile_exists and deploy:
```
oz123, line 522: To avoid confusion down the road, I suggest renaming variable `system` to `system_or_exists`.

oz123 (changes requested):

pipenv/core.py
willhoh: `system` quite often exists in the code and within many functions. Shall I replace it in everyone with `system_or_exists`?

oz123: No, just this one. Because that is what it determines here.

pipenv/core.py
willhoh: I didn't change this and would rather do it in another commit because in core.py alone the same thing occurs more than 30 times.

pipenv/core.py
```diff
@@ -506,7 +506,7 @@ def ensure_project(
     three=None,
     python=None,
     validate=True,
-    system=False,
+    in_system=False,
     warn=True,
```
oz123, line 510: You should not change the function signature. I only meant the local variable.
Also, I still think the name `system_or_exists` reflects the purpose better.

oz123 (changes requested):
Please change only the local variable name after:
`if project.s.PIPENV_USE_SYSTEM or project.virtualenv_exists:`.

pipenv/core.py
willhoh: VSCode has followed the local variable and changed it only where it is used. I had to change the one in the function signature as well, otherwise I would have a new one that is not used. Also, the `ensure_project` function is executed 2 times with `system` as a parameter in `core.py`, so I had to change it further down in the code as well.

Please forgive me if I don't quite understand what you mean.

I thought in_system fits better to a boolian, but you are right, yours describes better what is queried.

oz123: I believe VSCode is doing a disservice ("Bärendienst") here.
The `in_system` variable is defined in L521, and then referenced within the scope of this function in L525 and then L565.
These are the only places that need to be renamed.

willhoh: Then what is the system variable in the function signature for? If I replace system in L521, L525 and L566, the variable in L509 becomes superfluous and is not used.

oz123: It's used in case the if condition is not met, and it is used by external calling functions. I  would not touch it, because it can create havoc. Only in the scope of the function I would change it to be:
```python
    if project.s.PIPENV_USE_SYSTEM or project.virtualenv_exists:
        system_or_env = True
    else:
        system_or_env = system # defaults to False

    ....
       ensure_pipfile(
        project, validate=validate, skip_requirements=skip_requirements, system=system_or_env
    )
```
```
That keeps the orignal meaning of the variable, but make it clear in the scope of this function and this
fix.

willhoh: `...  
File "/opt/pipenv/pipenv/core.py", line 1861, in do_install
    ensure_project(
  File "/opt/pipenv/pipenv/core.py", line 525, in ensure_project
    if not system_or_exists:
UnboundLocalError: local variable 'system_or_exists' referenced before assignment
`

Unfortunately, that's what I get if I just change what you suggest.

oz123: Well, it was implicit. I am sorry for the confusion. We both mean to create good software. I know a PR review can be frustrating. Sorry it takes time.

oz123 (approved):
Thank you for your contribution and patience!

PR #4955, opened 2022-02-15
Windows: locate pyenv from PATH correctly

matteius:
### The issue

In debugging other Windows related pipenv issues, I had to patch this locally to get windows to find pyenv which was already on the PATH.  This combined with the latest fix:  https://github.com/pypa/pipenv/pull/4929 allowed pipenv to finally use pyenv-win on windows successfully to install other versions of python beyond what I had installed at a system level or manually using pyenv-win.  

Pipenv should be able to use the pyenv already on the path in windows, instead of printing:
```
$ pipenv sync
Warning: Python 3.9 was not found on your system...
Neither 'pyenv' nor 'asdf' could be found to install Python.
You can specify specific versions of Python with:
$ pipenv --python path\to\python
```

    Non-significant change.

### The fix

After chatting with @oz123 he suggested trying `shutil.which` and that resolved the issue.  Apparently `shutil` should be able to do job and is more modern than `distutils` which is scheduled to go away in Python 3.12

### The checklist

* [ ] Associated issue

### DIFF
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -14,7 +14,6 @@ import sys
 import warnings
 
 from contextlib import contextmanager
-from distutils.spawn import find_executable
 from pathlib import Path
 from urllib.parse import urlparse
 
--- a/pipenv/utils.py
+++ b/pipenv/utils.py
@@ -1623,7 +1623,7 @@ def find_windows_executable(bin_path, exe_name):
             if os.path.isfile(path):
                 return path
 
-    return find_executable(exe_name)
+    return shutil.which(exe_name)
 
 
 def path_to_url(path):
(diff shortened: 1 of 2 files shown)


oz123 (changes requested):
@matteius the tests break here:

https://github.com/pypa/pipenv/blob/9e1e7586bd93b3950b2404eb69f39223458d8f94/tests/unit/test_utils_windows_executable.py#L18,

You should change the mock to patch `shuitl.which,` and it's going to pass.

mungojam:
any chance of a new release with this in soon? We're holding off on upgrading from a very old python version until pyenv support on windows is available

matteius:
@mungojam `2022.3.23` has been released!

Issue #4926, opened 2022-01-19
Infinity lock


LEv145:
When I use `pipenv lock` pipenv infinitely generates a lock file
Maybe it is because `discord.py` required `aiohttp>=3.6.0,<3.8.0`, but last version of `aiohttp` is `3.8.1`

<details><summary>$ pipenv --support</summary>

Pipenv version: `'2021.11.23'`

Pipenv location: `'/usr/lib/python3.10/site-packages/pipenv'`

Python location: `'/usr/bin/python'`

Python installations found:

  - `3.10.1`: `/usr/bin/python`
  - `3.10.1`: `/usr/bin/python3`
  - `3.10.1`: `/bin/python`
  - `3.10.1`: `/bin/python3`
  - `3.9.8`: `/home/lev145/.local/share/virtualenvs/Flirty_Kitty-6UYt8Wcu/bin/python3.9`
  - `3.9.8`: `/home/lev145/.local/share/virtualenvs/Flirty_Kitty-6UYt8Wcu/bin/python`
  - `3.9.8`: `/home/lev145/.local/share/virtualenvs/Flirty_Kitty-6UYt8Wcu/bin/python3`
  - `3.9.8`: `/home/lev145/.local/share/virtualenvs/Flirty_Kitty-6UYt8Wcu/bin/python3.9`
  - `3.9.8`: `/home/lev145/.local/share/virtualenvs/Flirty_Kitty-6UYt8Wcu/bin/python`
  - `2.7.18`: `/usr/bin/python2`
  - `2.7.18`: `/usr/bin/python2.7`
  - `2.7.18`: `/bin/python2`
  - `2.7.18`: `/bin/python2.7`

PEP 508 Information:

```
{'implementation_name': 'cpython',
 'implementation_version': '3.10.1',
 'os_name': 'posix',
 'platform_machine': 'x86_64',
 'platform_python_implementation': 'CPython',
 'platform_release': '5.15.12',
 'platform_system': 'Linux',
 'platform_version': '#1 SMP PREEMPT Wed Dec 29 18:08:07 UTC 2021',
 'python_full_version': '3.10.1',
 'python_version': '3.10',
 'sys_platform': 'linux'}
```

System environment variables:

  - `SHELL`
  - `SESSION_MANAGER`
  - `PYENV_SHELL`
  - `XDG_SESSION_PATH`
  - `HISTCONTROL`
  - `GNOME_DESKTOP_SESSION_ID`
  - `PYCHARM_JDK`
[...]

PR #4964, opened 2022-02-19
fixes issue with infinite iteration, infinite locking. Makes lock deterministic, again

matteius:
### The issue
In debugging the infinity lock issue reports, this should solve for:
* https://github.com/pypa/pipenv/issues/4651
* https://github.com/pypa/pipenv/issues/4926

### The fix

This fixes the problem of of a series of methods that return a `Iterator[IndexCandidateInfo]` but this inner most method would sometimes have a empty list of `icans` and the function returned None, which lead to an infinite iteration when trying to evaluate `Criterion.candidates` for certain dependencies being resolved.    Before this change, the locking would eventually fail once it reached max iterations here: https://github.com/pypa/pipenv/blob/main/pipenv/patched/notpip/_internal/resolution/resolvelib/resolver.py#L93

### The checklist

* [X] Associated issue

### If this is a patch to the `vendor` directory...

> Please try to refrain from submitting patches directly to `vendor` or `patched`, but raise your issue to the upstream project instead, and inform Pipenv to upgrade when the upstream project accepts the fix.

This is the `patched/notpip/_internal` which seems to only be implementation specific to `pipenv`.  I doubt there is somewhere upstream to fix this.

### DIFF
--- a/pipenv/patched/notpip/_internal/resolution/resolvelib/factory.py
+++ b/pipenv/patched/notpip/_internal/resolution/resolvelib/factory.py
@@ -325,6 +325,7 @@ class Factory:
             base_cand = as_base_candidate(lookup_cand)
             assert base_cand is not None, "no extras here"
             yield self._make_extras_candidate(base_cand, extras)
+        raise StopIteration
 
     def _iter_candidates_from_constraints(
         self,
@@ -348,6 +349,7 @@ class Factory:
             )
             if candidate:
                 yield candidate
+        raise StopIteration
 
     def find_candidates(
         self,
--- a/pipenv/patched/notpip/_internal/resolution/resolvelib/factory.py
+++ b/pipenv/patched/notpip/_internal/resolution/resolvelib/factory.py
@@ -325,7 +325,6 @@ class Factory:
             base_cand = as_base_candidate(lookup_cand)
             assert base_cand is not None, "no extras here"
             yield self._make_extras_candidate(base_cand, extras)
-        raise StopIteration
 
     def _iter_candidates_from_constraints(
         self,
@@ -349,7 +348,6 @@ class Factory:
             )
             if candidate:
                 yield candidate
-        raise StopIteration
 
     def find_candidates(
         self,
--- a/pipenv/patched/notpip/_internal/resolution/resolvelib/factory.py
+++ b/pipenv/patched/notpip/_internal/resolution/resolvelib/factory.py
@@ -295,6 +295,8 @@ class Factory:
                 )
                 yield ican.version, func
 
+            raise StopIteration
+
         return FoundCandidates(
             iter_index_candidate_infos,
             _get_installed_candidate(),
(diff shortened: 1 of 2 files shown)


matteius:
Hmm, I am reading through https://www.python.org/dev/peps/pep-0479/ and it seems that I may be slightly mistaken by raising `StopIteration` and that a `return` may be sufficient and preferred way of doing it.  I am testing locally if that has the same effect as `raise StopIteration` did in fixing the bug.

frostming:
> Hmm, I am reading through https://www.python.org/dev/peps/pep-0479/ and it seems that I may be slightly mistaken by raising `StopIteration` and that a `return` may be sufficient and preferred way of doing it. I am testing locally if that has the same effect as `raise StopIteration` did in fixing the bug.

`return` is also redundant if it is the last statement.

pipenv/patched/notpip/_internal/resolution/resolvelib/factory.py
```diff
@@ -295,6 +295,8 @@ def iter_index_candidate_infos() -> Iterator[IndexCandidateInfo]:
                 )
                 yield ican.version, func
 
+            raise StopIteration
```
matteius, line 298: @frostming This one seems to need to `raise StopIteration` to prevent a bug where the `Criterion.candidates` is infinite.  What are your thoughts here?

frostming: Oh, you are fixing the code under the vendor directory of a vendored library(`pip`). It needs to be fixed upstream.

matteius: Actually this might not be the right approach, it bubbles as up as `RunTimError`.  Just happens to work to get out of the infinite loop, but probably isn't the right approach.

matteius: Maybe the issue is more inside of:  https://github.com/pypa/pipenv/blob/main/pipenv/patched/notpip/_vendor/resolvelib/resolvers.py#L242-L329
We never get to the final False case for some of these examples.  It just cycles through the same back-tracking results.  It is not clear to me how to fix.

matteius: If it is in `pip` why is the directory called `notpip`?  It seems this code is our own internal impementation of the `resolvelib` interfaces, but maybe I am misunderstanding.  That is why I asked in another thread what is `notpip`?

matteius:
Ok it seems that I've gained some insight from this excercise:  pip code uses the same resolver, it actually provides the code.  It doesn't fail when it gets the same requirements asked to install.  I think the issue is we instantiate our resolver differently than pip does.  
comparing: https://github.com/pypa/pipenv/blob/main/pipenv/utils.py#L662-L690
to: https://github.com/pypa/pip/blob/main/src/pip/_internal/cli/req_command.py#L329
Otherwise I am not sure how two code paths that are identical could be getting different results, except for the options are different.

matteius:
@frostming Can you advise on how to bring the latest pip into pipenv?  It appears the code is out of date and I am manually diffing in changes to check if it is fixed, because I suspect it is either fixed in latest pip or we are passing the wrong options.

matteius:
Closing in favor of trying to get someone to either educate me, or do it themselves, to upgrade the vendor'd pip that `pipenv` depends on.

PR #4969, opened 2022-02-23
Vendor pip 22.0.4 follow-up changes [Diff of actual changes to code]

matteius:
This is the result of many hours, debugging pip `22.0.3` working out the kinks to:  https://github.com/pypa/pipenv/pull/4966
However two days ago  `pip==22.0.4` was released and so I updated the parent branch vendoring to vendor to that, which resolved virtually all of the issues I had with `22.0.3` that lead to code changes.  I've since pulled out the code changes to the minimalist most changed which are:

**ALL TESTS NOW PASSING!!!**

**TODO**
- requirementslib doesn't yet support pip 22.x because there is a new require verbosity argument.  This requires a requirementslib update and some additional vendoring once that gets released.
  - Proposed:  https://github.com/sarugaku/requirementslib/pull/311
- Pipshims has to change as well to match the verbosity here is already logic in that shim to support prior versions of pip. 
  - Proposed: https://github.com/sarugaku/pip-shims/pull/79

In both cases I have a workaround to the TODOs in this PR until those packages can be releases and then I can update the vendoring to include them and remove my manual changes there.

This requires dropping support for python 3.6 because pip 22.x dropped support, which should be fine since its EOL.  However I recommend we do one more release of the other fixes in master before eventually doing a release that includes the pip vendoring updates.  

Once the `requirementslib` and `pipshims` have new releases I can vendor the parent PR and close down this one as it won't have any extra changes required, I think.


matteius:
The last two test failures are really perplexing.  I've deep dived the one for `test_lock_nested_direct_url`

```
Failure point running this command which runs fine from the CLI if you just use pip:
['/tmp/pipenv-wt8df5q7-project/.venv/bin/python', '/tmp/pip-standalone-pip-5jvzs58p/__env_pip__.zip/pip', 'install', '--ignore-installed', '--no-user', '--prefix', '/tmp/pip-build-env-o4i5kuwu/overlay', '--no-warn-script-location', '--no-binary', ':none:', '--only-binary', ':none:', '-i', 'http://127.0.0.1:38733/simple', '--pre', '--', 'setuptools>=40.8.0', 'wheel>=0.33.0']

["/tmp/pipenv-wt8df5q7-project/.venv/bin/python: can't find '__main__' module in '/tmp/pip-standalone-pip-5jvzs58p/__env_pip__.zip/pip'\n"]

ERROR:
pip subprocess to install build dependencies exited with 1
```

matteius:
This was not needed once the vendoring of new pip-shims and requirementslib were completed.

Issue #4975, opened 2022-03-07
Cannot use environment variables defined in .env in `pipenv run` and scripts in Pipfile


hmasdev:
### Issue description

I want to use environment variables defined in `.env` in `pipenv run` and scripts in `Pipfile`.
For example, when `FOO=SPAM` is in `.env` and `show-foo = "echo $FOO"` is in Pipfile as a script, I want to get `SPAM` by running `pipenv run echo '$FOO'` or `pipenv run show-foo`.

### Expected result

Running `pipenv run echo '$FOO'` or `pipenv run show-foo` returns `SPAM` with the following Pipfile and `.env`:

```ini:Pipfile
[[source]]
url = "https://pypi.org/simple"
verify_ssl = true
name = "pypi"

[packages]

[dev-packages]

[requires]
python_version = "3.10"

[scripts]
show-foo = "echo $FOO"
```

```bash:.env
FOO=SPAM
```

### Actual result

Running `pipenv run echo '$FOO'` or `pipenv run show-foo` behave as follows:

```bash
$ pipenv run echo '$FOO'
Loading .env environment variables...
$FOO
$ pipenv run show-foo
Loading .env environment variables...
$FOO
```

`$FOO` is not evaluated even if `FOO=SPAM` in `.env`.

### Steps to replicate

1. create the following Pipfile

   ```ini:Pipfile
   [[source]]
   url = "https://pypi.org/simple"
   verify_ssl = true
   name = "pypi"

   [packages]

   [dev-packages]

   [requires]
   python_version = "3.10"

   [scripts]
   show-foo = "echo $FOO"
   ```

2. create `.env`

   ```bash:.env
   FOO=SPAM
   ```

3. run `pipenv run echo '$FOO'` or `pipenv run show-foo`

   ```bash
   $ pipenv run echo '$FOO'
   Loading .env environment variables...
   $FOO
   $ pipenv run show-foo
   Loading .env environment variables...
   $FOO
   ```

---

<details><summary>$ pipenv --support</summary>

Pipenv version: `'2022.1.8'`

Pipenv location: `'/home/hmasu/.pyenv/versions/3.10.2/lib/python3.10/site-packages/pipenv'`

Python location: `'/home/hmasu/.pyenv/versions/3.10.2/bin/python3.10'`

Python installations found:

- `3.10.2`: `/home/hmasu/.pyenv/versions/3.10.2/bin/python3`
[...]

PR #4979, opened 2022-03-11
Fix for expanding dot env variables in scripts

matteius:
Ensure that dot env variables are actually in the real environment so they can be exapnded by os.path.expandvars

### The issue

https://github.com/pypa/pipenv/issues/4975
https://github.com/pypa/pipenv/issues/3901

### The fix

The problem is that in the current code we are loading the .env variables into a copy of the environment which is not available to `os.path.expandvars` so they are always treated as a string.  This fixes that by applying it to the environment before copying it.  Also refactor away some of the complexity of `load_dot_env`.

### The checklist

* [X] Associated issue

### DIFF
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -103,25 +103,23 @@ def load_dot_env(project, as_dict=False):
             [project_directory, ".env"]
         )
 
-        if os.path.isfile(dotenv_file):
+        if not os.path.isfile(dotenv_file) and project.s.PIPENV_DOTENV_LOCATION:
+            click.echo(
+                "{}: file {}={} does not exist!!\n{}".format(
+                    crayons.red("Warning", bold=True),
+                    crayons.normal("PIPENV_DOTENV_LOCATION", bold=True),
+                    crayons.normal(project.s.PIPENV_DOTENV_LOCATION, bold=True),
+                    crayons.red("Not loading environment variables.", bold=True),
+                ),
+                err=True,
+            )
+        if as_dict:
+            return dotenv.dotenv_values(dotenv_file)
+        else:
             click.echo(
                 crayons.normal(fix_utf8("Loading .env environment variables..."), bold=True),
                 err=True,
             )
-        else:
-            if project.s.PIPENV_DOTENV_LOCATION:
-                click.echo(
-                    "{}: file {}={} does not exist!!\n{}".format(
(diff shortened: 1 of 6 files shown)


pipenv/core.py
```diff
@@ -103,25 +103,23 @@ def load_dot_env(project, as_dict=False):
...
-        if os.path.isfile(dotenv_file):
+        if not os.path.isfile(dotenv_file) and project.s.PIPENV_DOTENV_LOCATION:
             click.echo(
-                crayons.normal(fix_utf8("Loading .env environment variables..."), bold=True),
+                "{}: file {}={} does not exist!!\n{}".format(
+                    crayons.red("Warning", bold=True),
+                    crayons.normal("PIPENV_DOTENV_LOCATION", bold=True),
+                    crayons.normal(project.s.PIPENV_DOTENV_LOCATION, bold=True),
+                    crayons.red("Not loading environment variables.", bold=True),
+                ),
```
oz123, line 113: Can `click.secho` be used here instead of crayons?

Crayons is kind of dead.

matteius: Sounds like a good idea, but also I'd prefer we do that in a follow-up PR that addresses the whole project space.

pipenv/core.py
```diff
@@ -2473,8 +2471,8 @@ def do_run(project, command, args, three=None, python=False, pypi_mirror=None):
         project, three=three, python=python, validate=False, pypi_mirror=pypi_mirror,
     )
 
+    load_dot_env(project)
```
frostming, line 2474: I am afraid this will break the isolation in test cases. dotenv files in one test case will contaminate other cases env.

matteius: @frostming Could you elaborate on this?  Right now the test suite is passing with this change so I am not sure I understand, but maybe I do.  If its a testing concern, shouldn't it be on a top level test fixture to cleanup the environment before or after each test?

matteius: The issue without this change is that environment that gets used for running the script is the environment without the dotenv variables.  Is there a better way to accomplish it?

frostming: > shouldn't it be on a top level test fixture to cleanup the environment before or after each test?

Oh yeah, it is cleaned up.

frostming (approved):

oz123 (approved):

Issue #4637, opened 2021-02-26
Pipenv Lock (Or Install) Does Not Respect Index Specified For A Package


janneronkko:
### Issue description

Index argument for a package is not respected

### Expected result

Pipenv only installs (locks) versions from specified package index.

### Actual result

Pipenv installs (locks) the package with greatest version from a package index not specified in the Pipfile.

### Steps to replicate

Run the following script:

```sh
#! /usr/bin/env bash

set -euxo pipefail


function finish {
  set +e

  kill -9 ${PYPI_SERVER1_PID} || true
  kill -9 ${PYPI_SERVER2_PID} || true
}

trap finish EXIT

cat > example.py <<EOF
import pkg_resources

def main():
    print(pkg_resources.get_distribution('example'))
EOF

cat > setup.py <<EOF
import os

import setuptools

setuptools.setup(
    name='example',
    version=os.environ['EXAMPLE_VERSION'],
    description='Example package',
    py_modules=['example'],
    python_requires='>=3.6',
    entry_points={
        'console_scripts': [
            'show-installed-package = example:main',
        ],
    },
)
EOF

EXAMPLE_VERSION=1 python setup.py bdist_wheel
EXAMPLE_VERSION=2 python setup.py bdist_wheel

python -m venv venv
venv/bin/pip install pypiserver

mkdir -p server1 server2
cp dist/example-1-py3-none-any.whl server1/
cp dist/example-2-py3-none-any.whl server2/

function run_pypi_server() {
  venv/bin/pypi-server \
    -p 808${1} \
    -i 127.0.0.1 \
    server${1}/ 2> pypi-server${1}.log &
}

run_pypi_server 1
PYPI_SERVER1_PID=$!

run_pypi_server 2
PYPI_SERVER2_PID=$!

rm -f Pipfile.lock

cat > Pipfile <<EOF
[[source]]
url = "http://127.0.0.1:8081"
name = "server1"

[[source]]
url = "http://127.0.0.1:8082"
name = "server2"

[packages]
example = {version="*", index="server1"}

[dev-packages]

[requires]
python_version = "$(python -c 'import sys; print(f"{sys.version_info.major}.{sys.version_info.minor}")')"
EOF

pipenv lock
pipenv sync
pipenv run show-installed-package

```

The script creates two versions of _example_ Python packge (version 1 and 2). Then it starts pypi server; server 1 contains example version 1 and server 2 contains example version 2.

[...]

PR #4983, opened 2022-03-14
Fix for: Pipenv Does Not Respect Index Specified For A Package

matteius:
Code changes that fix the indexes not being respected.   Acknowledged that this modifies code in pip, and so I will need extra considerations on how to proceed  I have opened a PR against Pip as well (https://github.com/pypa/pip/pull/10964) to propose my modifications of pip from here.   Yet without this logic, there is no way to enforce the index specifications are abided by during resolution and installation.   After several iterations of trying to understand this code and how to fix it, it feels like the `LinkCollector`''s `SearchScope` should have the knowledge it needs to return the right indexes at the right times.

### The issue
Pipenv Lock (Or Install) Does Not Respect Index Specified For A Package  https://github.com/pypa/pipenv/issues/4637

### The fix

Without this or similar logic, there is no way to enforce the index specifications are abided by during resolution and installation.   After several iterations of trying to understand this code and how to fix it, it feels like the `LinkCollector`''s `SearchScope` should have the knowledge it needs to return the right indexes at the right times.  Otherwise the pipenv resolver is totally disconnected from the pip resolver in terms of these specifications.

_The example test script is very non-trivial and I have gotten much of it into an integration test locally, but there are some stumbling blocks, such as it involves running two private pypi servers locally each pointing at one of the generated package versions of an `example` generated wheel package._

### The checklist

* [X] Associated issue

### DIFF
--- a/news/4637.bugfix.rst
+++ b/news/4637.bugfix.rst
@@ -1,0 +1,1 @@
+Patched our vendored Pip to fix: Pipenv Lock (Or Install) Does Not Respect Index Specified For A Package.
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -1437,7 +1437,7 @@ def pip_install(
         project, index, extra_indexes=extra_indexes, trusted_hosts=trusted_hosts,
         pypi_mirror=pypi_mirror
     )
-    sources = list(filter(lambda d: d['name'] == requirement.index, sources))
+    sources = list(filter(lambda d: d.get('name') == requirement.index, sources))
     if r:
         with open(r, "r") as fh:
             if "--hash" not in fh.read():
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -1413,9 +1413,8 @@ def pip_install(
             ignore_hashes = False
     line = None
     # Try installing for each source in project.sources.
-    if requirement.index:
+    if not index and requirement.index:
         index = requirement.index
-        extra_indexes = []
     if index and not extra_indexes:
         extra_indexes = list(project.sources)
         extra_indexes = list(filter(lambda d: d['name'] == requirement.index, extra_indexes))
@@ -1462,7 +1461,8 @@ def pip_install(
         pip_command.extend(["-r", vistir.path.normalize_path(r)])
     elif line:
         pip_command.extend(line)
-    pip_command.extend(prepare_pip_source_args(sources))
+    pip_source_args = prepare_pip_source_args(sources)
+    pip_command.extend(pip_source_args)
     if project.s.is_verbose():
         click.echo(f"$ {cmd_list_to_shell(pip_command)}", err=True)
     cache_dir = Path(project.s.PIPENV_CACHE_DIR)
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -1413,9 +1413,10 @@ def pip_install(
             ignore_hashes = False
     line = None
     # Try installing for each source in project.sources.
-    if not index and requirement.index:
+    if requirement.index:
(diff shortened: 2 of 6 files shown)


matteius:
Note:  There appears to be 1 failing test when they are run together, but that test passes in isolation.   I'll have to find more time to investigate that.
```
(pipenv) matteius@matteius-VirtualBox:~/pipenv$ python -m pytest -sv -k test_install_specifying_index_url
======================================================= test session starts ========================================================
platform linux -- Python 3.10.2, pytest-6.2.5, py-1.11.0, pluggy-1.0.0 -- /home/matteius/.virtualenvs/pipenv-wR0g04D--3.7/bin/python
cachedir: .pytest_cache
rootdir: /home/matteius/pipenv, configfile: setup.cfg, testpaths: tests
plugins: flaky-3.7.0, forked-1.3.0, xdist-2.4.0, timeout-2.0.1, pypi-0.1.1
collected 282 items / 281 deselected / 1 selected                                                                                  

tests/integration/test_install_uri.py::test_install_specifying_index_url $ pipenv install pipenv-test-private-package --index https://test.pypi.org/simple
Installing pipenv-test-private-package...
Installing dependencies from Pipfile.lock (b8219d)...

Pipfile: /tmp/pipenv-a70hf424-project/Pipfile
Using /home/matteius/.virtualenvs/pipenv-wR0g04D--3.7/bin/python3 (3.10.2) to create virtualenv...
created virtual environment CPython3.10.2.final.0-64 in 811ms
  creator CPython3Posix(dest=/tmp/pipenv-a70hf424-project/.venv, clear=False, no_vcs_ignore=False, global=False)
  seeder FromAppData(download=False, pip=bundle, setuptools=bundle, wheel=bundle, via=copy, app_data_dir=/tmp/pipenv-2zxclxjk-test/home/.local/share/virtualenv)
    added seed packages: pip==21.3.1, setuptools==58.3.0, wheel==0.37.0
  activators BashActivator,CShellActivator,FishActivator,NushellActivator,PowerShellActivator,PythonActivator

Successfully created virtual environment!
Virtualenv location: /tmp/pipenv-a70hf424-project/.venv
Adding pipenv-test-private-package to Pipfile's [packages]...
Installation Succeeded
Building requirements...
Resolving dependencies...
Success!
Updated Pipfile.lock (b8219d)!

PASSED

================================================ 1 passed, 281 deselected in 21.85s ================================================
```

matteius:
I narrowed down the test failure:  specifically which other test cannot be run with the failing test, if you run this pytest command, the current suite fails and I suspect it is because both of these tests pass in `--index` and that ends up contaminating something (the tests are not fully isolated)  However if you reverse the order of these two tests they both pass, which I find very odd.
`python -m pytest -sv tests/integration/test_install_uri.py::test_install_named_index_alias tests/integration/test_install_uri.py::test_install_specifying_index_url`

frostming (changes requested):
Need to update the patch files as well.

matteius:
> Need to update the patch files as well.

@frostming Thanks for checking on this 💯   I just opened a PR against Pip -- any idea how that group will react?  I am willing to make this into a patch but it seems like it would be better if it gets accepted upstream.   If it does not gain traction and needs to be patched instead, we might want to do it once 22.0.4 vendoring branch is merged in since the patches changed a lot between now and then.

frostming:
> it would be better if it gets accepted upstream

It isn't likely to be accepted since it isn't useful for `pip`.

matteius:
Thanks @frostming the patch file has been added.

oz123 (approved):

Issue #2222, opened 2018-05-20
Pipenv unexpectedly installs from requirements.txt (even from parent directories)


W7PEA:
This was a weird one and it was so odd that I thought I was losing my mind for a bit. ;-) Thanks for pipenv though, I'm very excited about this new direction!

Pipenv installs requirements.txt when you don't ask it to, and when it try to keep it from installing them, it still does. This is on Windows 10. Not sure if reproduces on Linux\OSx

`$ python -m pipenv.help`

<details><summary>$ python -m pipenv.help output</summary>

Pipenv version: `'2018.05.18'`

Pipenv location: `'C:\\Users\\patri\\AppData\\Local\\Programs\\Python\\Python36\\lib\\site-packages\\pipenv'`

Python location: `'C:\\Users\\patri\\AppData\\Local\\Programs\\Python\\Python36\\python.exe'`

Other Python installations in `PATH`:

  - `3.6`: `C:\Users\patri\AppData\Local\Programs\Python\Python36\python.exe`

  - `3.6.5`: `C:\Users\patri\AppData\Local\Programs\Python\Python36\python.exe`
  - `3.6.5`: `C:\Windows\py.exe`

PEP 508 Information:

```
{'implementation_name': 'cpython',
 'implementation_version': '3.6.5',
 'os_name': 'nt',
 'platform_machine': 'AMD64',
 'platform_python_implementation': 'CPython',
 'platform_release': '10',
 'platform_system': 'Windows',
 'platform_version': '10.0.17134',
 'python_full_version': '3.6.5',
 'python_version': '3.6',
 'sys_platform': 'win32'}
```

System environment variables:

  - `ALLUSERSPROFILE`
  - `APPDATA`
  - `COMMONPROGRAMFILES`
  - `COMMONPROGRAMFILES(X86)`
  - `COMMONPROGRAMW6432`
  - `COMPUTERNAME`
  - `COMSPEC`
  - `DRIVERDATA`
  - `GDAL_DATA`
  - `HOMEDRIVE`
  - `HOMEPATH`
  - `LOCALAPPDATA`
  - `LOGONSERVER`
  - `NUMBER_OF_PROCESSORS`
  - `ONEDRIVE`
  - `OS`
  - `OSGEO4W_ROOT`
  - `PATH`
  - `PATHEXT`
  - `POSTGIS_ENABLE_OUTDB_RASTERS`
  - `POSTGIS_GDAL_ENABLED_DRIVERS`
[...]

PR #4987, opened 2022-03-16
display the directory path of the found requirements.txt file

HosamAlmoghraby:

### The issue

**https://github.com/pypa/pipenv/issues/2222**
**https://github.com/pypa/pipenv/issues/2870**

### The fix

_When pipenv does not find Pipfile, it tries to find requirements.txt file, if the file is not in the same directory, then, pipenv looks in the parent directories (the default settings PIPENV_MAX_DEPTH=3)._
_when pipenv finds the requirements.txt file either in the same folder or in one of the parent folders, it prints the message: "**requirements.txt found, instead of Pipfile! Converting...**"._
_I think this message is cofusing for the user when there is no requirements.txt file stored in the same folder and the user might not know that pipenv looks in the parent folders as well._
_The confusion here could be avoided by displaying to the user the folder path of the requirements.txt file where pipenv found it._

_Here is my proposal: "**requirements.txt found in <requirements.txt-DIR-PATH> instead of Pipfile! Converting...**"
I bolded **requirements.txt** and **Pipfile** and colored the directory path to the requirements.txt file with Yellow._

### The checklist

* [ ] Associated issue

### DIFF
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -241,10 +241,12 @@ def ensure_pipfile(project, validate=True, skip_requirements=False, system=False
             )
         # If there's a requirements file, but no Pipfile...
         if project.requirements_exists and not skip_requirements:
+            requirements_dir_path = os.path.dirname(project.requirements_location)
             click.echo(
-                crayons.normal(
-                    fix_utf8("requirements.txt found, instead of Pipfile! Converting..."),
-                    bold=True,
+                "{0} found in {1} instead of {2}! Converting...".format(
+                    crayons.normal("requirements.txt", bold=True),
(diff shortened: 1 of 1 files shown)


HosamAlmoghraby:
here is a screenshot of how the message will look like:
image

`mkdir outer-dir`
`cd outer-dir`
`touch requirements.txt`
`mkdir inner-dir`
`cd inner-dir`
`pipenv shell`

matteius:
Personally I think it looks good, but also the current tests broke because of a git:// protocol deprecation, so not running this until we can merge:  https://github.com/pypa/pipenv/pull/4989 and your PR gets updated with the updated main branch at that time.

oz123 (approved):

Issue #4626, opened 2021-02-19
--selective-upgrade does not update hashes in Pipfile.lock


mogoh:
### Issue description

If I update a package with --selective-upgrade, I can not install the update with sync.

### Steps to replicate

pipenv install 'django==2.2.18'
pipenv install --selective-upgrade 'django==2.2.19'
pipenv --rm
pipenv sync

### Expected result

pipenv sync should install all packages from Pipfile.lock.

### Actual result

```
$ pipenv sync
Creating a virtualenv for this project...
Pipfile: /home/mogoh/temp/Pipfile
Using /usr/bin/python3.8 (3.8.6) to create virtualenv...
⠹ Creating virtual environment...created virtual environment CPython3.8.6.final.0-64 in 114ms
  creator CPython3Posix(dest=/home/mogoh/.local/share/virtualenvs/temp-O2KvBR8F, clear=False, global=False)
  seeder FromAppData(download=False, pip=bundle, setuptools=bundle, wheel=bundle, via=copy, app_data_dir=/home/mogoh/.local/share/virtualenv)
    added seed packages: pip==20.2.4, pkg_resources==0.0.0, setuptools==50.3.2, wheel==0.35.1
  activators BashActivator,CShellActivator,FishActivator,PowerShellActivator,PythonActivator,XonshActivator
✔ Successfully created virtual environment! 
Virtualenv location: /home/mogoh/.virtualenvs/temp-O2KvBR8F
Installing dependencies from Pipfile.lock (d1ee3e)...
An error occurred while installing django==2.2.19 --hash=sha256:c9c994f5e0a032cbd45089798b52e4080f4dea7241c58e3e0636c54146480bb4 --hash=sha256:0eaca08f236bf502a9773e53623f766cc3ceee6453cc41e6de1c8b80f07d2364! Will try again.
  🐍   ▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉ 3/3 — 00:00:00
Installing initially failed dependencies...
[InstallError]:   File "/home/mogoh/.local/lib/python3.8/site-packages/pipenv/cli/command.py", line 684, in sync
[InstallError]:       retcode = do_sync(
[InstallError]:   File "/home/mogoh/.local/lib/python3.8/site-packages/pipenv/core.py", line 2884, in do_sync
[InstallError]:       do_init(
[...]

Issue #4351, opened 2020-06-17
Pipenv install --selective-upgrade


saisankargochhayat:
### Issue description
Attempting just to upgrade one package, with out updating other packages. 

### Steps to replicate
Git clone - https://github.com/saisankargochhayat/pipenvTest
Couple of scenarios I tried out just to update one package requests - 
- `pipenv install requests==2.24.0 --keep-outdated`
  Just updates the package version, but not the hashes in the Pipfile.lock, pins down in Pipfile
- `pipenv install --selective-upgrade requests==2.24.0`
   Just updates the package version, but not the hashes in the Pipfile.lock, pins down in Pipfile
- `pipenv install requests==2.24.0` 
   Updates both requests and semver to their latest versions and the hashes. 
- `pipenv update --selective-upgrade requests`
   Updates both requests and semver to their latest versions and the hashes. 

Seems like `pipenv install {dependency}=={package_version} --keep-outdated` earlier did the trick, but now it works, but just misses out on the new hash. Is there a new way to achieve this?

-------------------------------------------------------------------------------
Pipenv support output - https://gist.github.com/saisankargochhayat/a7d09bdc1cea8eeadf902a1bae503b1d

saisankargochhayat:
`pipenv install requests==2.24.0 --keep-outdated` still doesn't seem to be working with Pipenv 2020.8.13 :disappointed: 
![image](https://user-images.githubusercontent.com/9396452/91492329-7afd4580-e883-11ea-8b6a-9710e5dae1ff.png)

danstewart:
I have been having this issue and have found a workaround.

```
# Install the package
# This will update the version number in Pipfile and the Pipfile.lock - but not the package hashes
# I'm not sure if --keep-outdated does anything here, I get the same results without it but I include it to be safe
pipenv install --keep-outdated --selective-upgrade package==0.0.1

# This will update the hashes for the modified package only
pipenv lock --keep-outdated
```

The only downside is that it seems to _add_ the new hashes and keeps the old ones, but that doesn't seem too big of a deal to me.

matteius:
`--keep-outdated` and likely `--selective-upgrade` are deprecated/slated for removal.   I've been working on an alternative command that solves the kind of problems these other flags caused.   Checkout out: https://github.com/pypa/pipenv/pull/5617

Issue #2412, opened 2018-06-25
Upgrading a dependency with --selective-upgrade doesn't seem to work with 2018.6.25


mark-adams:
Running using the latest release (2018.6.25), selective upgrades don't appear to work as expected. (If I'm doing something wrong, please let me know :-) )

This was originally filed as #2410 but was closed out because it was thought to have been fixed in the latest release.

/home/madams/.local/venvs/pipenv/bin/python -m 'pipenv.help'
<details><summary>$ python -m pipenv.help output</summary>

Pipenv version: `'2018.6.25'`

Pipenv location: `'/home/madams/.local/venvs/pipenv/local/lib/python2.7/site-packages/pipenv'`

Python location: `'/home/madams/.local/venvs/pipenv/bin/python'`

Other Python installations in `PATH`:

  - `2.7`: `/usr/bin/python2.7`
  - `2.7`: `/usr/bin/python2.7`
  - `3.6`: `/usr/bin/python3.6m`
  - `3.6`: `/usr/bin/python3.6`

  - `2.7.15`: `/usr/bin/python`
  - `2.7.15`: `/usr/bin/python2`
  - `3.6.5`: `/usr/bin/python3`

PEP 508 Information:

```
{'implementation_name': 'cpython',
 'implementation_version': '0',
 'os_name': 'posix',
 'platform_machine': 'x86_64',
 'platform_python_implementation': 'CPython',
 'platform_release': '4.15.0-23-generic',
 'platform_system': 'Linux',
 'platform_version': '#25-Ubuntu SMP Wed May 23 18:02:16 UTC 2018',
 'python_full_version': '2.7.14',
 'python_version': '2.7',
 'sys_platform': 'linux2'}
```

System environment variables:

  - `GOPATH`
  - `PYPI_PASSWORD`
  - `IM_CONFIG_PHASE`
  - `LESS`
  - `QT4_IM_MODULE`
  - `GJS_DEBUG_OUTPUT`
  - `PROJECT_HOME`
  - `LC_CTYPE`
  - `WINDOWPATH`
  - `XDG_CURRENT_DESKTOP`
  - `XDG_SESSION_TYPE`
  - `TERM_PROGRAM_VERSION`
  - `QT_IM_MODULE`
  - `LOGNAME`
  - `USER`
  - `PATH`
  - `XDG_VTNR`
  - `HOME`
  - `VSCODE_IPC_HOOK`
  - `VIRTUALENVWRAPPER_SCRIPT`
[...]

PR #4988, opened 2022-03-16
Issue 4650 Remove unsupported feature selective upgrade

matteius:
I am proposing via PR we remove the flag `--selective-upgrade` on several grounds:
* It doesn't work and hasn't for a long time
* The suggestion for people using it are to use `--keep-outdated` instead.
* There is no documentation for the code path.
* The code path doesn't make sense and exists in overly complex method we aim to simplify anyway.
* It does things like uses Pip feature 
  * to-satisfy-only (undocumented, please avoid) - packages are not upgraded (not even direct requirements) unless the currently installed version fails to satisfy a requirement (either explicitly specified or a dependency).
    * This is actually the “default” upgrade strategy when --upgrade is not set, i.e. pip install AlreadyInstalled and pip install --upgrade --upgrade-strategy=to-satisfy-only AlreadyInstalled yield the same behavior.
    * Ref:  https://pip.pypa.io/en/stable/development/architecture/upgrade-options/#controlling-what-gets-installed
* the https://pip.pypa.io/en/stable/cli/pip/#exists-action-option usage also doesn't seem to be anything special or make extra sense why its would be needed over `--keep-outdated`.   
* More reasons to follow if necessary ...

### The issue

https://github.com/pypa/pipenv/issues/4650
https://github.com/pypa/pipenv/issues/4626
https://github.com/pypa/pipenv/issues/4351
https://github.com/pypa/pipenv/issues/3461
https://github.com/pypa/pipenv/issues/2412
And more:  https://github.com/pypa/pipenv/issues?q=is%3Aissue+is%3Aopen+selective-upgrade

### The fix

I removed the code for `--selective-upgrade`

### The checklist

* [X] Associated issue

### DIFF
--- a/news/4650.removal.rst
+++ b/news/4650.removal.rst
@@ -1,0 +1,1 @@
+Removal of undocumented and broken install feature flag `--selective-upgrade` in favor of the flag `--keep-outdated`
--- a/pipenv/cli/command.py
+++ b/pipenv/cli/command.py
@@ -207,7 +207,6 @@ def install(
         code=state.installstate.code,
         deploy=state.installstate.deploy,
         keep_outdated=state.installstate.keep_outdated,
-        selective_upgrade=state.installstate.selective_upgrade,
         index_url=state.index,
         extra_index_url=state.extra_index_urls,
         packages=state.installstate.packages,
--- a/pipenv/cli/options.py
+++ b/pipenv/cli/options.py
@@ -75,7 +75,6 @@ class InstallState:
     def __init__(self):
         self.dev = False
         self.pre = False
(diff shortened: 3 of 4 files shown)


matteius:
Putting this one on hold until -- even though this flag isn't really supported, we probably need to support something that does the right thing for selective upgrades now that I realize what `--keep-outdated` is doing.  Will revisit this.

matteius:
Deferring this for now -- it needs to be revisited, but one user suggested that we keep the functionality and don't support it.  I don't love that approach either, but perhaps there is a way to cleanup or understand the implementation/implications in time.

PR #4990, opened 2022-03-17
Code reorg utils into utils module reduces complexity

matteius:
### The issue

The pipenv/utils.py has grown to 2338 lines of code, which makes it unwieldy and disorganized.  When trying to consider larger bugs and refactors this is problematic because of the logical complexity of navigating around and understanding what kind of util you are looking at.  I would like to do something similar with core.py at some point as well, but this is my first iteration of proposing reorg'ing some of the code.

https://github.com/pypa/pipenv/issues/4992

### The fix

Move the existing utils.py code into a utils module organized by functionality type.   This is a first pass at it, and I suspect there may be some tweaking to my naming conventions to make everyone happy.   The only draw back I see to doing this is loosing some of the git annotations around this code, but ultimately that would happen whenever we get around to organizing the code better and I think good code organization is more important than legacy annotations which would still ultimately be in the git history.

The only modifications to code being moved are the following methods have been dropped due to lack of use:
* `escape_grouped_arguments` + unit tests
* `parse_python_version`  + unit tests
* `add_to_set`
* `sys_version`
* `get_indexes_from_requirement`
* `interrupt_handled_subprocess`
* `rmtree`

Also two unused code lines were removed `pipenv/core.py` because I had already run the test suite showing they were not used after my IDE tipped me off about that.

### The checklist

* [x] Associated issue


pipenv/core.py
```diff
@@ -1031,10 +1031,7 @@ def do_lock(
         for k, v in lockfile[section].copy().items():
             if not hasattr(v, "keys"):
                 del lockfile[section][k]
-    # Ensure that develop inherits from default.
```
matteius, line 1034: These lines were unused.

pipenv/utils/dependencies.py
```diff
@@ -0,0 +1,393 @@
...
+
+
+def get_vcs_deps(
+    project=None,
+    dev=False,
+    pypi_mirror=None,
+    packages=None,
+    reqs=None
+):
+    from pipenv.vendor.requirementslib.models.requirements import Requirement
```
oz123, line 138: Why sad face here? Can this be moved one line up?

oz123 (commented):
It's a big one, I'll need a few more passes to review all of it.

pipenv/utils/dependencies.py
matteius: Sad face?  Can which be moved up a line?  The blank line 128 or the import statement to the top of the file?  Aside:  I'd like to do another PR soon that introduces pre-commit hooks and linting setup similar to what pip and many other projects have.

oz123: L136 looks like a sad face. IMHO it could be moved up.

matteius: Oh, I see it now lol.  I honestly suspect that the flake8 and pre-commit linting auto-formatting stuff I want to add to the code base would prefer this same for any method  whose parameters need to go on multiple lines though it does look like they could all fit on one line here.   The reality is I just moved existing code (https://github.com/pypa/pipenv/pull/4990/files#diff-c0f7c08428b42c6f1d1e153d4286ad792e8e26154e342718fc3e5b9bcbb39453L1883) here in this PR so don't want to get into the business of reformatting it when I have bigger plans to try and apply auto linting to just the pipenv files (and not the patched vendor ones) for that will be a separate PR/issue.

oz123: ok. I see.

news/4992.feature.rst
```diff
@@ -0,0 +1 @@
+Internal to pipenv, the utils.py was split into a utils module with unused code removed.
```
matteius, line 1: Note: that in the fast-follow linting PR I added a new type `process` same as Pip has, and this news fragment is converted to a `process` type in the subsequent PR.  I went back and forth on feature vs process for this one, but its not an end-user facing feature.

oz123 (approved):

oz123:
@matteius I think this can be merged.
@frostming want to have a look at this giant PR?

matteius:
I am going to merge this because I need to modify a util based on a new issue report from today, and I am tired of keeping those changes in sync with this PR.  It will also allow merging the linting PR.

Issue #4996, opened 2022-03-20
Add pytest coverage report and Review and remove more unused code


matteius:
Place holder ticket to continue work improving the readability of the code base. 

### Issue description

Noticed that `PIPENV_MAX_ROUNDS` is unused and the comments mention pip-tools which appears to be a prior requirement that we have moved away from. 

Going to start reviewing pip-tools comments in the code and go from there.

Afsan751:
#4906 #4775 #4990

PR #4997, opened 2022-03-20
4996 Remove reference to pip-tools, add pytest-cov and refactor some code based on findings

matteius:
This gives us coverage report so we can study what parts of the code are not being exercised by the existing tests, which can be indicators of unused code paths in some cases, such totally unused methods or hard to reach conditional branches that may not be relevant anymore.  It also helps finds code that should have been covered by tests.

### The issue

https://github.com/pypa/pipenv/issues/4996

### The fix

Let's take an example of a method that should be refactored and study its coverage report:  
image

At first glance it appears all tests pass and this method perhaps could be reduced to:
```
    def _build_package_list(self, package_section):
        """Returns the list of packages from the Project's Pipfile."""
        ps = {}
        for k, v in self.parsed_pipfile.get(package_section, {}).items():
            ps.update({k: v})
        return ps
```
But if that is true, then why iterate over the result to build up the same result, so that really becomes:
```
    def _build_package_list(self, package_section):
        """Returns the list of packages from the Project's Pipfile."""
        return self.parsed_pipfile.get(package_section, {})
```

So does this mean we have missed some test case that is important here to try and hit that super complex hard to reach conditional, or the later one to actually be excluded, or can we really simplify this 5 year old method from multiple authors that far?

Going to see what else the coverage report reveals ...

### The checklist

* [X] Associated issue


pipenv/project.py
```diff
@@ -537,8 +487,11 @@ def parsed_pipfile(self):
 
         (call clear_pipfile_cache() afterwards if mutating)"""
         contents = self.read_pipfile()
-        # use full contents to get around str/bytes 2/3 issues
-        cache_key = (self.pipfile_location, contents)
+        # This should be pretty fast; we need this pipfile a lot
```
matteius, line 490: This was the prior implementation of cache_key before someone realized it didn't work with py2 and made it be the entire contents.   Now that py2 is not supported, we can make this cache key short again.

pipenv/utils/shell.py
```diff
@@ -242,7 +242,7 @@ def is_virtual_environment(path):
     for bindir_name in ("bin", "Scripts"):
         for python in path.joinpath(bindir_name).glob("python*"):
             try:
-                exeness = is_file() and os.access(str(python), os.X_OK)
+                exeness = python.is_file() and os.access(str(python), os.X_OK)
```
matteius, line 245: This was a bug from my utils splitting (thanks pycharm import refactor) and I have ported this fix back to that branch which is still out for PR.

matteius:
As far as planning goes -- this one will be delayed until sometime after we update the pip vendoring to 22.0.4 since we will be dropping support for python 3.6 at that time, and these coverage requirements require at least python 3.7.   So this WIP PR could be considered in a later release once all that other pending work has been verified.

matteius:
Closing in favor of branch with fewer commits/merge conflicts:  https://github.com/pypa/pipenv/pull/5038

Issue #4998, opened 2022-03-21
Remove option --code which has been broken since 2018


matteius:
Remove option --code which has been broken since 2018

### Issue description

Our code in pipenv has been passing a boolean for a path value and silently swallowing sine 2018 for this code path.  The intention of the code path is to generate a requirements.txt from inspecting the project imports.   To do this, it uses pipreqs which has not been well maintained and contains a mapping of project names where the imports don't match the project name -- which is many projects these days.   I find the mapping file is largely incomplete and the CLI tool which can generate the requirements.txt is largely incomplete for the basic test projects I have tried this functionality on.  See also:

* https://github.com/bndr/pipreqs/issues/252
* https://github.com/bndr/pipreqs/blob/master/pipreqs/mapping

### The Fix

Rather than try to fix this under documented and broken feature, I believe it is time to remove it.  There are already better ways to generate a Pipfile from a requirements.txt and a requirements.txt from the Pipfile.lock.  It would be more than a matter than passing the path in to fix it in, because the method we call also raises exception if it cannot resolve a single dependency which is likely.

Recommendation:  If someone needs to generate a requirements.txt using pipreqs they can just install pipreqs and use their CLI to get started.  

We do not need to continue vendoring it into pipenv, which will also allow us to drop `yarg` vendored dependency as well.

PR #4999, opened 2022-03-21
4998 remove long broken pipreqs usage

matteius:
Remove option --code from install and --unused from check which has been broken since 2018.  Also removed unused args from check command, and removed the vendored dependencies:
* pipreqs
* yarg
* doctopt

### The issue
For more details on rationale:  
https://github.com/pypa/pipenv/issues/4998

### The checklist

* [X] Associated issue


tests/integration/test_cli.py
```diff
@@ -262,29 +262,6 @@ def test_install_parse_error(PipenvInstance):
...
-import flask
-            """.strip()
-            f.write(contents)
-        p.pipenv('install requests click flask')
-
-        assert all(pkg in p.pipfile['packages'] for pkg in ['requests', 'click', 'flask']), p.pipfile["packages"]
-
-        c = p.pipenv('check --unused .')
-        assert 'click' not in c.stdout
-        assert 'flask' not in c.stdout
```
matteius, line 285: Notice that it doesn't check that `requests` is actually in the `c.stdout` because if it had done that, it would have been obvious this hasn't been working in ages.

oz123 (changes requested):
I am totally fine with this.
Also notice, you can drop docopt!
https://github.com/bndr/pipreqs/pull/294

oz123 (approved):
I rebased your commits. All good now. Thanks for removing docopt too.

Issue #5000, opened 2022-03-22
Removed vendored requests and certifi that are unused


matteius:
Pipenv coverage report shows that  vendored requests and certifi that are unused.  Upon further examination, we utilize the certifi from the vendored pip and not the one that we vendor, and requests library remains unused by pipenv internals -- just the tests which call out it out as part of the dev dependencies.  

### Issue description

Placeholder for PR.

PR #5001, opened 2022-03-22
Removed vendored requests and certifi that are unused.

matteius:
Removes vendored requests and certifi because we can -- its not what we use in the project.

Edit:  so it appears we do use it, not sure why the coverage report has mislead me here.   Will revisit.

### The issue
https://github.com/pypa/pipenv/issues/5000


frostming:
Coverage report only tells there are not tests for it, rather than it isn't used. Using the coverage report as a mark may be misleading.

matteius:
> Coverage report only tells there are not tests for it, rather than it isn't used. Using the coverage report as a mark may be misleading.

Totally agree with the premise of your comment--in this case it is code that was exercised by tests but was showing red on the coverage report which seems odd.  At first I suspect that maybe because of where the import was happening or maybe it was using the `requests` from the Pipfile.lock requirements, but really if it was it wouldn't have blown up pulling out the vendor version.  

I was thinking there could be some value in being consistent in using the same version of requests for both pip and the remainder of the project but also still exploring this one.    @frostming  What are your thoughts about relying on the requests provided by pip rather than explicitly vendoring our own?

matteius:
I'll revisit this in a fresh PR.

PR #5010, opened 2022-03-25
Bugfix: Not loading .env when quiet

smac89:
### The issue

When the output is quieted, pipenv no longer loads the `.env` file

### The fix

Load .env but only print the message if quiet is False

### The checklist

* [ ] Associated issue


smac89:
fixes #5006

matteius (approved):

matteius:
@smac89 Nice catch!

PR #5012, opened 2022-03-26
5011 more tests

### LINKED ISSUE #5011: Add unit tests for recently introduced regressions
This ticket is to add tests for some of the recent regressions and feature changes:
* `pipenv run pip freeze` should have predictable and clean stdout logging
* `pipenv run script --quiet` should still load .env vars (and not log the message about it)
* Bonus would be: a test for the package index restriction.  This might be hard because we only have a single pypi server running, but I was thinking about an example that relies on pulling from our named pypi server when pypi.org is avialable and a newer package version exists there, so that might be a feasible test to add.

matteius:
fix #5011

### The checklist

* [X] Associated issue


matteius:
Too many attempts at setuptools and they published a new version 2 hours ago that hopefully fixes the bug:  https://pypi.org/project/setuptools/61.1.1/

Grafted the relevant commits to:  https://github.com/pypa/pipenv/pull/5015

Issue #4959, opened 2022-02-17
Feature request: pipenv should have a native way to generate requirements.txt directly from the lockfile


ImreC:
When you just run pipenv install [package] on your project the version in the Pipfile is not pinned, but the lockfile contains the installed package version. Unfortunately not all tooling around Python supports Pipfiles, which means that you very often need to generate a requirements.txt file based on your Pipfile.lock. I think all approaches for this are well described here: https://github.com/pypa/pipenv/issues/3493, but for simplicity I will outline them below:

1) Run `pipenv lock -r`. Why this does not work is that it runs `pipenv lock`, which generates a new lockfile and then generates the requirements. This leads to problematic cases if you want to keep the requirements exactly the same. 
2) Run `pipenv run pip freeze > requirements.txt` (potentially with `pipenv sync` before that to make sure that all the packages from Pipfile.lock` are installed). Why this is problematic is because running pip freeze very often leads to too many strict package dependencies being included which leads to dependency mismatches, rendering the whole file unusable. 
3) Some methods based on `jq` or some similar tool that literally parse the lockfile and then output a requirements.txt

I think the problem with option 1 and 2 is that they both do too much. Option 3 is closest to what most people need, but introduces a new dependency on `jq` which is also annoying to enforce across a larger group of developers with multiple environments. Which lead to that I am now including a simple python file containing this in almost every repo: 

```
import json

with open('Pipfile.lock', 'r') as f:
    reqs = f.read()

with open('requirements.txt', 'w') as f:
    f.writelines(f"{req_name}{value['version']}\n" for req_name, value in json.loads(reqs)['default'].items())
```

I feel that this is so mindbogglingly simple that it should be included somewhere in the pipenv package. My proposition is to include a command `pipenv requirements` which reads the Pipfile.lock json file and outputs a requirements.txt based on the default key. Adding `--dev` as an option also adds the requirements from the develop key. No locking, no syncing, no pip freezing. Just this. 

I am not only asking for this, but also happy to take some time to implement/maintain this if other people support this idea. 

Overall, I think pipenv is a vital tool in the Python ecosystem and will continue to use it in any case. Thanks for the work that went into this so far!  I think adding this seemingly trivial extra command would resolve some confusion around the requirements.txt (f.e. https://github.com/pypa/pipenv/issues/4731) and therefore would be a relatively low effort high value feature for pipenv.

matteius:
[...]

Issue #3751, opened 2019-05-18
pipenv injects secrets into Pipfile and lock file


TheFriendlyCoder:
### Issue description

When running pipenv in an environment which has a PIP_INDEX_URL environment variable defined, and that env var contains credentials to use when authenticating to the package index server, those credentials are injected as plain-text into the Pipfile and Pipefile.lock files.

### Expected result

pipenv should not inject index URLs containing credentials verbatim into the config file.

### Actual result

pipenv does inject index URLs containing credentials into the config file.

### Steps to replicate

We work behind a firewall with a mirror of the global PyPI repository in our lab. This mirror requires authentication to access. To make this work properly / easily with `pip` we generally expect users to define a custom PIP_INDEX_URL containing the users' credentials along with the URL for the mirror.

Further, when using pipenv to create a new environment or install a package (ie: `pipenv install requests`), we end up getting a Pipfile that looks something like this:

```
[[source]]
name = "pypi"
url = "https://pypi.org/simple"
verify_ssl = true

[dev-packages]

[packages]
requests = {index = "https://<credentials>@<server_url>/pypi/pypi-virtual/simple",version = "*"}

[requires]
python_version = "3.7"
```
I debated whether this should be reported as a "bug" or perhaps just a feature request to have this behavior changed, however given the severity of the impact of this default behavior I thought it was better suited as a bug. The tools should never inject secrets into files that may be commited to version control and this behavior could be very easily overlooked by users of the tool, which makes me very hesitant to roll out this tool for widespread use on my team. 

-------------------------------------------------------------------------------

<details><summary>$ pipenv --support</summary>

Pipenv version: `'2018.11.26'`

Pipenv location: `'/usr/local/lib/python3.7/site-packages/pipenv'`

Python location: `'/usr/local/opt/python/bin/python3.7'`

Python installations found:

  - `3.7.2`: `/usr/local/bin/python3`
  - `3.7.2`: `/usr/local/bin/python3.7m`
  - `3.6.1`: `/usr/local/bin/pypy3`
  - `3.5.2`: `/usr/local/bin/python3.5m`
  - `3.5.2`: `/usr/local/bin/python3.5`
  - `2.7.16`: `/usr/local/bin/python`
  - `2.7.16`: `/usr/local/bin/pythonw`
  - `2.7.13`: `/usr/local/bin/pypy`
  - `2.7.10`: `/usr/bin/python`
  - `2.7.10`: `/usr/bin/pythonw`
[...]

Issue #3138, opened 2018-10-31
Suppressing expansion of environment variables


spiderkeys:
### Issue description

I have a private repository configured in my Pipfile, which specifies credentials via environment variables in the manner described in the pipenv user guide:
```
[[source]]
url = "https://${JFROG_USERNAME}:${JFROG_PASSWORD}@myorg.jfrog.io/myorg/api/pypi/myrepo/simple"
verify_ssl = true
name = "myrepo"
```
Upon running 'pipenv install' and a package is unable to be found at the remote index, pipenv is printing the following message with environment variables fully expanded:
```
No versions found
Were https://pypi.org/simple or https://XXXXX:XXXXX@myorg.jfrog.io/myorg/api/pypi/myrepo/simple reachable?
```

### Expected result

I expected that credentials and other secrets specified via environment variables would remain unexpanded in stdout/stderr messages, such that secrets aren't leaked via CI logs when errors occur.

### Steps to replicate

- Create a pipfile that points to a remote index with credentials specified via environment variables
- Specify a package dependency that you know to not exist
- Run 'pipenv install'
- Observe expansion of environment variables in error output

-------------------------------------------------------------------------------

<details><summary>$ pipenv --support</summary>

Pipenv version: `'2018.10.13'`

Pipenv location: `'/home/spiderkeys/.local/lib/python3.6/site-packages/pipenv'`

Python location: `'/usr/bin/python3'`

Python installations found:

  - `3.7.1`: `/home/spiderkeys/.pyenv/versions/3.7.1/bin/python3.7`
  - `3.7.1`: `/home/spiderkeys/.pyenv/versions/3.7.1/bin/python3.7m`
  - `3.6.6`: `/usr/bin/python3.6`
  - `3.6.6`: `/usr/bin/python3.6m`
  - `2.7.15`: `/home/spiderkeys/.pyenv/versions/2.7.15/bin/python2.7`
  - `2.7.15rc1`: `/usr/bin/python2.7`

PEP 508 Information:

```
{'implementation_name': 'cpython',
 'implementation_version': '3.6.6',
 'os_name': 'posix',
 'platform_machine': 'x86_64',
 'platform_python_implementation': 'CPython',
 'platform_release': '4.15.0-23-generic',
 'platform_system': 'Linux',
 'platform_version': '#25-Ubuntu SMP Wed May 23 18:02:16 UTC 2018',
 'python_full_version': '3.6.6',
 'python_version': '3.6',
[...]

PR #5013, opened 2022-03-26
Implements requirements command as per #4959

ImreC:
This implements a feature according to https://github.com/pypa/pipenv/issues/4959.


matteius:
@ImreC Could you merge latest main branch into your branch so the CI should pass again?

pipenv/cli/command.py
```diff
@@ -709,5 +709,24 @@ def verify(state):
     sys.exit(0)
 
 
+@cli.command(
+    short_help="Generate a requirements.txt from Pipfile.lock.",
+    context_settings=CONTEXT_SETTINGS,
+)
+@option("--dev", is_flag=True, default=False, help="Also add development requirements.")
+@pass_state
+def reqs(state, dev=False):
```
matteius, line 718: Would you be open to calling this command `requirements` for clarity?

docs/advanced.rst
```diff
@@ -199,19 +199,45 @@ development dependencies::
...
-Finally, if you wish to generate a requirements file with only the
+If you wish to generate a requirements file with only the
 development requirements you can do that too, using the ``--dev-only``
 flag::
 
     $ pipenv lock -r --dev-only
     py==1.4.34
     pytest==3.2.3
 
+Often, you would want to generate a requirements file based on your current
```
matteius, line 210: Perhaps call this out as `Sometimes` instead of `Often` -- my concern here is that I've been realizing many folks are not taking full advantage of the dependency lock hash checking by just generating a requirements.txt and installing from that instead.   It would be good to make the documentation more clear about the benefits of using `pipenv install --deploy` or `pipenv sync` instead of generating requirements files because of the benefit of verification that the package version matches the same signature from when you locked/pinned to it.

tests/integration/test_reqs.py
```diff
@@ -0,0 +1,56 @@
...
+            {packages[0]}= "=={packages[1]}"
+            [dev-packages]
+            {dev_packages[0]}= "=={dev_packages[1]}"
+            """.strip()
+            f.write(contents)
+        p.pipenv('lock')
+        c = p.pipenv('reqs')
+        assert c.returncode == 0
+
+        assert '-i https://pypi.org/simple' in c.stdout
```
matteius, line 55: Users should be aware that generating a requirements.txt with pinned indexes will result in a pip command that will not fully respect the restricted indexes.   This was something that came up recently where we now have patched the pip internals within pipenv to ensure restricted indexes are obeyed, but this is another case where the end user only gets that benefit if they use pipenv to sync or install --deploy their dependencies, because pip itself will prefer a package in pypi and just use extra-indexes to search for things not in pypi, basically (it is probably a little more complicated than that but in a nutshell).

docs/advanced.rst
ImreC: I completely agree. This should only be used for the use case of interfacing with tooling that only supports requirements.txt.

pipenv/cli/command.py
ImreC: Most definitely. I wanted to call it that, but requirements is a long word to type. Will update

ImreC:
@matteius is the news file correct? I have never worked with this before

matteius:
> @matteius is the news file correct? I have never worked with this before

It should be name `4959.feature.rst`  is the main thing, and anything that is code should be in double backticks.  I have a linting PR out that will add pre-commit hooks to the project and so at that time it will be more obvious if you are naming the news fragments wrong among other things.   I think the content itself makes sense, just remember to update the command name there as well.

pipenv/cli/command.py
matteius: Thanks @ImreC for doing that, it will be more clear to non-English speakers I think why it exists.

tests/integration/test_reqs.py
ImreC: I also think that this should not be the main route, but if I don't add the indexes and you use multiple, the requirements file will fail completely. Anything required on my part currently?

matteius: I think you're good, I just was calling that out because the test reminded me of that fact.  I kicked off the test CI.

pipenv/cli/command.py
```diff
@@ -709,5 +709,24 @@ def verify(state):
     sys.exit(0)
 
 
+@cli.command(
+    short_help="Generate a requirements.txt from Pipfile.lock.",
+    context_settings=CONTEXT_SETTINGS,
+)
+@option("--dev", is_flag=True, default=False, help="Also add development requirements.")
```
oz123, line 716: Can you please add a flag for including the hashes?
Also, some people like keeping `requirements.txt` and `requirements-dev.txt`. I think this would be nice to have ``--dev-only`` too.

oz123 (changes requested):

pipenv/cli/command.py
matteius: Yeah we have another ticket that was just opened about getting hashes into the requirements.txt  -- https://github.com/pypa/pipenv/issues/5025

pipenv/cli/command.py
```diff
@@ -709,5 +709,24 @@ def verify(state):
...
+@pass_state
+def requirements(state, dev=False):
+    lockfile = state.project.lockfile_content
+    for i, package_index in enumerate(lockfile['_meta']['sources']):
+        prefix = '-i' if i == 0 else '--extra-index-url'
+        echo(crayons.normal(' '.join([prefix, package_index['url']])))
```
rittneje, line 722: This should output the index verbatim as specified in the Pipfile, _without_ expanding variables. It is fairly common to specify credentials for a private PyPI this way, and writing those into the resulting requirements.txt file is counter-productive.

I'm not sure whether your existing code is already doing this or not, just wanted to mention this need.

pipenv/cli/command.py
ImreC: I will add this today

pipenv/cli/command.py
ImreC: @rittneje interesting point. I did a couple of tests and for me the $ENV_VAR gets substituted in the output and ${ENV_VAR} does not. I am currently just echoing the value from Pipfile.lock to stdout. Do you have any suggestions how to prevent this substitution in a smart way? I am still thinking about it

pipenv/cli/command.py
```diff
@@ -714,17 +714,22 @@ def verify(state):
...
-        echo(crayons.normal(f"{req_name}{value['version']}"))
-    if dev:
+    if not dev_only:
+        for req_name, value in lockfile['default'].items():
+            hashes = [f' \\\n    --hash={h}' for h in value['hashes'] if hash]
+            echo(crayons.normal(''.join([req_name, value['version'], *hashes])))
+    if dev or dev_only:
         for req_name, value in lockfile['develop'].items():
-            echo(crayons.normal(f"{req_name}{value['version']}"))
+            hashes = [f' \\\n    --hash={h}' for h in value['hashes'] if hash]
```
oz123, line 731: I might be wrong here, but it seems to me you are checking if hash for every `value['hashes']`.
Can you rewrite this as:
```
if hash:
    hashes = [f' \\\n    --hash={h}' for h in value['hashes'] if hash]
else:
    hashes = []
```

oz123 (changes requested):

oz123:
@ImreC thank you for the latest changes! really good to see this, just a small change still, and I think this can be merged.

pipenv/cli/command.py
ImreC: Fixed

ImreC:
For now I have changed the test back so it passes. All checks that I have performed indicates that this works as expected (the literal index URL including env var keys gets outputted with no expansion)

matteius:
@ImreC I suspect for that test that shows the environment variables were not expanded, I think they are expanded.  It has been in known issue reports, here are just some:
https://github.com/pypa/pipenv/issues/3751
https://github.com/pypa/pipenv/issues/3138

matteius:
@ImreC I think it does explain your test failing and unless @rittneje has some inputs on this one, we may have to defer the environment variable in the pip indexes as a separate issue/enhancement.

For your second question, I am of the view that we should try to figure out how to support pinned indexes such as `requests = {index = "pypi",version = "*"}).` -- It is interesting though, because you are generating pip commands and I have patched pip in pipenv to be able to pin indexes correctly, meaning that the pip command is likely going to do it funky, potentially even wrong, by searching pypi and other indexes URLs for something that is index restricted.

matteius:
Ok so `requests = {index = "pypi",version = "*"})` isn't the best example because pypi is the default index and that basically gets pinned to pypi when when you leave off the index such as ` requests = {version = "*"})` ... a better example is torch:

```
[[source]]
verify_ssl = true
name = "pypi"

[[source]]
url = "https://download.pytorch.org/whl/cu113/"
verify_ssl = true
name = "downloadpytorch"

[packages]
torch = {version = "==1.11.0+cu113", index = "downloadpytorch"}

[dev-packages]

[requires]
python_version = "3.10"
```

ImreC:
I would definitely leave it in for now as the behavior is different than `pipenv requirements`, but it might make more sense to at some point take it out. Maybe `pipenv lock` does too much with the options currently in it. Conceptually it is maybe cleaner to use `pipenv lock` for creating a lockfile and doing that well, and then using `pipenv requirements` to generate requirements.txt from that, instead of trying to do everything from a single command with flags. I have not looked into the implementation, but happy to have a look at some point to see what it currently does.

matteius:
@ImreC Looks like your build is failing on the new linting checks I added recently.  Can you run `pre-commit install` on your branch and cleanup any errors generated during: `pre-commit run --all-files --verbose`.   `pip install pre-commit` if you don't have it already.

ImreC:
@matteius should be fixed

matteius (approved):

oz123:
@ImreC you might want to change your author email. It seems currently wrong. If you are fine with the current email, I can merge it like this.

ImreC:
@oz123 I changed my email address mid-PR so I think it is fine. Just for my understanding. Where is this visible?

ImreC:
> @matteius suppression expansion of variables is already merged into pip and I believe also into pipenv. So this should not happen here. Also, as @ImreC already mentioned index url are not part of `requirements.txt`. I believe pip can handle the following:
> 
> ```
> $ cat requirements.txt
> package_a==1.0.1 # from private pypi
> package_b==2.0.2 # from another pypi
> requests  # from pypi
> ```
> 
> can be resolved with:
> 
> ```
> pip install --extra-index-url https://private_a.org l --extra-index-url https://private_b.org
> ```
> 
> See also: https://stackoverflow.com/a/41926975

@oz123 I think this case will work fine. The only problem that I can see is where you have packages with clashing names, f.e. you have a private package index with a package called `requests` in it. This will definitely not work with requirements.txt, even though I think pipenv offers the option to specifically define that this package needs to come from the private index and not pypi. Or maybe similarly named packages exist in multiple locations. We could solve this by defining multiple requirements.txt outputs, but it seems like a really exotic case to me. In case you control the private index a more obvious thing to do would be to not name your package `requests`, or add prefixes or something.

matteius:
>  The only problem that I can see is where you have packages with clashing names, f.e. you have a private package index with a package called requests in it. This will definitely not work with requirements.txt, even though I think pipenv offers the option to specifically define that this package needs to come from the private index and not pypi. 

@ImreC Actually this shouldn't be a big issue anymore, please see:   https://github.com/pypa/pipenv/pull/5029

TLDR:  Basically all packages have to be pinned now to an index, and by default the first index in the file is the primary index that unspecified packages utilize.

oz123 (approved):

Issue #4027, opened 2019-11-14
Allow more control over .env loading message


mjpieters:
I'm running automated `pipenv run ...` commands, and each and every time there is a `Loading .env environment variables` message.

Please, let me disable that message. `pipenv run` doesn't take switches nor is there a `PIPENV_BARE` environment variable, and so [output of the message is unconditional](https://github.com/pypa/pipenv/blob/master/pipenv/core.py#L148-L152).

Either support `--bare` for `pipenv run`, or better still, let me control pipenv output with an environment variable like `PIPENV_BARE` or `PIPENV_QUIET`.

What I don't want:

* Having to disable .env loading with `PIPENV_DONT_LOAD_ENV`
* Having to redirect stderr to /dev/null; the commands I run with `pipenv run` have to be able to use stderr too, and if the .env is corrupted or `PIPENV_DOTENV_LOCATION` is set I don't want error messages about those issues to be missed.

264nm:
Second this one...

teroyks:
A third from me.

codl:
It seems that since 2022.3.23 the message is printed regardless of the presence of a `.env` file, and on stdout instead of stderr, making this even more of an issue

Project0:
Super annoying, this just broke some CI pipelines for us.

matteius:
@codl thanks for pointing this out--would be happy to accept a PR that doesn't output this message when quiet is `PIPENV_QUIET` however the change to stdout from stderr is because it is not an error message, it is proper functioning of pipenv to load the .env file.   I'd also be fine with it not outputting the message when there is no .env file to load.   You can see the PR that changed this behavior to fix a different issue:  https://github.com/pypa/pipenv/pull/4979/files

@Project0 When you are using pipenv for CI pipelines you should probably have the version pinned to prevent a new release from breaking you, especially if you are dependent on the values of stdout not changing.  For example, we will be removing crayons library and converting everything to use `click.echo` in a subsequent release, and there is a big chance this will again chance some stdout messaging, so wouldn't want you to break when this happens.

We plan to have another release in a month, so if its something we can get approved/merged before then it will be included.   Let me know if no one from this group is able to make such a change and we can indicate it is something that should be done, maybe I can do it too when I free up.

matteius:
[...]

PR #5014, opened 2022-03-26
Fix issue where .env file is not loaded when quiet mode is used (#4027, #5006)

nomis:
### The issue

#4027 is incomplete because it stops the `.env` file from being loaded at all in quiet mode

### The checklist

* [x] Associated issue


matteius:
Thanks @nomis but this is a duplicate PR of https://github.com/pypa/pipenv/pull/5010

PR #5017, opened 2022-03-27
Subdirectory vcs installs are never passing the subdirectory in requirementslib which is now required by setuptools

### LINKED ISSUE #5019: new version of setuptools broke CI test for test_lock_nested_vcs_direct_url
I have spent 12+ hours debugging this and have come to the conclusion that new version of setuptools broke CI test for `test_lock_nested_vcs_direct_url` and it was because `requirementslib` isn't passing in the subdirectory as the directory, and this used to work for `setuptools` but now creates a different style package.  

### Issue description

PR out for review addresses this with a patch to `requirementslib` until we can address the CI failures in requirementslib and get a proper new build out there.

### Expected result

Expected existing test `test_lock_nested_vcs_direct_url` that passed 4 days ago to continue passing.
`
### Actual result

Well you know how it goes ...

matteius:
It would appear that potentially a recent change to setuptools revealed that we weren't actually passing in the subdirectory properly to generate the pep517 wheels, and that was revealed ~3 days ago when the CI started failing on a change that was merged in but had passed the CI the day before. 

Fixes #5019
Fixes #5014

### DIFF
--- a/tests/integration/test_cli.py
+++ b/tests/integration/test_cli.py
@@ -211,27 +211,6 @@ pyver = "which python"
 
 
 @pytest.mark.cli
-def test_scripts_resolve_dot_env_vars(PipenvInstance):
-    with PipenvInstance() as p:
-        with open(".env", "w") as f:
-            contents = """
-HELLO=WORLD
-            """.strip()
-            f.write(contents)
-
-        with open(p.pipfile_path, "w") as f:
-            contents = """
-[scripts]
-hello = "echo $HELLO"
-            """.strip()
-            f.write(contents)
-        c = p.pipenv('run hello')
-        print(c)
-        print(c.stdout)
-        assert 'WORLD' in c.stdout
-
-
-@pytest.mark.cli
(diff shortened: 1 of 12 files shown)


tests/integration/test_cli.py
```diff
@@ -210,27 +210,6 @@ def test_scripts(PipenvInstance):
         assert 'which python' in c.stdout
 
 
-@pytest.mark.cli
-def test_scripts_resolve_dot_env_vars(PipenvInstance):
```
matteius, line 214: I wrote this test recently but technically put it in the wrong file -- relocated it in this PR and extended it's use case.

frostming (approved):
LGTM, do you try to update `requirementslib` as well?

matteius:
Thanks @frostming, I have now access to requirementslib but unfortunately the build of main is already failing there by the time I have seen the project, so it will take me more time to patch upstream.   Its on my list since I already have another PR over there for the future pip vendoring as well.

PR #5026, opened 2022-03-30
More granular control over PIPENV_VENV_IN_PROJECT variable.

### LINKED ISSUE #2763: pipenv installs into .venv even though PIPENV_VENV_IN_PROJECT is not set
##### Issue description

If a folder exists called .venv, pipenv will automatically install into it even though PIPENV_VENV_IN_PROJECT is not set.

##### Expected result

I expected it to install into the global virtual environment location (usually `/home/moo/.local/share/virtualenvs/`)

##### Actual result

It installed into `./.venv/`

##### Extra information

It is worth to note what I am trying to achieve.

I am running Windows and VS Code installed in Windows. But I am also running WSL. What I am trying to do is to use WSL (bash) as my terminal in VS Code. There are few things to consider here.

1) I have installed python both in Windows and WSL/bash (Ubuntu).
2) I have set PIPENV_VENV_IN_PROJECT in Windows, but not in WSL.
3) The VS Code extension is configured like `"python.pythonPath": "${workspaceFolder}/.venv",`. For the VS Code extension to work, I need to execute `pipenv install` in Windows Powershell or cmd (i.e. in Windows). This is since VS Code is installed in Windows.
4) However, when I open the integrated terminal in VS Code I want to use WSL. As such, it cannot use .venv since everything installed there is Windows-specific. So I try to do `pipenv install` in the integrated terminal (running WSL/bash) in the hopes that it will install to the global venv location. This way, I could properly use both VS Code with the python extension, and at the same time use the integrated terminal to run things there as well.

##### Steps to replicate

```bash
14:44:36 DESKTOP-41D0GKF ~/test$ ll
total 0
drwxrwxrwx 1 moo moo 512 Aug 20 14:44 ./
drwxr-xr-x 1 moo moo 512 Aug 20 14:43 ../
14:44:39 DESKTOP-41D0GKF ~/test$ pipenv install requests
Pipfile: /home/moo/test/Pipfile
Using /usr/bin/python3 (3.5.2) to create virtualenv...
⠋Already using interpreter /usr/bin/python3
Using base prefix '/usr'
New python executable in /home/moo/.local/share/virtualenvs/test-O3XhBtkW/bin/python3
Also creating executable in /home/moo/.local/share/virtualenvs/test-O3XhBtkW/bin/python
Setting project for test-O3XhBtkW to /home/moo/test

Virtualenv location: /home/moo/.local/share/virtualenvs/test-O3XhBtkW
Creating a Pipfile for this project...                                                                                                                                                                                                      Installing requests...
Collecting requests
  Using cached https://files.pythonhosted.org/packages/65/47/7e02164a2a3db50ed6d8a6ab1d6d60b69c4c3fdf57a284257925dfc12bda/requests-2.19.1-py2.py3-none-any.whl
Collecting certifi>=2017.4.17 (from requests)
  Using cached https://files.pythonhosted.org/packages/16/1f/50d729c104b21c1042aa51560da6141d1cab476ba7015d92b2111c8db841/certifi-2018.8.13-py2.py3-none-any.whl
Collecting chardet<3.1.0,>=3.0.2 (from requests)
  Using cached https://files.pythonhosted.org/packages/bc/a9/01ffebfb562e4274b6487b4bb1ddec7ca55ec7510b22e4c51f14098443b8/chardet-3.0.4-py2.py3-none-any.whl
Collecting idna<2.8,>=2.5 (from requests)
  Using cached https://files.pythonhosted.org/packages/4b/2a/0276479a4b3caeb8a8c1af2f8e4355746a97fab05a372e4a2c6a6b876165/idna-2.7-py2.py3-none-any.whl
Collecting urllib3<1.24,>=1.21.1 (from requests)
  Using cached https://files.pythonhosted.org/packages/bd/c9/6fdd990019071a4a32a5e7cb78a1d92c53851ef4f56f62a3486e6a7d8ffb/urllib3-1.23-py2.py3-none-any.whl
Installing collected packages: certifi, chardet, idna, urllib3, requests
Successfully installed certifi-2018.8.13 chardet-3.0.4 idna-2.7 requests-2.19.1 urllib3-1.23

Adding requests to Pipfile's [packages]...
Updated Pipfile.lock (0b4483)!
Installing dependencies from Pipfile.lock (0b4483)...
  🐍   ▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉ 5/5 — 00:00:00
14:45:07 DESKTOP-41D0GKF ~/test$
14:48:39 DESKTOP-41D0GKF ~/test$
14:48:39 DESKTOP-41D0GKF ~/test$ ll
total 4
drwxrwxrwx 1 moo moo  512 Aug 20 14:45 ./
drwxr-xr-x 1 moo moo  512 Aug 20 14:43 ../
-rw-rw-rw- 1 moo moo  153 Aug 20 14:45 Pipfile
-rw-r--r-- 1 moo moo 2106 Aug 20 14:45 Pipfile.lock
14:48:40 DESKTOP-41D0GKF ~/test$ pipenv --venv
/home/moo/.local/share/virtualenvs/test-O3XhBtkW
14:48:45 DESKTOP-41D0GKF ~/test$ pipenv --rm
Removing virtualenv (/home/moo/.local/share/virtualenvs/test-O3XhBtkW)...
14:48:49 DESKTOP-41D0GKF ~/test$ mkdir .venv
14:48:51 DESKTOP-41D0GKF ~/test$ pipenv install requests
Pipfile: /home/moo/test/Pipfile
Using /usr/bin/python3.5m (3.5.2) to create virtualenv...
⠋Running virtualenv with interpreter /usr/bin/python3.5m
Using base prefix '/usr'
New python executable in /home/moo/test/.venv/bin/python3.5m
Also creating executable in /home/moo/test/.venv/bin/python

Virtualenv location: /home/moo/test/.venv                                                                                                                                                                                                Installing requests...
Collecting requests
  Using cached https://files.pythonhosted.org/packages/65/47/7e02164a2a3db50ed6d8a6ab1d6d60b69c4c3fdf57a284257925dfc12bda/requests-2.19.1-py2.py3-none-any.whl
Collecting idna<2.8,>=2.5 (from requests)
  Using cached https://files.pythonhosted.org/packages/4b/2a/0276479a4b3caeb8a8c1af2f8e4355746a97fab05a372e4a2c6a6b876165/idna-2.7-py2.py3-none-any.whl
Collecting certifi>=2017.4.17 (from requests)
  Using cached https://files.pythonhosted.org/packages/16/1f/50d729c104b21c1042aa51560da6141d1cab476ba7015d92b2111c8db841/certifi-2018.8.13-py2.py3-none-any.whl
Collecting chardet<3.1.0,>=3.0.2 (from requests)
  Using cached https://files.pythonhosted.org/packages/bc/a9/01ffebfb562e4274b6487b4bb1ddec7ca55ec7510b22e4c51f14098443b8/chardet-3.0.4-py2.py3-none-any.whl
Collecting urllib3<1.24,>=1.21.1 (from requests)
  Using cached https://files.pythonhosted.org/packages/bd/c9/6fdd990019071a4a32a5e7cb78a1d92c53851ef4f56f62a3486e6a7d8ffb/urllib3-1.23-py2.py3-none-any.whl
Installing collected packages: idna, certifi, chardet, urllib3, requests
Successfully installed certifi-2018.8.13 chardet-3.0.4 idna-2.7 requests-2.19.1 urllib3-1.23

Adding requests to Pipfile's [packages]...
Installing dependencies from Pipfile.lock (0b4483)...
  🐍   ▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉ 5/5 — 00:00:00
14:49:10 DESKTOP-41D0GKF ~/test$ pipenv install requests^C
14:50:24 DESKTOP-41D0GKF ~/test$ ll
total 4
drwxrwxrwx 1 moo moo  512 Aug 20 14:48 ./
drwxr-xr-x 1 moo moo  512 Aug 20 14:43 ../
-rw-rw-rw- 1 moo moo  153 Aug 20 14:45 Pipfile
-rw-r--r-- 1 moo moo 2106 Aug 20 14:45 Pipfile.lock
drwxrwxrwx 1 moo moo  512 Aug 20 14:49 .venv/
14:50:25 DESKTOP-41D0GKF ~/test$ pipenv --venv
/home/moo/test/.venv
14:50:29 DESKTOP-41D0GKF ~/test$
```

Pipenv version: `'2018.7.1'`

Pipenv location: `'/usr/local/lib/python3.5/dist-packages/pipenv'`

Python location: `'/usr/bin/python3'`

  - `2.7`: `/usr/bin/python2.7`
  - `2.7`: `/usr/bin/python2.7`
  - `3.5`: `/usr/bin/python3.5m`
  - `3.5`: `/usr/bin/python3.5`
  - `3.6`: `/usr/bin/python3.6m`
  - `3.6`: `/usr/bin/python3.6`

  - `2.7.12`: `/usr/bin/python`
  - `2.7.12`: `/usr/bin/python2`
  - `3.6.6`: `/usr/local/bin/python3`
  - `3.5.2`: `/usr/bin/python3`

PEP 508 Information:

```
 'implementation_version': '3.5.2',
 'os_name': 'posix',
 'platform_machine': 'x86_64',
 'platform_release': '4.4.0-17134-Microsoft',
 'platform_system': 'Linux',
 'platform_version': '#137-Microsoft Thu Jun 14 18:46:00 PST 2018',
 'python_full_version': '3.5.2',
 'python_version': '3.5',
 'sys_platform': 'linux'}
```

System environment variables:

  - `XDG_DATA_DIRS`
  - `LESSCLOSE`
  - `USER`
  - `LANG`
  - `TERM`
  - `_`
  - `LOGNAME`
  - `SHLVL`
  - `OLDPWD`
  - `LS_COLORS`
  - `LESSOPEN`
  - `PIP_PYTHON_PATH`
  - `SHELL`
  - `NAME`
  - `PS1`
  - `HOSTTYPE`
  - `PWD`
  - `PYTHONDONTWRITEBYTECODE`
  - `PATH`
  - `HOME`

  - `PATH`: `/home/moo/bin:/home/moo/.local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/mnt/c/Program Files/Python36/Scripts:/mnt/c/Program Files/Python36:/mnt/c/Program Files/Docker/Docker/resources/bin:/mnt/c/Windows/System32:/mnt/c/Windows:/mnt/c/Windows/System32/wbem:/mnt/c/Windows/System32/WindowsPowerShell/v1.0:/mnt/c/Program Files/ctags58:/mnt/c/Program Files (x86)/Windows Kits/8.1/Windows Performance Toolkit:/mnt/c/Program Files/Microsoft VS Code/bin:/mnt/c/Program Files/Git/cmd:/mnt/c/Windows/System32/OpenSSH:/mnt/c/Users/moo/AppData/Local/Microsoft/WindowsApps:/snap/bin`
  - `SHELL`: `/bin/bash`
  - `LANG`: `en_US.UTF-8`
  - `PWD`: `/home/moo/test`

---------------------------

Contents of `Pipfile` ('/home/moo/test/Pipfile'):

```toml
[[source]]
verify_ssl = true
name = "pypi"

[packages]
requests = "*"

[dev-packages]

[requires]
python_version = "3.5"

```

Contents of `Pipfile.lock` ('/home/moo/test/Pipfile.lock'):

```json
{
    "_meta": {
        "hash": {
            "sha256": "50b8935d42537b1d392d8fe9a921302d58566b8b364c66af879714c0220b4483"
        },
        "pipfile-spec": 6,
        "requires": {
            "python_version": "3.5"
        },
        "sources": [
            {
                "name": "pypi",
                "verify_ssl": true
            }
        ]
    },
    "default": {
        "certifi": {
            "hashes": [
                "sha256:4c1d68a1408dd090d2f3a869aa94c3947cc1d967821d1ed303208c9f41f0f2f4",
                "sha256:b6e8b28b2b7e771a41ecdd12d4d43262ecab52adebbafa42c77d6b57fb6ad3a4"
            ],
            "version": "==2018.8.13"
        },
        "chardet": {
            "hashes": [
                "sha256:84ab92ed1c4d4f16916e05906b6b75a6c0fb5db821cc65e70cbd64a3e2a5eaae",
                "sha256:fc323ffcaeaed0e0a02bf4d117757b98aed530d9ed4531e3e15460124c106691"
            ],
            "version": "==3.0.4"
        },
        "idna": {
            "hashes": [
                "sha256:156a6814fb5ac1fc6850fb002e0852d56c0c8d2531923a51032d1b70760e186e",
                "sha256:684a38a6f903c1d71d6d5fac066b58d7768af4de2b832e426ec79c30daa94a16"
            ],
            "version": "==2.7"
        },
        "requests": {
            "hashes": [
                "sha256:63b52e3c866428a224f97cab011de738c36aec0185aa91cfacd418b5d58911d1",
                "sha256:ec22d826a36ed72a7358ff3fe56cbd4ba69dd7a6718ffd450ff0e9df7a47ce6a"
            ],
            "index": "pypi",
            "version": "==2.19.1"
        },
        "urllib3": {
            "hashes": [
                "sha256:a68ac5e15e76e7e5dd2b8f94007233e01effe3e50e8daddf69acfd81cb686baf",
                "sha256:b5725a0bd4ba422ab0e66e89e030c806576753ea3ee08554382c14e685d117b5"
            ],
            "markers": "python_version >= '2.6' and python_version != '3.0.*' and python_version < '4' and python_version != '3.3.*' and python_version != '3.2.*' and python_version != '3.1.*'",
            "version": "==1.23"
        }
    },
    "develop": {}
}

```
</details>

matteius:
Allow PIPENV_VENV_IN_PROJECT to be read in as None, and ensure if it is set to False that it does not use .venv directory.

### The issue

Fixes  #2763 

### The fix

The variable was being cast to a boolean so it would be False even if it was not set, this allows a user to explicitly set the variable to a non-truthy value and ensure that the .venv directory gets ignored.

### The checklist

* [X] Associated issue

### DIFF
--- a/news/4992.process.rst
+++ b/news/4992.process.rst
@@ -1,1 +1,0 @@
-Internal to pipenv, the utils.py was split into a utils module with unused code removed.
--- a/pipenv/environments.py
+++ b/pipenv/environments.py
@@ -299,12 +299,16 @@ class Setting:
 
         self.PIPENV_VENV_IN_PROJECT = os.environ.get("PIPENV_VENV_IN_PROJECT")
         if self.PIPENV_VENV_IN_PROJECT is not None:
-            self.PIPENV_VENV_IN_PROJECT = (
-                self.PIPENV_VENV_IN_PROJECT.lower() in _true_values
-            )
-        """If set, creates ``.venv`` in your project directory.
-
-        Default is to create new virtual environments in a global location.
+            if self.PIPENV_VENV_IN_PROJECT.lower() in _true_values:
+                self.PIPENV_VENV_IN_PROJECT = True
+            elif self.PIPENV_VENV_IN_PROJECT.lower() in _false_values:
+                self.PIPENV_VENV_IN_PROJECT = False
+            else:
+                self.PIPENV_VENV_IN_PROJECT = None
+        """ When set True, will create or use the ``.venv`` in your project directory.
+        When Set False, will ignore the .venv in your project directory even if it exists.
+        Default is None will use the .venv of project directory should it exist, otherwise
+          will create new virtual environments in a global location.
         """
 
         self.PIPENV_VERBOSE = bool(os.environ.get("PIPENV_VERBOSE"))
--- a/pipenv/environments.py
+++ b/pipenv/environments.py
@@ -299,10 +299,8 @@ class Setting:
 
         self.PIPENV_VENV_IN_PROJECT = os.environ.get("PIPENV_VENV_IN_PROJECT")
         if self.PIPENV_VENV_IN_PROJECT is not None:
-            self.PIPENV_VENV_IN_PROJECT = self.PIPENV_VENV_IN_PROJECT.lower() in (
-                "true",
-                "1",
-                "t",
+            self.PIPENV_VENV_IN_PROJECT = (
+                self.PIPENV_VENV_IN_PROJECT.lower() in _true_values
             )
         """If set, creates ``.venv`` in your project directory.
 
--- a/pipenv/environments.py
+++ b/pipenv/environments.py
@@ -297,7 +297,13 @@ class Setting:
         Default is 120 seconds, an arbitrary number that seems to work.
         """
 
-        self.PIPENV_VENV_IN_PROJECT = bool(os.environ.get("PIPENV_VENV_IN_PROJECT"))
+        self.PIPENV_VENV_IN_PROJECT = os.environ.get("PIPENV_VENV_IN_PROJECT")
+        if self.PIPENV_VENV_IN_PROJECT is not None:
+            self.PIPENV_VENV_IN_PROJECT = self.PIPENV_VENV_IN_PROJECT.lower() in (
+                "true",
+                "1",
+                "t",
+            )
         """If set, creates ``.venv`` in your project directory.
 
         Default is to create new virtual environments in a global location.
--- a/tests/integration/test_install_basic.py
+++ b/tests/integration/test_install_basic.py
@@ -405,8 +405,6 @@ def test_install_venv_project_directory(PipenvInstance):
             prefix="pipenv-", suffix="temp_workon_home"
         ) as workon_home:
             os.environ["WORKON_HOME"] = workon_home
-            if "PIPENV_VENV_IN_PROJECT" in os.environ:
-                del os.environ["PIPENV_VENV_IN_PROJECT"]
 
             c = p.pipenv("install six")
             assert c.returncode == 0
--- a/tests/integration/test_pipenv.py
+++ b/tests/integration/test_pipenv.py
@@ -92,14 +92,9 @@ def test_proper_names_unamanged_virtualenv(PipenvInstance):
 def test_directory_with_leading_dash(raw_venv, PipenvInstance):
     with temp_environ():
         with PipenvInstance(chdir=True, venv_in_project=False, name="-project-with-dash") as p:
-            if "PIPENV_VENV_IN_PROJECT" in os.environ:
-                del os.environ['PIPENV_VENV_IN_PROJECT']
             c = p.pipenv('run pip freeze')
             assert c.returncode == 0
             c = p.pipenv('--venv')
             assert c.returncode == 0
             venv_path = c.stdout.strip()
             assert os.path.isdir(venv_path)
-            # Manually clean up environment, since PipenvInstance assumes that
-            # the virutalenv is in the project directory.
-            p.pipenv('--rm')
--- a/news/2763.bugfix.rst
+++ b/news/2763.bugfix.rst
@@ -1,0 +1,2 @@
+Pipenv will now ignore ``.venv`` in the project when ``PIPENV_VENV_IN_PROJECT`` variable is False.
+Unset variable maintains the existing behavior of preferring to use the project's ``.venv`` should it exist.
--- a/pipenv/project.py
+++ b/pipenv/project.py
@@ -242,6 +242,8 @@ class Project:
 
     def is_venv_in_project(self):
         # type: () -> bool
+        if self.s.PIPENV_VENV_IN_PROJECT is False:
+            return False
         return self.s.PIPENV_VENV_IN_PROJECT or (
             self.project_directory
             and os.path.isdir(os.path.join(self.project_directory, ".venv"))
@@ -288,7 +290,7 @@ class Project:
             return str(get_workon_home().joinpath(self.virtualenv_name))
 
         # If content looks like a path, use it as a relative path.
-        # Otherwise use directory named after content in WORKON_HOME.
+        # Otherwise, use directory named after content in WORKON_HOME.
         if looks_like_dir(name):
             path = Path(self.project_directory, name)
             return path.absolute().as_posix()
--- a/tests/integration/test_dot_venv.py
+++ b/tests/integration/test_dot_venv.py
@@ -7,13 +7,15 @@ from tempfile import TemporaryDirectory
 
 import pytest
 
+from pipenv.environments import _true_values, _false_values
 from pipenv.utils.shell import normalize_drive, temp_environ
 
 
 @pytest.mark.dotvenv
-def test_venv_in_project(PipenvInstance):
+@pytest.mark.parametrize("true_value", _true_values)
+def test_venv_in_project(true_value, PipenvInstance):
     with temp_environ():
-        os.environ['PIPENV_VENV_IN_PROJECT'] = '1'
+        os.environ['PIPENV_VENV_IN_PROJECT'] = true_value
         with PipenvInstance() as p:
             c = p.pipenv('install requests')
             assert c.returncode == 0
@@ -21,10 +23,38 @@ def test_venv_in_project(PipenvInstance):
 
 
 @pytest.mark.dotvenv
-def test_venv_at_project_root(PipenvInstance):
+@pytest.mark.parametrize("false_value", _false_values)
+def test_venv_in_project_disabled_ignores_venv(false_value, PipenvInstance):
+    venv_name = "my_project"
+    with temp_environ():
+        os.environ['PIPENV_VENV_IN_PROJECT'] = false_value
+        with PipenvInstance() as p:
+            file_path = os.path.join(p.path, '.venv')
+            with open(file_path, 'w') as f:
+                f.write(venv_name)
+
+            with temp_environ(), TemporaryDirectory(
+                prefix='pipenv-', suffix='temp_workon_home'
+            ) as workon_home:
+                os.environ['WORKON_HOME'] = workon_home
+                c = p.pipenv('install requests')
+                assert c.returncode == 0
+                c = p.pipenv('--venv')
+                assert c.returncode == 0
+                venv_loc = Path(c.stdout.strip()).absolute()
+                assert venv_loc.exists()
+                assert venv_loc.joinpath('.project').exists()
+                venv_path = normalize_drive(venv_loc.as_posix())
+                venv_expected_path = Path(workon_home).joinpath(venv_name).absolute().as_posix()
+                assert venv_path == normalize_drive(venv_expected_path)
+
+
+@pytest.mark.dotvenv
+@pytest.mark.parametrize("true_value", _true_values)
+def test_venv_at_project_root(true_value, PipenvInstance):
     with temp_environ():
         with PipenvInstance(chdir=True) as p:
-            os.environ['PIPENV_VENV_IN_PROJECT'] = '1'
+            os.environ['PIPENV_VENV_IN_PROJECT'] = true_value
             c = p.pipenv('install')
             assert c.returncode == 0
             assert normalize_drive(p.path) in p.pipenv('--venv').stdout
@@ -80,7 +110,7 @@ def test_venv_file(venv_name, PipenvInstance):
 
 @pytest.mark.dotvenv
 def test_empty_venv_file(PipenvInstance):
-    """Tests virtualenv creation when a empty .venv file exists at the project root
+    """Tests virtualenv creation when an empty .venv file exists at the project root
     """
     with PipenvInstance(chdir=True) as p:
         file_path = os.path.join(p.path, '.venv')
@@ -109,9 +139,8 @@ def test_empty_venv_file(PipenvInstance):
 
 
 @pytest.mark.dotvenv
-def test_venv_file_with_path(PipenvInstance):
-    """Tests virtualenv creation when a .venv file exists at the project root
-    and contains an absolute path.
+def test_venv_in_project_default_when_venv_exists(PipenvInstance):
+    """Tests virtualenv creation when a .venv file exists at the project root.
     """
     with temp_environ(), PipenvInstance(chdir=True) as p:
         with TemporaryDirectory(


smac89:
@matteius I'm not able to test atm, but from a casual glance, it looks like that should fix the issue.

Moreover, I think switching from booleans to constants may be a better approach to avoid confusion (but this may introduce breakages):

I propose the following values:
- enable
- disable
- auto

The `PIPENV_VENV_IN_PROJECT` environment variable can take any of these values, and when it is set to `auto` (which should be the default), then pipenv will behave the way it currently does.

Actually, this may not break anything, because `enable` and `disable` may just be more aliases for truthy and falsey.

Anyways, yea that looks like a good fix for the issue

news/4992.process.rst
```diff
@@ -1 +0,0 @@
-Internal to pipenv, the utils.py was split into a utils module with unused code removed.
```
matteius, line 1: I duplicated this news fragment in main.

tests/integration/test_pipenv.py
```diff
@@ -92,14 +92,9 @@ def test_proper_names_unamanged_virtualenv(PipenvInstance):
 def test_directory_with_leading_dash(raw_venv, PipenvInstance):
     with temp_environ():
         with PipenvInstance(chdir=True, venv_in_project=False, name="-project-with-dash") as p:
-            if "PIPENV_VENV_IN_PROJECT" in os.environ:
```
matteius, line 95: These tests aren't about this variable and shouldn't impact the test run since it use `with temp_environ()`.

tests/integration/test_install_basic.py
```diff
@@ -405,8 +405,6 @@ def test_install_venv_project_directory(PipenvInstance):
             prefix="pipenv-", suffix="temp_workon_home"
         ) as workon_home:
             os.environ["WORKON_HOME"] = workon_home
-            if "PIPENV_VENV_IN_PROJECT" in os.environ:
```
matteius, line 408: These tests aren't about this variable and shouldn't impact the test run since it use `with temp_environ()`.

tests/integration/test_pipenv.py
```diff
@@ -92,14 +92,9 @@ def test_proper_names_unamanged_virtualenv(PipenvInstance):
...
-                del os.environ['PIPENV_VENV_IN_PROJECT']
             c = p.pipenv('run pip freeze')
             assert c.returncode == 0
             c = p.pipenv('--venv')
             assert c.returncode == 0
             venv_path = c.stdout.strip()
             assert os.path.isdir(venv_path)
-            # Manually clean up environment, since PipenvInstance assumes that
-            # the virutalenv is in the project directory.
-            p.pipenv('--rm')
```
matteius, line 105: None of the other tests do this kind of cleanup.

matteius:
I made some updates and made it safer with respect to the allowed true and false values.  If we want to expand that list at some point to include enable(d)/disable(d) we can but I mostly just wanted to put out a patch for the issue at hand.

pipenv/environments.py
```diff
@@ -299,12 +299,16 @@ def initialize(self):
 
         self.PIPENV_VENV_IN_PROJECT = os.environ.get("PIPENV_VENV_IN_PROJECT")
         if self.PIPENV_VENV_IN_PROJECT is not None:
-            self.PIPENV_VENV_IN_PROJECT = (
-                self.PIPENV_VENV_IN_PROJECT.lower() in _true_values
-            )
-        """If set, creates ``.venv`` in your project directory.
```
oz123, line 301: Good good! Those `()` around the value were confusing, and I am glad you removed it.

pipenv/environments.py
```diff
@@ -299,12 +299,16 @@ def initialize(self):
...
-
-        Default is to create new virtual environments in a global location.
+            if self.PIPENV_VENV_IN_PROJECT.lower() in _true_values:
+                self.PIPENV_VENV_IN_PROJECT = True
+            elif self.PIPENV_VENV_IN_PROJECT.lower() in _false_values:
+                self.PIPENV_VENV_IN_PROJECT = False
```
oz123, line 305: Why does one need both `False` and `None` here? I see you documented it, but why do we need this feature?
If this is only for backward compatibility, I suggest considering supporting it but also to issue a warning that only `false`
should be used in the future.
If someone needs `.venv` they could specify it specifically. Otherwise, IMHO `.venv` is like magic. It's hidden at first when you ` ls` a directory, and this behavior is kind of opposite to Python's own: Explicit it better than implicit.
This is a great principle for any software project too...

oz123 (approved):

pipenv/environments.py
matteius: None indicates that you haven't set the variable, so in this case the behavior is use the .venv if it is present.  However if you want to definitely ignore a present .venv you would se it to be False.   This is helpful for those doing development locally and building docker containers, that sometimes the .venv directory exists locally and gets copied into their container but is different system installed versions than the OS the container is running on.  They want to be able to explicitly state in the .env not to ever use a .venv directory in the project, even if it exists because it got copied into the container.  For additional context:  https://github.com/pypa/pipenv/issues/2763#issuecomment-1083715772

oz123: > if you want to definitely ignore a present .venv you would se it to be False.

I thought this should be the default, always ignore. Even if the variable isn't set.

A solution to @smac89's problem could done with `.dockerignore`. They probably have something like this:
in project source:
```
$ ls -a
setup.py
Pipfile
Pipfile.lock
src/<many-python-files-here>
.venv
Dockerfile
```

Now, they (naively) do something in `Dockerfile`
```
ADD  .  /usr/src/app
WORKDIR  /usr/src/app
RUN pipenv install
```
And viola, this thing is broken, because it uses an existing `venv`. The solution is to create `.dockerignore` file, and then add `.venv` in it. This will prevent the ADD directive from copying `.venv` into the docker container.

The real problem is however, the decision to [default  to the name](https://github.com/pypa/pipenv/blob/9a3b3ce70621af6f9adaa9eeac9cf83fa149319c/pipenv/environments.py#L301) `.venv`. This should have been named either `pipenv` `venv` without the leading dot.

NeroVanbiervliet:
From [a quick search on github](https://github.com/search?p=1&q=PIPENV_VENV_IN_PROJECT&type=Code) it seems that most people are using `1` or `true` as a way to set the variable. So they should be fine :)

markmiscavage:
> This change is not 100% backwards compatible and broke one of our docker builds:
> 
> ```dockerfile
> # Dockerfile
> RUN PIPENV_VENV_IN_PROJECT="enabled" pipenv sync
> ```
> 
> Previously, "enabled" was parsed as `True`, but with this change, you need to supply one of `_true_values` for it to work.

We were also using "enabled" and this broke all our project builds (40+ projects in Bitbucket) and took me a long while to debug as the error I was initially receiving was `Error: the command gunicorn could not be found within PATH or Pipfile's [scripts].`  on container start, rather than any issues during container build.

oz123:
While this is certainly annoying that so many builds are broken, I think this 'enabled' work was wrong to use. This probably would also work if you wrote 'disabled', since the previous code only checked if the variable was set, not if it was truthy.
```
>>> import os
>>> bool(os.environ.get("PIPENV_VENV_IN_PROJECT", "disabled"))
True
>>> 
```

This was an accident waiting to happen. Sorry, for breaking your builds, maybe we could have done this in a more graceful way.

pipenv/environments.py
smac89: As I mentioned in my comment, we use `docker-compose`, and as far as I can tell, docker-compose does not give one the option of filtering out certain files from a mounted volume. Therefore, once we mount the volume inside the build container, it includes everything in that volume.

smac89:
> > This change is not 100% backwards compatible and broke one of our docker builds:
> > ```dockerfile
> > # Dockerfile
> > RUN PIPENV_VENV_IN_PROJECT="enabled" pipenv sync
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Previously, "enabled" was parsed as `True`, but with this change, you need to supply one of `_true_values` for it to work.
> 
> We were also using "enabled" and this broke all our project builds (40+ projects in Bitbucket) and took me a long while to debug as the error I was initially receiving was `Error: the command gunicorn could not be found within PATH or Pipfile's [scripts].` on container start, rather than any issues during container build.

I suggest you pin your version of `pipenv` to avoid these potential breaks in the future.

Issue #4706, opened 2021-05-27
Using: keyrings.google-artifactregistry-auth pipenv update -d fails hard.


klausmyrseth:
Be sure to check the existing issues (both open and closed!), and make sure you are running the latest version of Pipenv.

Check the [diagnose documentation](https://pipenv.pypa.io/en/latest/diagnose/) for common issues before posting! We may close your issue if it is very similar to one of them. Please be considerate, or be on your way.

Make sure to mention your debugging experience if the documented solution failed.

### Issue description

Google Artifact Repositories use keyrings with their own addon to make git and twine work perfectly, but when using pipenv it fails hard to a point where we cant use pipenv.

Note I also tried with 20.11.15 with same result

### Expected result

pipenv update -d should update dependencies and make an update Pipfile.lock

### Actual result

(applied_mapper)  klaus@new-server ~/src/applied_mapper master> pipenv update -d -v
Running $ pipenv lock then $ pipenv sync.
Locking [dev-packages] dependencies…
Building requirements...
Resolving dependencies...
⠧ Locking...
                          ROUND 1                           
✘ Locking Failed! 
Current constraints:
  aadomain (from -r /tmp/pipenvg2_zajwhrequirements/pipenv-2_qs5aav-constraints.txt (line 10))
  applied-core (from -r /tmp/pipenvg2_zajwhrequirements/pipenv-2_qs5aav-constraints.txt (line 14))
  applied-domain (from -r /tmp/pipenvg2_zajwhrequirements/pipenv-2_qs5aav-constraints.txt (line 9))
  geojson (from -r /tmp/pipenvg2_zajwhrequirements/pipenv-2_qs5aav-constraints.txt (line 13))
  keyring (from -r /tmp/pipenvg2_zajwhrequirements/pipenv-2_qs5aav-constraints.txt (line 7))
  keyrings.google-artifactregistry-auth (from -r /tmp/pipenvg2_zajwhrequirements/pipenv-2_qs5aav-constraints.txt (line 15))
  pbr (from -r /tmp/pipenvg2_zajwhrequirements/pipenv-2_qs5aav-constraints.txt (line 5))
  pylint (from -r /tmp/pipenvg2_zajwhrequirements/pipenv-2_qs5aav-constraints.txt (line 4))
  pytest (from -r /tmp/pipenvg2_zajwhrequirements/pipenv-2_qs5aav-constraints.txt (line 8))
[...]

PR #5036, opened 2022-04-08
[fix] allow the user to overwrite the `no_input` flag to pip

allcloud-jonathan:
this is required if you want to e.g. use a keyring to authenticate
against Google Artifact Registry. fixes https://github.com/pypa/pipenv/issues/4706

### The issue

### The fix

### The checklist

* [x] Associated issue: https://github.com/pypa/pipenv/issues/4706

### DIFF
--- a/pipenv/utils/resolver.py
+++ b/pipenv/utils/resolver.py
@@ -518,7 +518,7 @@ class Resolver:
             pip_options, _ = self.pip_command.parser.parse_args(self.pip_args)
             pip_options.cache_dir = self.project.s.PIPENV_CACHE_DIR
             pip_options.no_python_version_warning = True
-            pip_options.no_input = True
+            pip_options.no_input = self.project.settings.get("disable_pip_input", True)
             pip_options.progress_bar = "off"
             pip_options.ignore_requires_python = True
             pip_options.pre = self.pre or self.project.settings.get(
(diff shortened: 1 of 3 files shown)


pipenv/utils/resolver.py
```diff
@@ -518,7 +518,7 @@ def pip_options(self):
             pip_options, _ = self.pip_command.parser.parse_args(self.pip_args)
             pip_options.cache_dir = self.project.s.PIPENV_CACHE_DIR
             pip_options.no_python_version_warning = True
-            pip_options.no_input = True
+            pip_options.no_input = self.project.settings.get("disable_pip_input", True)
```
matteius, line 521: I think this is going to at a minimum require some documentation around how to specify this in the Pipfile and what it does.

matteius: Sure -- so you have to go into the `docs` directory and there is a separate `requirements.txt` in there to install using `pip -r requirements.txt`.  I think after that you should just have to run `make html` to build the html docs.  Going from memory here but I did it not so long ago ...

allcloud-jonathan: I can't get the pipfile to show as syntax. But the doc is added. Maybe you can take it from here, @matteius ?

matteius: since its in your fork, best I can do is address any formatting in a followup PR.  I'll at least get this merged now.

PR #5038, opened 2022-04-09
Added pytest-cov and addressed outdated references to pip-tools

matteius:
### The issue

#4996 

### The fix

We don't actually vendor in pip-tools and we have a laundry list of TODOs in this method `_build_package_list` and I think this correctly addresses them.  We end up dropping `is_installable_file` but based on my grepping, that method exists in requirementslib as well so I think the code still accounts for `is_installable_file` and we have tests that install from files, so I think its a non-issue.

### The checklist

* [X] Associated issue

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -70,7 +70,6 @@ filterwarnings =
 
 [coverage:run]
 parallel = true
-omit = pipenv/vendor*, pipenv/patched*
 
 [bdist_wheel]
 universal = 1
--- a/setup.cfg
+++ b/setup.cfg
@@ -44,7 +44,7 @@ python_version=3.6
 mypy_path=typeshed/pyi:typeshed/imports
 
 [tool:pytest]
-addopts = -ra --cov=pipenv --cov-report html
+addopts = -ra
 testpaths = tests
 ; Add vendor and patched in addition to the default list of ignored dirs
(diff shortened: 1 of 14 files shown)


matteius:
Since this one addresses problems with the lockfile, but has to drop py3.6 to do it, going to defer it until right before the pip 22.0.4 vendoring so that I can get these Pipfile/lock file improvements ahead of that PR.

Note:  it works now to just run `pipenv lock` from Windows, since there are windows specific requirements in the lock file.

setup.cfg
```diff
@@ -68,5 +68,9 @@ filterwarnings =
     ignore::DeprecationWarning
     ignore::PendingDeprecationWarning
 
+[coverage:run]
+parallel = true
+omit = pipenv/vendor*, pipenv/patched*
+
```
oz123, line 74: Omitting `vendor` and `patched` might hide unused patched and vendor packages.
Do you have a solution on how to avoid this?

matteius: I forget why I had originally excluded those, but thinking about it now I don't see a strong reason to.  It might have been just because I excluded them from the linting that I copied that same pattern to exclude them from coverage, but I don't think there is a good reason to like there was for the linting.

oz123 (approved):

PR #5039, opened 2022-04-10
Allow pipenv option to consider all listed package sources when syncing the Pipfile.lock

matteius:
### The issue

Maybe we can allow an option `install_search_all_sources` where its possible to still consider all sources eligible for install, since the hashes of the lock file were generated from index restricted packages and are pinned to specific versions and hashes in the lockfile.

### The fix

Fix https://github.com/pypa/pipenv/issues/5041

### The checklist

* [X] Associated issue


pipenv/core.py
```diff
@@ -1484,23 +1482,26 @@ def pip_install(
             ignore_hashes = False
     line = None
     # Try installing for each source in project.sources.
+    search_all_sources = project.settings.get("install_search_all_sources", False)
```
oz123, line 1485: Should we add public facing documentation for this option?

matteius: Yes -- I am going to add that once we decide if this should be what its name is.

oz123: I think the name `install_search_all_sources` is OK. Once you write the docs for this, you can figure if this works or not.

matteius: Squeezing this into tonight's release -- let me know if there are any follow up tweaks you want to see to the docs or this PR.

PR #5059, opened 2022-04-19
Removed usage of fs_str from vistir

oz123:
As discussed, I am suggesting to remove `fs_str`.

### DIFF
--- a/pipenv/__init__.py
+++ b/pipenv/__init__.py
@@ -18,15 +18,14 @@ sys.path.insert(0, PIPENV_VENDOR)
 sys.path.insert(0, PIPENV_PATCHED)
 
 from pipenv.vendor.urllib3.exceptions import DependencyWarning
-from pipenv.vendor.vistir.compat import fs_str
 
 warnings.filterwarnings("ignore", category=DependencyWarning)
 warnings.filterwarnings("ignore", category=ResourceWarning)
 warnings.filterwarnings("ignore", category=UserWarning)
 
 # Load patched pip instead of system pip
-os.environ["PIP_SHIMS_BASE_MODULE"] = fs_str("pipenv.patched.notpip")
-os.environ["PIP_DISABLE_PIP_VERSION_CHECK"] = fs_str("1")
+os.environ["PIP_SHIMS_BASE_MODULE"] = "pipenv.patched.notpip"
+os.environ["PIP_DISABLE_PIP_VERSION_CHECK"] = "1"
 
 # Hack to make things work better.
 try:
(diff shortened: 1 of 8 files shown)


.github/workflows/ci.yaml
```diff
@@ -29,8 +29,22 @@ on:
       - ".gitmodules"
       - ".gitattributes"
       - ".editorconfig"
-
 jobs:
+  lint:
```
matteius, line 33: I think it is a decent enough idea to make this its own job, but can we make it run on python "3.10" and then remove it from the individual test runs:  https://github.com/pypa/pipenv/pull/5059/files#diff-944291df2c9c06359d37cc8833d182d705c9e8c3108e7cfe132d61a06e9133ddR81

.github/workflows/ci.yaml
```diff
@@ -29,8 +29,25 @@ on:
...
+      PYTHONIOENCODING: "utf-8"
+      GIT_ASK_YESNO: "false"
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: 3.9
+      - run: |
+          python -m pip install --upgrade pip setuptools wheel --upgrade-strategy=eager
+          python -m pip install -e . --upgrade
```
matteius, line 48: Probably don't have to install pipenv to lint it.

oz123: Yes, maybe it's enough to just install pre-commit...

oz123:
OK, The latest commits don't have much to do with the title, but here is what they do:
We migrate from a CI pipeline the looks like this:

image

To this:

image

There is an explicit step for linting, only if it passes the tests will run. If these pass, the vendoring will trigger, and once this is done we build a package with update vendoring code.

.github/workflows/ci.yaml
oz123: Was working on that while you wrote the comment ...

.github/workflows/ci.yaml
```diff
@@ -87,28 +90,28 @@ jobs:
           GIT_SSH_COMMAND: ssh -o StrictHostKeyChecking=accept-new -o CheckHostIP=no
         run: |
           pipenv run pytest -ra -n auto -v --fulltrace tests
-
-  build:
-    name: Build Package
+  vendor:
+    name: Vendoring
+    needs: tests
```
matteius, line 95: I actually really depend on the vendoring build running in parallel and being one of the first things to complete, when working on patches.  It seems the current actions run doesn't show the vendor step anymore.

oz123: The current actions moved the vendor step to after the tests. It should be before the tests, if the new code relies on that. I can fix this.

PR #5062, opened 2022-04-20
Removed usage of fs_str from vistir

oz123:
This function was all about compatability of py2-py3.3 versions.
Later versions don't need it.

This PR replaces #5059 as I did too much havoc there with all the CI suggestions.

### DIFF
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -754,7 +754,7 @@ def batch_install(
 
         with vistir.contextmanagers.temp_environ():
             if not allow_global:
-                os.environ["PIP_USER"] = vistir.compat.fs_str("0")
+                os.environ["PIP_USER"] = "0"
                 if "PYTHONHOME" in os.environ:
                     del os.environ["PYTHONHOME"]
             if "GIT_CONFIG" in os.environ and dep.is_vcs:
@@ -967,7 +967,7 @@ def do_create_virtualenv(project, python=None, site_packages=None, pypi_mirror=N
         cmd.append("--system-site-packages")
 
     if pypi_mirror:
-        pip_config = {"PIP_INDEX_URL": vistir.misc.fs_str(pypi_mirror)}
+        pip_config = {"PIP_INDEX_URL": pypi_mirror}
     else:
         pip_config = {}
 
@@ -998,7 +998,7 @@ def do_create_virtualenv(project, python=None, site_packages=None, pypi_mirror=N
     # This mimics Pew's "setproject".
     project_file_name = os.path.join(project.virtualenv_location, ".project")
     with open(project_file_name, "w") as f:
-        f.write(vistir.misc.fs_str(project.project_directory))
+        f.write(project.project_directory)
     from .environment import Environment
 
     sources = project.pipfile_sources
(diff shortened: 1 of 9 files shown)


pipenv/core.py
```diff
@@ -1561,20 +1561,18 @@ def pip_install(
     DEFAULT_EXISTS_ACTION = "w"
     if selective_upgrade:
         DEFAULT_EXISTS_ACTION = "i"
-    exists_action = vistir.misc.fs_str(
-        project.s.PIP_EXISTS_ACTION or DEFAULT_EXISTS_ACTION
-    )
+    exists_action = project.s.PIP_EXISTS_ACTION or DEFAULT_EXISTS_ACTION
```
matteius, line 1564: I think you might need to cast this one to a str otherwise it would be a boolean whereas it was a string before?

pipenv/utils/resolver.py
```diff
@@ -980,19 +979,19 @@ def venv_resolve_deps(
     target_file.close()
     cmd.extend(["--write", make_posix(target_file.name)])
     with temp_environ():
-        os.environ.update({fs_str(k): fs_str(val) for k, val in os.environ.items()})
+        os.environ.update({k: val for k, val in os.environ.items()})
```
matteius, line 982: This statement seems to have no effect after the refactor -- we should consider if val needs to be cast to a `str` or if this line can be removed entirely.

matteius (changes requested):
Looks good -- made a couple of notes to look into first.

pipenv/utils/resolver.py
oz123: Good catch!
```
In [4]: os.environ.update({"foo": Foo()})
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-4-b63c173a9c62> in <module>
----> 1 os.environ.update({"foo": Foo()})

/usr/lib/python3.9/_collections_abc.py in update(self, other, **kwds)
    939         if isinstance(other, Mapping):
    940             for key in other:
--> 941                 self[key] = other[key]
    942         elif hasattr(other, "keys"):
    943             for key in other.keys():

/usr/lib/python3.9/os.py in __setitem__(self, key, value)
    682     def __setitem__(self, key, value):
    683         key = self.encodekey(key)
--> 684         value = self.encodevalue(value)
    685         putenv(key, value)
    686         self._data[key] = value

/usr/lib/python3.9/os.py in encode(value)
    754         def encode(value):
    755             if not isinstance(value, str):
--> 756                 raise TypeError("str expected, not %s" % type(value).__name__)
    757             return value.encode(encoding, 'surrogateescape')
    758         def decode(value):

TypeError: str expected, not Foo

```
This works:
```
os.environ.update({"foo": str(Foo())})
```

pipenv/core.py
oz123: I think we should explicitly look for a thruty string, since any string casted to bool will be True.

matteius: It defaults to a string "w" https://github.com/pypa/pipenv/blob/main/pipenv/environments.py#L341

So if we leave it as a boolean where it is a string before then we should also review all usages of `PIP_EXISTS_ACTION`

matteius: Neverminded, to my surprise this will apparently still be a string.
```
>>> "i" or "w"
'i'
```

matteius (approved):

pipenv/core.py
oz123: For now, leaving untouched. Will fix in #5063

matteius (approved):

matteius (approved):

PR #5067, opened 2022-04-21
Remove some more vistir usage

oz123:
* Removed usage of vistir.compat.JSONDecodeError
  `json.JSONDecodeError` is Python 3.6+
* Removed vistir.path.rmtree in favour of shutil.rmtree
  Should work, but we might have to revert this.

### DIFF
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -2652,7 +2652,7 @@ def do_check(
     quiet=False,
     pypi_mirror=None,
 ):
-    from pipenv.vendor.vistir.compat import JSONDecodeError
+    import json
 
     if not system:
         # Ensure that virtualenv is available.
@@ -2689,7 +2689,7 @@ def do_check(
     if c.returncode is not None:
         try:
             results = simplejson.loads(c.stdout.strip())
-        except JSONDecodeError:
+        except json.JSONDecodeError:
             click.echo(
                 "{}\n{}\n{}".format(
                     crayons.white(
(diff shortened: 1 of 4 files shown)


oz123:
@matteius I've seen this error for MacOS Python3.10 happen already a few times (in the main branch too). No idea how to fix it, but it seems unrelated to the changes I proposed in the PR.

pipenv/core.py
```diff
@@ -2791,8 +2792,9 @@ def do_check(
 
 
 def do_graph(project, bare=False, json=False, json_tree=False, reverse=False):
+    import json as jsonlib
```
matteius, line 2795: Can we just `import json` at the top level now?  I don't really see why to namespace it to `jsonlib` either.

pipenv/utils/shell.py
```diff
@@ -252,17 +252,10 @@ def is_virtual_environment(path):
...
-        - already exists, silently complete
-        - regular file in the way, raise an exception
-        - parent directory(ies) does not exist, make them as well
-        From: http://code.activestate.com/recipes/82465-a-friendly-mkdir/
-    =======
-            - already exists, silently complete
-            - regular file in the way, raise an exception
-            - parent directory(ies) does not exist, make them as well
-            From: http://code.activestate.com/recipes/82465-a-friendly-mkdir/
-    >>>>>>> main
```
matteius, line 265: Good catch

matteius (approved):

matteius:
> @matteius I've seen this error for MacOS Python3.10 happen already a few times (in the main branch too). No idea how to fix it, but it seems unrelated to the changes I proposed in the PR.

I re-ran just the one failed job and it succeeded.  Looks like a 1-off GitHub data transfer failure

pipenv/utils/shell.py
oz123: pure luck :-)

PR #5069, opened 2022-04-22
Update docs and add deprecation warning for lock -r

ImreC:
@matteius I gave `pipenv lock -r` some more thought and I think we should over time deprecate it in favor of `pipenv requirements`. This PR contains a deprecation warning and an update to the docs. Please let me know your thoughts

### DIFF
--- a/pipenv/cli/command.py
+++ b/pipenv/cli/command.py
@@ -323,6 +323,11 @@ def lock(ctx, state, **kwargs):
     dev_only = state.lockoptions.dev_only
     pre = state.installstate.pre
     if emit_requirements:
+        echo(crayons.yellow("""
+            Warning: The lock flags -r/--requirements, along with it's options (--keep-outdated, --dev, --dev-only)
+            will be deprecated in a future version of pipenv in favor of the new requirements command.
+            For more info see https://pipenv.pypa.io/en/latest/advanced/#generating-a-requirements-txt
+            """))
         # Emit requirements file header (unless turned off with --no-header)
         if state.lockoptions.emit_requirements_header:
             header_options = ["--requirements"]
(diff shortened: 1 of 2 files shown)


pipenv/cli/command.py
```diff
@@ -323,6 +323,15 @@ def lock(ctx, state, **kwargs):
     dev_only = state.lockoptions.dev_only
     pre = state.installstate.pre
     if emit_requirements:
+        echo(
+            crayons.yellow(
+                """
+            Warning: The lock flags -r/--requirements, along with it's options (--keep-outdated, --dev, --dev-only)
```
matteius, line 329: I think you cannot deprecate `--keep-outdated` because it is used change the behavior of locking and thus install as well.   I think the `--dev` or `--dev-only` may also be used similarly .. isn't `--dev` how you specify to only install the development requirements?  I am not sure about `--dev-only` actually.

Otherwise I think this is the right direction to head in.

Unrelated, can you also have a look at this PR:  https://github.com/pypa/pipenv/pull/5071

matteius: Oh one more thing, we are trying to not add any more crayons usage -- use click coloring instead.  @oz123 will thank you.  I have a recent example of doing just that:  https://github.com/pypa/pipenv/pull/5046/files#diff-ef852c4ac364f946819f765a6bc26f04f1b0968f31fc512949a60fa2ab0685e8R903-R910

oz123: Does `requirements` subcommand updates the packages? Is the default `keep-outdated` or update?
This should be clarified in the deperecation message.

ImreC: I've updated the message. It indeed was very confusing. Apologies. Please check if it is okay now @matteius @oz123

matteius (approved):

oz123:
@ImreC  thank you for this contribution!

PR #5078, opened 2022-04-23
More vistir dropping

oz123:
Moving to Python version 3.7+ allows us to remove more usage of vistir.
Eventually we can completely drop it, but I prefer doing this gradually instead of a giant PR.

Also, I am not done yet. Will add 2-3 more commits, and we'll continue chopping this vistir with b[i,y]te size steps.

Pun intended :-)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,4 +1,10 @@
 name: CI
+concurrency:
+  group: >-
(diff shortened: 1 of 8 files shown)


oz123:
@matteius when (and if) merging, please do not squash the commits. Doing so will lead to losing of alot of information found in the commit messages.

matteius (approved):

PR #5080, opened 2022-04-24
Drop crayons in exceptions.py

### LINKED ISSUE #5057: Turn off color via PY_COLORS=0
Please support a more standard way to turn off colors. It is tedious to turn off color for every tool separately, especially since with PY_COLORS there exists a general way for the Python ecosystem.

Even better would be to respect NO_COLOR=1, which IMHO is the closest thing to a general standard. General support for this would avoid playing whack-a-mole with every new tool that comes along.

Also, there are many valid reasons not to want colors on the command line, so calling the environment variable PIPENV_COLORBLIND is in rather bad taste.

oz123:
As discussed, we should remove crayons in favor of `click.echo` and `click.secho`.
`click` has a large user base and thus a better support across all major platform.
This PR is a test balloon. If it is merged, it allows us to carefully test that no drastic UI changes have
been introduced, without completely removing crayons.

Once we completely removed crayons, it's easy to fix #5057.

### DIFF
--- a/pipenv/exceptions.py
+++ b/pipenv/exceptions.py
@@ -1,18 +1,13 @@
 import itertools
-import re
 import sys
 from collections import namedtuple
 from traceback import format_tb
 
 from pipenv import environments
 from pipenv._compat import decode_for_output
-from pipenv.patched import crayons
-from pipenv.vendor import vistir
+from pipenv.vendor import click, vistir
 from pipenv.vendor.click.exceptions import ClickException, FileError, UsageError
 from pipenv.vendor.vistir.misc import echo as click_echo
-
-ANSI_REMOVAL_RE = re.compile(r"\033\[((?:\d|;)*)([a-zA-Z])", re.MULTILINE)
-STRING_TYPES = ((str,), crayons.ColoredString)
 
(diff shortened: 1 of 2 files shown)


pipenv/exceptions.py
```diff
@@ -296,14 +291,11 @@ def __init__(self, message=None, **kwargs):
             message = "Failed to create virtual environment."
         self.message = message
         extra = kwargs.pop("extra", None)
-        if extra is not None and isinstance(extra, STRING_TYPES):
-            # note we need the format interpolation because ``crayons.ColoredString``
-            # is not an actual string type but is only a preparation for interpolation
-            # so replacement or parsing requires this step
-            extra = ANSI_REMOVAL_RE.sub("", f"{extra}")
+        if extra is not None and isinstance(extra, str):
```
matteius, line 294: Well this is a big improvement!

matteius (approved):

pipenv/exceptions.py
oz123: This is also working without checking for `not None`. It's also shorter.

Issue #5075, opened 2022-04-22
 FileNotFoundError: [Errno 2] when installing pipenv from pipfile


limkin092:
Hi all, 

I am experiencing some issues while building my docker file on the last shown step (1):

```
COPY Pipfile* ./
RUN pip3 install pipenv
(1) RUN PIPENV_VENV_IN_PROJECT=1 pipenv install --deploy --clear --keep-outdated.
```
I get the following Stacktrace:
```
 > [builder-image 5/5] RUN pipenv install --deploy --clear --keep-outdated:                                                                                                                                                    
#10 1.527 Creating a virtualenv for this project...                                                                                                                                                                            
#10 1.527 Pipfile: /Pipfile                                                                                                                                                                                                    
#10 1.543 Using /usr/bin/python3 (3.10.4) to create virtualenv...                                                                                                                                                              
⠼ Creating virtual environment...created virtual environment CPython3.10.4.final.0-64 in 491ms                                                                                                                                 
#10 2.680   creator CPython3Posix(dest=/.venv, clear=False, no_vcs_ignore=False, global=False)
#10 2.680   seeder FromAppData(download=False, pip=bundle, setuptools=bundle, wheel=bundle, via=copy, app_data_dir=/root/.local/share/virtualenv)
#10 2.680     added seed packages: pip==22.0.4, setuptools==62.1.0, wheel==0.37.1
#10 2.680   activators BashActivator,CShellActivator,FishActivator,NushellActivator,PowerShellActivator,PythonActivator
#10 2.680 
#10 2.758✔ Successfully created virtual environment! 
#10 2.760 Traceback (most recent call last):
#10 2.760   File "/usr/local/bin/pipenv", line 8, in <module>
#10 2.760     sys.exit(cli())
#10 2.760   File "/usr/local/lib/python3.10/dist-packages/pipenv/vendor/click/core.py", line 1128, in __call__
#10 2.761     return self.main(*args, **kwargs)
#10 2.761   File "/usr/local/lib/python3.10/dist-packages/pipenv/cli/options.py", line 56, in main
#10 2.761     return super().main(*args, **kwargs, windows_expand_args=False)
#10 2.761   File "/usr/local/lib/python3.10/dist-packages/pipenv/vendor/click/core.py", line 1053, in main
#10 2.761     rv = self.invoke(ctx)
#10 2.761   File "/usr/local/lib/python3.10/dist-packages/pipenv/vendor/click/core.py", line 1659, in invoke
[...]

PR #5087, opened 2022-04-28
revert setuptools required upgrade.

matteius:
Setuptools upgrade in setup.py causing issues for some users.

### The fix

Fixes: https://github.com/pypa/pipenv/issues/5075

### The checklist

* [X] Associated issue

### DIFF
--- a/news/5075.bugfix.rst
+++ b/news/5075.bugfix.rst
@@ -1,0 +1,1 @@
+Revert specifier of ``setuptools`` requirement in ``setup.py`` back to what it was in order to fix ``FileNotFoundError: [Errno 2]`` issue report.
--- a/setup.py
+++ b/setup.py
@@ -23,7 +23,7 @@ if sys.argv[-1] == "publish":
 required = [
     "pip>=22.0.4",
     "certifi",
-    "setuptools>=60.0.0",
+    "setuptools>=36.2.1",
     "virtualenv-clone>=0.2.5",
     "virtualenv",
 ]


matteius:
> Which version is actually required?
@nomis It is a good question but from my experience many parts of pip and requirementslib pull in the latest setuptools and use it isolation anyway, and there is nothing that can be done about _that_.

> If pipenv doesn't need a newer version (and can be tested with the older version) it shouldn't be specifying it as a dependency.

It always specified this version as a dependency so this is rolling back my prior change, I don't believe we can drop it altogether as it is required.

oz123:
Won't this still pull setup tools >=60?
Can we Block larger than 60 versions?

matteius:
> Won't this still pull setup tools >=60? Can we Block larger than 60 versions?

This is just reverting it to what it was before, and it seems to work for the folks having the issue, which might just be ubuntu 22.0.4 specific.  I am not convinced that restrict it to less than the latest isn't going to cause different issues for folks that already have a newer version installed at the system level.  Doesn't seem worth it, and internally pipenv is going to pull in the latest setuptools anyway for use in isolation in some spots of requirementslib and pip.

oz123:
Ok. But pulling this requires a new release (which I'm fine with) but I'm still not sure it's necessary. The issue has a documented solution in the comments. Won't this still be an issue when we upgrade setup tools later?

matteius:
@oz123 I think it is what @nomis pointed out in the issue comments:

> When the setuptools/virtualenv/distutils bug is resolved then pipenv can depend on the newer version of those.

I did take a gamble on updating setuptools in the setup.py when dropping 3.6 support because it seemed like that version is really old, but now it seems this causes issues with whatever the system packaged setuptools or currently installed version is supposed to be so I think we should not restrict it newer yet.  I picked the `60.0.0` version based on which one they dropped 3.6 support in, but alas.   

Since pip and requirementslib seem to pull in the latest for some tasks anyway I find it is a complex relationship with setuptools. 

Is there a specific issue report in one of those other project @nomis that we could track on if its safe to try forcing the setuptools upgrade at that time?

oz123:
Ok. Then let's merge this. But we need to keep track of the upstream bug somehow.

oz123 (approved):

Issue #5092, opened 2022-05-01
`pipenv requirements` does not include markers


thomasf:
I just tried to use `pipenv reqirements` instead of `pipenv lock --requirements` but it lacks markers so windows only dependencies from my lock file makes `pip install -r ...` fail.

example transitive dependency form lock file that at some point was locked on a windows machine:

```
        "twisted-iocpsupport": {
            "hashes": [
                "sha256:306becd6e22ab6e8e4f36b6bdafd9c92e867c98a5ce517b27fdd27760ee7ae41",
                "sha256:3c61742cb0bc6c1ac117a7e5f422c129832f0c295af49e01d8a6066df8cfc04d",
                "sha256:72068b206ee809c9c596b57b5287259ea41ddb4774d86725b19f35bf56aa32a9",
                "sha256:7d972cfa8439bdcb35a7be78b7ef86d73b34b808c74be56dfa785c8a93b851bf",
                "sha256:81b3abe3527b367da0220482820cb12a16c661672b7bcfcde328902890d63323",
                "sha256:851b3735ca7e8102e661872390e3bce88f8901bece95c25a0c8bb9ecb8a23d32",
                "sha256:985c06a33f5c0dae92c71a036d1ea63872ee86a21dd9b01e1f287486f15524b4",
                "sha256:9dbb8823b49f06d4de52721b47de4d3b3026064ef4788ce62b1a21c57c3fff6f",
                "sha256:b435857b9efcbfc12f8c326ef0383f26416272260455bbca2cd8d8eca470c546",
                "sha256:b76b4eed9b27fd63ddb0877efdd2d15835fdcb6baa745cb85b66e5d016ac2878",
                "sha256:b9fed67cf0f951573f06d560ac2f10f2a4bbdc6697770113a2fc396ea2cb2565",
                "sha256:bf4133139d77fc706d8f572e6b7d82871d82ec7ef25d685c2351bdacfb701415"
            ],
            "markers": "platform_system == 'Windows'",
            "version": "==1.0.2"
        },
```

Running `pipenv lock --requirements --keep-outdated`  (on linux) results in:

```
twisted-iocpsupport==1.0.2; platform_system == 'Windows'
```
Running `pipenv requirements` (on linux) has no marker so the requirements file fails to install on linux.

```
twisted-iocpsupport==1.0.2
```

pipenv, version 2022.4.30

MrBones757:
We are also having this issue, and have reverted back to the lock mechanism.

[...]

PR #5114, opened 2022-05-30
Include markers issue #5092

ImreC:
@matteius @oz123 I think this is literally what needs to happen for this, after the changes that were done a month ago. Or am I overlooking something?


oz123:
@ImreC thanks for contributing again.
I would appreciate if it, if you could include a test case and the usual news fragment.

oz123 (changes requested):
See my other comment about test and news.

ImreC:
Yeah sure. Was just checking if this was literally it. Will do that later today

matteius:
Yeah that seems like the logical change.  @ImreC Not sure you timeline, but I'd like to do another pipenv release due to another issue in the next week or so and would love if we could include this one as well.

ImreC:
Hi @matteius, this completely fell off the map on my end. I'm having a very busy couple of weeks. I will find some time on Friday to do this

ImreC:
@oz123 @matteius I have now implemented it as an optional flag. Do you agree with this approach? Happy to make it default if that's better

matteius:
@ImreC I am thinking it should be the default behavior to include markers, but perhaps a flag like you have to exclude them (non-default) makes sense?  Especially with reports like this, having the ability to exclude markers may be relevant for some:

Also, Not sure if you saw my comment the other day I CC'd you on another issue report, the TLDR is (having trouble finding the ticket link)
> I do wonder if in the code I linked to for the requirements command, if we should have the logic first add the dev dependencies so that the production dependencies can overwrite any of those? -- that seems more correct to me than allowing the dev dependencies to break the production specifiers.

oz123:
I tend to agree here, the flag should actually be `--exclude-markers` and the default behaviour should include them.

ImreC:
@oz123 @matteius I flipped the markers flag. I really have been short for time the last couple of weeks. Hope this covers all.

docs/advanced.rst
```diff
@@ -281,6 +281,7 @@ flag::
     pytest==3.2.3
 
 Adding the ``--hash`` flag will add package hashes to the output for extra security.
+Adding the ``--exclude-markers`` flagwill exclude the markers from the output.
```
matteius, line 284: typo need space `flag will`  -- I tried to edit this but I don't think I can directly without some hoop jumping.

pipenv/cli/command.py
```diff
@@ -776,18 +777,18 @@ def requirements(state, dev=False, dev_only=False, hash=False):
 
     deps = {}
 
-    if not dev_only:
-        deps.update(lockfile["default"])
     if dev or dev_only:
         deps.update(lockfile["develop"])
+    if not dev_only:
```
matteius, line 782: Thanks for flipping the ordering here!

matteius (approved):
I approve @ImreC nice work -- just if you can please add the space for the typo so we can merge soon.

oz123:
@ImreC @matteius please also fix the merge conflict.

matteius:
@oz123 I am not seeing what you mean by a potential merge conflict ... If there is not one, I am ready to merge this and will push a commit to main to fix the documentation typo.

oz123:
Ay! Sorry depending on the type of merge, there will be a conflict or there won't be.
image

image

PR #5150, opened 2022-06-27
Remove importlib_metadata and importlib_resources

### LINKED ISSUE #5106: Vendored importlib_resources uses obsolete typing.io import
### Issue description

The vendored version of importlib_resources uses the `typing.io` namespace, which is deprecated since Python 3.8 and will be removed in Python 3.12 (cf. python/cpython#92871). This was seemingly fixed in importlib_resources v5.2.1 (according to the changelog).

oz123:
Remove these libraries as they seem unused.

Fix  #5106


matteius:
I don't think we can remove `importlib_metadata` being that click and pep517 and also the vendor version of pip `notpip` use it.  we also reference it in the `Pipfile.lock` to support Python versions < 3.10 for the same reason(s).

```
        "importlib-metadata": {
            "hashes": [
                "sha256:65a9576a5b2d58ca44d133c42a241905cc45e34d2c06fd5ba2bafa221e5d7b5e",
                "sha256:766abffff765960fcc18003801f7044eb6755ffae4521c8e8ce8e83b9c9b0668"
            ],
            "markers": "python_version < '3.10'",
            "version": "==4.8.3"
        },
```

matteius (changes requested):
Don't think we can drop `importlib_metadata` yet.  It also appears that some tests and files still reference `importlib_resources` but those references can be updated to use internal python because we dropped support for python 3.6 which was why it was included in the first place.

Pipfile.lock
```diff
@@ -585,14 +548,6 @@
             ],
             "version": "==2022.1"
         },
-        "pywin32-ctypes": {
```
matteius, line 588: I think this one is problematic and why you'll want me to run `pipenv lock` from windows on this branch.

matteius: I re-ran lock on windows (because I know that until the marker bug gets fixed, we would loose the windows specific requirements in dev) and pushed that out so this is resolved.

matteius (approved):

PR #5162, opened 2022-07-05
Wexpect trial

matteius:

### The issue

Trying to get some improved windows support and figure out why those 3 tests are failing in the CI.

### The fix

Not sure if it does.

### The checklist

* [ ] Associated issue


oz123:
We have to think on conditionally vendoring this module on Windows only packages. Otherwise, everyone has this vendored regardless of the OS.

matteius:
Nothing really indicates to me that this is necessary.

PR #5164, opened 2022-07-06
Refactor and cleannup some cruft code around no_deps flag.

matteius:
### The issue

Replacement PR for really old PR:  https://github.com/pypa/pipenv/pull/3985

### The fix

Cleans up some impossible code paths and a logical bug where a flag gets overwritten once in a loop and then retries are caused.

### The checklist

* [X] Associated issue

### DIFF
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -720,7 +720,6 @@ def batch_install(
     if sequential_deps is None:
         sequential_deps = []
     failed = not retry
-    install_deps = not no_deps
     if not failed:
         label = INSTALL_LABEL if not environments.PIPENV_HIDE_EMOJIS else ""
     else:
@@ -755,8 +754,9 @@ def batch_install(
         elif dep.is_vcs:
             is_artifact = True
         if not project.s.PIPENV_RESOLVE_VCS and is_artifact and not dep.editable:
-            install_deps = True
-            no_deps = False
+            skip_dependencies = False
+        else:
+            skip_dependencies = no_deps
 
         with vistir.contextmanagers.temp_environ():
             if not allow_global:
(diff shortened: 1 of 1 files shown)


pipenv/core.py
```diff
@@ -1459,7 +1457,7 @@ def pip_install(
     r=None,
     allow_global=False,
     ignore_hashes=False,
-    no_deps=None,
```
matteius, line 1462: Nothing that calls this ever passed anything other than True/False and the default behavior is `False` -- meaning install all dependencies.  Skipping dependency installs is the special case of `True` being passed.

pipenv/core.py
```diff
@@ -720,7 +720,6 @@ def batch_install(
     if sequential_deps is None:
         sequential_deps = []
     failed = not retry
-    install_deps = not no_deps
```
matteius, line 723: Essentially moving this to line 759 is what the original PR did that this replaces -- the reason why is because line 757 can override this for a specific artifact, but we should not use that flag for all iterations of the loop.

pipenv/core.py
```diff
@@ -1503,14 +1501,6 @@ def pip_install(
                 extra_indexes = list(project.sources)
     if requirement and requirement.vcs or requirement.editable:
         requirement.index = None
-        # Install dependencies when a package is a non-editable VCS dependency.
-        # Don't specify a source directory when using --system.
-        if not requirement.editable and no_deps is not True:
```
matteius, line 1508: The comment indicated this first block was due to really really old Pipfile.lock and no tests fail around this.

pipenv/core.py
```diff
@@ -1503,14 +1501,6 @@ def pip_install(
...
-            # Leave this off because old Pipfile.lock don't have all deps included
-            # TODO: When can it be turned back on?
-            no_deps = False
-        elif requirement.editable and no_deps is None:
```
matteius, line 1512: `no_deps` was never actually `None`.

oz123 (approved):

PR #5166, opened 2022-07-07
Vendor in pip 22.1.2

### LINKED ISSUE #5147: Vendor in the latest pip 22.1.2
Looks like there have been some new versions of pip since 22.0.4, the version we use in our current stable release.  This ticket is to track the work to vendor in `22.1.2` or latest version at the time the ticket can be accomplished

matteius:
Vendor in pip 22.1.2 -- the latest pip.  Also does:
Vendor in latest `requirementslib` which has a fix for Annotated variables and rewriting imports of packaging library, as well as vendor in latest `pip-shims` which rewrites imports of packaging library.  
The net benefit is I was able to drop the vendor'd packaging and use the one from pip _internals across the board.  Prior behavior is some things were using pip _internals packaging and some were using packaging that was installed in the virtualenv as a side-effect of creating the virtualenv.

### The issue

Fixes #5147

### The fix

By keeping pip updated in this project we maintain the latest fixes and have an easier time with future vendoring tasks.

### The checklist

* [X] Associated issue


tasks/vendoring/__init__.py
```diff
@@ -743,9 +741,14 @@ def main(ctx, package=None, type=None):
     else:
         target_dirs = [vendor_dir, patched_dir]
     if package:
-        log("Using vendor dir: %s" % vendor_dir)
-        vendor(ctx, vendor_dir, package=package)
-        download_licenses(ctx, vendor_dir, package=package)
+        if type is None or type == "vendor":
```
matteius, line 744: This is a change I was making time and time again to the vendoring script to actually respect the argument being passed.

PR #5168, opened 2022-07-08
Remove funcsigs as it is only for Python 3.3 and earlier

oz123:
``funcsigs`` is a backport of the `PEP 362`_ function signature features from
Python 3.3's `inspect`_ module. The backport is compatible with Python 2.6, 2.7
as well as 3.3 and up. 3.2

There is only one place which is used in pip-shims, but the exception
will never happen on current supported versions of pipenv and Python.

https://github.com/sarugaku/pip-shims/blob/d165d55a53065d51aaaebde4124a98c6a8c9fcba/src/pip_shims/utils.py#L387

```
    try:
        signature = inspect.signature(fn_or_class)
    except AttributeError: # never reached on python 3.3 or later.
        import funcsigs

```


matteius:
Thanks for pointing this out @oz123  I created:  https://github.com/sarugaku/pip-shims/pull/84/files
and: https://github.com/sarugaku/pip-shims/issues/83

matteius (approved):

matteius:
Can you add a news fragment?  If you merge main in you'll get the fix I applied for the windows tests failing.

PR #5172, opened 2022-07-12
Use distribution to find path of installed pipenv.

matteius:
### The issue

Now that we dropped support for python 3.6 we should migrate away from `pkg_resoruces` as

>Use of pkg_resources is discouraged in favor of [importlib.resources](https://docs.python.org/3/library/importlib.html#module-importlib.resources), [importlib.metadata](https://docs.python.org/3/library/importlib.metadata.html)

### The fix

Start chipping away at our direct `pkg_resoruces` usages.

### The checklist

* [ ] Associated issue

### DIFF
--- a/pipenv/utils/resolver.py
+++ b/pipenv/utils/resolver.py
@@ -7,6 +7,7 @@ import subprocess
 import sys
 import warnings
 from functools import lru_cache
+from importlib.metadata import PackageNotFoundError, distribution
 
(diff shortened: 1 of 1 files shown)


matteius:
The time does not seem right for this -- it needs to be revisited, but we depend on `pip` which is using `pkg_resources` from `setuptools` so this change may not make sense.

Issue #3741, opened 2019-05-10
Remove crayons from pipenv in favor of click.styles


PatrikKopkan:
Hi, I am trying to devendor packaged pipenv in Fedora. The reason is that it would be easier for us track and solve bugs, CVEs... etc.

I'm working on the pythonfinder package (not patched). It apparently uses the patched version of crayons in pipenv.

I want to ask if you are planning to move from patched crayons to upstream crayons, upstreaming your changes.

I see that the creator of crayons is also the pipenv's BDFL and pipenv uses 7 months old patched version of it.
The rationale listed is very vague: "Crayons is upgraded".

techalchemy:
Hi, someone could upstream patches to crayons but it is definitely not very high priority.  I also am not sure if they would be accepted.  Kenneth also has mostly stepped back from the BDFL role.  We could consider other options besides patching here, but it just isn't on the priority list right now so it would have to be done by someone in the community

oz123:
The patches seem like issues fixes, but also black formatting.
I would vote for completely removing crayons since everything that can be done with crayons can be done with click.

PR #5178, opened 2022-07-18
Remove Crayons per issue 3741, use click echo functions instead

otherjake:

### The issue

Remove crayons per issue 3741
Fixes https://github.com/pypa/pipenv/issues/3741

### The fix

Use click.echo, click.secho, click.style instead

### The checklist

* [ ] support colorblind envV with click.style if possible


matteius:
I think in the long run we need to deprecate the colorblind environment variable and replace it with a more appropriately named variable.  However we do need to consider that we likely had already broken that variable with the current usages of click styling.  @oz123 can you think of a way to support disabling the click styling at a global level in pipenv based on enabling a specific flag?

oz123:
> I think in the long run we need to deprecate the colorblind environment variable and replace it with a more appropriately named variable. However we do need to consider that we likely had already broken that variable with the current usages of click styling. @oz123 can you think of a way to support disabling the click styling at a global level in pipenv based on enabling a specific flag?

The appropriate name would be NO_COLOR, see here: https://no-color.org/

rich, which is already used by pip supports it. In the long run, I would like us to use the same infrastructure as pip, hence I would vote in favor using coloring, and progress bars from rich. It's more work, but it is more sustainable for us.

As for globally disabling click colors as long we  use it, we can steal some code from the [click-extras ](https://github.com/kdeldycke/click-extra/blob/9a5a7e3570eb91a0edb419e1efb15fb9549a1ece/click_extra/colorize.py#L163) project and monkey patch click to disable the color function.

oz123 (approved):

pipenv/core.py
```diff
@@ -664,10 +671,10 @@ def _cleanup_procs(project, procs, failed_deps_queue, retry=True):
                 project.environment.expand_egg_links()
                 click.echo(
                     "{}".format(
-                        crayons.yellow(
+                        click.style(
                             "Failed initial installation: Failed to overwrite existing "
                             "package, likely due to path aliasing. Expanding and trying "
-                            "again!"
+                            "again!",
```
matteius, line 677: missing `fg="yellow"`

oz123: Fixed.

pipenv/core.py
```diff
@@ -75,20 +74,22 @@
...
 else:
     INSTALL_LABEL = "   "
     INSTALL_LABEL2 = "   "
     STARTING_LABEL = "   "
 
 # Disable colors, for the color blind and others who do not prefer colors.
-if environments.PIPENV_COLORBLIND:
-    crayons.disable()
+#if environments.PIPENV_COLORBLIND:
+
```
otherjake, line 86: this is the spot where it previously switched the NO_COLOR behavior

oz123: image

otherjake: I am implementing a conditional monkeypatch in this spot on style and secho, they are used separately for composing messages that aren't all one color, also tests for each for the old and new variables.  Should an issue be created for deprecating PIPENV_COLORBLIND?

PR #5182, opened 2022-07-22
Vendoring improvements

oz123:
This PR continues the cleaning of vendored package.

 * Removed YAML
 * Removed funcsigs
 * Removed copy of crayons
 * Patch dparse to only locally import yaml
 * Bump version of tomlkit we vendor

Additionally, it minor changes to the vendoring script itself (mostly minor cleanup related to Python2 and legacy backports).


docs/dev/contributing.rst
```diff
@@ -230,9 +230,11 @@ be aware of the following things when filing bug reports:
 Run the tests
 -------------
 
-Two ways of running the tests are as follows:
+There are a few ways of running the tests:
 
-1. ``./run-tests.sh`` or ``run-tests.bat``
+1. run-tests.sh
```
matteius, line 235: Might be worth noting how the CI runs the tests -- that is the file I always reference for running them.
One time setup:
```
git submodule sync
git submodule update --init --recursive
python -m pip install -e . --upgrade
pipenv install --deploy --dev
```

Run tests: `pipenv run pytest -ra -n auto -v --fulltrace tests`

matteius: Nevermind, it is right below this already:  https://github.com/pypa/pipenv/pull/5182/files#diff-b423d06524f8dec680323ad0294641ed796dfa5b2c876be7d84df83d21d374f5R264-R268

pipenv/vendor/pip_shims/models.py
```diff
@@ -76,7 +76,7 @@ def __init__(
         version,
         round_prereleases_up=True,
         base_import_path=None,
-        vendor_import_path="pip._vendor",
+        vendor_import_path="pipenv.patched.notpip._vendor",
```
matteius, line 79: This should not get changed I don't think.   It would require a patch being its not an actual import that can be re-written, but I suspect it is not required because there is already a concept in `pip-shims` of specifying what the pip module is, so I think this line should not have to change.

oz123: Fixed.

matteius (approved):

PR #5184, opened 2022-07-23
Heed no color requests

otherjake:
### The issue
Make Click heed requests for no color per https://no-color.org
### The fix
Monkey patch click.secho in core.py
Add new test for monkey patch doing the job correctly


matteius:
Declining in favor of: https://github.com/pypa/pipenv/pull/5185

Thanks for working with me on this @otherjake !

PR #5185, opened 2022-07-23
Refactor to support the NO_COLOR argument and deprecate PIPENV_COLORBLIND

matteius:
Follow-up to make NO_COLOR work and deprecated PIPENV_COLORBLIND.

### The issue

There was some issues with the prior monkey patch suggested approach:  https://github.com/pypa/pipenv/pull/5184

So I opened this PR instead.

### DIFF
--- a/pipenv/cli/command.py
+++ b/pipenv/cli/command.py
@@ -25,7 +25,6 @@ from pipenv.cli.options import (
 )
 from pipenv.exceptions import PipenvOptionsError
 from pipenv.utils.processes import subprocess_run
-from pipenv.utils.shell import env_to_bool
 from pipenv.vendor.click import (
     Choice,
     argument,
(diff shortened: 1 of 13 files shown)


pipenv/cli/command.py
```diff
@@ -93,6 +94,20 @@ def cli(
...
+            "PIPENV_COLORBLIND is deprecated, use NO_COLOR instead"
+            "Per https://no-color.org/",
+            err=True,
+        )
+
+    if env_to_bool(os.getenv("NO_COLOR")) or env_to_bool(os.getenv("PIPENV_COLORBLIND")):
+        from pipenv.utils.shell import style_no_color
+        from pipenv.vendor import click
+
+        click.original_style = click.style
```
oz123: Good to keep around `click.style` with `click.original_style`.

oz123 (changes requested):

PR #5190, opened 2022-07-25
Never rely on user installed pip--rely on vendored pip

matteius:

### The issue

I noticed this through a number of issue reports that sometimes a user's local pip conflicts with pipenv's expected behavior.   To me, it does not make sense that we do not always use our patched vendored pip to install certain types of packages.   With this change it should be possible for a user to only install `pipenv` and not have pip installed at all.   There is nothing wrong with having both installed, but why have `pipenv` require it and rely on it when we go to the work to patch an vendor in the latest and greatest `pip`?


matteius:
Possibly the last test failure hinges on: https://github.com/pypa/pip/issues/11120

matteius:
This is where the requirements are coming from that fail to install in the standalone pip in `test_lock_nested_direct_url`:  https://github.com/sarugaku/vistir/blob/master/pyproject.toml#L2

matteius:
Closing in favor of:  https://github.com/pypa/pipenv/pull/5193

PR #5193, opened 2022-07-26
minimum set of changes for using internal vendor pip.

matteius:

### The issue

I noticed this through a number of issue reports that sometimes a user's local pip conflicts with pipenv's expected behavior.   To me, it does not make sense that we do not always use our patched vendored pip to install certain types of packages.   With this change it should be possible for a user to only install `pipenv` and not have pip installed at all.   There is nothing wrong with having both installed, but why have `pipenv` require it and rely on it when we go to the work to patch an vendor in the latest and greatest `pip`?

I will rename `notpip` to `pip` in a follow-up PR.


matteius:
Declining in favor of branch based on this one that also had to vendor in pip 22.2 to get this last failing test passing:  https://github.com/pypa/pipenv/pull/5196

PR #5197, opened 2022-07-27
We need to be able to modify vistir to fix this bug, and we should work towards its complete removal.

matteius:
### The issue

Vistir has broken distutil code and probably other bugs.  Fix the known bug here, and note in the vendoring that its deprecated and our copy has diverged from the parent pypi `vistir` -- our goal is complete removal of this library.   An effort will be made to begin converting usages in `requirementslib` since that will be the biggest blocker to completely removing it from `pipenv`.

### DIFF
--- a/pipenv/vendor/vendor.txt
+++ b/pipenv/vendor/vendor.txt
@@ -31,7 +31,7 @@ toml==0.10.2
 tomli==1.1.0
 tomlkit==0.9.2
 urllib3==1.26.6
-vistir==0.5.2
+vistir==0.5.2  # Deprecated: our copy is modified to fix bugs
 wheel==0.36.2
 yaspin==2.0.0
 zipp==3.5.0
--- a/news/5197.vendor.rst
+++ b/news/5197.vendor.rst
@@ -1,0 +1,1 @@
+Fix ``distutils`` usage in our copy of ``vistir`` and note the library is deprecated and marked for removal.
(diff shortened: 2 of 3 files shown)


pipenv/vendor/vendor.txt
```diff
@@ -31,7 +31,7 @@ toml==0.10.2
 tomli==1.1.0
 tomlkit==0.9.2
 urllib3==1.26.6
-vistir==0.5.2
+vistir==0.5.2  # Deprecated: our copy is modified to fix bugs
```
matteius, line 34: we should try commenting this out entirely and see if vendoring still passes -- I think it won't though.  Future vendoring would overwrite this bugfix.  Maybe we should move vistir out of the vendor namespace and chip away at it from within `pipenv` module at same level as `utils`?

oz123: If we can remove vistir completely, I will be extremely happy. But I am afraid it won't be so easy.
However, can you add a patch in the patches directory?

matteius: I'd prefer to not patch it and move it out of vendor into a the level of utils so we can whack away at it.

matteius: Or -- I just got access to all of https://github.com/sarugaku so I can upstream this as well.

oz123: That's what I meant. I wrote uranusjar today and he was very quick to respond.

matteius:
This is resolved by this PR:  https://github.com/pypa/pipenv/pull/5199

PR #5196, opened 2022-07-27
Vendor in pip==22.2, rename notpip -> pip and exclusively use vendored pip

matteius:
Was branched from and supersedes this PR: https://github.com/pypa/pipenv/pull/5193
 which had one failing test that I decided to see if upgrading to the latest pip 22.2 would fix, and it has fixed it locally -- woo-hoo!

Fixes #5188

### The issue

This accomplishes the following:
* rename patched `notpip` to `pip` to be clear that its a patched version of `pip`.
* Remove the part of the `_post_pip_import.patch` that overrode the standalone pip to be the user installed pip -- and instead go with the patched import to use our vendor'd pip.
* Vendor in the next newest version of `pip==22.2`

PR #5199, opened 2022-07-28
notpip is pip, Vendor in pip==22.2.1, standalone pipenv

### LINKED ISSUE #5188: Vendor in Pip 22.2.x
On July 21st `pip==22.2` was released.    https://pypi.org/project/pip/22.2/

matteius:
Was branched from and supersedes this PR: https://github.com/pypa/pipenv/pull/5193
Also supersedes this PR: https://github.com/pypa/pipenv/pull/5196
 which had one failing test that I decided to see if upgrading to the latest pip 22.2 would fix, and it has fixed it locally -- woo-hoo! 

Fixes #5188

### The issue

This accomplishes the following:
* rename patched `notpip` to `pip` to be clear that its a patched version of `pip`.
* Remove the part of the `_post_pip_import.patch` that overrode the standalone pip to be the user installed pip -- and instead go with the patched import to use our vendor'd pip.
* Vendor in the next newest version of `pip==22.2.1`
* Vendor in latest requirementslib
* Vendor in latest vistir

This will allow for conversion off of some pip-shims, not needing pip installed to use pipenv (or having conflicting versions), and several bug fixes from the vendored dependencies


pipenv/vendor/click/decorators.py
```diff
@@ -372,7 +372,7 @@ def callback(ctx: Context, param: Parameter, value: bool) -> None:
                 from importlib import metadata  # type: ignore
             except ImportError:
                 # Python < 3.8
-                import pipenv.vendor.importlib_metadata as metadata  # type: ignore
```
oz123, line 375: Should this be changed? I don't think so, if we still support Python3.7.

pipenv/vendor/markupsafe/_speedups.c
```diff
@@ -1,339 +0,0 @@
-#include <Python.h>
```
oz123, line 1: Do we really want to delete this file?

pipenv/vendor/pep517/meta.py
```diff
@@ -9,7 +9,7 @@
 try:
     import importlib.metadata as imp_meta
 except ImportError:
-    import pipenv.vendor.importlib_metadata as imp_meta
+    import importlib_metadata as imp_meta
```
oz123, line 12: We never fail here with ImportError. But it's still confusing that this was changed.

matteius: I think I recall pondering about it because `import pipenv.vendor.importlib_metadata` doesn't exist -- It was confusing to me we would reference a fallabck import that does not exist.

PR #5203, opened 2022-07-28
Check environment variable PIPENV_CUSTOM_VENV_NAME

### LINKED ISSUE #1226: How to set the full name of the virtualenv created

otherjake:
Resolves #1226

Allows user to set a custom venv name

### The fix
Check for PIPENV_CUSTOM_VENV_NAME environment variable, accept it as venv name at creation if it is set


oz123 (approved):

oz123 (changes requested):
Instead of asking `os.getenv` directly, we should go through:
```
self.s.PIPENV_CUSTOM_VENV_NAME
```
All settings are found in `environments.Settings` class.
Using this class conventions is also to document the variable, as other variables are documented there.

oz123:
Setting `PIPENV_CUSTOM_VENV_NAME` is going to trick a lot of people if we don't preserve it in Pipefile. 

Suppose I did:
`export PIPENV_CUSTOM_VENV_NAME=foobar`

Now there is:
```
home/myuser/.virtualenvs/foobar`
```

If I run pipenv in a new shell it's not going to have PIPENV_CUSTOM_VENV_NAME, hence it will look for an environment
in the hash location, and not in  `foobar`. Hence, I suggest we preserve this in the Pipfile, and read it from there.

@matteius do you have an opinion here?

matteius:
>  Setting PIPENV_CUSTOM_VENV_NAME is going to trick a lot of people if we don't preserve it in Pipefile.

@oz123 I am not sure I agree with that approach when its much easier to have a `.env` file to store the environment variable you want to use for that project -- I fear that if its in the `Pipfile` it forces everyone on the project to have the same virtualenv name, and some may want the default generated one whereas others want to customize it -- a local `.env` solves this.  Otherwise I might already have a virtualenv setup and my teammate pushes that change to the `Pipfile` which forces my machine to generate a new virtualenv under this other name.

otherjake:
I am considering putting self.s.PIPENV_CUSTOM_VENV_NAME into activate in environment.py, that seems like the creating a venv context level that this is appropriate to.  Waiting for more consensus around the pipfile bit.

matteius:
> I am considering putting self.s.PIPENV_CUSTOM_VENV_NAME into activate in environment.py

@otherjake The reading of the environment variable should happen in `environments.py` where the other environment variables are read from.

oz123:
@matteius true. Putting it in `.env` is the correct way. Totally forgot we support this.

docs/advanced.rst
```diff
@@ -558,6 +558,10 @@ wherever you want, e.g.::
 
 In addition, you can also have Pipenv stick the virtualenv in ``project/.venv`` by setting the ``PIPENV_VENV_IN_PROJECT`` environment variable.
 
+☤ Custom Virtual Environment Name
+-------------------------------------
+
+Pipenv supports a custom name for the virtual environment set at ``PIPENV_CUSTOM_VENV_NAME``.  by default the venv will be named after the project directory.
```
matteius, line 564: Would be better to say: 
```
Pipenv supports the ability to specify a custom name for the virtual environment by setting the environment variable ``PIPENV_CUSTOM_VENV_NAME`` to be whatever you want.  The logical place to specify this would be in a user's `.env` file in the root of the project, which gets loaded by pipenv when it is invoked.   The default behavior is unchanged: default name is generated by using the project directory combined with a dash and encoded hash value .
```

oz123:
This obsoletes, https://github.com/pypa/pipenv/pull/4974. This is a reminder for us to  make sure we close that one too.

matteius (approved):
Nice @otherjake !

oz123:
@otherjake Thank you for your contribution! I added a small change to your test. Didn't want to bother you again with changes.
Please take a look at it.

Besides that, as usual :100: points  and a big thanks!

oz123 (approved):

PR #5207, opened 2022-07-29
Regen patch hash

otherjake:
Patch Hash Didn't Match


matteius:
Git was showing the patch file as corrupt on Jakes computer, so had him regenerate it and it seems to apply find on vendoring here and on my computer.  Noting that the hash length has increased by one character.

matteius (approved):

oz123 (approved):

Issue #5180, opened 2022-07-20
pipenv update on linux using python 3.10 leaves files in /tmp/ 


dbaxa:
Be sure to check the existing issues (both open and closed!), and make sure you are running the latest version of Pipenv.

### Issue description

`pipenv update` leaves temp files in /tmp/ which should not persist/remain after python exits.

### Expected result

No temp files related to pipenv remain in /tmp/

### Actual result

pipenv temp files are left in `/tmp`

### Steps to replicate

`pipenv update`

-------------------------------------------------------------------------------

Please run `$ pipenv --support`, and paste the results here. Don't put backticks (`` ` ``) around it! The output already contains Markdown formatting.
pipenv --support

<details><summary>$ pipenv --support</summary>

Pipenv version: `'2022.7.4'`

Pipenv location: `'/home/user/.virtualenvs/myvenv/lib/python3.10/site-packages/pipenv'`

Python location: `'/home/user/.virtualenvs/myvenv/bin/python'`

Python installations found:

  - `3.10.4`: `/home/user/.virtualenvs/myvenv/bin/python`
  - `3.10.4`: `/home/user/.virtualenvs/myvenv/bin/python3`
  - `3.10.4`: `/home/user/.virtualenvs/myvenv/bin/python`
  - `3.10.4`: `/home/user/.virtualenvs/myvenv/bin/python3`
  - `3.10.4`: `/usr/bin/python3`
  - `3.10.4`: `/usr/bin/python`
  - `3.10.4`: `/bin/python3`
  - `3.10.4`: `/bin/python`
  - `2.7.18`: `/usr/bin/python2`
  - `2.7.18`: `/usr/bin/python2.7`
  - `2.7.18`: `/bin/python2`
  - `2.7.18`: `/bin/python2.7`

PEP 508 Information:

```
{'implementation_name': 'cpython',
 'implementation_version': '3.10.4',
 'os_name': 'posix',
 'platform_machine': 'x86_64',
 'platform_python_implementation': 'CPython',
 'platform_release': '5.15.0-41-generic',
 'platform_system': 'Linux',
 'platform_version': '#44-Ubuntu SMP Wed Jun 22 14:20:53 UTC 2022',
 'python_full_version': '3.10.4',
 'python_version': '3.10',
 'sys_platform': 'linux'}
```

System environment variables:

  - `SHELL`
  - `LANGUAGE`
  - `PWD`
  - `LOGNAME`
  - `XDG_SESSION_TYPE`
[...]

PR #5210, opened 2022-07-30
Issue 5180 tempfiles are leak in tmp

dqkqd:
Thanks for creating such beautiful project!

### The issue
I'm trying to fix https://github.com/pypa/pipenv/issues/5180

`get_paths`, `get_lib_paths`, `get_include_path` functions create new .json files to store output called from `subprocess_run` without closing it.

### The fix

Because the .json file is used only inside the function call. I have two solutions for this.
- Delete the .json file at the end of each function.
- Don't create a file, write output directly into `stdout`.

As these functions contain many returns, there will be a lot of duplicated codes to delete the file. So I tend to the second solution.

As far as I know, `subprocess_run` execute `py_command`, and if it failed to execute, `c.stdout` will be printed out to console. But the `py_command` does not print anything out. Therefore `c.stdout` is always empty. 
Instead of creating new temporary file, I think it is better to use `c.stdout` to write directly to `paths` variable.

### The checklist

* [ ] Associated issue


pipenv/environment.py
```diff
@@ -476,22 +465,17 @@ def get_include_path(self) -> Optional[Dict[str, str]]:
...
-        tmpfile = vistir.path.create_tracked_tempfile(suffix=".json")
-        tmpfile.close()
-        tmpfile_path = make_posix(tmpfile.name)
         py_command = (
-            "import distutils.sysconfig, io, json, sys; paths = {{u'include': "
-            "u'{{0}}'.format(distutils.sysconfig.get_python_inc(plat_specific=0)), "
-            "u'platinclude': u'{{0}}'.format(distutils.sysconfig.get_python_inc("
-            "plat_specific=1)) }}; value = u'{{0}}'.format(json.dumps(paths));"
-            "fh = io.open('{0}', 'w'); fh.write(value); fh.close()"
+            "import distutils.sysconfig, io, json, sys; paths = {u'include': "
```
matteius, line 469: I realize you are just refactoring existing code, but now I notice this usage of distutils which is totally deprecated.  I wonder what the correct transformation of this usage would be.

dqkqd: Instead of writing to a file, I refactor so it print out to the `stdout` directly. 
Because the filename should not be passed to string as variable anymore, I deleted some unnecessary curly braces.

dqkqd: As you said, I tried to remove distutils.sysconfig and use sysconfig instead.

dqkqd:
I noticed that `self.get_paths()` does not raise any exception. So I think exception handling is unnecessary, and we could safely remove `self.get_include_path()`, `self.get_lib_paths()`, `self._replace_parent_version()` functions.
https://github.com/pypa/pipenv/blob/75b77ebebf5559c75f3ece9eba068afecba3769b/pipenv/environment.py#L217-L240

pipenv/environment.py
dqkqd: Since `sysconfig.get_path` does not Exception like `distutils.sysconfig`, the `c.returncode` is always `0`. Therefore, those lines below could be refactored
https://github.com/pypa/pipenv/blob/75b77ebebf5559c75f3ece9eba068afecba3769b/pipenv/environment.py#L393-L415

matteius:
@dqkqd Looks good, nice work! -- do you mind adding a news fragment explaining your change?

matteius (approved):

matteius:
@dqkqd Thanks for adding a news fragment -- the linter wants there to be a newline at the end of that news fragment.

dqkqd:
> @dqkqd Thanks for adding a news fragment -- the linter wants there to be a newline at the end of that news fragment.

Sorry, I forgot to run `pre-commit`

dqkqd:
Sorry for the mess with the fragment file. I fixed it again.
I also remove exception handling when trying to run `self.get_paths()`.
I'm not sure `returncode == 0`  inside `self.get_paths()`should be removed or not. Since `py_command` run inside `suprocess_run`, might be it would be better to check in case `py_command` has error.

matteius (changes requested):
@dqkqd It would seem this latest commit is problematic for windows: https://github.com/pypa/pipenv/pull/5210/commits/3ca7c96107bf589634f9e2e2e400effcae1b437b 

Can it be reverted?

dqkqd:
> @dqkqd It would seem this latest commit is problematic for windows: [3ca7c96](https://github.com/pypa/pipenv/commit/3ca7c96107bf589634f9e2e2e400effcae1b437b)
> 
> Can it be reverted?
Yes, I reverted it. Not sure why it didn't work since I don't have windows.

matteius (approved):

oz123 (approved):

matteius:
@dqkqd I just want to say again what awesome work you did with this change -- to be able to eliminate some file usage altogether is amazing for performance, and I think we see that in the new measurements that @oz123 pointed me at this evening, which are provided by the head of Lincoln Loop:  https://lincolnloop.github.io/python-package-manager-shootout/
Its hard to look back on the historical data presently, but its in the github actions and from the tweet I saw before, the variance of the mean time to lock was reduced as was the overall time.   Nice work!

dqkqd:
@matteius Thank you so much. This is my first time contributing to open sources. Sorry for the mess up with git.

PR #5220, opened 2022-08-04
Fix extra indexes

aronbierbaum:
### The issue

When installing a requirement that explicitly references an index all indexes are searched even though `install_search_all_sources = False`.

### The fix

There were two different locations where the code appears to not have the expected behavior.

### The checklist

* [ ] Associated issue

### DIFF
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -1503,7 +1503,8 @@ def pip_install(
         trusted_hosts=trusted_hosts,
         pypi_mirror=pypi_mirror,
     )
-    if not search_all_sources and requirement.index in sources:
+    source_names = {src.get("name") for src in sources}
+    if not search_all_sources and requirement.index in source_names:
         sources = list(filter(lambda d: d.get("name") == requirement.index, sources))
     if r:
         with open(r, "r") as fh:
--- a/pipenv/utils/indexes.py
+++ b/pipenv/utils/indexes.py
@@ -45,9 +45,9 @@ def prepare_pip_source_args(sources, pip_args=None):
 
 
 def get_project_index(
-    project: Optional[Union[str, TSource]],
-    index: Optional[List[str]] = None,
-    trusted_hosts: Optional[Project] = None,
+    project: Project,
+    index: Optional[Union[str, TSource]] = None,
+    trusted_hosts: Optional[List[str]] = None,
 ) -> TSource:
     from pipenv.project import SourceNotFound
 
@@ -68,7 +68,7 @@ def get_project_index(
 def get_source_list(
     project: Project,
     index: Optional[Union[str, TSource]] = None,
-    extra_indexes: Optional[List[str]] = None,
+    extra_indexes: Optional[Union[str, List[str]]] = None,
     trusted_hosts: Optional[List[str]] = None,
     pypi_mirror: Optional[str] = None,
 ) -> List[TSource]:
--- a/pipenv/utils/indexes.py
+++ b/pipenv/utils/indexes.py
@@ -78,14 +78,16 @@ def get_source_list(
     if extra_indexes:
         if isinstance(extra_indexes, str):
             extra_indexes = [extra_indexes]
+
         for source in extra_indexes:
             extra_src = get_project_index(project, source)
             if not sources or extra_src["url"] != sources[0]["url"]:
                 sources.append(extra_src)
-        else:
-            for source in project.pipfile_sources:
-                if not sources or source["url"] != sources[0]["url"]:
-                    sources.append(source)
+
+        for source in project.pipfile_sources:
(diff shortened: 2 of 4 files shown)


pipenv/utils/indexes.py
```diff
@@ -82,10 +82,6 @@ def get_source_list(
             extra_src = get_project_index(project, source)
             if not sources or extra_src["url"] != sources[0]["url"]:
                 sources.append(extra_src)
-        else:
```
oz123, line 85: How are you handling the supposed expected behavior?
Should it not be addressed?

oz123:
This seems like a good improvement, but I still would like to get a second opinion from @matteius .
Could you please also add a news file?

oz123 (changes requested):

pipenv/utils/indexes.py
matteius: Yeah, I don't think this part should be removed because I think it might be what adds a new index to the `Pipfile` when specifying one that does not exist via the CLI.

aronbierbaum: I updated the pull to simply move the code outside of the `for`/`else`.

matteius: The change is still problematic, witthout the else you could be adding the sources twice -- I am not sure what the intent here is moving this outside the else.  There are two failing tests though:
> FAILED tests/integration/test_lock.py::test_private_index_skip_lock - assert ...
FAILED tests/integration/test_install_uri.py::test_install_specifying_index_url

aronbierbaum: Unless I'm missing something, which I definitely could be, the `else` was always executed because the `for source in extra_indexes` loop was never terminated by a `break` statement. As a result I simply moved the second `for` loop outside the `else` statement to make it obvious that it was always executed.

https://docs.python.org/3/tutorial/controlflow.html#break-and-continue-statements-and-else-clauses-on-loops

I think the test failures are caused by a different commit on this pull.

pipenv/core.py
```diff
@@ -1511,7 +1511,8 @@ def pip_install(
         trusted_hosts=trusted_hosts,
         pypi_mirror=pypi_mirror,
     )
-    if not search_all_sources and requirement.index in sources:
+    source_names = {src.get("name") for src in sources}
+    if not search_all_sources and requirement.index in source_names:
```
aronbierbaum, line 1515: I think this change might have broken some tests. The proposed change only shows that the previous code would never find `requirement.index` in `sources` because the types didn't match. `str` and `List[TSource]`

matteius: @aronbierbaum This patch fixes the other test that does fail:
```
diff --git a/tests/integration/test_lock.py b/tests/integration/test_lock.py
index 28e7f21f..d571a8d0 100644
--- a/tests/integration/test_lock.py
+++ b/tests/integration/test_lock.py
@@ -335,10 +335,11 @@ name = "testpypi"

 [packages]
 pipenv-test-private-package = {version = "*", index = "testpypi"}
+pipenv-test-public-package = {version = "*", index = "pypi"}
 requests = "*"
             """.strip()
             f.write(contents)
```

matteius: The reason that is important is because it doesn't search all sources and `pipenv-test-private-package` is in testpypi but requires `pipenv-test-public-package` is in `pypi`.   When we initially switched to index restricted packages, there were some users that had to do similar things for their Pipfile index restrictions to work.

pipenv/utils/indexes.py
matteius: I see, this change makes sense then.

matteius:
@aronbierbaum I am not sure why earlier I could not get that second test failure, because I was able to reproduce it now.   I think this would be a good way the fix the other test, since we showed both methods here, one where we rely on index restrictions, and this test where we allow all indexes to be searched:
```
diff --git a/tests/integration/test_install_uri.py b/tests/integration/test_install_uri.py
index f21da0f4..470274ef 100644
--- a/tests/integration/test_install_uri.py
+++ b/tests/integration/test_install_uri.py
@@ -209,6 +209,9 @@ name = "pypi"
 six = "*"

 [dev-packages]
+
+[pipenv]
+install_search_all_sources = true
             """.strip()
             f.write(contents)
         c = p.pipenv("install pipenv-test-private-package --index https://test.pypi.org/simple")
```

matteius (approved):

AndreasBBS:
@aronbierbaum Thanks for this fix! :+1:  I've been pulling hairs for hours trying to understand why I pipenv is installing the public "utils" package instead of my private one. You're awesome man! :1st_place_medal:

oz123 (approved):

Issue #5128, opened 2022-06-13
conda-feedstock broken?


drewgilliam:
I've haven't gotten any response over at conda-forge, so I'm hoping someone here can help.  The conda-forge feedstock has failed for almost all 2022 pipenv releases (latest pipenv version on conda-forge is 2022.1.8).  Would anyone on the pipenv team be able to help fix the conda-forge feedstock?

https://github.com/conda-forge/pipenv-feedstock/issues/24
https://anaconda.org/conda-forge/pipenv/files

matteius:
@drewgilliam  Can you kick off some new builds over there?  All of the old builds are too old to have the log files still available in Github Actions.

drewgilliam:
Thanks for the quick response!  I don't think I have permissions to do anything in pipenv-feedstock - I'm just a user of conda-forge not an admin of the pipenv-feedstock.  There seem to only be two human contributors in the history of pipenv-feedstock., @roccqqck and @beckermr, maybe they could kick something off?

beckermr:
I don't recall making any non-admin changes to the feedstock. (I did accidentally make commits to most feedstocks on conda forge a while back.)

Feel free to push a fix.

matteius:
@beckermr Well for one, I don't know what the `pipenv-feedstock` project is about, and the actual problem has been lost to the logs being no longer available, so unless you can kick off a run that demonstrates the failure within the logs, then I won't be spending time on it.

matteius:
Sorry for the delay in responding @drewgilliam, yes I believe this is expected that CI should be set to 1 or unset or 0.  However at first glance, that code doesn't look correct.  Shouldn't this line:
`PIPENV_IS_CI = env_to_bool(os.environ.get("CI") or os.environ.get("TF_BUILD") or False)`
Actually be more correct as this:
`PIPENV_IS_CI = env_to_bool(os.environ.get("CI")) or env_to_bool(os.environ.get("TF_BUILD"))`

cc: @oz123

matteius:
@drewgilliam Clarifying question -- does your environment need CI to be set to `azure` for other reasons in the build pipeline besides `pipenv`?

drewgilliam:
I don't know too much about the conda-forge feedstock environment - in my limited understanding, its a fairly automated environment where a lot of stuff is autogenerated for each feedstock package by the "conda smithy".  The `CI` variable appears to be set in one of these auto-generated files, I assume/hope for good reason:
[...]

PR #5222, opened 2022-08-05
Improve CI variable handling

matteius:
Fixes: #5128


pipenv/environments.py
```diff
@@ -81,7 +81,15 @@ def normalize_pipfile_path(p):
 os.environ.pop("__PYVENV_LAUNCHER__", None)
 # Internal, to tell whether the command line session is interactive.
 SESSION_IS_INTERACTIVE = _isatty(sys.stdout)
-PIPENV_IS_CI = env_to_bool(os.environ.get("CI") or os.environ.get("TF_BUILD") or False)
+
+# TF_BUILD indicates to Azure pipelines it is a build step
+PIPENV_IS_CI = os.environ.get("CI") or os.environ.get("TF_BUILD")
```
oz123, line 86: What is TF_BUILD used for? 
We should document this for the future.
One line above should suffice.

pipenv/environments.py
```diff
@@ -81,7 +81,15 @@ def normalize_pipfile_path(p):
...
+try:
+    PIPENV_IS_CI = env_to_bool(PIPENV_IS_CI)
```
oz123, line 88: It should suffice to use `os.getenv` instead of `env_to_bool` if you just want to check it the variable was set.

oz123 (changes requested):

pipenv/environments.py
matteius: I tried to do that in line 85.

oz123: I missed that. Sorry.

Issue #5214, opened 2022-07-31
KeyError: 'packages' was thrown in shell complettion


dqkqd:
### Issue description

I enabled shell complation for ZSH. When I press tab after `$ pipenv install`, KeyError: 'packages' was thrown.

```
$ pipenv install Traceback (most recent call last):
  File "/home/user/.local/bin/pipenv", line 8, in <module>
    sys.exit(cli())
  File "/home/user/.local/lib/python3.10/site-packages/pipenv/vendor/click/core.py", line 1128, in __call__
    return self.main(*args, **kwargs)
  File "/home/user/.local/lib/python3.10/site-packages/pipenv/cli/options.py", line 56, in main
    return super().main(*args, **kwargs, windows_expand_args=False)
  File "/home/user/.local/lib/python3.10/site-packages/pipenv/vendor/click/core.py", line 1048, in main
    self._main_shell_completion(extra, prog_name, complete_var)
  File "/home/user/.local/lib/python3.10/site-packages/pipenv/vendor/click/core.py", line 1123, in _main_shell_completion
    rv = shell_complete(self, ctx_args, prog_name, complete_var, instruction)
  File "/home/user/.local/lib/python3.10/site-packages/pipenv/vendor/click/shell_completion.py", line 49, in shell_complete
    echo(comp.complete())
  File "/home/user/.local/lib/python3.10/site-packages/pipenv/vendor/click/shell_completion.py", line 291, in complete
    completions = self.get_completions(args, incomplete)
  File "/home/user/.local/lib/python3.10/site-packages/pipenv/vendor/click/shell_completion.py", line 272, in get_completions
    obj, incomplete = _resolve_incomplete(ctx, args, incomplete)
  File "/home/user/.local/lib/python3.10/site-packages/pipenv/vendor/click/shell_completion.py", line 576, in _resolve_incomplete
    if _is_incomplete_argument(ctx, param):
  File "/home/user/.local/lib/python3.10/site-packages/pipenv/vendor/click/shell_completion.py", line 439, in _is_incomplete_argument
    value = ctx.params[param.name]
KeyError: 'packages'
pipenv         

```

### Expected result

No exceptions thrown.

### Steps to replicate
Enable [shell-completion](https://github.com/pypa/pipenv#shell-completion)
Press tab after `$ pipenv install`.

-------------------------------------------------------------------------------

[...]

PR #5225, opened 2022-08-06
fix bug when auto completing install and uninstall

bakhtiary:
### The issue

https://github.com/pypa/pipenv/issues/5214

This commit fixes the auto completing crash mentioned in the above command.

### The fix

The bug is related to click and I have opened a corresponding issue in their github:
https://github.com/pallets/click/issues/2336

The main problem is that expose_value=False when used for arguments breaks the autocomplete.
The best solution is to fix click and then vendor the new version. However "expose_value=False" doesn't do anything for us as far as I can see. So removing it or setting it to true is a viable option.

### The checklist

* [X] Associated issue

### DIFF
--- a/news/5214.bugfix.rst
+++ b/news/5214.bugfix.rst
@@ -1,0 +1,2 @@
+Fix auto-complete crashing on 'install' and 'uninstall' keywords
+
--- a/news/5214.bugfix.rst
+++ b/news/5214.bugfix.rst
@@ -1,2 +1,1 @@
 Fix auto-complete crashing on 'install' and 'uninstall' keywords
-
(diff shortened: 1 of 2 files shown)


matteius (approved):

matteius:
@bakhtiary The linter step failed on the `pre-commit` check for the news fragment -- could you run this and commit the changes?
`pre-commit run --all-files --verbose --show-diff-on-failure`

bakhtiary:
done!

Amir H. Bakhtiary
http://amirhb.com/

oz123 (approved):

PR #5226, opened 2022-08-06
Convert this test off pip-shims, it became flakey recently.

matteius:
### The issue

This test sporadically fails on some of the CI runs -- looking more closely, we can convert off a vistir usage and update this test to not use pip-shims.  Hopefully this helps the CI not fail this test randomly.

### DIFF
--- a/pipenv/utils/dependencies.py
+++ b/pipenv/utils/dependencies.py
@@ -1,5 +1,6 @@
 import os
 from contextlib import contextmanager
+from tempfile import NamedTemporaryFile
 from typing import Mapping, Sequence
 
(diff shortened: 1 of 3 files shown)


pipenv/utils/dependencies.py
```diff
@@ -273,9 +274,7 @@ def convert_deps_to_pip(
         return dependencies
 
     # Write requirements.txt to tmp directory.
-    from pipenv.vendor.vistir.path import create_tracked_tempfile
-
-    f = create_tracked_tempfile(suffix="-requirements.txt", delete=False)
```
oz123, line 278: You always get bonus points for removing usage of vistir :-)

oz123 (approved):

PR #5227, opened 2022-08-06
Use pep517 from pip own _vendor

oz123:
Continuing to remove duplications, we now use pep517 bundled with pip.
Also, cleanup some unused code in the vendor task.


oz123:
These freaky Mac Tests, these breakages continue. However, they have nothing to do with the proposed changes.

matteius (approved):

Issue #4371, opened 2020-06-29
Pipenv installs spurious dependencies


J-M0:
### Issue description

When installing any version of Django < 3.0 as a dependency and django-debug-toolbar as a dev dependency, `pipenv lock` determines that asgiref is a required package when it in fact is not.

### Expected result

```
$ pipenv graph
django-debug-toolbar==2.2
  - Django [required: >=1.11, installed: 2.2.13]
    - pytz [required: Any, installed: 2020.1]
    - sqlparse [required: Any, installed: 0.3.1]
  - sqlparse [required: >=0.2.0, installed: 0.3.1]
```

### Actual result

```
$ pipenv graph
asgiref==3.2.10
django-debug-toolbar==2.2
  - Django [required: >=1.11, installed: 2.2.13]
    - pytz [required: Any, installed: 2020.1]
    - sqlparse [required: Any, installed: 0.3.1]
  - sqlparse [required: >=0.2.0, installed: 0.3.1]
```

### Steps to replicate

In a new project, run the following commands:

```
$ pipenv install 'Django<3.0'
$ pipenv install --dev django-debug-toolbar
```

-------------------------------------------------------------------------------

<details><summary>$ pipenv --support</summary>

Pipenv version: `'2020.6.2'`

Pipenv location: `'/usr/local/Cellar/pipenv/2020.6.2/libexec/lib/python3.8/site-packages/pipenv'`

Python location: `'/usr/local/Cellar/pipenv/2020.6.2/libexec/bin/python3.8'`

Python installations found:

  - `3.8.0`: `/Users/jmorris/.pyenv/versions/3.8.0/bin/python3`
  - `3.7.7`: `/usr/local/bin/python3`
  - `3.7.7`: `/usr/local/bin/python3.7m`
  - `3.7.7`: `/usr/local/bin/python3.7`
  - `3.6.8`: `/Users/jmorris/.pyenv/versions/3.6.8/bin/python3`
  - `3.5.9`: `/Users/jmorris/.pyenv/versions/3.5.9/bin/python3`
  - `3.5.1`: `/Users/jmorris/.pyenv/versions/3.5.1/bin/python3`
  - `2.7.18`: `/Users/jmorris/.pyenv/versions/2.7.18/bin/python2`
  - `2.7.17`: `/usr/local/bin/python2`
  - `2.7.17`: `/usr/local/bin/python2.7`
  - `2.7.16`: `/usr/bin/python2.7`

PEP 508 Information:

```
[...]

Issue #5169, opened 2022-07-08
Remove vendord cached_property


oz123:
This is built in `functool.cached_property` since Python 3.8 and is redefined in many places we can import from without
fearing it will be removed:
```
$ git grep -r "class cached_property"
pipenv/patched/notpip/_vendor/distlib/util.py:class cached_property(object):
pipenv/patched/notpip/_vendor/distro/distro.py:    class cached_property:  # type: ignore
pipenv/patched/notpip/_vendor/pyparsing/core.py:    class cached_property:
pipenv/vendor/cached_property.py:class cached_property(object):
pipenv/vendor/cached_property.py:class cached_property_with_ttl(object):
pipenv/vendor/distlib/util.py:class cached_property(object):
```

PR #5234, opened 2022-08-08
Issue 4371 incorrect dependencies when install dev packages

dqkqd:
### The issue

I'm trying to fix https://github.com/pypa/pipenv/issues/4371. I think it's better to consider 2 cases below:
- Installing from Pipfile.
- Installing from provided packages (the issue itself).

#### Installing from Pipfile

I read through the code, and here is what it does when installing packages from Pipfile:
1. Resolve `dev-packages` dependencies. Write to lockfile as default.
2. Resolve `packages` dependencies. Write to lockfile as develop.
3. Overwrite duplicated develop packages by default packages.
4. Install default and develop packages in the lockfile.

Because, `dev-packages` is resolved separately, there might be a chance where `dev-packages` is overwritten but its dependencies are not (they aren't duplicated with `packages`). So I think the result dependency graph would be more accurate if `packages` is considered as constraint while resolving `dev-packages` dependencies. Luckily pip provided [constraints files](https://pip.pypa.io/en/stable/user_guide/#constraints-files).

I modified the code to do the followings:
1. Resolve `packages` dependencies. Write to lockfile.
2. Create constraints file from `packages`. Skip editable and extras packages as describe in [pip constraints files](https://pip.pypa.io/en/stable/user_guide/#constraints-files).
3. Resolve `dev-packages` dependencies using constraints file above. Write to lockfile as default.
4. Overwrite duplicated packages. Although `dev-packages` was resolved with constraints, this step is also important because there some cases constraints might be skipped (editable and extras packages).
5. Install resolved packages.

#### Installing from provided packages

When packages are passed as arguments, pipenv does the followings:
1. Install passed packages without resolving dependencies. Write to Pipfile and lockfile.
2. Resolving `dev-packages` and `packages` independently. Modify the lockfile.
3. Overwrite duplicated packages.
4. Install all the packages in the lockfile again. There might be a case it will uninstall the packages from step 1 before installing.

Similar to the first problem, I tried to use default packages as constraints to resolve packages dependencies. I think it is enough to just resolve `dev-packages` at step 1. Because default packages should be installed directly, and resolved packages at step 2 don't need to be resolved again.

### The checklist

* [ ] Associated issue

### DIFF
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -28,6 +28,8 @@ from pipenv.utils.constants import MYPY_RUNNING
 from pipenv.utils.dependencies import (
     convert_deps_to_pip,
     get_canonical_names,
+    has_extras,
+    is_editable,
     is_pinned,
     is_required_version,
     is_star,
@@ -788,6 +790,7 @@ def batch_install(
                 trusted_hosts=trusted_hosts,
                 extra_indexes=extra_indexes,
                 use_pep517=use_pep517,
+                use_constraint=False,  # no need to use constraints, it's written in lockfile
             )
             c.dep = dep
 
@@ -1443,6 +1446,40 @@ def write_requirement_to_file(
     return r
 
 
+def write_constraint_to_file(
+    project: Project,
+    requirements_dir: Optional[str] = None,
+) -> str:
+    if not requirements_dir:
+        requirements_dir = vistir.path.create_tracked_tempdir(
+            prefix="pipenv", suffix="requirements"
+        )
+
+    f = tempfile.NamedTemporaryFile(
+        prefix="pipenv-", suffix="-constraint.txt", dir=requirements_dir, delete=False
+    )
+
+    # duplicated in pipenv.utils.resolver
+    # https://pip.pypa.io/en/stable/user_guide/#constraints-files
+    # constraints must have a name, they cannot be editable, and they cannot specify extras.
+    constraints = {
+        k: v
+        for k, v in project.packages.items()
+        if not is_editable(k) and not is_editable(v) and not has_extras(v)
+    }
+    constraints = convert_deps_to_pip(constraints, project=project, r=False)
+
+    for line in constraints:
+        if project.s.is_verbose():
+            click.echo(
+                f"Writing default constraints to temporary file: {line!r}", err=True
+            )
+        f.write(vistir.misc.to_bytes(line))
+    r = f.name
+    f.close()
+    return r
+
+
 def pip_install(
     project,
     requirement=None,
@@ -1453,12 +1490,14 @@ def pip_install(
     block=True,
     index=None,
     pre=False,
+    dev=False,
     selective_upgrade=False,
     requirements_dir=None,
     extra_indexes=None,
     pypi_mirror=None,
     trusted_hosts=None,
     use_pep517=True,
+    use_constraint=False,
 ):
     piplogger = logging.getLogger("pipenv.patched.pip._internal.commands.install")
     if not trusted_hosts:
@@ -1542,6 +1581,12 @@ def pip_install(
         pip_command.extend(["-r", vistir.path.normalize_path(r)])
     elif line:
         pip_command.extend(line)
+    if dev and use_constraint:
+        constraint_filename = write_constraint_to_file(
+            project,
+            requirements_dir=requirements_dir,
+        )
+        pip_command.extend(["-c", vistir.path.normalize_path(constraint_filename)])
     pip_command.extend(prepare_pip_source_args(sources))
     if project.s.is_verbose():
         click.echo(f"$ {cmd_list_to_shell(pip_command)}", err=True)
@@ -2128,9 +2173,11 @@ def do_install(
                         selective_upgrade=selective_upgrade,
                         no_deps=False,
                         pre=pre,
+                        dev=dev,
                         requirements_dir=requirements_directory,
                         index=index_url,
                         pypi_mirror=pypi_mirror,
+                        use_constraint=True,
                     )
                     if c.returncode:
                         sp.write_err(
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -43,6 +43,7 @@ from pipenv.utils.shell import (
     cmd_list_to_shell,
     find_python,
     is_python_command,
+    normalize_path,
     project_python,
     subprocess_run,
     system_which,
@@ -1578,7 +1579,7 @@ def pip_install(
     )
     pip_command.extend(pip_args)
     if r:
-        pip_command.extend(["-r", vistir.path.normalize_path(r)])
+        pip_command.extend(["-r", normalize_path(r)])
     elif line:
         pip_command.extend(line)
     if dev and use_constraint:
@@ -1586,7 +1587,7 @@ def pip_install(
             project,
             requirements_dir=requirements_dir,
         )
-        pip_command.extend(["-c", vistir.path.normalize_path(constraint_filename)])
+        pip_command.extend(["-c", normalize_path(constraint_filename)])
     pip_command.extend(prepare_pip_source_args(sources))
     if project.s.is_verbose():
         click.echo(f"$ {cmd_list_to_shell(pip_command)}", err=True)
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -1550,7 +1550,7 @@ def pip_install(
     if dev and use_constraint:
         constraint_filename = prepare_default_constraint_file(
             project,
-            dir=requirements_dir,
+            directory=requirements_dir,
         )
         pip_command.extend(["-c", normalize_path(constraint_filename)])
     pip_command.extend(prepare_pip_source_args(sources))
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -28,11 +28,12 @@ from pipenv.utils.constants import MYPY_RUNNING
 from pipenv.utils.dependencies import (
     convert_deps_to_pip,
     get_canonical_names,
+    get_constraints_from_deps,
     is_pinned,
     is_required_version,
     is_star,
     pep423_name,
-    prepare_default_constraint_file,
+    prepare_constraint_file,
     python_version,
 )
 from pipenv.utils.indexes import get_source_list, parse_indexes, prepare_pip_source_args
@@ -1548,9 +1549,12 @@ def pip_install(
     elif line:
         pip_command.extend(line)
     if dev and use_constraint:
-        constraint_filename = prepare_default_constraint_file(
-            project,
+        default_constraints = get_constraints_from_deps(project.packages)
+        constraint_filename = prepare_constraint_file(
+            default_constraints,
             directory=requirements_dir,
+            sources=None,
+            pip_args=None,
         )
         pip_command.extend(["-c", normalize_path(constraint_filename)])
     pip_command.extend(prepare_pip_source_args(sources))
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -28,8 +28,7 @@ from pipenv.utils.constants import MYPY_RUNNING
 from pipenv.utils.dependencies import (
     convert_deps_to_pip,
     get_canonical_names,
-    has_extras,
-    is_editable,
+    get_constraints_from_deps,
     is_pinned,
     is_required_version,
     is_star,
@@ -1460,16 +1459,7 @@ def write_constraint_to_file(
         prefix="pipenv-", suffix="-constraint.txt", dir=requirements_dir, delete=False
     )
 
-    # duplicated in pipenv.utils.resolver
-    # https://pip.pypa.io/en/stable/user_guide/#constraints-files
-    # constraints must have a name, they cannot be editable, and they cannot specify extras.
-    constraints = {
-        k: v
-        for k, v in project.packages.items()
-        if not is_editable(k) and not is_editable(v) and not has_extras(v)
-    }
-    constraints = convert_deps_to_pip(constraints, project=project, r=False)
-
+    constraints = get_constraints_from_deps(project.packages)
     for line in constraints:
         if project.s.is_verbose():
             click.echo(
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -1095,8 +1095,9 @@ def do_lock(
         for k, v in lockfile[section].copy().items():
             if not hasattr(v, "keys"):
                 del lockfile[section][k]
-    # Resolve dev-package dependencies followed by packages dependencies.
-    for is_dev in [True, False]:
+
+    # Resolve package to generate constraints before resolving dev-packages
+    for is_dev in [False, True]:
         pipfile_section = "dev-packages" if is_dev else "packages"
         if project.pipfile_exists:
             packages = project.parsed_pipfile.get(pipfile_section, {})
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -28,11 +28,11 @@ from pipenv.utils.constants import MYPY_RUNNING
 from pipenv.utils.dependencies import (
     convert_deps_to_pip,
     get_canonical_names,
-    get_constraints_from_deps,
     is_pinned,
     is_required_version,
     is_star,
     pep423_name,
+    prepare_default_constraint_file,
     python_version,
 )
 from pipenv.utils.indexes import get_source_list, parse_indexes, prepare_pip_source_args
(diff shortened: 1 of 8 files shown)


pipenv/core.py
```diff
@@ -1541,6 +1581,12 @@ def pip_install(
         pip_command.extend(["-r", vistir.path.normalize_path(r)])
     elif line:
         pip_command.extend(line)
+    if dev and use_constraint:
+        constraint_filename = write_constraint_to_file(
+            project,
+            requirements_dir=requirements_dir,
+        )
+        pip_command.extend(["-c", vistir.path.normalize_path(constraint_filename)])
```
matteius, line 1589: Is there an alternative function that would make sense other than calling something from vistir?  Also, is there a way to do this without writing to a file?

pipenv/core.py
```diff
@@ -1442,6 +1446,40 @@ def write_requirement_to_file(
...
+
+    # duplicated in pipenv.utils.resolver
+    # https://pip.pypa.io/en/stable/user_guide/#constraints-files
+    # constraints must have a name, they cannot be editable, and they cannot specify extras.
+    constraints = {
+        k: v
+        for k, v in project.packages.items()
+        if not is_editable(k) and not is_editable(v) and not has_extras(v)
+    }
+    constraints = convert_deps_to_pip(constraints, project=project, r=False)
```
matteius, line 1470: `convert_deps_to_pip` calls quite a bit of code in `requirementslib`.  I've been studying this code base for 10 months now and I still haven't wrapped my head around all of the resolver/dependency logic, but from spending some time in the debugger last night I know that this method calls a lot of code that we need help unraveling in requirementslib.      I am concerned that this PR may be adding to the amount of logic we have to understand around this, and could possibly be duplicating some code that we should be refactoring elsewhere.   I'll definitely need more time to study this change, but I am still glad you are pushing forward with a solution to the original issue.

tests/integration/test_lock.py
```diff
@@ -791,3 +791,39 @@ def test_pipenv_respects_package_index_restrictions(PipenvInstance):
                                       'sha256:ec22d826a36ed72a7358ff3fe56cbd4ba69dd7a6718ffd450ff0e9df7a47ce6a'],
                            'index': 'local', 'version': '==2.19.1'}
         assert p.lockfile['default']['requests'] == expected_result
+
+
+@flaky
```
matteius, line 796: We shouldn't be adding any new flaky tests.

tests/integration/test_lock.py
```diff
@@ -791,3 +791,39 @@ def test_pipenv_respects_package_index_restrictions(PipenvInstance):
...
+@pytest.mark.dev
+@pytest.mark.lock
+@pytest.mark.install
+def test_dev_lock_use_default_packages_as_constraint(PipenvInstance):
```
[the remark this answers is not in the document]
dqkqd: Actually there are 2 constraints files. The first one is inside resolver.py
https://github.com/pypa/pipenv/blob/30a6b1a5b82f0d04be14274b189467f10be2c3a2/pipenv/utils/resolver.py#L591-L595
The second one is `write_constraint_to_file` as mention above.

- The first constraints file is created and used when: Locking or Installing with Pipfile using `--dev` flag (user don't pass any packages args).
- The second constrains file is created and used when: User pass packages as argument using `--dev` flag.

When locking, the code does the following things:
- Resolve dependencies with `venv_resolve_deps` (this create and use the first constraints file).
- Running `batch_install`, install all packages in the lock file. This one calls `pip_install`, this could create the second constraints file (if both `dev` and `use_constraint` equal True). But since all the packages were resolved in step 1 and written to lockfile, so resolving isn't necessary.
https://github.com/pypa/pipenv/blob/30a6b1a5b82f0d04be14274b189467f10be2c3a2/pipenv/core.py#L780-L793

tests/integration/test_lock.py
dqkqd: I will remove it. Frankly, I don't know what it does, I use it because I saw other tests functions use it as well.

tests/integration/test_lock.py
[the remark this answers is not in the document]
matteius: Thanks @dqkqd -- I was able to hit that breakpoint in the test `test_install_dev_use_default_constraints` after I posted the first message.   Can you explain more why to have two different methods of similar code for generating the two different constraints files?

pipenv/core.py
dqkqd: Thanks for reviewing.

There are similar function to normalize path inside shell.py, I will replace vistir with this.
https://github.com/pypa/pipenv/blob/6c0c451b81fe08e4a11f275931b9952ec1182154/pipenv/utils/shell.py#L98-L101


After reading [constraints files](https://pip.pypa.io/en/stable/user_guide/#constraints-files), I think currently there are no way to pass constraint as argument to pip.

- The constraint file created from `write_to_constrain_file` is passed to subprocess, so it might be hard to not use the file.
https://github.com/pypa/pipenv/blob/30a6b1a5b82f0d04be14274b189467f10be2c3a2/pipenv/core.py#L1609-L1610

- The constraints files create from `default_constraint_file` inside resolver could be removed I think. But this might take time reading pip code base to see how **constraints files** in convert into **constraint line**.

tests/integration/test_lock.py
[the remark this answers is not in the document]
dqkqd: We have `lock` and `install` method. The first constraint file in `resolver.py` is used for `lock`, the second one in `core.py` is used for `install`.
1. The first one should not be removed, because it would cause lockfile is not generated correctly.
2. The second one could be remove. Actually, this is my first approach.
Suppose I remove this one. Then `pip_install` inside this loop should not be called, because it will install packages without resolving dependencies.
https://github.com/pypa/pipenv/blob/30a6b1a5b82f0d04be14274b189467f10be2c3a2/pipenv/core.py#L2136-L2271
So `pip_install` should be replaced by another `check_installable_package` method. After that, resolving all the packages with `do_lock` (which use the first constraints file) will give us the same result.
I didn't use this approach because:
    - `$ pip search` is not working, I could not find a way to check if a package is installable without calling `$ pip install`
    - The loop locks scary.

pipenv/core.py
dqkqd: Sorry, the duplicated code I mentioned is actually myself.
https://github.com/pypa/pipenv/blob/30a6b1a5b82f0d04be14274b189467f10be2c3a2/pipenv/core.py#L1462-L1470
https://github.com/pypa/pipenv/blob/30a6b1a5b82f0d04be14274b189467f10be2c3a2/pipenv/utils/resolver.py#L565-L574

I'm not sure should I write a attribute which generate constraints packages inside `project.py` or not.

tests/integration/test_lock.py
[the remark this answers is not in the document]
dqkqd: Also, there is another case where resolving dependencies inside `pip_install` is necessary.
For example, I have packages installed with this Pipfile:
```toml
[packages]
django = "<3.0"
```
Suppose I install new dev package using command `$ pipenv install --dev "django>3.0,<4.0"`. Clearly the package is installable but it shouldn't be, because it would confliect with default packages.

pipenv/core.py
matteius: I would suggest extracting it from resolver.py to a standalone util, that doesn't depend on a class and re-use it in both places.   Instead of `self.project` it can just take a project as a parameter.  I am not sure if it makes more sense to live in "resolver.py" or "dependencies.py" but dependencies.py already has the methods `convert_deps_to_pip` and `is_editable` so I am leaning towards putting it there, plus its a smaller file than resolver.py which is quite large still--would be good to chip away at the code organization improvements.

dqkqd: Thank you, it seems reasonable to have it inside "dependencies.py". I updated the code.

pipenv/utils/resolver.py
```diff
@@ -552,6 +560,28 @@ def constraint_file(self):
...
+            suffix="-default-constraints.txt",
+            dir=self.req_dir,
+            delete=False,
+        )
+        default_constraints = get_constraints_from_deps(self.project.packages)
+        default_constraints_file.write("\n".join([c for c in default_constraints]))
+        default_constraints_file.close()
+        return default_constraints_file.name
+
+    @property
```
matteius, line 579: Can we make this a `@cached_property` instead so that we don't have to implement that ourselves with `if self._default_constraint_file is None:` logic?

pipenv/utils/resolver.py
```diff
@@ -630,6 +660,37 @@ def parsed_constraints(self):
             )
         return self._parsed_constraints
 
+    @property
```
matteius, line 663: Recommend also making this a @cached_property

pipenv/utils/resolver.py
```diff
@@ -630,6 +660,37 @@ def parsed_constraints(self):
...
+            self._parsed_default_constraints = parse_requirements(
+                self.default_constraint_file,
+                constraint=True,
+                finder=self.finder,
+                session=self.session,
+                options=pip_options,
+            )
+        return self._parsed_default_constraints
+
+    @property
```
matteius, line 677: And this one also I think would be better as a cached_property.

pipenv/utils/resolver.py
```diff
@@ -630,6 +660,37 @@ def parsed_constraints(self):
...
+    def default_constraints(self):
+        from pipenv.patched.pip._internal.req.constructors import (
```
matteius, line 679: Would be better if these imports were at the top of the files -- shouldn't be a circular import in this case.

pipenv/utils/resolver.py
```diff
@@ -646,6 +707,8 @@ def constraints(self):
                 )
                 for c in self.parsed_constraints
             ]
+            if self.dev:
```
matteius, line 710: Could you explain more why we only include the default_constraints when dev is specified?  I don't think I quite understand out of the gate.  I got to looking at this more because this is the only place that uses self.dev, and I am noticing how many parameters `pip_install` takes already.

pipenv/utils/resolver.py
```diff
@@ -552,6 +560,28 @@ def constraint_file(self):
             self._constraint_file = self.prepare_constraint_file()
         return self._constraint_file
 
+    def prepare_default_constraint_file(self):
```
matteius, line 563: This feels like something that could be a standalone utility in dependencies.py that is not method of the class that might be helpful towards future refactors.  something like `def prepare_default_constraint_file(directory, project)`

matteius: Actually, on second thought, this feels very similar to the method `get_constraints_from_deps` -- Perhaps we could extract out the creation of the tracked tempfile and the writing to it of the constraints out to a shared method that takes the directory and the constraints to be written.  Like `def prepare_default_constraint_file(directory, constraints)`.  It could be re-used by `write_constraint_to_file`

pipenv/utils/resolver.py
dqkqd: I think constraint is used only when installing dev-packages. If it does with new normal packages, the installing package could not overwrite existing packages, because it use them as contraint, and therefore might not be installed.

pipenv/utils/resolver.py
dqkqd: Yeah, I will try this. Btw, I might not be able to response to this issue in the next few days. I’m sorry. Will get back on to this as soon as possible.

matteius: It could also be re-used by `prepare_constraint_file` which has similar logic right now.

pipenv/utils/resolver.py
matteius: Ah ok thanks, that makes sense.

matteius:
@dqkqd I left a few more notes about areas I think we could further refine the code and reduce duplication.   I think if we can factor out the three spots we write constraint files into a common utility and some of the places we can use `@cached_property` to allow for simpler and more readable properties.  I think the change makes sense and could be merged, but chipping away at the some more of this type of technical debt would be good.

pipenv/utils/resolver.py
dqkqd: I used ``@cached_property`` but falling some tests on MacOS.

dqkqd: Those tests also failed without ``@cached_property`` as well, so maybe ``@cached_property`` is not the issue. Because ``@cached_property`` does not work with python 3.7, I used ``@property`` and ``@lru_cache`` instead.

pipenv/utils/resolver.py
dqkqd: I added that at the top of the file.

pipenv/utils/resolver.py
matteius: I think the problem is you need to import `@cached_property` from the vendor'd dependencies, go with `from pipenv.vendor.cached_property import cached_property` until https://github.com/pypa/pipenv/issues/5169 is completed.

matteius: The only tests that are expected could fail, and likely on mac OS right now is `test_convert_deps_to_pip` and `test_convert_deps_to_pip_one_way`.  Still investigating how these failures started in the main branch, so if you see that specifically its safe to ignore.  The other tests should all pass.

dqkqd: > I think the problem is you need to import `@cached_property` from the vendor'd dependencies, go with `from pipenv.vendor.cached_property import cached_property` until #5169 is completed.

Thank you, I'll try that.

dqkqd: I ran with the main branch, but there are more failed tests on macOS other than ``test_convert_deps_to_pip`` and ``test_convert_deps_to_pip_one_way``.

matteius: @dqkqd It is possible that the new version of setuptools that was released yesterday has affected things.  I will re-run the  main branch actions, but you can see the last time it ran upon merge ony those `*_depts_to_pip` tests had failed:  https://github.com/pypa/pipenv/runs/7758310446?check_suite_focus=true

matteius: @dqkqd I fixed the unexpected test failures upstream in main branch, in this PR:  https://github.com/pypa/pipenv/pull/5241

I am still not sure why those *_depts_to_pip tests have sporadic failures in the CI and Mac OS, but thats another can of worms.

tests/integration/test_lock.py
matteius: I think it was considered for tests that would sometimes fail to have them re-run, but we expect our tests to pass so its just an odd annotation.

pipenv/utils/dependencies.py
```diff
@@ -298,6 +298,31 @@ def is_constraint(dep):
     return constraints
 
 
+def prepare_default_constraint_file(
+    project,
+    dir=None,
```
matteius, line 303: We should avoid shadowing built in python function name `dir` -- I recommend naming it `directory`.

pipenv/utils/resolver.py
dqkqd: @matteius I updated as you said. Not sure why it passed ``*_depts_to_pip`` cases. I didn't even modify ``convert_deps_to_pip``.

pipenv/utils/dependencies.py
dqkqd: I fixed it.

pipenv/utils/resolver.py
dqkqd: I wrote standalone utility in dependencies.py, but not sure how to use it in ``prepare_constraint_file``. Maybe my implementation is not good. Please tell me if there are any room for improvement.

pipenv/utils/resolver.py
matteius: @dqkqd It is a "flakey" test case, maybe I'll mark it as such and it won't fail the build as often.  I haven't gotten to the bottom of why sometimes it behaves differently, it has failed on Mac OS most frequently, but I have seen it once in a while fail on other systems.

pipenv/utils/resolver.py
matteius: @dqkqd Thanks, it looks great, basically I think you need to add a third parameter which is the `constraints` and have that computed outside the helper utility file, because that is what is different about the three places that write this file.  I think if we can solve for that, it will be ready for merge :-D

matteius: Oh -- and please add a news fragment :-)

dqkqd: I modified it and add news fragment.

matteius (approved):

matteius:
@dqkqd Nice work!

dqkqd:
@matteius Thank you.

PR #5238, opened 2022-08-10
Fix typo: the latest version of pipenv is 2022.8.5

jfly:
I don't know if this file is autogenerated. If it is, then I think
there's a bug somewhere. If it's not, then this was just human error.


matteius:
It was human error, the file gets generated by vendoring scripts which are invoked by humans.

PR #5260, opened 2022-08-15
Remove yaspin and termcolor

oz123:
Termcolor is a dependency of yaspin. Yaspin isn't used
anywhere in pipenv. There is a patch for it, and there is also
vistir dependency for it, but we have our own spinner.

Side note, we can consider using rich, which pip uses.


matteius:
@oz123 yaspin is used by `vistir` and the `run` command we do use, but we always pass `no_spinner=False` so I think this is safe.  I think longer term, lets convert away from `vistir` `run` as well if possible.

matteius (approved):

PR #5268, opened 2022-08-17
Revert "Revert "Remove yaspin and termcolor""

matteius:
Not sure why the CI passes on this, for me on windows and linux I am unable to run the tests because I get `ModuleNotFoundError` whereas vistir catches `ImportError` only.   I think this needs more followup, but this is a placeholder of it.

I really think we need to patch vistir to make this change work better.  I've tried multiple things to get the tests to run locally -- if I switch back to main branch there is no issue.


matteius:
Depends on the new `vistir` release that @oz123 is working on.

oz123:
I found out how pipenv is dependent on Yaspin.. Which is a bit bummer. We have 3 libraries in the code for doing spinner:
Yaspin
Spinner (in pipenv utils)
Rich 

We should first decide which spinner we stick too , and gradually move to it, then we can drop Yaspin and termcolor.
My obvious choice is to use the spinner from rich, which is bundled with pip._vendor.

oz123:
@matteius closing this PR in favor replacement of all spinners.

Issue #5239, opened 2022-08-10
`pipenv lock` nondeterminism with environment markers (again)


jfly:
I reported this bug about pipenv behaving non-deterministically a while ago: https://github.com/pypa/pipenv/issues/4967. [@matteius](https://github.com/pypa/pipenv/issues/4967#issuecomment-1104663451) and [I](https://github.com/pypa/pipenv/issues/4967#issuecomment-1112793695) both thought that it was fixed by upgrading to a vendored pip 22.x, but I just checked, and it looks like the problem is back. Actually, it looks like maybe it never went away!

The setup is the same as before: add `gevent` and `sqlalchemy` to a Pipfile, and observe that repeated `pipenv lock`s result in the different lockfiles.

# Proof/demo

### pipenv 2022.8.5

    $ docker run $(docker build -q --build-arg=PIP_VERSION=22.2.2 --build-arg=PIPENV_VERSION=2022.8.5 https://github.com/jfly/2022-02-22-pipenv-nondeterminism.git#main)
    attempt 1: greenlet markers="platform_python_implementation == 'CPython'"
    attempt 2: greenlet markers="platform_python_implementation == 'CPython'"
    attempt 3: greenlet markers="platform_python_implementation == 'CPython'"
    attempt 4: greenlet markers="python_version >= '3' and platform_machine == 'aarch64' or (platform_machine == 'ppc64le' or (platform_machine == 'x86_64' or (platform_machine == 'amd64' or (platform_machine == 'AMD64' or (platform_machine == 'win32' or platform_machine == 'WIN32')))))"

### pipenv 2022.7.24

    $ docker run $(docker build -q --build-arg=PIP_VERSION=22.2.2 --build-arg=PIPENV_VERSION=2022.7.24 https://github.com/jfly/2022-02-22-pipenv-nondeterminism.git#main)
    attempt 1: greenlet markers="platform_python_implementation == 'CPython'"
    attempt 2: greenlet markers="python_version >= '3' and platform_machine == 'aarch64' or (platform_machine == 'ppc64le' or (platform_machine == 'x86_64' or (platform_machine == 'amd64' or (platform_machine == 'AMD64' or (platform_machine == 'win32' or platform_machine == 'WIN32')))))"
    attempt 3: greenlet markers="platform_python_implementation == 'CPython'"

### pipenv 2022.4.21

[...]

PR #5279, opened 2022-08-22
Fix nondeterminism with environment markers #5239

bakhtiary:
No problem

### The issue

Fixing #2927 5239

### The fix

We were using set to keep track of constraints, this changed their order leading to nondeterministic behaviour.

### The checklist

* [X] Associated issue


matteius (approved):
The change makes sense to me; let's see how the test runner does and give @oz123  a chance to review too.

oz123:
I'm not able to say much about the change, because the tests break. Should the test also be modified?

bakhtiary:
hmmm, let me check the tests. I thought they might flaky, but this seems more than flake.

matteius (changes requested):
Looks like there are some other places in the code that are still calling `.add()` to add a constraint.   The test failures have more details.

bakhtiary:
> I'm not able to say much about the change, because the tests break. Should the test also be modified?

@oz123 I think it is Fixed now. most of the tests are passing, but I am not sure if what I have done is in line with what the algorithm expects.

bakhtiary:
> Looks like there are some other places in the code that are still calling `.add()` to add a constraint. The test failures have more details.

fixed :D. I did actually run the tests, but then pushed when they started taking too long to finish :D.

oz123:
The changes proposed look alright for me. However, I am wondering if ordered-set is a more appropriate solution.
It keeps the set, but adds order like a list.
An ordered set recipe can be floating around, and we can simply add it to our utils.

matteius:
What about a sorted set vs an ordered set so that we get a consistent ordering regardless of the order things are added in?

bakhtiary:
> The changes proposed look alright for me. However, I am wondering if ordered-set is a more appropriate solution.
> It keeps the set, but adds order like a list.
> An ordered set recipe can be floating around, and we can simply add it to our utils.

It is a good idea, specially that then it would maintain the functionality of the code the same. The main draw back of the list is that it can have duplicates and I don't have a clue what that would entail.

The con is that we'd have to add an extra library for the ordered set.

@matteius I don't think sorted set would work.

bakhtiary:
One more option is to use a set and a list together in these 2 places and just check the set before inserting in the list.

matteius:
@bakhtiary Well technically there is a SortedSet implementation that would work, but it would probably be simpler to use "use [collections.OrderedDict](https://docs.python.org/3/library/collections.html) from the Python standard library with just keys (and values as None) for the same purpose."  https://stackoverflow.com/questions/1653970/does-python-have-an-ordered-set

bakhtiary:
But as @oz123 said we can also make our own ordered-set. We can make it using a list and a set or using the dict which is ordered since python 3.6. 

Or we can vendor either the OrderedSet or the ordered-set package.

These are all easy, just tell me which one and I'll get it done in a second :smiley_cat: I personally like the orderedset based on the dict Idea.

matteius:
@bakhtiary I was thinking we could use` collections.OrderedDict` and then just `sorted(d.keys())` without having to add any new code.

bakhtiary:
> @bakhtiary I was thinking we could use` collections.OrderedDict` and then just `sorted(d.keys())` without having to add any new code.

Oh I see. you are right. I believe the orderedDict will give us the keys in the order they were inserted. 

I think it would work. should I do it?

matteius:
@bakhtiary Yeah `OrderedDict` feels like the cleanest solution to me in terms of least amount of memory and least additional code added.   We can probably sort the keys to be safe before we pass them as constraints, seeing as how the ordering upstream might change and I would be surprised if the ordering mattered for the actual correct application of the constraints.  We can debate this point about the sorting -- but what I really don't understand and I think to be true, is if we add a new package the markers for an existing package could completely convert to a different type of marker because technically the set and ordering could and would change when adding new dependencies.  Determinism is good though; consistency will probably be the next thing we will want to figure out in a separate effort.

oz123:
I'm glad I triggered such a good discussion.
The package ordered-set has very unpleasant code to read imo. Compared to the original recipe 
Therehttps://github.com/ActiveState/code/blob/3b27230f418b714bc9a0f897cb8ea189c3515e99/recipes/Python/576696_OrderedSet_with_Weakrefs/README.md.

However, the idea of using a dictionary with keys only as a set replacement is my favorite.
It might not be straight forward for people who read the code in the future so it's worth adding a comment.

Also, we don't need ordereddict. All dictionaries in python 3.6+ are ordered. 

@bakhtiary sorry for sending you for another round of work on this. Thank you for your patience and work on this issue. 🤗

bakhtiary:
I somehow accidentally closed this. I have opened another PR here: https://github.com/pypa/pipenv/pull/5286. feel free to revive this PR and delete the other one if there is a way.

bakhtiary:
@matteius This PR some how was accidentally closed. It was not merged. I opened a new one here: https://github.com/pypa/pipenv/pull/5286

PR #5284, opened 2022-08-23
Convert test runner to use pypiserver package as standalone process

matteius:
* Converts the pipenv test runner to use `pypiserver`.
* Invokes `pypiserver` correctly for all supported OS types 
* Uses `waitress` for Windows only because it reduced the run time of windows test runner by 10 minutes from the default wsgi bottle.
* Does not use `waitress` for Ubuntu because it slowed it down significantly, nor for mac OS which showed no difference using it.
* Refactor test naming and cleanup conftest file to be more certain about temp directory cleanup
* Tableflip on the `dep_to_pip` style tests and realize that there is still some cross tests contamination going on that sometimes confused requirementslib into adding a specifier for requests without a specifier --  I ultimately concluded that I don't have time currently to figure out how to prevent this other than to convert those requests and requests with extras pip_to_deps tests  to use uvicorn as its test subject instead, since it too has extras and we don't use it in other tests that pull in the vcs setup.py metadata artifacts.  We can see on the latest main build this failed again too, so we basically have been getting lucky and this change to `uvicorn` at least gurantees the build won't be flaky anymore for this particular reason:  https://github.com/pypa/pipenv/runs/8179118655?check_suite_focus=true
* Updated dev documentation and the scripts for invoking pytest to startup the pypiserver ahead of time.   
* I did some speed analysis and I feel this is on par with our current test runner, but is way better because we eliminate the fixtures that invoke several threads of an old unsupported pypiserver and instead the tests are CPU lighter now when they run even if they take a similar amount of time.
* This really opens the door to better debugging because it becomes easy to access the same private pypi server that the tests are referencing and check its behavior and the available packages.   The options we can pass to this pypiserver are way more expansive, and likely to improve overtime as that project has a new maintainer that appears interested in accepting contributions.  For example, right now there is no way to specify thread count to the bottle wsgi server, and so for waitress it defaults to 4, but the test runner has 2 cores so this is overkill -- we can probably get in the future an ability to specify this value to experiment with runtimes.
* We are closer than ever with this PR to being able to support private pypi tests that involve credentials :tada:
* Probably other benefits to doing this.

 **Deferred for Future**
* ~Extends the hash lookup method to consider if the index being used supports the same pypi API for retrieving package hashes.~ -- Deferred to future PR since I shouldn't add a functional change to pipenv in this re-work of the testing framework.
* Associated PR for `pypiserver` and when I subbed out feature it shaved almost a 30-60 seconds off my test runner locally:  https://github.com/pypiserver/pypiserver/pull/440 

### Artifacts and Stats
image

image

### First run where this PR refactor was really successful:
image

### Second run where I pushed doc/news changes and let it re-run:
image

###Latest main branch run (for possible comparison)
image

### The checklist


tests/integration/test_install_uri.py
```diff
@@ -72,17 +72,17 @@ def test_ssh_vcs_install(PipenvInstance):
 def test_urls_work(PipenvInstance):
     with PipenvInstance(chdir=True) as p:
         # the library this installs is "django-cms"
-        path = p._pipfile.get_url("django", "3.4.x.zip")
+        url = "https://github.com/lidatong/dataclasses-json/archive/refs/tags/v0.5.7.zip"
```
matteius, line 75: Go with a lighter-weight test and use a real URL (rather than the old pypi server's artifact's URLs feature).

tests/integration/test_lock.py
```diff
@@ -402,64 +401,11 @@ def test_private_index_lock_requirements(PipenvInstance_NoPyPI):
         assert c.returncode == 0
 
 
-@pytest.mark.lock
-@pytest.mark.install
-@pytest.mark.skip_windows
-@pytest.mark.skipif(sys.version_info >= (3, 9), reason="old setuptools doesn't work")
-@pytest.mark.needs_internet
-def test_outdated_setuptools_with_pep517_legacy_build_meta_is_updated(PipenvInstance):
```
matteius, line 410: These tests are the black magic of the devil and old setuptools -- how they ever passed I have no clue and vote we should drop them now.   Just read the comments, it feels like a very old problem it was checking for, and with time we would drop these tests anyway.

oz123:
I believe, though I haven't tested it, that adding gunicorn on Ubuntu and Mac might improve the speed of the tests.
WSGIRef is darn slow, especially when running in parallel

matteius:
> I believe, though I haven't tested it, that adding gunicorn on Ubuntu and Mac might improve the speed of the tests. WSGIRef is darn slow, especially when running in parallel

It might have helped ubuntu but I think it hurt Mac OS run time so I am going to change gunicorn to just be installed for linux.
image

PR #5286, opened 2022-08-23
Fix nondeterminism with environment markers #5239

bakhtiary:
this is a continuation of the following PR:
https://github.com/pypa/pipenv/pull/5279

### The issue

### The fix

### The checklist

* [x] Associated issue

### DIFF
--- a/pipenv/utils/resolver.py
+++ b/pipenv/utils/resolver.py
@@ -5,7 +5,7 @@ import subprocess
 import sys
 import warnings
 from functools import lru_cache
-from typing import Dict, List, Optional, Set, Tuple, Union
+from typing import Dict, List, Optional, Tuple, Union
 
 from pipenv import environments
 from pipenv.exceptions import RequirementError, ResolutionFailure
@@ -86,7 +86,6 @@ def get_package_finder(
 
 
 class HashCacheMixin:
-
     """Caches hashes of PyPI artifacts so we do not need to re-download them.
 
     Hashes are only cached when the URL appears to contain a hash in it and the
@@ -189,12 +188,14 @@ class Resolver:
         pre: bool = False,
         clear: bool = False,
     ) -> Tuple[
-        Set[str],
+        List[str],
         Dict[str, Dict[str, Union[str, bool, List[str]]]],
         Dict[str, str],
         Dict[str, str],
     ]:
-        constraints: Set[str] = set()
+        constraints: Dict[
+            str, None
+        ] = {}  # Used Dict instead of Set because Dict is ordered and is hence stable
         skipped: Dict[str, Dict[str, Union[str, bool, List[str]]]] = {}
         if index_lookup is None:
             index_lookup = {}
@@ -247,9 +248,9 @@ class Resolver:
             constraint_update, lockfile_update = self.get_deps_from_req(
                 req, resolver=transient_resolver, resolve_vcs=project.s.PIPENV_RESOLVE_VCS
             )
-            constraints |= constraint_update
+            constraints.update(constraint_update)
             skipped.update(lockfile_update)
-        return constraints, skipped, index_lookup, markers_lookup
+        return list(constraints.keys()), skipped, index_lookup, markers_lookup
 
     def parse_line(
         self,
@@ -309,7 +310,7 @@ class Resolver:
         req: Requirement,
         resolver: Optional["Resolver"] = None,
         resolve_vcs: bool = True,
-    ) -> Tuple[Set[str], Dict[str, Dict[str, Union[str, bool, List[str]]]]]:
+    ) -> Tuple[Dict[str, None], Dict[str, Dict[str, Union[str, bool, List[str]]]]]:
         from pipenv.vendor.requirementslib.models.requirements import Requirement
         from pipenv.vendor.requirementslib.models.utils import (
             _requirement_to_str_lowercase_name,
@@ -317,7 +318,9 @@ class Resolver:
         from pipenv.vendor.requirementslib.utils import is_installable_dir
 
         # TODO: this is way too complex, refactor this
-        constraints: Set[str] = set()
+        constraints: Dict[
+            str, None
+        ] = {}  # Used Dict instead of Set because Dict is ordered and is hence stable
         locked_deps: Dict[str, Dict[str, Union[str, bool, List[str]]]] = {}
         editable_packages = self.project.get_editable_packages(dev=self.dev)
         if (req.is_file_or_url or req.is_vcs) and not req.is_wheel:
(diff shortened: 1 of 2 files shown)


bakhtiary:
@matteius @oz123 the PR https://github.com/pypa/pipenv/pull/5279 has moved here,

pipenv/utils/resolver.py
```diff
@@ -418,7 +418,7 @@ def get_deps_from_req(
                         err=True,
                     )
                 return constraints, locked_deps
-            constraints.add(req.constraint_line)
+            constraints[req.constraint_line] = None
             return constraints, locked_deps
         return constraints, locked_deps
```
matteius, line 423: Shouldn't this return `constraints.keys(), locked_deps` or whatever calls `.keys()` on the constraints dictionary?

oz123: It depends on how it used down the road I guess. If we are just iterating it, like a list then it is ok. Since, iterating is over keys only.

bakhtiary: @matteius You are right, it would be nicer if we convert this set to a list when we are returning it. This particular function is actually internal to this class, but the function get_metadata is public I will fix this ...

dqkqd:
I think if we want to keep all the packages sorted before putting them into pip's resolver. We can simply sort `self._constraints` instead.
https://github.com/pypa/pipenv/blob/b6d9ad6a63e04998b742bbd167abdec3e4b7a117/pipenv/utils/resolver.py#L632-L647
Like this.
```python
self._constraints.sort(key=lambda ireq: ireq.name)
```
Those sorted `self._constraints` will be put directly into `resolver`.
https://github.com/pypa/pipenv/blob/b6d9ad6a63e04998b742bbd167abdec3e4b7a117/pipenv/utils/resolver.py#L680-L685

matteius:
I think the only thing to decide on left is if we sort the constraints -- I personally think we should sort them where @dqkqd points to the code, because sorting them seems to be the most consistent way we could present them to the resolver.  I suspect there is no rhyme or reason to the order the constraints get added, so sorting also seems to make sense in that regards.

bakhtiary:
@matteius There are some tests failing should I look into them or can we merge this as it is?

matteius:
@bakhtiary These test failures are expected as possible for the time being until they can be figured out why they sometimes fail and usually on mac os:  
```
FAILED tests/unit/test_utils.py::test_convert_deps_to_pip[deps0-requests] - A...
FAILED tests/unit/test_utils.py::test_convert_deps_to_pip[deps1-requests[socks]]
```

The windows failure was an issue to install before the test runner ran, so I kicked off the failed jobs again.

Let's take up the issue of sorting them constraints in a separate issue/PR then.  I suspect we can merge this later today.

dqkqd:
@bakhtiary,  `self._constraints` is actually those packages which we are trying to reorder. So sorting those packages is the same as sorting `self._constraints`. The variable name seems confuse though.

dqkqd:
The issue is not fixed on main branch. Maybe packages's order is changing somewhere.
```bash
   ~/Doc/workspace/pipenv    main !2 ?2 ❯ ./stress-lock.sh
attempt 1: greenlet markers="python_version >= '3' and (platform_machine == 'aarch64' or (platform_machine == 'ppc64le' or (platform_machine == 'x86_64' or (platform_machine == 'amd64' or (platform_machine == 'AMD64' or (platform_machine == 'win32' or platform_machine == 'WIN32'))))))"
attempt 2: greenlet markers="platform_python_implementation == 'CPython'"
attempt 3: greenlet markers="python_version >= '3' and platform_machine == 'aarch64' or (platform_machine == 'ppc64le' or (platform_machine == 'x86_64' or (platform_machine == 'amd64' or (platform_machine == 'AMD64' or (platform_machine == 'win32' or platform_machine == 'WIN32')))))"
attempt 4: greenlet markers="python_version >= '3' and platform_machine == 'aarch64' or (platform_machine == 'ppc64le' or (platform_machine == 'x86_64' or (platform_machine == 'amd64' or (platform_machine == 'AMD64' or (platform_machine == 'win32' or platform_machine == 'WIN32')))))"
attempt 5: greenlet markers="python_version >= '3' and platform_machine == 'aarch64' or (platform_machine == 'ppc64le' or (platform_machine == 'x86_64' or (platform_machine == 'amd64' or (platform_machine == 'AMD64' or (platform_machine == 'win32' or platform_machine == 'WIN32')))))"
attempt 6: greenlet markers="python_version >= '3' and platform_machine == 'aarch64' or (platform_machine == 'ppc64le' or (platform_machine == 'x86_64' or (platform_machine == 'amd64' or (platform_machine == 'AMD64' or (platform_machine == 'win32' or platform_machine == 'WIN32')))))"
attempt 7: greenlet markers="python_version >= '3' and platform_machine == 'aarch64' or (platform_machine == 'ppc64le' or (platform_machine == 'x86_64' or (platform_machine == 'amd64' or (platform_machine == 'AMD64' or (platform_machine == 'win32' or platform_machine == 'WIN32')))))"
attempt 8: greenlet markers="platform_python_implementation == 'CPython'"
attempt 9: greenlet markers="platform_python_implementation == 'CPython'"
```

pipenv/utils/resolver.py
```diff
@@ -189,12 +188,14 @@ def get_metadata(
...
-        Set[str],
+        List[str],
         Dict[str, Dict[str, Union[str, bool, List[str]]]],
         Dict[str, str],
         Dict[str, str],
     ]:
-        constraints: Set[str] = set()
+        constraints: Dict[
+            str, None
+        ] = {}  # Used Dict instead of Set because Dict is ordered and is hence stable
```
cclauss, line 198: Why not a [frozenset](https://docs.python.org/3/library/stdtypes.html#frozenset)?

pipenv/utils/resolver.py
```diff
@@ -247,9 +248,9 @@ def get_metadata(
             constraint_update, lockfile_update = self.get_deps_from_req(
                 req, resolver=transient_resolver, resolve_vcs=project.s.PIPENV_RESOLVE_VCS
             )
-            constraints |= constraint_update
+            constraints.update(constraint_update)
             skipped.update(lockfile_update)
-        return constraints, skipped, index_lookup, markers_lookup
+        return list(constraints.keys()), skipped, index_lookup, markers_lookup
```
cclauss, line 253: `.keys()` is not required here.

pipenv/utils/resolver.py
```diff
@@ -309,15 +310,17 @@ def get_deps_from_req(
...
         from pipenv.vendor.requirementslib.models.utils import (
             _requirement_to_str_lowercase_name,
         )
         from pipenv.vendor.requirementslib.utils import is_installable_dir
 
         # TODO: this is way too complex, refactor this
-        constraints: Set[str] = set()
+        constraints: Dict[
+            str, None
+        ] = {}  # Used Dict instead of Set because Dict is ordered and is hence stable
```
cclauss, line 323: Why not a [frozenset](https://docs.python.org/3/library/stdtypes.html#frozenset)?

pipenv/utils/resolver.py
cclauss: `list[constraints]` is an list while `constraints.keys()` is an iterator.

dqkqd:
As I mention in https://github.com/pypa/pipenv/pull/5299#issuecomment-1228731246. I think this should be revert because it does not solve linked issue.

PR #5228, opened 2022-08-07
Vendor in `requirementslib==2.0.0` & drop `pip-shims` & revert sysconfig patch w/ safer setuptools usage

matteius:
Vendor in requirementslib==2.0.0 and drop pip-shims entirely.

Some systems have patched sysconfig differently, and from this discussion (https://github.com/pypa/setuptools/discussions/3544) it would appear that importing setuptools packages is problematic

PR #5289, opened 2022-08-23
issue-5285 Remove eager and unnecessary importing of setuptools and pkg_resources

matteius:

Will require a requirementslib change to actually roll this out, but this is a branch that prototypes fixing #5285 and #5290  
The underlying reason why we patched Pip to use sysconfig by default for all python versions (https://github.com/pypa/pipenv/pull/5275/files#diff-e154923a3c93acd48d40dd714127c4b90cac776ebd71930770919d539d137f77R44) ~though that appears to be compatible with this fix and so far that patch isn't causing a specific problem so I lean towards leaving it in for now.~  Actually #5290 came in right after I wrote that, so now I am in favor of dropping that patch with this change.

### The issue

Some systems have patched sysconfig differently, and from this discussion (https://github.com/pypa/setuptools/discussions/3544) it would appear that importing setuptools packages is problematic


matteius:
Replaced by:  https://github.com/pypa/pipenv/pull/5228

PR #5291, opened 2022-08-23
Downgrade pyparsing -- testing for the sporadic failing deps_to_pip tests

matteius:
I began to wonder if upgrading `pyparsing` isn't what set off the sporadic test failure on those two deps_to_pip tests.  The reason I wonder about this is the `requirementslib` build never did pass when I unpinned `pyparsing`:  https://github.com/sarugaku/requirementslib/pull/327

Also, when I re-ran vendoring the first time it reverted the `zipp` import and so I fixed that in the vendoring script + the old version of pyparsing we used doesn't have cached_property so I adjusted that part of vendoring too.

### DIFF
--- a/pipenv/vendor/pyparsing/__init__.py
+++ b/pipenv/vendor/pyparsing/__init__.py
@@ -1,6 +1,6 @@
 # module pyparsing.py
 #
-# Copyright (c) 2003-2022  Paul T. McGuire
+# Copyright (c) 2003-2021  Paul T. McGuire
 #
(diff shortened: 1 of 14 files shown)


matteius:
Well that ended up not being the issue -- I'll patch the zipp thing next time I do vendoring which will probably be soon with requirementslib.

matteius:
Ok for sure now we know that `pyparsing` isn't the culprit for that failing test:  https://github.com/pypa/pipenv/runs/7984979349?check_suite_focus=true

Issue #5296, opened 2022-08-25
pipenv install --system fails when `python` points to Python 2.


enku:
### Issue description

Since 2022.8.15 `pipenv install --system` will fail on systems that have both Python 2 and Python 3 installed but `which python` resolves to Python 2.  In commit dbea3f5 it switched from using `project._which()` which uses `sys.executable` to determine which python to run to using `python_project()` which, when using `--system` first checks for `python` and then `python3`.  So if, e.g. your project/system is using Python 3 which is installed as `/usr/local/bin/python3` but your system Python package us Python 2 which is at `/usr/bin/python` then it will try to execute using Python 2, resulting in

```
[pipenv.exceptions.InstallError]: File "/git-test/pipenv/patched/pip/__pip-runner__.py", line 21
[pipenv.exceptions.InstallError]:     fullname: str,
[pipenv.exceptions.InstallError]:             ^
[pipenv.exceptions.InstallError]: SyntaxError: invalid syntax
```

### Expected result

Expected result is `pipenv install --system` installs the dependencies using Python 3.  Although I'm not sure it's really a bug since technically it is using the "system" Python.  Nevertheless it is a regression.

### Actual result

```
# python3 -m pipenv install  --system --verbose
Installing dependencies from Pipfile.lock (89d195)...
Writing supplied requirement line to temporary file: "dataclasses==0.8; python_version < '3.7' --hash=sha256:8479067f342acf957dc82ec415d355ab5edb7e7646b90dc6e2fd1d96ad084c97 --hash=sha256:0201d89fa866f68c8ebd9d08ee6ff50c0b255f8ec63a71c16fda7af82bb887bf"
Installing 'dataclasses'
$ /usr/bin/python /usr/local/lib/python3.9/site-packages/pipenv/patched/pip/__pip-runner__.py install --verbose --upgrade --require-hashes --no-deps --exists-action=i -r /tmp/pipenv-v0jrsbzs-requirements/pipenv-_x_y9mvb-requirement.txt
Writing supplied requirement line to temporary file: "importlib-metadata==4.8.2; python_version < '3.8' --hash=sha256:53ccfd5c134223e497627b9815d5030edf77d2ed573922f7a0b8f8bb81a1c100 --hash=sha256:75bdec14c397f528724c1bfd9709d660b33a4d2e77387a3358f20b848bb5e5fb"
Installing 'importlib-metadata'
[...]

PR #5300, opened 2022-08-26
Fix: try python3 before python in `install --system` (#5296)

enku:
Fix a regression from commit dbea3f5 where `pip install --system` tries
to install using the `python` executable first which, on some systems
may point to Python 2.  Instead try `python3` first.

### The issue

As explained in #5296 the is a regression whereby `pipenv install --system` will try `$(which python)` if it exists instead of `$(which python3)` and on some systems `python` points to Python 2, which will fail with `SyntaxError`.

### The fix

The proposed fix is to try `python3` first, then `python` if `python3` fails.

### The checklist

* [x] Associated issue

### DIFF
--- a/news/5296.bugfix.rst
+++ b/news/5296.bugfix.rst
@@ -1,0 +1,1 @@
+Fix an issue when using ``pipenv install --system`` on systems that having the ``python`` executable pointing to Python 2 and a Python 3 executable being ``python3``.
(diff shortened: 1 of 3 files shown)


enku:
I think the failed checks were the flaky tests that @matteius spoke of.

matteius (approved):

oz123 (approved):

Issue #2207, opened 2018-05-16
Checking if installed packages are up to date is slow and use lots of CPU


Tenzer:
I've been trying to migrate a project to use Pipenv but I'm slightly blocked on how much longer it takes for Pipenv to check if the installed dependencies are up to date, compared to pointing `pip install` at a requirements file.

In our setup we run tests inside Docker containers. The image we run the tests on is one that comes pre-installed with the dependencies our project has at the time the image is built. Before we run any tests we then make sure the dependencies are up to date, in case any new dependencies are needed for any new code/tests that might have been added. For this we have just been using `pip install -r requirements.txt`, which normally completes in around 30 seconds when there's no new dependencies to install.

I then tried to switch this to Pipenv and pre-installed the dependences in the image using a `Pipfile` and `Pipfile.lock` and then running `pipenv install --deploy --dev --system` against the files. That works fine and I got an image created, but the problem comes to when we want to run tests and want to check if dependencies are up to date first. I've done this using the same `pipenv install --deploy --dev --system` command and instead of 30 seconds it now takes 5 minutes and 30 seconds! On top of that the CPU usage it much, much higher.

I've made a small test with the `Pipfile` and `Pipfile.lock` we are using (only slightly modified): https://github.com/Tenzer/pipenv-test. Some simple tests that can be run with it is for instance to first install the dependencies and then afterwards check that they are up to date in the local environment, than then see how long time and CPU the second operation takes:

```
$ docker run -it --rm -v $(pwd):/test python bash
root@9f6ecaf12cf8:/# cd /test
root@9f6ecaf12cf8:/test# pip install pipenv
[...]
root@9f6ecaf12cf8:/test# pipenv install --deploy --dev --system
Installing dependencies from Pipfile.lock (f4e26d)…
Ignoring appnope: markers 'sys_platform == "darwin"' don't match your environment
  🐍   ▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉ 212/212 — 00:02:25
root@9f6ecaf12cf8:/test# time pipenv install --deploy --dev --system
Installing dependencies from Pipfile.lock (f4e26d)…
Ignoring appnope: markers 'sys_platform == "darwin"' don't match your environment
[...]

PR #5301, opened 2022-08-27
Reduce the amount of calls to pip and the number of tempfiles in batch_install.

matteius:
This is a performance oriented PR -- the more packages you have to install, the better this change is.  The one possible downside to this change is it becomes necessary to drop the progress bar indicator as the installs are performed by `pip` in up to two large batches (packages with hashes, and packages without hashes).  I did some work to ensure that the verbose flag prints the pip output progress in real time such that progress can still be observed, though it is not the traditional `pipenv` progress bar, that gets dropped.

When doing a regular `pipenv sync --dev` of the pipenv project when the current dependencies are already installed, the running time is about equal between `main` and `explore-batch-install` to re-install the four editable dependencies.

0m34.686s for `explore-batch-install` and real    0m33.596s on `main` for example

However it seems to really shine when installing a lot of dependencies for the first time.

## main `pipenv --rm && time pipenv sync --dev`
```

real    1m17.746s
user    0m0.000s
sys     0m0.015s
```

## explore-batch-install `pipenv --rm && time pipenv sync --dev`
```
$ pipenv --rm && time pipenv sync --dev
Removing virtualenv (\c\Users\matte\.virtualenvs\pipenv-ZKcqGg2F)...
Pipfile: C:\Users\matte\Projects\pipenv\Pipfile
Using default python from C:\Program Files\WindowsApps\PythonSoftwareFoundation.Python.3.10_3.10.1520.0_x64__qbz5n2kfra8p0\python.exe (3.10.5) to create virtualenv...
[ ===] Creating virtual environment...created virtual environment CPython3.10.5.final.0-64 in 398ms
  creator CPython3Windows(dest=C:\c\users\matte\.virtualenvs\pipenv-ZKcqGg2F, clear=False, no_vcs_ignore=False, global=False)
  seeder FromAppData(download=False, pip=bundle, setuptools=bundle, wheel=bundle, via=copy, app_data_dir=C:\Users\matte\AppData\Local\pypa\virtualenv)
    added seed packages: pip==22.2.2, setuptools==63.2.0, wheel==0.37.1
  activators BashActivator,BatchActivator,FishActivator,NushellActivator,PowerShellActivator,PythonActivator

Successfully created virtual environment!
Virtualenv location: \c\Users\matte\.virtualenvs\pipenv-ZKcqGg2F
Installing dependencies from Pipfile.lock (a4c141)...

real    1m13.979s
user    0m0.030s
sys     0m0.046s

```

## For the sentry benchmark test:  main branch:
```
matteius@matteius-VirtualBox:~/pipenv-triage/sentry$ pipenv --rm
Removing virtualenv (/home/matteius/.virtualenvs/sentry-LGSyi0g8)...
matteius@matteius-VirtualBox:~/pipenv-triage/sentry$ rm Pipfile
matteius@matteius-VirtualBox:~/pipenv-triage/sentry$ rm Pipfile.lock 
matteius@matteius-VirtualBox:~/pipenv-triage/sentry$ time pipenv install -r requirements-base.txt 
Pipfile: /home/matteius/pipenv-triage/sentry/Pipfile
Using /home/matteius/.pyenv/versions/3.10.2/bin/python3 (3.10.2) to create virtualenv...
⠧ Creating virtual environment...created virtual environment CPython3.10.2.final.0-64 in 191ms
  creator CPython3Posix(dest=/home/matteius/.virtualenvs/sentry-LGSyi0g8, clear=False, no_vcs_ignore=False, global=False)
  seeder FromAppData(download=False, pip=bundle, setuptools=bundle, wheel=bundle, via=copy, app_data_dir=/home/matteius/.local/share/virtualenv)
    added seed packages: pip==22.2.2, setuptools==63.2.0, wheel==0.37.1
  activators BashActivator,CShellActivator,FishActivator,NushellActivator,PowerShellActivator,PythonActivator

Virtualenv location: /home/matteius/.virtualenvs/sentry-LGSyi0g8
Requirements file provided! Importing into Pipfile...
Building requirements...
Resolving dependencies...
✔ Success! 
Updated Pipfile.lock (03a7b9)!
Installing dependencies from Pipfile.lock (03a7b9)...
  🐍   ▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉ 123/123 — 00:03:36

real	4m39.155s
user	7m19.092s
sys	0m21.095s
matteius@matteius-VirtualBox:~/pipenv-triage/sentry$ pipenv --version
pipenv, version 2022.8.24
```

## explore-batch-install (this development branch)
```
matteius@matteius-VirtualBox:~/pipenv-triage/sentry$ pipenv --rm
Removing virtualenv (/home/matteius/.virtualenvs/sentry-LGSyi0g8)...
matteius@matteius-VirtualBox:~/pipenv-triage/sentry$ rm Pipfile.lock 
matteius@matteius-VirtualBox:~/pipenv-triage/sentry$ rm Pipfile 
matteius@matteius-VirtualBox:~/pipenv-triage/sentry$ pi^C
matteius@matteius-VirtualBox:~/pipenv-triage/sentry$ time pipenv install -r requirements-base.txt 
Pipfile: /home/matteius/pipenv-triage/sentry/Pipfile
Using /home/matteius/.pyenv/versions/3.10.2/bin/python3 (3.10.2) to create virtualenv...
⠦ Creating virtual environment...created virtual environment CPython3.10.2.final.0-64 in 170ms
  creator CPython3Posix(dest=/home/matteius/.virtualenvs/sentry-LGSyi0g8, clear=False, no_vcs_ignore=False, global=False)
  seeder FromAppData(download=False, pip=bundle, setuptools=bundle, wheel=bundle, via=copy, app_data_dir=/home/matteius/.local/share/virtualenv)
    added seed packages: pip==22.2.2, setuptools==63.2.0, wheel==0.37.1
  activators BashActivator,CShellActivator,FishActivator,NushellActivator,PowerShellActivator,PythonActivator

Virtualenv location: /home/matteius/.virtualenvs/sentry-LGSyi0g8
Requirements file provided! Importing into Pipfile...
Building requirements...
Resolving dependencies...
✔ Success! 
Updated Pipfile.lock (03a7b9)!
Installing dependencies from Pipfile.lock (03a7b9)...
  🐍   ▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉▉ 123/123 — 00:00:00

real	1m53.793s
user	1m33.928s
sys	0m7.316s
matteius@matteius-VirtualBox:~/pipenv-triage/sentry$ pipenv --version
pipenv, version 2022.8.25.dev0
```

# Sentry Benchmark -- just the sync
```
matteius@matteius-VirtualBox:~/pipenv$ git checkout main
matteius@matteius-VirtualBox:~/pipenv$ cd ~/pipenv-triage/sentry
matteius@matteius-VirtualBox:~/pipenv-triage/sentry$ pipenv --rm
Removing virtualenv (/home/matteius/.virtualenvs/sentry-LGSyi0g8)...
matteius@matteius-VirtualBox:~/pipenv-triage/sentry$ time pipenv sync
Pipfile: /home/matteius/pipenv-triage/sentry/Pipfile
Using /home/matteius/.pyenv/versions/3.10.2/bin/python3 (3.10.2) to create virtualenv...
created virtual environment CPython3.10.2.final.0-64 in 185ms
  creator CPython3Posix(dest=/home/matteius/.virtualenvs/sentry-LGSyi0g8, clear=False, no_vcs_ignore=False, global=False)
  seeder FromAppData(download=False, pip=bundle, setuptools=bundle, wheel=bundle, via=copy, app_data_dir=/home/matteius/.local/share/virtualenv)
    added seed packages: pip==22.2.2, setuptools==63.2.0, wheel==0.37.1
  activators BashActivator,CShellActivator,FishActivator,NushellActivator,PowerShellActivator,PythonActivator

Successfully created virtual environment!
Virtualenv location: /home/matteius/.virtualenvs/sentry-LGSyi0g8
Installing dependencies from Pipfile.lock (03a7b9)...

real	3m56.202s
user	6m54.377s
sys	0m30.768s
matteius@matteius-VirtualBox:~/pipenv-triage/sentry$ cd ~/pipenv
matteius@matteius-VirtualBox:~/pipenv$ git checkout explore-batch_install
Switched to branch 'explore-batch_install'
Your branch is up to date with 'origin/explore-batch_install'.
matteius@matteius-VirtualBox:~/pipenv$ cd ~/pipenv-triage/sentry
matteius@matteius-VirtualBox:~/pipenv-triage/sentry$ pipenv --rm
Removing virtualenv (/home/matteius/.virtualenvs/sentry-LGSyi0g8)...
^[[Amatteius@matteius-VirtualBox:~/pipenv-triage/sentry$ time pipenv sync
Pipfile: /home/matteius/pipenv-triage/sentry/Pipfile
Using /home/matteius/.pyenv/versions/3.10.2/bin/python3 (3.10.2) to create virtualenv...
created virtual environment CPython3.10.2.final.0-64 in 180ms
  creator CPython3Posix(dest=/home/matteius/.virtualenvs/sentry-LGSyi0g8, clear=False, no_vcs_ignore=False, global=False)
  seeder FromAppData(download=False, pip=bundle, setuptools=bundle, wheel=bundle, via=copy, app_data_dir=/home/matteius/.local/share/virtualenv)
    added seed packages: pip==22.2.2, setuptools==63.2.0, wheel==0.37.1
  activators BashActivator,CShellActivator,FishActivator,NushellActivator,PowerShellActivator,PythonActivator

Successfully created virtual environment!
Virtualenv location: /home/matteius/.virtualenvs/sentry-LGSyi0g8
Installing dependencies from Pipfile.lock (03a7b9)...

real	1m7.600s
user	1m9.126s
sys	0m6.198s
```

Based on an estimate of this running at ~30% of the time of the main branch, I estimate our mean times for the pipenv install benchmark would wind up 45-55 seconds; it would be really cool to see!


tests/integration/test_install_basic.py
```diff
@@ -316,13 +316,10 @@ def test_skip_requirements_when_pipfile(PipenvInstance):
             contents = """
 [packages]
 six = "*"
-fake_package = "<0.12"
             """.strip()
             f.write(contents)
         c = p.pipenv("install")
         assert c.returncode == 0
-        assert "fake_package" in p.pipfile["packages"]
```
matteius, line 324: Fake package has been problematic in many different refactors, and it isn't relevant to this test case anyway.   I've made an effort to replace those other problem usages with a different relevant package for testing as well.

tests/integration/test_install_markers.py
```diff
@@ -37,14 +36,7 @@ def test_platform_python_implementation_marker(PipenvInstance):
...
     with PipenvInstance() as p:
-        with open(p.pipfile_path, 'w') as f:
-            contents = """
-[packages]
-depends-on-marked-package = "*"
-            """.strip()
-            f.write(contents)
-
-        c = p.pipenv('install')
+        c = p.pipenv('install depends-on-marked-package')
```
matteius, line 39: This was problematic because this only exists in our local pypi, but that Pipfile doesn't have the sources -- not sure why this hadn't cropped up before now but I think it has to do with I ported the intended fix for the searching of all indexes to this PR: https://github.com/pypa/pipenv/pull/5220 as well since it started as a copy of `pip_install` which had this bug.

tests/integration/test_lock.py
```diff
@@ -376,13 +376,11 @@ def test_private_index_lock_requirements(PipenvInstance_NoPyPI):
 @pytest.mark.install  # private indexes need to be uncached for resolution
 @pytest.mark.requirements
 @pytest.mark.needs_internet
-def test_private_index_mirror_lock_requirements(PipenvInstance_NoPyPI):
+def test_private_index_lock_requirements(PipenvInstance_NoPyPI):
     # Don't use the local fake pypi
     with temp_environ(), PipenvInstance_NoPyPI(chdir=True) as p:
         # Using pypi.python.org as pipenv-test-public-package is not
         # included in the local pypi mirror
-        mirror_url = os.environ.pop('PIPENV_TEST_INDEX', "https://pypi.kennethreitz.org/simple")
```
matteius, line 384: This mirror is dead in all actuality, and was giving problems in a couple recent refactors.

tests/integration/test_sync.py
```diff
@@ -94,25 +94,6 @@ def test_sync_sequential_detect_errors(PipenvInstance):
         assert c.returncode != 0
 
 
-@pytest.mark.sync
-@pytest.mark.lock
-def test_sync_sequential_verbose(PipenvInstance):
```
matteius, line 99: The verbose output is driven by the output of pip, no sense having a test depend on what it or what it did before since pip is subject to future upgrades.

pipenv/core.py
```diff
@@ -1574,6 +1548,166 @@ def pip_install(
     return c
 
 
+def pip_install_deps(
```
matteius, line 1551: This started as a copy of `pip_install` and as modified to take a list of dependencies instead of a single package.

pipenv/core.py
```diff
@@ -1574,6 +1548,166 @@ def pip_install(
...
+            "PIP_SRC", os.getenv("PIP_SRC_DIR", project.virtualenv_src_location)
+        )
+    else:
+        src_dir = os.getenv("PIP_SRC", os.getenv("PIP_SRC_DIR"))
+    if not requirements_dir:
+        requirements_dir = vistir.path.create_tracked_tempdir(
+            prefix="pipenv", suffix="requirements"
+        )
+
+    hashed_requirements = tempfile.NamedTemporaryFile(
```
matteius, line 1579: From the pip documentation: https://pip.pypa.io/en/stable/topics/secure-installs/#hash-checking-mode
> Note that hash-checking is an all-or-nothing proposition. Specifying --hash against any requirement will activate this mode globally.
We have two types of requirements -- those normal once with hashes, and the editable, vcs, local types that do not and it is not possible to do this in a single call to pip, but we can do it in just two!

pipenv/core.py
```diff
@@ -1574,6 +1548,166 @@ def pip_install(
...
+        )
+        sources = get_source_list(
+            project,
+            requirement.index,
+            extra_indexes=list(project.sources),
+            trusted_hosts=trusted_hosts,
+            pypi_mirror=pypi_mirror,
+        )
+        source_names = {src.get("name") for src in sources}
+        if not search_all_sources and requirement.index in source_names:
```
matteius, line 1609: This includes the fix from that other PR I referenced around package index restrictions and whether or not to search all sources.

pipenv/core.py
```diff
@@ -1574,6 +1548,166 @@ def pip_install(
...
+            sources = list(filter(lambda d: d.get("name") == requirement.index, sources))
+        if sources and not vcs_or_editable:
```
matteius, line 1611: This bit me the first go round because vcs and editable dependencies need to not specify the index, because they won't get resolved.

pipenv/core.py
```diff
@@ -1574,6 +1548,166 @@ def pip_install(
...
+        c = subprocess_run(pip_command, block=False, capture_output=True, env=pip_config)
+        if file == hashed_requirements:
+            c.deps = hashed_deps
+        else:
+            c.deps = not_hashed_deps
+        c.env = pip_config
+        cmds.append(c)
+        while True:
+            line = c.stdout.readline()
+            if line == "":
```
matteius, line 1704: This was really problematic when the pip logger was overriden or verbose was passed to pip, we never recieved the EOF and no matter what the process.poll() or return_code are never available, as if the process doesn't get cleaned up until the cleanup routine.   Though the cleanup routine uses `p.communicate()` which doesn't get the output until the process returns at the end so there was no indicator of progress without this logic here.   This gives real time progress logic when verbose is passed.

pipenv/core.py
```diff
@@ -1634,26 +1768,6 @@ def fallback_which(command, location=None, allow_global=False, system=False):
     return ""
 
 
-def which_pip(project):
```
matteius, line 1637: We stopped using this method in the prior effort to only reference the internal vendor'd `pip`.

matteius:
@dqkqd I would greatly appreciate your review of this as well if you have the time to do so.

dqkqd:
Yes, I would love to.

dqkqd:
There are some [global options](https://pip.pypa.io/en/stable/reference/requirements-file-format/#global-options) used by `pip install`. So I think dividing packages into `hash` and `no-hash` group is not enough.

matteius:
> There are some [global options](https://pip.pypa.io/en/stable/reference/requirements-file-format/#global-options) used by `pip install`. So I think dividing packages into `hash` and `no-hash` group is not enough.

@dqkqd many of the global options though are per line such as the --hash or the --index -- for example "--require-hashes
Require a hash to check each requirement against, for repeatable installs. This option is implied when any package in a requirements file has a --hash option." -- but the hash arguments are different per line.   Index are also different per liine, and the constraints file would apply the same.   Which option are you worried about that would conflict within the same file?

matteius:
@dqkqd Hmm, you might be right about the indexes though -- I think we may not even need them in the requirements.txt the way it is now, because the install command also gets the index and extra-index arguments:
```
Installing package: torch
Writing supplied requirement line to temporary file: 'torch'
Installing 'torch'
$ /c/Users/matte/.virtualenvs/pytorch-1gDgs60Q/Scripts/python.exe 'C:\Users\matte\Projects\pipenv\pipenv\patched\pip\__pip-runner__.py' install --no-input --verbose --upgrade --exists-action=i -r 'c
:\users\matte\appdata\local\temp\pipenv-njscw9uc-requirements\pipenv-lglrxyzh-requirement.txt' -i https://download.pytorch.org/whl/cu111 --extra-index-url https://pypi.org/simple --extra-index-url h
ttps://pypi.org/simple
```

dqkqd:
So every packages with the same index should go into a same batch. I think other global options also work that way as well.
For example if we divide packages using index, editable, hashes. We would need at most 8 batches to cover all edge cases.

dqkqd:
> @dqkqd Hmm, you might be right about the indexes though -- I think we may not even need them in the requirements.txt the way it is now, because the install command also gets the index and extra-index arguments:

If we don't use index-url, `pip` will use the default one, and break some cases where user wants to use their private pypi. So I think we must not discard it.

matteius:
> If we don't use index-url, `pip` will use the default one, and break some cases where user wants to use their private pypi. So I think we must not discard it.

There was a prior bug I discovered from my conversation with you and from testing with my pytorch + pypi example, I changed it to be how the regular `pip_install` works so we don't have to manage so many requirements files with indexes.   Basically it passes in the extra-index-urls and the hash restrictions ensure you get the exact package you are looking for.
```
$ /c/Users/matte/.virtualenvs/pytorch-1gDgs60Q/Scripts/python.exe 'C:\Users\matte\Projects\pipenv\pipenv\patched\pip\__pip-runner__.py' install -i https://pypi.org/simple --extra-index-url
https://download.pytorch.org/whl/cu111 --extra-index-url https://download.pytorch.org/whl/cu113 --no-input --upgrade --no-deps --exists-action=i -r 'c:\users\matte\appdata\local\temp\pipenv
-ijg8hyy6-requirements\pipenv-1e5k9dw4-hashed-reqs.txt'
```

This latest commit solved the issue I was having installing with this 2-index example, plus it is inline with what `pip_install` is already doing.  So I think we can still have just two:  the hashed and not-hashed.   

`--pre` releases is another example but currently that is an all or nothing flag (so not an issue), though there have been requests to be able to specify --pre on a per package basis ... I am not sure if that is a good idea or not, when it is possible to install a specific pre-release with == specifier.  We'll cross that bridge when we revisit if it makes sense to support per package pre releases though.

dqkqd:
So packages with hashes and ``extra-index-url`` should be correctly installed. Then we just need to care about packages without hashes.

> This latest commit solved the issue I was having installing with this 2-index example, plus it is inline with what pip_install is already doing. So I think we can still have just two: the hashed and not-hashed.

If `pip_install` have package and ``index-url`` in the same line, I think it might be a bug. `pip` doesn't grab ``index-url`` if they are same line. I tested it using 2 requirements file.

**Dockerfile**
```Dockerfile
FROM archlinux
ENV LANG C.UTF-8
RUN pacman -Syu python3 python-pip --noconfirm
COPY requirements.txt .
RUN python -m pip install -r requirements.txt -v -v &> result.txt
RUN cat result.txt | grep "Found index url"
```

**Requirements file with index in the same line:**

```bash
# requirements.txt
pytest -i https://pypi.tuna.tsinghua.edu.cn/simple 
```
result
```
Found index url https://pypi.org/simple
Found index url https://pypi.org/simple
Found index url https://pypi.org/simple
Found index url https://pypi.org/simple
Found index url https://pypi.org/simple
Found index url https://pypi.org/simple
```

**Requirements file with index in different line:**

```bash
# requirements.txt
-i https://pypi.tuna.tsinghua.edu.cn/simple 
pytest
```
result:
```
Found index url https://pypi.tuna.tsinghua.edu.cn/simple
Found index url https://pypi.tuna.tsinghua.edu.cn/simple
Found index url https://pypi.tuna.tsinghua.edu.cn/simple
Found index url https://pypi.tuna.tsinghua.edu.cn/simple
Found index url https://pypi.tuna.tsinghua.edu.cn/simple
Found index url https://pypi.org/simple
```

matteius:
> If pip_install have package and index-url in the same line, I think it might be a bug. pip doesn't grab index-url if they are same line. I tested it using 2 requirements file.

@dqkqd Thanks for the insights -- `pip_install` does not do what I tried doing with adding the `-i <index_url>` to the requirements.txt -- which as you describe won't work the way I intended.  Both `pip_install` and `pip_install` and `pip_install_deps` both provide the same mechanism to provide the sources during install, so we should have parity with what it was doing before for that with the current changes.

dqkqd:
Thanks for explaining, I read doc and code again. You were right.
- I were wrong from the start when mention about global options of pip requirements file, we are installing from Pipfile, not from pip requirements.
- All sources from Pipfile will be use and pass into `pip_command`, at first I thought we only use source which we define for each packages.

There are somethings I noticed during experiments, not sure they are helpful though:
- `$ pipenv install --dev --verbose --skip-lock` fails to install package, `$ pipenv graph` does not show anything. 
```toml
# Pipfile
[package]
pytest = "*"
```
- Editable is global option, so editable packages should not go into the same batch with non-editable packages, otherwise installed non-packages will be re-installed again.
- Editable packages are always non-hashed.
- Non-editable packages are always hashed if we have lockfile or `--skip-lock` isn't used.

pipenv/core.py
```diff
@@ -658,47 +656,49 @@ def _cleanup_procs(project, procs, failed_deps_queue, retry=True):
...
-                project.environment.expand_egg_links()
-                click.echo(
-                    "{}".format(
-                        click.style(
-                            "Failed initial installation: Failed to overwrite existing "
-                            "package, likely due to path aliasing. Expanding and trying "
-                            "again!",
-                            fg="yellow",
+            deps = c.deps.copy() if hasattr(c, "deps") else None
+            for dep in deps:
```
dqkqd, line 660: I think if `deps is None`, this loop will cause error.

matteius:
@dqkqd Thank you for your observations and comments--I have pushed some additional changes to make the distinction between the two file types more clear and fix the bug with the `--skip-lock` behavior and cleanup the method signatures from unused parameters, and default to empty list instead of None for the case of `c.deps` not existing.

pipenv/core.py
oz123: Yes, that's correct. If deps is `None` there will be an error. However, it's always setup to an empty list `[]`.

oz123 (changes requested):
Can we change the access to `c.deps`?

pipenv/progress.py
```diff
@@ -1,222 +0,0 @@
-"""
```
oz123, line 1: Can you extract this commit to a separte PR?

oz123 (changes requested):

pipenv/core.py
matteius: Do you mean `deps = getattr(c, "deps", []).copy()` ?

oz123: A `[]` won't have a copy method,  so it will break.

oz123 (approved):

Tenzer:
I'm happy to see this made it into Pipenv. I pointed out the performance overhead of using a separate `pip` process for the sake of the progress bar four years ago in https://github.com/pypa/pipenv/issues/2207, but back then, there was too much resistance to removing the progress bar that nothing happened to it.

matteius:
Oh thank you @Tenzer -- I am happy to hear others may be as excited about this change as I am.  I am sorry that I didn't quite understand that issue back when I commented on it in July and locked discussions -- It was mainly because there are so many old issues with the original contributors that can be impossible to parse and make sense of sometimes, that I just prefer in some cases people open new issue reports but now that I re-read that one all the same conclusions I came to you had come to there as well.  

I had not even gotten to performance profiling pip, so its interesting you had found that most of the time was with importlib, I find that kind of surprising.   Anyway, it became clear when I was trying out such an optimization that it used significantly less CPU time and less or similar wall clock time to do the same installs.    In the case of the Python Package Manager benchmark, we cut our install time of the sentry requirements in half!   https://lincolnloop.github.io/python-package-manager-shootout/

Fortunately I had the support of @oz123 to merge such a change--we both saw the performance improvement as way more important than the progress bar, especially when the `--verbose` flag can give us real time progress of the pip process.

PR #5304, opened 2022-08-28
Remove obsolete flag --three

oz123:
Since we only support python3 now, this flag is obsolete. It's been long enough since the removal of Python2, hence
I am guessing, no one is going to have its CI break because of this flag removal.


matteius:
Let's add a deprecation warning for this flag first in a different PR.

PR #5299, opened 2022-08-26
Fixes pipenv lock nondeterminism with environment markers

dqkqd:
Thank you for contributing to Pipenv!

### The issue
Fixes #5239

### The fix

Packages are added into resolver in this order:
1. Load from Pipfile
2. Pass into environment variables ``PIPENV_PACKAGES``
3. Pass into `Resolver` with the name `constraints`.
4. Parsed by pip's `parsed_requirements`
5. Convert into `InstallRequirement` by `install_req_from_parsed_requirement`
6. Pass into pip's resolver

After that, we would have all package's markers from resolver. 

Thanks @bakhtiary for providing a nice way to overcome this issue by sorting packages.
I think if we want to sort packages, we better sort them after step 5, which is the last step we could intervene.

But even if those packages are sorted, this might not be the best solution. Because markers are generated by pip's resolver, which is something we couldn't control (for example those packages might be reordered inside resolver).

### The checklist

* [x] Associated issue
* [x] A news fragment in the `news/` directory to describe this fix with the extension `.bugfix.rst`, `.feature.rst`, `.behavior.rst`, `.doc.rst`. `.vendor.rst`. or `.trivial.rst` (this will appear in the release changelog). Use semantic line breaks and name the file after the issue number or the PR #.

<!--

PR #5306, opened 2022-08-28
Revert "Fix nondeterminism with environment markers #5239 (#5286)"

matteius:
This reverts commit 4b9fc02a2561db47585e0fd822ada6dae6a49943.

@oz123 @bakhtiary -- this reverts: https://github.com/pypa/pipenv/pull/5286

because the change made here is actually fixing the issue and @dqkqd did some testing without this changeset and it was not required for fixing the issue:  https://github.com/pypa/pipenv/pull/5299#issuecomment-1228731246

I appreciate your effort to make pipenv better, please keep it up!   In this case, given some additional comments that came in on that PR and the added complexity that was not required, I think we should revert it.

### DIFF
--- a/news/5239.bugfix.rst
+++ b/news/5239.bugfix.rst
@@ -1,1 +1,0 @@
-Fix nondeterminism with environment markers #5239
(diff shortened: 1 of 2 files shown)


oz123:
In a away I think it's ok to revert. However, I liked the idea of using the dictionary keys for an ordered-set. 

We will have to watch it for a while. I suspect that working with ordered set of depends is better.

oz123 (approved):

Issue #5326, opened 2022-09-02
Installing from tarball fails on 2022.09.02


alexkahn-tempus:
### Issue description

Whilst installing via `RUN PIPENV_VENV_IN_PROJECT=1 pipenv install --deploy --verbose`

I have a requirement: `en-core-sci-md = {file = "https://s3-us-west-2.amazonaws.com/ai2-s2-scispacy/releases/v0.4.0/en_core_sci_md-0.4.0.tar.gz"}`

But it does not appear in my virtualenv upon:

```py
import spacy
spacy.load("en_core_sci_md")
```

### Expected result

I expected my docker image to have a dependency in the appropriate site-packages folder when installation finished.

### Actual result

The dependency was not available and did not exist in the site-packages folder.

### Steps to replicate
Pipfile entry:
```
en-core-sci-md = {file = "https://s3-us-west-2.amazonaws.com/ai2-s2-scispacy/releases/v0.4.0/en_core_sci_md-0.4.0.tar.gz"}
```

Corresponding lockfile entry:

```json
        "en-core-sci-md": {
            "file": "https://s3-us-west-2.amazonaws.com/ai2-s2-scispacy/releases/v0.4.0/en_core_sci_md-0.4.0.tar.gz",
            "hashes": [
                "sha256:9ca2cf8f3b066d0a7e69c10d9cca8822bcfa7fb8c9772bc90f8780629a81e974"
            ],
            "version": "==0.4.0"
        },
```
Running pipenv install --verbose caused this entry:
```
Writing supplied requirement line to temporary file: 'https://s3-us-west-2.amazonaws.com/ai2-s2-scispacy/releases/v0.4.0/en_core_sci_md-0.4.0.tar.gz'
```
But it did not appear later in the output.

Reverting to 2022.08.31 solved the issue temporarily.

❤️ The project and the maintainers, thanks so much for all your work!
-------------------------------------------------------------------------------

matteius:
Thanks for the report @alexkahn-tempus -- I see what the issues are, and opened a PR.   Gitlab is doing mainteance so I'll try to run the checks again later.  https://github.com/pypa/pipenv/pull/5327

oz123:
This PR is merged.

matteius:
Hey @alexkahn-tempus  `2022.9.4` was made specifically for this fix.

alexkahn-tempus:
Thank so much @matteius and @oz123 - I'll update my production systems and loop back if there is anything else.

PR #5327, opened 2022-09-03
fix for issue where requirement is a file or URL but does conntain hashes.

matteius:
there are actually a couple of fixes here, one is that sometimes the file URL contains hashes so we should include that with the group with hashes like we did in `2022.8.31` by accident.  Also the filters after need to detect this additional part of the conditional and that change was missing in #5326.

### The issue

#5306

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,7 +22,7 @@ repos:
   hooks:
   - id: black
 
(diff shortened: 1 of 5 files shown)


matteius:
Gitlab is having mainteance so the repository for flake8 is unavailable for the next ~180 minutes while they upgrade their database.

oz123 (approved):

PR #5328, opened 2022-09-03
Deprecate the --three flag

oz123:
pipenv now uses python3 by default.

Following our discussion in #5304

### DIFF
--- a/pipenv/cli/options.py
+++ b/pipenv/cli/options.py
@@ -11,6 +11,7 @@ from pipenv.vendor.click import (
     echo,
     make_pass_decorator,
     option,
+    secho,
 )
 from pipenv.vendor.click import types as click_types
 from pipenv.vendor.click_didyoumean import DYMMixin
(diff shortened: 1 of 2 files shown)


pipenv/cli/options.py
```diff
@@ -288,6 +289,11 @@ def callback(ctx, param, value):
 
 def three_option(f):
     def callback(ctx, param, value):
+        secho(
```
matteius, line 292: You should move this down inside the `if value is not None:` and then your tests will pass.  Right now it will always alert users, even if not passing this flag.

oz123: Done.

oz123:
@matteius the breaks on Windows are unrelated to the change. It's that requests test again.

matteius:
> @matteius the breaks on Windows are unrelated to the change. It's that requests test again.

The pypiserver PR will make those a non-issue, because they change to tests uvicorn and uvicorn[standard] instead of requests until we can sort out why requirementslib infers the specifier from setup metadata when none is specified.  It was clearly an edge case of that expected behavior and cross test contamination.

matteius (approved):

PR #5331, opened 2022-09-06
Examine the 20 slowest tests and make improvements.

matteius:
On my fast laptop these were the 20 slowest tests:

```
259.54s call     tests/integration/test_lock.py::test_complex_lock_with_vcs_deps
254.14s call     tests/integration/test_install_uri.py::test_get_vcs_refs
253.65s call     tests/integration/test_install_twists.py::test_multiple_editable_packages_should_not_race
152.11s call     tests/integration/test_install_twists.py::test_local_zipfiles
137.95s call     tests/integration/test_lock.py::test_lock_editable_vcs_with_extras_without_install
133.73s call     tests/integration/test_lock.py::test_lock_nested_direct_url
133.52s call     tests/integration/test_lock.py::test_lock_editable_vcs_without_install
132.87s call     tests/integration/test_install_uri.py::test_install_editable_git_tag
130.86s call     tests/integration/test_lock.py::test_vcs_lock_respects_top_level_pins
130.70s call     tests/integration/test_uninstall.py::test_uninstall_all_local_files
127.40s call     tests/integration/test_install_twists.py::TestDirectDependencies::test_https_dependency_links_install
127.36s call     tests/integration/test_install_uri.py::test_basic_vcs_install_with_env_var
125.63s call     tests/integration/test_install_uri.py::test_basic_vcs_install
125.47s call     tests/integration/test_project.py::test_include_editable_packages
116.35s call     tests/integration/test_install_uri.py::test_editable_vcs_install
113.39s call     tests/integration/test_lock.py::test_lock_editable_vcs_with_markers_without_install
112.68s call     tests/integration/test_uninstall.py::test_uninstall_all_dev
111.80s call     tests/integration/test_lock.py::test_lock_editable_vcs_with_ref_in_git
109.92s call     tests/integration/test_cli.py::test_pipenv_clean_pip_no_warnings
109.60s call     tests/integration/test_install_twists.py::test_relative_paths
```

Here I examine them, make adjustments to what they should be doing to test (or if they should even exist in a couple cases, due to repeated test cases).   After running the tests again:

### DIFF
--- a/tests/integration/test_install_basic.py
+++ b/tests/integration/test_install_basic.py
@@ -395,7 +395,7 @@ def test_install_venv_project_directory(pipenv_instance_pypi):
 @pytest.mark.system
 def test_system_and_deploy_work(pipenv_instance_private_pypi):
     with pipenv_instance_private_pypi(chdir=True) as p:
-        c = p.pipenv("install tablib")
+        c = p.pipenv("install urllib3")
         assert c.returncode == 0
         c = p.pipenv("--rm")
         assert c.returncode == 0
@@ -403,17 +403,6 @@ def test_system_and_deploy_work(pipenv_instance_private_pypi):
         assert c.returncode == 0
         c = p.pipenv("install --system --deploy")
         assert c.returncode == 0
-        c = p.pipenv("--rm")
-        assert c.returncode == 0
-        Path(p.pipfile_path).write_text(
-            """
-[packages]
-tablib = "*"
-        """.strip()
-        )
-        c = p.pipenv("install --system")
-        assert c.returncode == 0
-
 
 @pytest.mark.basic
 @pytest.mark.install
--- a/tests/integration/test_install_basic.py
+++ b/tests/integration/test_install_basic.py
@@ -1,50 +1,23 @@
 import os
-import mock
-
 from pathlib import Path
 from tempfile import TemporaryDirectory
 
 import pytest
 
-from flaky import flaky
-
 from pipenv.utils.processes import subprocess_run
 from pipenv.utils.shell import temp_environ
 
 
-@pytest.mark.setup
-@pytest.mark.basic
-@pytest.mark.install
-def test_basic_setup(pipenv_instance_private_pypi):
-    with pipenv_instance_private_pypi() as p:
-        with pipenv_instance_private_pypi(pipfile=False) as p:
-            c = p.pipenv("install requests")
-            assert c.returncode == 0
-
-            assert "requests" in p.pipfile["packages"]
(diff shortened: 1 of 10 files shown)


pipenv/core.py
```diff
@@ -2730,6 +2730,9 @@ def do_run(
     """
     from .cmdparse import ScriptEmptyError
 
+    load_dot_env(project, quiet=quiet)
+    env = os.environ.copy()
+
```
oz123, line 2735: There was a past discussion about when we should parse dotenv. I hope this doesn't break any user expectations.

tests/integration/test_install_basic.py
```diff
@@ -1,50 +1,23 @@
 import os
-import mock
-
 from pathlib import Path
 from tempfile import TemporaryDirectory
 
 import pytest
```
oz123, line 6: By removing flaky we aren't trying again on flaky internet connections. I believe this was the original intent.
Does it really slow things that badly?

tests/integration/test_install_basic.py
```diff
@@ -1,50 +1,23 @@
...
-            assert "chardet" in p.lockfile["default"]
-            assert "idna" in p.lockfile["default"]
-            assert "urllib3" in p.lockfile["default"]
-            assert "certifi" in p.lockfile["default"]
-
-
-@flaky
 @pytest.mark.basic
 @pytest.mark.install
 def test_basic_install(pipenv_instance_private_pypi):
```
oz123, line 13: OK, that is really similar to the above similar you removed.

pipenv/core.py
matteius: Oh I need to move this to a different PR.   This change was because right now it does not work to set PIPENV_CUSTOM_VENV_NAME in the .env file because it gets loaded too late.

tests/integration/test_install_basic.py
matteius: its is a long running test case covered/repeated by other test cases.

tests/integration/test_install_basic.py
matteius: Well my problem with it is the tests seem randomly annotated with flaky -- Its almost like every test should be labeled flaky -- in which case let's not annotate them and just always retry it once.  But the other side of my brain thinks -- well we are either connecting to pypi which is fronted by Fastly, or our private pypi, or github.   I haven't really seen flaky ever kick in a way that isn't become someone broke a test and so it retires it -- but maybe some people like it?  I personally think it adds confusion.

oz123 (approved):

PR #5339, opened 2022-09-08
Replace usage of  pipfile

### LINKED ISSUE #5310: Replace package pipfile with plette
[plette](https://github.com/sarugaku/plette/) has all the abilities of [pipfile](https://github.com/pypa/pipfile) and more.
The API is identical.
Also, pipfile isn't really maintained anymore. Hence, we should drop it, and use plette instead.

oz123:
This is because pipfile is quite dormant.

This will help with removing six. Also, worth noting:
plette's own Pipfile class uses a much stricter checking. I don't think it's going to be a problem
since 99% of the user don't craft their own `Pipfile`s. However, I needed to fix some tests to
pass again.
If people did craft their own `Pipfile` the errors from Plette are pretty straight forward to understand
so they can fix their files.  
Also, a nice bonus: it seems that the test suite now runs ~20 min, compared to ~23 min. Still too long, 
Just reminding ourselves ... it used to be 50 minutes ...
Fixes #5310


tests/integration/test_install_basic.py
```diff
@@ -416,6 +416,7 @@ def test_rewrite_outline_table(pipenv_instance_private_pypi):
         assert 'colorama = "*"' in contents
 
 
+@pytest.mark.skip(reason="Plette is more strict then pipfile ...")
```
matteius, line 419: Let's just axe this test case -- I believe I did in my other PR around categories.

tests/integration/test_install_markers.py
```diff
@@ -167,8 +165,9 @@ def test_environment_variable_value_does_not_change_hash(pipenv_instance_private
             assert lock_hash is not None
             assert lock_hash == project.calculate_pipfile_hash()
 
-            # sanity check on pytest
-            assert 'PYPI_USERNAME' not in str(pipfile.load(p.pipfile_path))
+            # This line seems wrong. Why would pipenv install expand the variable and
```
matteius, line 168: I agree -- lets just axe those assertions and not leave in the comments.

tests/integration/test_lock.py
```diff
@@ -527,6 +531,7 @@ def test_lockfile_with_empty_dict(pipenv_instance_pypi):
         assert p.lockfile['_meta']
 
 
+@pytest.mark.skip(reason="Plette is more strict then pipfile ...")
```
matteius, line 534: Let's just axe this test.

tests/integration/test_install_markers.py
oz123: Sure. Just wanted you to be aware of these things. Now that you noticed we can
remove it.

tests/integration/test_lock.py
oz123: done

matteius (approved):

PR #5343, opened 2022-09-08
Bump pipdeptree

oz123:
This is required for enabling Python3.11


oz123:
> Perhaps add the changes from #5245 so that we are assured that the tests pass.

I prefer doing this in a separate PR. Let's merge this one, and then one can merge the main branch to #5245 .

tasks/vendoring/patches/vendor/pipdeptree-update-pip-import.patch
```diff
@@ -1,13 +1,18 @@
...
+diff --git a/pipenv/vendor/pipdeptree/__main__.py b/pipenv/vendor/pipdeptree/__main__.py
+index 85cca3c..a002019 100644
+--- a/pipenv/vendor/pipdeptree/__main__.py
++++ b/pipenv/vendor/pipdeptree/__main__.py
+@@ -1,5 +1,13 @@
++import os
+ import sys
+ 
++pardir = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
++# for finding pipdeptree itself
```
matteius, line 10: I bet we could drop this patch and add `pipdeptree` to the vendoring tasks `LIBRARY_RENAMES` and re-run vendoring to get the imports corrected.

oz123: I tried this, without much success. We call this tool with a subshell, so all the paths are weird. 

I'd rather not invest more time on this now.

matteius (approved):

cclauss:
After this was merged, I rebased the Python 3.11 tests in #5245 and they fail. :-(

matteius:
Same failures?

cclauss:
It appears that the vendored-in `pytest` is trying to import `tomli` instead of using the `tomllib` that is built into Py3.11.

Is the vendored-in pytest current?  https://github.com/pytest-dev/pytest/releases
```
lib/python3.11/site-packages/_pytest/config/findpaths.py", line 67, in load_config_dict_from_file
    import tomli
ModuleNotFoundError: No module named 'tomli'
```
https://docs.python.org/3.11/library/tomllib.html?highlight=tomllib

PR #5348, opened 2022-09-09
Remove --sequential

oz123:


matteius:
I think we may need to deprecate the flag before we remove it in case people use it in their CI.

oz123:
I believe that people who use the flag in CI won't notice the difference, since it's a warning... A one off breakage is a risk we can take. If someone relied on the functionality, it's already gone. ... 
So, let's merge this with a warning. I have a strong gut feeling there won't be any rush of issues because of this.

matteius (approved):

tconkling:
Hey, just a heads up that this _does_ break CI. If you're using the `--sequential` flag, this new version of Pipenv parses that flag as a dependency name, and fails with 

```
pipenv.vendor.requirementslib.exceptions.RequirementError: Failed parsing requirement from '--sequential'
```

PR #5350, opened 2022-09-10
Safer usages of pkg_resources

matteius:
Fixes #5349
Fixes #5208

### The issue

* Cleanup some patching to be consistent with how we patch pipenv to be available internal to pip `__main__.py` which is a more modern way using importlib that does not rely on pkg_resoruces.
* Utilizing the internal pkg_resources, and make that part of the `vendoring.update` task by adding it to the `LIBRARY_RENAMES` and reinvoke the  `vendoring.update` task which rewrote the imports.

PR #5351, opened 2022-09-10
Begin converting away from pkg_resources

matteius:
Starts fixing #5349

### The issue

* Cleanup some patching to be consistent with how we patch pipenv to be available internal to pip `__main__.py` which is a more modern way using importlib that does not rely on pkg_resoruces.

There is more work to do to fully drop `pkg_resources` which seems to be what we are supposed to do:  https://setuptools.pypa.io/en/latest/pkg_resources.html

image

The problem enters with `pipdeptree` is expecting the `pkg_resources` Distribution.  I opened https://github.com/tox-dev/pipdeptree/issues/175 to have that discussion.

### DIFF
--- a/news/5349.bugfix.rst
+++ b/news/5349.bugfix.rst
@@ -1,0 +1,1 @@
+Modernize ``pipenv`` path patch with ``importlib.util`` to eliminate import of ``pkg_resources``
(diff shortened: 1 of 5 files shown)


matteius:
Ok, figured out my blockers with the prior PR, and so closing this in favor of:  https://github.com/pypa/pipenv/pull/5350
which will depend on:  https://github.com/sarugaku/requirementslib/pull/343

Issue #3478, opened 2019-01-24
Fish completions don't tab complete file paths


bram2000:
### Issue description

When using fish shell, entering something like `pipenv run nvim ` should allow tab-completion of file paths (this is just one example, obviously this would be the case for other commands run under pipenv too).

### Expected result

Tab completion works to complete filenames to edit. Ideally we should treat anything after `pipenv run` as another full fish command for completion (because it could be any valid command).

### Actual result

Nothing - tab completion just does nothing after entering `pipenv run nvim`.

### Steps to replicate

try to tab complete file paths after `pipenv run <command>`, in my case it's neovim, so `pipenv run nvim ./<tab>` just does nothing where it should complete the path.
-------------------------------------------------------------------------------

Please run `$ pipenv --support`, and paste the results here. Don't put backticks (`` ` ``) around it!

<details><summary>$ pipenv --support</summary>

Pipenv version: `'2018.11.26'`

Pipenv location: `'/usr/local/Cellar/pipenv/2018.11.26/libexec/lib/python3.7/site-packages/pipenv'`

Python location: `'/usr/local/Cellar/pipenv/2018.11.26/libexec/bin/python3.7'`

Python installations found:

  - `3.8.0a0`: `/Users/jbramley/.pyenv/versions/3.8-dev/bin/python3`
  - `3.8.0a0`: `/Users/jbramley/.pyenv/versions/3.8-dev/bin/python3.8m`
  - `3.7.2`: `/usr/local/bin/python3`
  - `3.7.2`: `/usr/local/bin/python3.7m`
  - `3.7.1`: `/Users/jbramley/.pyenv/versions/shopping_cart_3.7/bin/python3`
  - `3.7.1`: `/Users/jbramley/.pyenv/versions/3.7.1/bin/python3.7m`
  - `3.7.0`: `/Users/jbramley/.pyenv/versions/3.7.0/bin/python3`
  - `3.7.0`: `/Users/jbramley/.pyenv/versions/3.7.0/bin/python3.7m`
  - `3.6.5`: `/Users/jbramley/.pyenv/versions/TMP2/bin/python3`
  - `3.6.5`: `/Users/jbramley/.pyenv/versions/3.6.5/bin/python3.6m`
  - `2.7.15`: `/usr/local/bin/python`
  - `2.7.15`: `/usr/local/bin/pythonw`
  - `2.7.14`: `/Users/jbramley/.pyenv/versions/neovim2/bin/python`
[...]

PR #5352, opened 2022-09-11
cli: set `click_types.Path` for `-r` and `-e` options

exploide:
### The issue

pipenv can generate its own shell completions using the click framework. Prior to this PR, there are no file/path completions for the `-r` and `-e` options. See #3478 for a discussion.

### The fix

This PR changes the option type to `click_types.Path` for `-r` and `-e`. This is enough for click to know that file/path completions are desired here.
And this isn't only beneficial for fish shell, it automatically applies to bash and zsh also.

This fixes partially issue #3478 but not the part concerning `pipenv run` completions.

### The checklist

* [x] Associated issue

I'm not sure whether the news fragment is mandatory and which type would be appropriate. If you tell me, I could add one.

### DIFF
--- a/news/5352.trivial.rst
+++ b/news/5352.trivial.rst
@@ -1,0 +1,1 @@
+emit file-based shell completions for -r and -e options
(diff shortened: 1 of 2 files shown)


oz123:
Can you please add a news file?

exploide:
> Can you please add a news file?

done

oz123 (approved):

matteius (approved):

Issue #5167, opened 2022-07-08
Version of pipenv > 2020.11.15 will produce error AttributeError: 'JoinedStr' object has no attribute 's'


alexanderdevm:
### Issue description

Version of pipenv > `2020.11.15` will produce error `AttributeError: 'JoinedStr' object has no attribute 's'` when it installs a package from private repo that contains `setup.py` with dependencies from private repo. 

### Expected result

Packages and its dependencies should install from private repo (bitbucket)

### Actual result

```
pipenv install --verbose
Installing dependencies from Pipfile.lock (19667c)...
/home/SYS/comp1/.pyenv/versions/3.7.13/lib/python3.7/site-packages/pipenv/patched/notpip/_internal/operations/prepare.py:226: PipDeprecationWarning: DEPRECATION: A future pip version will change local packages to be built in-place without first copying to a temporary directory. We recommend you use --use-feature=in-tree-build to test your packages with this new behavior before it becomes the default.
 pip 21.3 will remove support for this functionality. You can find discussion regarding this at https://github.com/pypa/pip/issues/7555.
  issue=7555
Traceback (most recent call last):
  File "/home/SYS/comp1/.pyenv/versions/3.7.13/bin/pipenv", line 8, in <module>
    sys.exit(cli())
  File "/home/SYS/comp1/.pyenv/versions/3.7.13/lib/python3.7/site-packages/pipenv/vendor/click/core.py", line 1128, in __call__
    return self.main(*args, **kwargs)
  File "/home/SYS/comp1/.pyenv/versions/3.7.13/lib/python3.7/site-packages/pipenv/vendor/click/core.py", line 1053, in main
    rv = self.invoke(ctx)
  File "/home/SYS/comp1/.pyenv/versions/3.7.13/lib/python3.7/site-packages/pipenv/vendor/click/core.py", line 1659, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/SYS/comp1/.pyenv/versions/3.7.13/lib/python3.7/site-packages/pipenv/vendor/click/core.py", line 1395, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/SYS/comp1/.pyenv/versions/3.7.13/lib/python3.7/site-packages/pipenv/vendor/click/core.py", line 754, in invoke
    return __callback(*args, **kwargs)
[...]

PR #5357, opened 2022-09-15
Improve detection of unparsable ASTs

albertoandreottiATgmail:

### The issue

While running pipenv lock, I got this,

```
File "/usr/local/lib/python3.7/site-packages/pipenv/vendor/requirementslib/models/setup_info.py", line 198, in caller
    return func(*args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/pipenv/vendor/requirementslib/models/setup_info.py", line 366, in _find_install_requires
    return [el.s for el in variable.elts]
  File "/usr/local/lib/python3.7/site-packages/pipenv/vendor/requirementslib/models/setup_info.py", line 366, in <listcomp>
    return [el.s for el in variable.elts]
AttributeError: 'BinOp' object has no attribute 's'
```
It seems that some situations in which the AST cannot be parsed are passing undetected.

### The fix

### The checklist

* [ ] Associated issue

### DIFF
--- a/pipenv/vendor/requirementslib/models/setup_info.py
+++ b/pipenv/vendor/requirementslib/models/setup_info.py
@@ -356,7 +356,9 @@ class SetupReader:
             variable = cls._find_variable_in_body(body, value.id)
 
             if variable is not None and isinstance(variable, ast.List):
-                return [el.s for el in variable.elts]
+                r = [el.s for el in variable.elts if hasattr(el, 's')]
+                if r:
+                    return r
 
         raise Unparsable()
 


oz123:
Thank you! This PR should be opened against https://github.com/sarugaku/requirementslib.

matteius:
@oz123  is correct and there is a ticket over there too for this already.  I started a PR once around it, but never finished it.  We should alert the user when we were unable to parse the AST (likely because it used string interpolation).

matteius:
Also this:  https://github.com/pypa/pipenv/issues/5167

albertoandreottiATgmail:
Hello, thank you so much for the comments. Any advice on how to avoid running into this problem?

matteius:
@albertoandreottiATgmail Yes -- the problem is the requirementslib setup.py parser is designed to not execute any code, so likely in the install_requires, or other string being parsed, its an f-string or using format string, which is interpolation and would require executing arbitrary code to evaluate.   If the affected library could change to use a basic string without interpolation then you would not get this error.   In requirementslib we should detect it and alert the user as to this cause.  Once we do that, we will vendor in the new requirementslib and pipenv would output the more reasonable error.

matteius:
@albertoandreottiATgmail This branch outdates this PR and should also make the desired improvement, if you have spare cycles and could check:  https://github.com/pypa/pipenv/pull/5793

PR #5383, opened 2022-10-02
Update documentation around requirements command and provide current examples.

### LINKED ISSUE #5266: `pipenv requirements` mode considers pipfile.lock but docs reference Pipfile
Apologies if I've missed an existing feature request, I had a search through and couldn't find one.

### Is your feature request related to a problem? Please describe.

The [serverless-wsgi](https://www.serverless.com/plugins/serverless-wsgi) plugin generates a requirements.txt from pipfile, before installing those requirements. There are other tools that also rely on requirements.txt. But one limitation of requirements.txt vs. pipenv is that it doesn't provide any guarantees of sub-dependency versions.

If there were a `pipenv requirements` mode that would generate a requirements.txt file with all locked dependencies specified, then that would bridge the divide and allow us to ensure that the locked dependencies in pipfile.lock were used by these external tools.

### Describe the solution you'd like

Much like `pipenv sync` only considers the pipenv.lock file, a new argument to `pipenv requirements` would make it generate a requirements.txt file from all the dependencies in `pipfile.lock`. I assume it would need to list the lowest level dependencies first in order to ensure that they didn't get installed twice which might happen if top level dependencies were listed first.

matteius:
Fixes #5266 

### The checklist

* [X] Associated issue

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -263,99 +263,49 @@ Let's take this ``Pipfile``::
     [dev-packages]
     pytest = {version="==3.2.3"}
 
-Which generated the following ``Pipfile.lock`` upon completion of running ``pipenv lock```::
+Which generates a ``Pipfile.lock`` upon completion of running ``pipenv lock``` similar to::
 
     {
-		"_meta": {
-			"hash": {
-				"sha256": "4b81df812babd4e54ba5a4086714d7d303c1c3f00d725c76e38dd58cbd360f4e"
-			},
-			"pipfile-spec": 6,
-			"requires": {},
-			"sources": [
-				{
-					"name": "pypi",
-					"url": "https://pypi.python.org/simple",
-					"verify_ssl": true
-				}
-			]
-		},
-		"default": {
-			"certifi": {
-				"hashes": [
-					"sha256:0d9c601124e5a6ba9712dbc60d9c53c21e34f5f641fe83002317394311bdce14",
-					"sha256:90c1a32f1d68f940488354e36370f6cca89f0f106db09518524c88d6ed83f382"
-				],
-				"markers": "python_version >= '3.6'",
-				"version": "==2022.9.24"
-			},
-			"chardet": {
-				"hashes": [
-					"sha256:84ab92ed1c4d4f16916e05906b6b75a6c0fb5db821cc65e70cbd64a3e2a5eaae",
-					"sha256:fc323ffcaeaed0e0a02bf4d117757b98aed530d9ed4531e3e15460124c106691"
-				],
-				"version": "==3.0.4"
-			},
-			"idna": {
-				"hashes": [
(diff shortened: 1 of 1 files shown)


docs/advanced.rst
```diff
@@ -247,52 +247,149 @@ Anaconda uses Conda to manage packages. To reuse Conda–installed Python packag
...
+				"hashes": [
+					"sha256:0d9c601124e5a6ba9712dbc60d9c53c21e34f5f641fe83002317394311bdce14",
+					"sha256:90c1a32f1d68f940488354e36370f6cca89f0f106db09518524c88d6ed83f382"
+				],
+				"markers": "python_version >= '3.6'",
+				"version": "==2022.9.24"
+			},
+			"chardet": {
+				"hashes": [
+					"sha256:84ab92ed1c4d4f16916e05906b6b75a6c0fb5db821cc65e70cbd64a3e2a5eaae",
```
oz123, line 294: I think you can trim part of the contents of the lock file from here to the develop section, and then remove also parts of the develop section.

This will be more readable IMHO.

oz123 (changes requested):
I think including the complete lock file isn't necessary.
One could just do:

```
content
...snipped...
content
```

So users can see that only parts of the file are included.

matteius:
@oz123 -- The problem I ran into is which part of the lockfile to not show?   Since all of it gets consumed by the subsequent requirements command examples, I thought it might be useful for people to see that the markers are brought over and that the output represents all of the packages from those sections.

oz123:
I would just include this: 
```
{
            "_meta": {
                    "hash": {
                            "sha256": "4b81df812babd4e54ba5a4086714d7d303c1c3f00d725c76e38dd58cbd360f4e"
                    },
                    "pipfile-spec": 6,
                    "requires": {},
                    "sources": [
                            {
                                    "name": "pypi",
                                    "verify_ssl": true
                            }
                    ]
            },
            "default": {
			... snipped ...
                    "requests": {
                            "hashes": [
                                    "sha256:6a1b267aa90cac58ac3a765d067950e7dbbf75b1da07e895d1f594193a40a38b",
                                    "sha256:9c443e7324ba5b85070c4a818ade28bfabedf16ea10206da1132edaa6dda237e"
                            ],
                            "index": "pypi",
                            "version": "==2.18.4"
                    },
			... snipped ...
            },
            "develop": {
                    ... snipped ...
                    "pytest": {
                            "hashes": [
                                    "sha256:27fa6617efc2869d3e969a3e75ec060375bfb28831ade8b5cdd68da3a741dc3c",
                                    "sha256:81a25f36a97da3313e1125fce9e7bbbba565bc7fec3c5beb14c262ddab238ac1"
                            ],
                            "index": "pypi",
                            "version": "==3.2.3"
                    }
                    ... snipped ...
    }
```

For the sake of the example, it's fine.

oz123 (approved):

Issue #5345, opened 2022-09-09
Feature: Populate python_full_version instead of python_version in Pipfile when full python version is specified in "--python" arg


internetofjames:
I verified that I am using the latest version of pipenv at the time of writing (`2022.9.8`).

### Is your feature request related to a problem? Please describe.

Not related to a problem, just observed a behavior I find odd given current functionality of pipenv. When creating a virtual environment, specifying a full python version (e.g., `pipenv install --python 3.10.4`) generates a Pipfile with `python_version = "3.10"`. Given that pipenv supports a `python_full_version` key in the \[requires\] table, it seems odd to me that it doesn't use the full version field when a full python version is specified. Instead, it only uses the `python_version` field and specifies the feature release (e.g., `3.10`).

### Describe the solution you'd like

When using a full python version qualifier in the `pipenv install` command (e.g., `pipenv install --python 3.10.4`), the resulting Pipfile will use the `python_full_version` field instead:
```toml
[requires]
python_full_version = "3.10.4"
```

If the full version is not specified, the current expected behavior should apply:
```toml
[requires]
python_version = "3.10"
```

### Describe alternatives you've considered

An obvious workaround is to manually fill out the field, but if others agree with me, I believe this behavior should happen automatically.

jerempy:
I'd like to take a shot on this one

dojutsu-user:
@matteius I believe this can be closed.

PR #5387, opened 2022-10-03
Python full version in pipfile

jerempy:

### The issue
https://github.com/pypa/pipenv/issues/5345

A user requested that when the full version is specified that it gets added to a new pipfile when it is created so that he/she wouldn't need to manually add it themselves.

### The fix

Found the spot where the pipefile [requires] section gets created and modified it slightly to save based on if full version was specified or not.

### The checklist

* [x] Associated issue

### DIFF
--- a/news/5345.behavior.rst
+++ b/news/5345.behavior.rst
@@ -1,0 +1,1 @@
+New pipfiles show python_full_version under [requires] if specified
--- a/news/5345.behavior.rst
+++ b/news/5345.behavior.rst
@@ -1,1 +1,1 @@
-New pipfiles show python_full_version under [requires] if specified
+New pipfiles show python_full_version under [requires] if specified. Previously creating a new pipenv project would only specify in the Pipfile the major and minor version, i.e. "python_version = 3.7". Now if you create a new project with a fully named python version it will record both in the Pipfile. So: "python_version = 3.7" and "python_full_version = 3.7.2"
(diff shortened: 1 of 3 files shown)


oz123 (changes requested):
Thank you for your PR.

Please fix the linting issue, and add a news fragment.

This is then good to be merged!

oz123:
Please rename the news snippet to:   5345.behavior.rst

oz123:
As I suspected this change leads to breakage of CI.
```
Successfully set up CPython (3.7.14)

```

I think that pinning to a minor version is not a good thing. Unless it's additional. What I mean is, we should have both, e.g.:

```
python_version=3.7
python_full_verison=3.7.14
```

jerempy:
@oz123 I updated it so now the pipfile will record both the python version and the full version in addition if specified. Also updated tests so should work for whatever python version the test is running. Sorry for the back and forth and hopefully we good to go now 🤞

news/5345.behavior.rst
```diff
@@ -0,0 +1 @@
+New pipfiles show python_full_version under [requires] if specified
```
oz123, line 1: Can you please add a longer explanation here?

oz123 (changes requested):

news/5345.behavior.rst
jerempy: ok cool. good call. Hows it now?

oz123: Much better. Thank you.

matteius:
What happens if a new minor version is released?   Will users have to manually update the `python_full_version`?

matteius (approved):

Issue #5294, opened 2022-08-25
Upgrade scripts to run package functions too


mgzenitech:
Be sure to check the existing issues (both open and closed!), and make sure you are running the latest version of Pipenv.

Check the [diagnose documentation](https://pipenv.pypa.io/en/latest/diagnose/) for common issues and the [PEEP list](https://github.com/pypa/pipenv/blob/master/peeps/) before posting! We may close your issue if it is very similar to one of them. Please be considerate and follow the PEEP process, or be on your way.

Make sure to mention your debugging experience if the documented solution failed.

### Is your feature request related to a problem? Please describe.

No

### Describe the solution you'd like

It would be very useful to be able to call package function as script (very similar as PDM does that https://pdm.fming.dev/latest/usage/scripts/). That would be the only way to have 100% testable code.

```
[scripts]
main = {call = "foo_package.bar_module:main('dev')"}
```

### Describe alternatives you've considered

PDM

### Additional context

<details><summary>$ pipenv --support</summary>

Pipenv version: `'2022.8.24'`

Pipenv location: `'/home/mariusg/.local/pipx/venvs/pipenv/lib/python3.9/site-packages/pipenv'`

Python location: `'/home/mariusg/.local/pipx/venvs/pipenv/bin/python'`

OS Name: `'posix'`

User pip version: `'22.2.2'`

user Python installations found:

  - `3.10.5`: `/usr/bin/python`
  - `3.10.5`: `/usr/bin/python3`
  - `3.9.9`: `/home/mariusg/.pyenv/versions/3.9.9/bin/python`
  - `3.9.8`: `/home/mariusg/.pyenv/versions/3.9.8/bin/python`

PEP 508 Information:

```
{'implementation_name': 'cpython',
 'implementation_version': '3.9.9',
 'os_name': 'posix',
 'platform_machine': 'x86_64',
 'platform_python_implementation': 'CPython',
 'platform_release': '5.4.210-1-MANJARO',
 'platform_system': 'Linux',
 'platform_version': '#1 SMP PREEMPT Thu Aug 11 13:15:22 UTC 2022',
 'python_full_version': '3.9.9',
 'python_version': '3.9',
 'sys_platform': 'linux'}
```

System environment variables:

  - `SHELL`
  - `SESSION_MANAGER`
  - `PYENV_SHELL`
  - `XDG_MENU_PREFIX`
  - `LC_ADDRESS`
  - `DOTNET_ROOT`
  - `LC_NAME`
[...]

PR #5390, opened 2022-10-06
Scripts run package functions

jerempy:
### The issue

Requested new feature to add the ability to include something like:
```[scripts]
main = {call = "foo_package.bar_module:main('dev')"}
```

   https://github.com/pypa/pipenv/issues/5294

### The fix

This adds the ability to parse a callable with module.func in scripts

### The checklist

* [x] Associated issue

### DIFF
--- a/pipenv/cmdparse.py
+++ b/pipenv/cmdparse.py
@@ -1,9 +1,14 @@
 import itertools
 import re
 import shlex
+import tomlkit
 
 
 class ScriptEmptyError(ValueError):
+    pass
+
+
+class ScriptParseError(ValueError):
     pass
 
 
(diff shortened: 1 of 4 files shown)


jerempy:
Also, should probably get an update to: https://pipenv.pypa.io/en/latest/advanced/#custom-script-shortcuts

pipenv/cmdparse.py
```diff
@@ -2,31 +2,61 @@
 import re
 import shlex
 
+import tomlkit
```
matteius, line 5: This should be `from pipenv.vendor import tomlkit` to ensure that imports will work for all users -- its probably that tomlkit was just installed in your environment for this to work, but we have it vendor'd in.

jerempy: ok cool. there were some other places not using the vendored version and i updated those as well

oz123 (approved):

oz123:
@matteius can you please resolve the merge conflicts?

matteius:
@jerempy Hey -- looks like there is merge conflict with the named package categories work that got merged -- could you rebase?

PR #5396, opened 2022-10-10
Update documentation for callable scripts

jerempy:
### The issue

Missed the docs on this one: https://github.com/pypa/pipenv/pull/5390
So updated it. Also noticed could improve the error message if command parsed wrongly so updated that too.

### The fix

Updated docs

### The checklist

* [x] Associated issue

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -581,8 +581,8 @@ For example:
     $ pipenv run echospam "indeed"
     I am really a very silly example indeed
 
-You can also specify pacakge functions as callables such as: ``<pathed.module>:<func>``. These can also take arguments. 
-For exaple: 
+You can also specify pacakge functions as callables such as: ``<pathed.module>:<func>``. These can also take arguments.
+For exaple:
 
 .. code-block:: toml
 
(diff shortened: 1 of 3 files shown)


docs/advanced.rst
```diff
@@ -581,7 +581,20 @@ For example:
...
+You can also specify pacakge functions as callables such as: ``<pathed.module>:<func>``. These can also take arguments.
+For exaple:
+
+.. code-block:: toml
+
+    [scripts]
+    my_func_with_args = {call = "package.module:func('arg1', 'arg2')"}
+    my_func_no_args = {call = "package.module:func()"}
+
+::
```
oz123, line 593: should this be:
```
.. code-block:: shell
```
?

oz123 (changes requested):

docs/advanced.rst
jerempy: Maybe, but I was just following the same format as the rest of this file to match the other parts of this section (they don't use `..code-block:: shell` and just use `::`
If you want me to change it I can change it, should i change it for the whole section?

PR #5402, opened 2022-10-12
Remove usage of mkdir_p in favour of os.makedirs

oz123:
As the last commit says, this function is obsolete.

### DIFF
--- a/pipenv/project.py
+++ b/pipenv/project.py
@@ -18,7 +18,6 @@ from pipenv.environment import Environment
 from pipenv.environments import Setting, is_in_virtualenv, normalize_pipfile_path
 from pipenv.patched.pip._internal.commands.install import InstallCommand
 from pipenv.patched.pip._vendor import pkg_resources
-from pipenv.utils.constants import is_type_checking
 from pipenv.utils.dependencies import (
     get_canonical_names,
     is_editable,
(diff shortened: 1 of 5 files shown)


pipenv/project.py
```diff
@@ -1,5 +1,3 @@
-from __future__ import annotations
```
matteius, line 1: I think let's not remove this and the `if is_type_checking():` block below, since those are only needed for the mypy checking.

oz123: This is needed if type hints are used in the code. Otherwise, there is NameError raised.
E.G, If I re-enable the guard, this line 
https://github.com/pypa/pipenv/blob/16b82949ecb2b3dd6fc18a7e79532ea6182eba67/pipenv/project.py#L266
will explode.
See the failed actions log here:

https://github.com/pypa/pipenv/actions/runs/3233835071/jobs/5296208253
image

This guard was really a must when supporting for Python2 was required. However, right now, it is not necessary.

matteius: That's because List needs to be imported from collections.   I think we should not be undoing the type annotations.  I have a ticket to get my checks working again.

matteius: It only List depends on being available outside of the annotations.  That is probably it.

oz123: I re-introduced the guard, and will do a separate PR to remove this guard in  all the places we use it later.

matteius (approved):

PR #5448, opened 2022-10-30
Fixes issue with unexpected package index resolution after upgrading pip to 22.3

### LINKED ISSUE #5444: Download source regression in 10.25
### Issue description

When installing packages from the same lock file, 10.25 will fail with hash check error, while 10.12 will succeed.

### Expected result

Both should install the packages.

### Actual result

Based on what I can tell, the lockfile includes the SHAs for the .whl and the .tar.gz from PyPi.  However, it actually downloads the wheel from PiWheels instead, which has a different SHA, not listed in the lock file.

So it's kind these steps happening

1. The index listed in the Pipfile.lock is `pypi` ([ref](https://github.com/stumpylog/pipenv-issue-repro/blob/main/Pipfile.lock#L27))
2. The SHAs are from PyPi
3. The download happens from PiWheels ([ref](https://github.com/stumpylog/pipenv-issue-repro/actions/runs/3332871227/jobs/5514234634#step:5:36))

```
[pipenv.exceptions.InstallError]:     asgiref==3.5.2 from https://www.piwheels.org/simple/asgiref/asgiref-3.5.2-py3-none-any.whl#sha256=dda3f0b6b921471f50c266c63f0a8a6eb2c65e66359915db18b69b8d629ca594 (from -r /tmp/pipenv-yap_1oy1-requirements/pipenv-u0qmzuvp-hashed-reqs.txt (line 1)):
[pipenv.exceptions.InstallError]:         Expected sha256 1d2880b792ae8757289136f1db2b7b99100ce959b2aa57fd69dab783d05afac4
[pipenv.exceptions.InstallError]:         Expected     or 4a29362a6acebe09bf1d6640db38c1dc3d9217c68e6f9f6204d72667fc19a424
[pipenv.exceptions.InstallError]:              Got        dda3f0b6b921471f50c266c63f0a8a6eb2c65e66[35](https://github.com/stumpylog/pipenv-issue-repro/actions/runs/3332756300/jobs/5513993154#step:5:36)9915db18b69b8d629ca594
```

### Steps to replicate

I threw together a basic repo here: https://github.com/stumpylog/pipenv-issue-repro

Please run `$ pipenv --support`, and paste the results here. Don't put backticks (`` ` ``) around it! The output already contains Markdown formatting.

If you're on macOS, run the following:

    $ pipenv --support | pbcopy

If you're on Windows, run the following:

    > pipenv --support | clip

If you're on Linux, run the following:

    $ pipenv --support | xclip

matteius:
### The issue

After upgrading pip to 22.3 there were cases where users had multiple index defined and the resolution of which index to use changed during install causing hash validation to fail.

Fixes #5444 

### The fix

Technically it as never that correct to be passing all sources using extra index URLs when user has default behavior of `install_search_all_sources = false` -- 
**note:** this won't fix what I believe to be a change in behavior to pip that may still cause issues for users that have `install_search_all_sources = true` however this brings the pipenv behaviors more in line with whats expected by ensuring complete isolation of the index assigned to each package.

### The checklist

* [X] Associated issue


pipenv/core.py
```diff
@@ -678,6 +679,12 @@ def _cleanup_procs(project, procs, failed_deps_queue, retry=True):
                 failed_deps_queue.put(dep)
 
 
+def get_trusted_hosts():
+    trusted_hosts = []
+    trusted_hosts.extend(os.environ.get("PIP_TRUSTED_HOSTS", []))
+    return trusted_hosts
```
oz123, line 685: PIP_TRSUTED_HOSTS is a space separated list, so we should do here:
```
try:
    return os.environ.get("PIP_TRUSTED_HOSTS", []).split(" ")
execpt AttributeErorr:
    return []
```
```
And we can also remove line 683.

**Review by oz123 — CHANGES_REQUESTED** [review]

File: pipenv/core.py [hunk 2]
```diff
@@ -692,19 +699,78 @@ def batch_install(
...
         sequential_deps = []
-    failed = not retry
     deps_to_install = deps_list[:]
     deps_to_install.extend(sequential_deps)
     deps_to_install = [
         dep for dep in deps_to_install if not project.environment.is_satisfied(dep)
     ]
-    trusted_hosts = []
+    search_all_sources = project.settings.get("install_search_all_sources", False)
+    sources = get_source_list(
```
oz123, line 710: Should this be documented in plette?
```
install_search_all_sources
```

pipenv/core.py
```diff
@@ -1656,33 +1710,17 @@ def pip_install_deps(
         ]
         pip_args = get_pip_args(
             project,
-            pre=pre,
+            pre=project.settings.get("allow_prereleases", False),
```
oz123, line 1713: Same as above: If this is this something which is part of the Pipefile, then it should be documented in plette.

pipenv/core.py
matteius: It is not a new feature or anything, but I am sure the plette documentation lacks it -- but are we really documenting `pipenv` implementation points in `plette` now?

oz123 (changes requested):
Great that this is fixed, and along the way we uncover some undocumented Pipefile stuff.

pipenv/core.py
matteius: This was the existing code just relocated, but perhaps there was a bug with it?

pipenv/core.py
matteius: I made a ticket in plette to handle any documentation updates there separately -- I have some idea that we would document the categories like `pipenv` but not the specific pipenv settings that can be supplied to the category, which are already documented in pipenv.   Let's not let it block this fix though:  https://github.com/sarugaku/plette/issues/34

pipenv/core.py
matteius: Pushed the refactor change.

oz123:
Please cherry pick the following commit to fix the CI:
be92e26606fe750210c391f17318c8eaf06c51ef 
and 
6f8831316c21aeeb6929bfbb1b4460bd3731d4e5
from https://github.com/pypa/pipenv/pull/5450

pipenv/core.py
oz123: I don’t think we should document this only in plette, but it makes sense for me to add it to the spec there. Thanks for opening the issue there.

PR #5451, opened 2022-11-02
Consolidate pipenv settings to use get_from_env

micahjsmith:
### The issue

See #5407

* some pipenv settings are insufficiently documented, especially in terms of how to disable them
* implementation of pipenv settings detection is inconsistent, even though `pipenv.environments.get_from_env` exists as a pretty hardened utility

### The fix

* refactor all relevant settings to use get_from_env, which more explicitly allows settings to be disabled
    * add default kwarg to get_from_env
* update docs to indicate how to explicitly disable a setting

Other changes
- Fix rst in contributing guide
- Refactor out is_env_truthy

### The checklist

* [x] Associated issue
* [x] A news fragment in the `news/` directory to describe this fix with the extension `.bugfix.rst`, `.feature.rst`, `.behavior.rst`, `.doc.rst`. `.vendor.rst`. or `.trivial.rst` (this will appear in the release changelog). Use semantic line breaks and name the file after the issue number or the PR #.

PR #5461, opened 2022-11-07
fix PIPENV_MAX_DEPTH environment variable

bjorand:
### The issue

I upgraded  from 2022.10.25 to 2022.11.5 and I figured out that my PIPENV_MAX_DEPTH env variable was not taken into account anymore. I had to use PIPENV_PIPENV_MAX_DEPTH. So here is a PR to fix the issue.

Bug introduces here: https://github.com/pypa/pipenv/pull/5451/files#diff-f74d4ef3c891c83b08a2b15b21602d0c31859892478cc4f56bf5be1a2c5092ecL185

### DIFF
--- a/pipenv/environments.py
+++ b/pipenv/environments.py
@@ -184,7 +184,7 @@ class Setting:
         """
 
         # NOTE: +1 because of a temporary bug in Pipenv.
-        self.PIPENV_MAX_DEPTH = int(get_from_env("PIPENV_MAX_DEPTH", default=3)) + 1
+        self.PIPENV_MAX_DEPTH = int(get_from_env("MAX_DEPTH", default=3)) + 1
         """Maximum number of directories to recursively search for a Pipfile.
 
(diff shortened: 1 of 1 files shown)


matteius:
That seems like an odd fix -- I wonder why the variables would be automatically prefixed with `PIPENV_` now?

oz123:
@matteius that is because the earlier code was inconsistent, and the PR @micahjsmith fixed things. He introduced a method
to get all the variables consistently. This added a PREFIX, and there was a one off that slipped through:
[
](https://github.com/micahjsmith/pipenv/blob/93e97b75104f7b72017f7bf8e1cfd61b518f1a01/pipenv/environments.py#L20)

oz123 (approved):

matteius:
Awesome, thanks!

micahjsmith:
My bad, thanks for catching this @bjorand !

PR #5480, opened 2022-11-19
New pipenv check with minimal output

yeisonvargasf:
### The issue

- The pipenv new Safety implementation is using a subprocess.

- There isn't a non-verbose output.

This pull request is related to https://github.com/pypa/pipenv/pull/5218

### The fix

- `pipenv check` uses Safety directly, replacing the Safety implementation with a subprocess.

- Added a new `minimal` output, a non-verbose output.

### DIFF
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -1,3 +1,4 @@
+import io
 import json as simplejson
 import logging
 import os
@@ -2922,7 +2923,8 @@ def do_check(
     if not quiet and not project.s.is_quiet():
         click.echo(
             click.style(
-                decode_for_output("Checking installed package safety..."), bold=True
+                decode_for_output("Checking installed packages for vulnerabilities..."),
+                bold=True,
             )
         )
     if ignore:
@@ -2947,6 +2949,8 @@ def do_check(
 
     if output == "full-report":
         options.append("--full-report")
+    elif output == "minimal":
+        options.append("--json")
     elif output not in ["screen", "default"]:
         options.append(f"--output={output}")
 
@@ -2959,7 +2963,7 @@ def do_check(
     if safety_project:
         options.append(f"--project={safety_project}")
 
-    cmd = _cmd + [safety_path, "--debug", "check"] + options
+    cmd = _cmd + [safety_path, "check"] + options
 
     if db:
         if not quiet and not project.s.is_quiet():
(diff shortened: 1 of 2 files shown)


yeisonvargasf:
@matteius and @oz123, please review it and let me know if you need any further changes so this PR can be merged into the @matteius branch (`vendor-latest-safety2`).

Note: CI will probably fail as the `main` branch CI status is also falling.

pipenv/core.py
```diff
@@ -2922,7 +2923,8 @@ def do_check(
     if not quiet and not project.s.is_quiet():
         click.echo(
             click.style(
-                decode_for_output("Checking installed package safety..."), bold=True
+                decode_for_output("Checking installed packages for vulnerabilities..."),
```
oz123, line 2926: Please don't use decode for output. It's not needed any more.

oz123 (changes requested):

pipenv/core.py
yeisonvargasf: Done! 👍

pipenv/core.py
```diff
@@ -16,7 +16,7 @@
 from typing import Dict, List, Optional, Union
 
 from pipenv import environments, exceptions, pep508checker
-from pipenv._compat import decode_for_output, fix_utf8
+from pipenv._compat import fix_utf8
```
oz123, line 19: also, no need for fix_utf8 ...

pipenv/core.py
```diff
@@ -2866,9 +2866,7 @@ def do_check(
             pypi_mirror=pypi_mirror,
         )
     if not quiet and not project.s.is_quiet():
-        click.echo(
-            click.style(decode_for_output("Checking PEP 508 requirements..."), bold=True)
-        )
+        click.echo(click.style("Checking PEP 508 requirements...", bold=True))
```
oz123, line 2869: And finally, please don't use `click.echo(click.style(...))`, instead use `click.secho`.

oz123 (approved):

pipenv/core.py
yeisonvargasf: Is it ok if I update `fix_utf8` in the whole `core.py` file? (Other functions use it in the same `core.py` file)

oz123: No, need too. Just where you touch the code, you should replace it.

yeisonvargasf: Okay, `fix_utf8` wasn't used in the `do_check` function, so I left unchanged the other functions using `fix_utf8`.

pipenv/core.py
yeisonvargasf: Thanks! I just updated those references, and I kept others that are composed of styled and unstyled strings.

PR #5481, opened 2022-11-19
Bump version requirementslib and more vendor clean up

oz123:
As we now bumped requirementslib, we could now also bump vistir to 0.7.4.
This in turn forces us to make some tiny changes, and allows us to finally drop yaspin.


oz123:
@matteius all tests pass, besides Python3.7, but this is unrelated to the changes I did.

pipenv/utils/resolver.py
```diff
@@ -63,6 +63,16 @@
 err = rich.console.Console(stderr=True)
 
 
+class FakePipTempDirdirectory:
```
matteius, line 66: I think having the name Fake in a class that is used in the actual production code is confusing, though I see why you added this.  Also it has `Dirdirectory`  Can we name it  `PipenvTempDirectoryWrapper` or something similar instead?

oz123: Can we use  `QuasiPipTempDirectory`  instead? Wrapper is misleading because we aren't wrapping anything.

matteius: Well we are wrapping the directory name 🤷   I am not sure it is Quasi either?

matteius: Ah its fine -- better than Fake.

matteius (approved):

Issue #5488, opened 2022-11-23
Followup on skipped test `test_pipenv_check`


matteius:
Had to skip this test to get the latest safety work across the line -- ideally @yeisonvargasf can help us look at how to make safety in the unit test look at the virtualenv that the test creates.

### Issue description

Test is marked skipped for now

### Expected result

To have a test of the integration.

yeisonvargasf:
This issue is probably caused by #4819. I'm going to take a look.

PR #5501, opened 2022-11-24
Get packages for `pipenv check` from the target venv.

yeisonvargasf:
### The issue

#4819
#5488

### The fix

Now, this command gets the packages to be analyzed from the target venv.

### DIFF
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -2963,9 +2963,14 @@ def do_check(
         _cmd + ["-m", "pip", "list", "--format=freeze"], is_verbose=project.s.is_verbose()
     )
 
-    temp_requirements = tempfile.NamedTemporaryFile(prefix=f"{project.virtualenv_name}_")
-    temp_requirements.write(target_venv_packages.stdout.strip().encode("utf-8"))
-    temp_requirements.seek(0)
+    temp_requirements = tempfile.NamedTemporaryFile(
+        mode="w+",
+        prefix=f"{project.virtualenv_name}",
+        suffix="_requirements.txt",
+        delete=False,
+    )
+    temp_requirements.write(target_venv_packages.stdout.strip())
+    temp_requirements.close()
 
     options.extend(["--file", temp_requirements.name])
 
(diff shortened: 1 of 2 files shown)


matteius:
Something not quite right with this change and the Windows CI.

matteius:
@yeisonvargasf A variation of these changes work for windows -- the primary issue is that the file gets-re-opened so it needs to be closed first.
```
PS C:\Users\matte\Projects\pipenv> git diff
diff --git a/pipenv/core.py b/pipenv/core.py
index acebcc20..494191cb 100644
--- a/pipenv/core.py
+++ b/pipenv/core.py
@@ -2959,6 +2959,16 @@ def do_check(
     if safety_project:
         options.append(f"--project={safety_project}")

+    target_venv_packages = run_command(
+        _cmd + ["-m", "pip", "list", "--format=freeze"], is_verbose=project.s.is_verbose()
+    )
+
+    temp_requirements = tempfile.NamedTemporaryFile(mode="w+", prefix=f"{project.virtualenv_name}", suffix="_requirements.txt", delete=False)
+    temp_requirements.write(target_venv_packages.stdout.strip())
+    temp_requirements.close()
+
+    options.extend(["--file", temp_requirements.name])
+
     cmd = _cmd + [safety_path, "check"] + options

     if db:
@@ -3044,6 +3054,8 @@ def do_check(

     cli(prog_name="pipenv")

+    temp_requirements.remove()
+

     import json as jsonlib
diff --git a/tests/integration/test_cli.py b/tests/integration/test_cli.py
index f086cdfe..b6009d85 100644
--- a/tests/integration/test_cli.py
+++ b/tests/integration/test_cli.py
@@ -140,7 +140,6 @@ def test_pipenv_graph_reverse(pipenv_instance_private_pypi):

 @pytest.mark.cli
-@pytest.mark.skip("Safety 2 ends up scanning the project virtualenv and not the instance created by this test.")
 def test_pipenv_check(pipenv_instance_private_pypi):
     with pipenv_instance_private_pypi() as p:
         c = p.pipenv('install pyyaml')
```

yeisonvargasf:
@matteius, thank you! I pushed the fix for that case; let me know if there is something more to fix.

PR #5523, opened 2022-12-07
Fix regression where wheel path is not included in the lockfile

### LINKED ISSUE #5479: Install local wheel package missing path in lock file
In my program I use a local wheel, I'm in the process to upgrading from py 3.7 to py 3.11 and I had some trouble installing a local wheel (hash mismatch) bu I was able to install it after the first lock.

Then I noticed wheel path is missing in the lock file, is that a normal behavior?

immagine

This is my actual pipfile 
```
[[source]]
verify_ssl = true
name = "pypi"

[dev-packages]
pyinstaller = "5.6.2"

[packages]
python-dotenv = "==0.21.0"
requests = "==2.28.1"
flask = "==2.2.2"
flask-cors = "==3.0.10"
flask-caching = "==2.0.1"
flask-jwt-extended = "==4.4.4"
python-dateutil = "==2.8.2"
peewee = "==3.15.4"
pysqlite3 = {path = "./../_res/wheels/pysqlite3-0.4.7-cp311-cp311-win_amd64.whl"}

[requires]
python_version = "3.11"
python_full_version = "3.11.0"
```
vs old one

immagine

matteius:

### The issue

Fixes #5479

### DIFF
--- a/news/5479.bugfix.rst
+++ b/news/5479.bugfix.rst
@@ -1,0 +1,1 @@
+Fix regression where ``path`` is not propagated to the ``Pipfile.lock``.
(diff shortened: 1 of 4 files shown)


oz123 (approved):

oz123 (changes requested):
Can we add a test for this to prevent future regressions?

oz123 (approved):

PR #5544, opened 2022-12-17
Deprecate --keep-outdated flag.

matteius:
### The issue

This flag, which was never fully explained in the documentation, leads to a source of a lot of confusion and issue because the underlying concept is flawed.   It cannot be used until a consistent lock file is generated with `pipenv lock` or `pipenv install` and then when applied it will bring in any new package from a new consistent lock phase which is not already defined in the `Pipfile.lock` thus often creating inconsistent and unexpected locks. 

### The fix

This was always a workaround to avoid managing the Pipfile specifiers and packages listed/pinned in the Pipfile, which is the inputs into the resolver for the new lock phase whenever modifying package requirements.  Instead, users that do not want packages updated while upgrading other packages should be pinning those in the Pipfile so they can be used as proper inputs to the pip resolver.

I have considered for a year now how to make `--keep-outdated` better, but the reality is it was a flawed idea that existed before the switch to using the pip resolver, and it no longer makes sense to keep around because it leads to confusion and support tickets that shouldn't exist.

### The checklist

* [ ] Associated issue

### DIFF
--- a/pipenv/cli/options.py
+++ b/pipenv/cli/options.py
@@ -158,14 +158,12 @@ def keep_outdated_option(f):
         state = ctx.ensure_object(State)
         state.installstate.keep_outdated = value
         if value:
-            click.echo(
-                click.style(
-                    "Use of --keep-outdated has been deprecated for removal."
-                    "The flag does not respect package resolver results and leads to inconsistent lock files.  "
-                    "Please pin relevant requirements in your Pipfile and discontinue use of this flag.",
-                    fg="yellow",
-                    bold=True,
-                ),
+            click.secho(
+                "The flag --keep-outdated has been deprecated for removal."
+                "The flag does not respect package resolver results and leads to inconsistent lock files.  "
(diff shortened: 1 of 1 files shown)


pipenv/cli/options.py
```diff
@@ -156,6 +157,17 @@ def keep_outdated_option(f):
     def callback(ctx, param, value):
         state = ctx.ensure_object(State)
         state.installstate.keep_outdated = value
+        if value:
+            click.echo(
+                click.style(
+                    "Use of --keep-outdated has been deprecated for removal."
+                    "The flag does not respect package resolver results and leads to inconsistent lock files.  "
```
oz123, line 164: I prefer
"The flag --keep-outdated has been deprecated and marked for removal.
It does not respect ..."

oz123 (approved):

oz123 (changes requested):
@matteius please see comments. Small changes desired.

oz123 (approved):

PR #5556, opened 2022-12-28
Remove pytest pypi take 2

oz123:
pytest-pypi is no longer maintained by us, and it is no longer used as we now use pypi-server package.


Pipfile.lock
```diff
@@ -132,73 +122,65 @@
             ],
             "version": "==1.2.2"
         },
-        "colorama": {
```
matteius, line 135: Problem is we loose this lock unless we lock on windows.   I think if we add `colorama= {version = "*", markers="sys_platform == 'win32'"}` to the `Pipfile` and relock it would solve this.

oz123: Ok. I'll add this to.

oz123: Fixed.

matteius (approved):

Issue #5582, opened 2023-01-29
Unable to create environment with system Python


PythonCoderAS:
Be sure to check the existing issues (both open and closed!), and make sure you are running the latest version of Pipenv.

Check the [diagnose documentation](https://pipenv.pypa.io/en/latest/diagnose/) for common issues before posting! We may close your issue if it is very similar to one of them. Please be considerate, or be on your way.

Make sure to mention your debugging experience if the documented solution failed.

### Issue description

Whenever I try to `pipenv install`, I get a `FileNotFoundError: [Errno 2] No such file or directory: '/home/user/.local/share/virtualenvs/wikidata-mix-n-match-bot-DcUD2uRB/bin/python'`

### Expected result

The install to work

### Actual result

```
Loading .env environment variables...
Using python: 3.10
Path to python: /usr/bin/python3
Creating a virtualenv for this project...
Pipfile: /home/user/wikidata-mix-n-match-bot/Pipfile
Using /usr/bin/python3 (3.10.6) to create virtualenv...
⠦ Creating virtual environment...created virtual environment CPython3.10.6.final.0-64 in 326ms
  creator CPython3Posix(dest=/home/user/.local/share/virtualenvs/wikidata-mix-n-match-bot-DcUD2uRB, clear=False, no_vcs_ignore=False, global=False)
  seeder FromAppData(download=False, pip=bundle, setuptools=bundle, wheel=bundle, via=copy, app_data_dir=/home/user/.local/share/virtualenv)
    added seed packages: pip==22.0.4, setuptools==62.1.0, wheel==0.37.1
  activators BashActivator,CShellActivator,FishActivator,NushellActivator,PowerShellActivator,PythonActivator

✔ Successfully created virtual environment!
Traceback (most recent call last):
  File "/usr/local/bin/pipenv", line 8, in <module>
    sys.exit(cli())
  File "/usr/local/lib/python3.10/dist-packages/pipenv/vendor/click/core.py", line 1128, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/pipenv/cli/options.py", line 57, in main
    return super().main(*args, **kwargs, windows_expand_args=False)
  File "/usr/local/lib/python3.10/dist-packages/pipenv/vendor/click/core.py", line 1053, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.10/dist-packages/pipenv/vendor/click/core.py", line 1659, in invoke
[...]

PR #5593, opened 2023-02-06
Bump required setuptools and virtualenv

oz123:
Following the discussion in #5582, it seems it's about time to update these versions.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,9 +21,9 @@ if sys.argv[-1] == "publish":
 
 required = [
     "certifi",
-    "setuptools>=36.2.1",
+    "setuptools>=67.0.0",
     "virtualenv-clone>=0.2.5",
-    "virtualenv",
+    "virtualenv>=20.17.1",
 ]
 extras = {
     "dev": [


matteius (approved):

setup.py
```diff
@@ -21,9 +21,9 @@
 
 required = [
     "certifi",
-    "setuptools>=36.2.1",
+    "setuptools>=67.0.0",
```
edmorley, line 24: Hi :-)

This setuptools version is very very new - and contains many breaking changes meaning that some projects are not yet able to adopt it.

Please can the minimum version be lowered slightly? Ideally to say `63.x` or older? 

(setuptools v64 contained a number of editable mode breaking changes, then v66 contained the PEP 440 breaking changes)

oz123: Why does this affect your project?

edmorley: The Heroku Python buildpack bootstraps pip, setuptools and wheel in the environment (to curated known-reliable versions), then installs pipenv using pip. Installing pipenv upgrades the curated setuptools version (currently `63.4.3`), since pipenv's minimum setuptools version is newer.

This means we either can't use the version of setuptools we want, or else will have to stay on older pipenv for now.

If pipenv actually needed setuptools 67.x then setting a very new minimum version is fine, however, I'm presuming pipenv does not actually need such a new version (and instead just something newer than the previous minimum of setuptools 36.x).

In general, if a package doesn't need the latest of a dependency, it's more friendly to others consuming the package is minimum version ranges are set slightly more conservatively. 

After all, users who want to use the latest of everything will still be able to do that, given that new installs of pipenv will get latest of everything without having to do anything, and even existing installs will get upgraded, if they are on a version older than 63.x (if the minimum is relaxed slightly).

I'm happy to open a PR if you would be open to it? :-)

matteius: > I'm happy to open a PR if you would be open to it? :-)

@edmorley That would be great -- I actually was worried about something like this going with the newest setuptools because I know a lot has been changing in that ecosystem.

Issue #5600, opened 2023-02-07
pipenv check looks at installed packages, not Pipfile.lock


jacobfelknor:
## Use Case

I would like to run `pipenv check` as a separate job from the build/test job inside a CI pipeline without rebuilding environment. I discovered that I must actually install all packages to a `pipenv` environment before using `pipenv check`. Ideally, I should be able to scan the dependencies inside `Pipfile.lock` without actually installing the whole environment. 

I believe its misleading that right now `pipenv` is just acting as a "proxy" to `safety`,  and by default checks an environment that may not match `Pipfile.lock`.  By using `pipenv check` the assumption should be that it is checking the environment specified in `Pipfile.lock` and if you need to check an environment that deviates, you use `safety` directly.

I've traced the behavior down to these lines:
https://github.com/pypa/pipenv/blob/8939c863464b23b5503569669d1c3f9ad31a498f/pipenv/core.py#L2900-L2902

Instead of generating the temp `requirements.txt` file from the current environment using `pip list`, can we instead generate the temp `requirements.txt` from `Pipfile.lock`? Something like

```python
# this command should also respect the wishes of the --dev argument, if provided. Unsure on specifics of implementation
target_venv_packages = run_command(
        _cmd + ["-m", "pipenv", "requirements"], is_verbose=project.s.is_verbose()
    )
```

## Workaround

I'm currently using the following workaround in my CI job, but would like to go through `pipenv` directly. 

```bash
pipenv requirements --dev | safety check --stdin
```

matteius:
`pipenv=2022.2.18` has made this the default behavior for `pipenv check`

jacobfelknor:
@matteius 

Thanks for taking this suggestion! I appreciate it. I have two quick pieces of feedback for even further improvement

1. The `--dev` argument has no effect on the `pipenv check` command. If user specifies the `--dev` flag, we should generate the temp requirements file from `pipenv requirements --dev`. 

2. `pipenv check` still prints "Checking installed packages for vulnerabilities...", regardless of it's using the installed environment with `--use-installed` or if it's using the dependencies from `Pipfile.lock`. If using default behavior, this should read something like "Checking Pipfile.lock packages for vulnerabilities..."

ryan-rozario:
@matteius Could I create a PR for the second suggestion since it is a simple condition that needs to be added.

matteius:
@ryan-rozario Yes please, that would be a big help.

PR #5607, opened 2023-02-12
Issue 5600 check command should be able to pull requirements from lock file

matteius:
### The issue

Fixes issue #5600 

### The fix

Supply the options to pull the requirements from the lock file categories.

### The checklist

* [X] Associated issue

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -390,57 +390,48 @@ It can be used to specify multiple categories also.
 Pipenv includes the `safety <https://github.com/pyupio/safety>`_ package, and will use it to scan your dependency graph
 for known security vulnerabilities!
 
+By default ``pipenv check`` will scan the virtualenv packages for what is installed and use this as the input to the safety command.
+To have ``pipenv check`` use the ``Pipfile.lock`` to determine package inputs simply pass ``--use-lock`` to use the default packages
+category, or pass the specific ``--categories`` you want to check against.
+
 Example::
 
-    $ cat Pipfile
-    [packages]
-    django = "==1.10.1"
+    $ pipenv install wheel==0.37.1
+    $ cat Pipfile.lock
+    ...
+    "default": {
+        "wheel": {
+            "hashes": [
+                "sha256:4bdcd7d840138086126cd09254dc6195fb4fc6f01c050a1d7236f2630db1d22a",
+                "sha256:e9a504e793efbca1b8e0e9cb979a249cf4a0a7b5b8c9e8b65a5e39d49529c1c4"
+            ],
+            "index": "pypi",
+            "version": "==0.37.1"
+        }
+    },
+    ...
 
-    $ pipenv check
-    Checking PEP 508 requirements...
-    Passed!
-    Checking installed package safety...
+    $ pipenv check --use-lock
+    ...
+    -> Vulnerability found in wheel version 0.37.1
+       Vulnerability ID: 51499
+       Affected spec: <0.38.1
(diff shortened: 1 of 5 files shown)


docs/advanced.rst
```diff
@@ -390,57 +390,48 @@ It can be used to specify multiple categories also.
...
-    Checking installed package safety...
-
-    33075: django >=1.10,<1.10.3 resolved (1.10.1 installed)!
-    Django before 1.8.x before 1.8.16, 1.9.x before 1.9.11, and 1.10.x before 1.10.3, when settings.DEBUG is True, allow remote attackers to conduct DNS rebinding attacks by leveraging failure to validate the HTTP Host header against settings.ALLOWED_HOSTS.
-
-    33076: django >=1.10,<1.10.3 resolved (1.10.1 installed)!
-    Django 1.8.x before 1.8.16, 1.9.x before 1.9.11, and 1.10.x before 1.10.3 use a hardcoded password for a temporary database user created when running tests with an Oracle database, which makes it easier for remote attackers to obtain access to the database server by leveraging failure to manually specify a password in the database settings TEST dictionary.
+By default ``pipenv check`` will scan the virtualenv packages for what is installed and use this as the input to the safety command.
+To have ``pipenv check`` use the ``Pipfile.lock`` to determine package inputs simply pass ``--use-lock`` to use the default packages
+category, or pass the specific ``--categories`` you want to check against.
```
oz123, line 395: IMO This new behaviour should be the default.
scanning of installed packages, means to me,  everything we install in a project.
Everything includes implicit packages which found in the lock file, and explicit packages we
specify in the Pipfile.

oz123 (changes requested):

docs/advanced.rst
matteius: Well the lock file covers also what it is in the `Pipfile`, but my concern with making this the default behavior is any existing workflows that may depend on scanning what is installed, since its totally possible and reasonable to sometimes install an extra package outside the view of Pipfile/Pipfile.lock   so I was reluctant to change the default behavior here.   Plus it would only scan the default packages and none of the other named categories without supplying the `--categories` argument.  

I do plan to support the ability to specify an env var in the future that would set the `--categories` implicitly, but I haven't gotten there yet, if specified it would complicate this further since there wouldn't be a way to fall back to what is actually installed.  Let me think on this some more.

news/5600.feature.rst
```diff
@@ -1 +1,3 @@
-Passing ``--use-lock`` to ``pipenv check`` will check again lock file default packages, and specifies ``--categories`` allows specifying which lockfile categories to use.
+Behavior change for ``pipenv check`` now checks the default packages group of the lockfile.
+Specifying ``--categories`` to override which categories to check against.
+Pass ``--use-installed`` to get the prior behavior of checking the packages actually installed into the environment.
```
oz123, line 3: How is categories specified?
Is it a quoted list, e.g.:
```
--categories="foo bar baz" 
```
And of course what is the separtor? space or coma, e.g. `--categories="foo, bar, baz"`
Or do we pass the flag multiple times?
```
--categories foo  --categories bar --categories= baz 
```
Looking at the code, it's `" "` separated, but I think it is appropriate to document this again in the news snippet.

oz123 (changes requested):

news/5600.feature.rst
matteius: @oz123 This PR made it more flexible so you can pass spaces or commas -- https://github.com/pypa/pipenv/pull/5606

oz123: OK. This is already documented in a past news item.

PR #5616, opened 2023-02-19
Split apart core.py

matteius:

### The issue

There are many issues related to code in core.py -- and the problem for me is the file is too unweidly to make sense of it, like utils.py once was.

### The fix

Use pycharm refactor to safely break apart the core.py

Removes 3 unused utilities and some constants in the process.

### The checklist

* [X] Associated issue


oz123 (approved):
Awesome work.
This will definitely help us maintain pipenv.

PR #5621, opened 2023-02-26
Doc revisions1

matteius:

### The issue

The docs are long in need of work -- this is revision1 which tackles cleaning up the contents from the `Basics` and `Advanced` sections and splitting them apart.  There were some bits left in advanced that I wasn't sure where to put them, so I left that section but renamed it to be other since most of the advanced information moved to other sections.  I also added some context on the `pipenv update` and `pipenv upgrade` commands rework.

### The fix

This is not the end-all be-all fix to the docs, no this is just progress.   I plan to iterate on this more over time -- splitting apart advanced as well.

I was getting frustrated with rst files and found I was having better luck making the new sections markdown.

* Removes the super old animation (we can make a new one, but I didn't want that old one holding us back longer)
* Removes references to python3.6 and python 2 -- also just old weird misinformation in general removed.
* Language updated to have a cohesive voice.

I am sure I missed some things and its definitely not perfect, but I think its an improvement.

### The checklist

* [ ] Associated issue


Pipfile
```diff
@@ -21,6 +21,7 @@ gunicorn = {version = "*", markers="sys_platform == 'linux'"}
 parse = "*"
 importlib-metadata = {version = "*", markers="python_version < '3.8'"}
 colorama= {version = "*", markers="sys_platform == 'win32'"}
+myst-parser = "*"
```
oz123, line 25: We also need to update requirements.txt in docs/. 
Otherwise the build in Read the docs will fail.

That is ofcourse only a middle step.
What I would really like is to have `tests`  and `docs` categories in Pipfile, and then we can only install `docs` category in the build.

oz123 (changes requested):

Pipfile
oz123: Also - there is a missing dependency:
```
myst-parser = {extras = ["linkify"], version = "*"}
```

oz123 (changes requested):

docs/dev/indexes.md
```diff
@@ -0,0 +1,81 @@
...
+
+    $ pipenv install --pypi-mirror <mirror_url>
+
+    $ pipenv update --pypi-mirror <mirror_url>
+
+    $ pipenv sync --pypi-mirror <mirror_url>
+
+    $ pipenv lock --pypi-mirror <mirror_url>
+
+    $ pipenv uninstall --pypi-mirror <mirror_url>
```
oz123, line 79: What is the difference between a “private index” and a “pypi-mirror”?
This is has to potential to be misunderstood.

matteius: Updated PR -- this was docs I moved from advanced, but I made it more clear.

docs/commands.md
```diff
@@ -0,0 +1,73 @@
...
+
+# Pipenv Commands
+
+The commands reference for pipenv (incomplete)
+
+## install
+
+``$ pipenv install`` is used for installing packages into the pipenv virtual environment
+and updating your Pipfile.
+
```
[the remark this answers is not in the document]
matteius: Updated.

oz123 (changes requested):

oz123 (approved):

PR #5650, opened 2023-04-12
Update docs for tilde-equals in package versions

### LINKED ISSUE #5594: installing package with ~= leads to installing the latest version of that package
### Issue description

pipenv install "tensorflow~=2.9" leads to the installation of the latest version of tensorflow which is 2.11.x I would expect that it would not update the major version

### Expected result

a version of tensorflow equal to 2.9.3 (for example)

### Actual result

tensorflow 2.11.x

### Steps to replicate
python 3.7
run on cli
pipenv install "tensorflow~=2.9"

ryan-rozario:
### The issue

Fixes #5594 

### The fix

Added clarity to the example about tilde-equals operator for package versions

### The checklist

* [x] Associated issue


docs/specifiers.md
```diff
@@ -8,9 +8,9 @@ You can specify versions of a package using the [Semantic Versioning scheme](htt
 
 For example, to install requests you can use:
 
-    $ pipenv install requests~=1.2
+    $ pipenv install requests~=1.1
 
-Pipenv will install version `1.2` and any minor update, but not `2.0`.
+Pipenv will install version `1.2` as it is a minor update, but not `2.0`.
```
matteius, line 13: I wonder if we should also provide an example of using tilde to allow updating the major version, example ~1 would allow upgrading to I guess, anything?   It seems like a strange use case now that I mention it.

ryan-rozario: I have added an example that allows micro version updates. Which would be more common usecase. 
I dont think we need an example for updating major version.

PR #5664, opened 2023-04-22
Reqlib pydantic

matteius:
### The issue

* Requirementslib is a nightmare library whose maintenance and code paths are very hard to detangle.
* It is based on an old attrs library which was a precursor to dataclasses, but there are better alternatives.
* The type annotations across pipenv are a mess, and that is true of its vendor'd libraries as well.   Yet there is zero type enforcement, so who knows if the annotations are meaningful, and they follow the style of type hints in comments before it became accepted into the method signatures.
* Did I mention Requirementslib is a nightmare library whose maintenance and code paths are very hard to detangle?  There are circular references with how the dataclasses were designed or manifested, and due to a long history of patching the library, there are many inefficiencies (ie: setup being run multiple times)  

### The fix

I am a big fan of pydantic's flavor of Python dataclasses https://docs.pydantic.dev/ and have been using them in other projects with great success.  Pydantic is also the dataclass library that powers FastAPI, and it is fast efficient and has a lot of bells and whistles for as light as the code base it.  So I took a shot at using it here as well and I have 40 hours into detangling larges parts of requirementslib in addition to converting it to pydantic and speeding up its execution.   I believe I have fixed some edge cases as well such as relative "." paths.

After chatting with @oz123 we have the goal of converting pythonfinder and plette to also use the pydantic dataclasses -- doing so should further help detangle requirementslib in followup future refactors as well since, it interacts with plette and the dictionary-like view its based on has some limiting factors.

This PR also adds the pythonfinder refactor to pydantic and totally drops the attrs libraries from pipenv.

### Next Steps
* [ ] Get peer review of this PR and factor in any feedback (where possible) -- requirementslib is still a bit delicate to modify even after this work, but its the start of ripping the bandaids off.
* [ ] Once sufficient buy in, port the changes over from this PR to requirementslib and Pythonfinder; release new versions.
* [ ] Vendor in the new versions to a PR and ship a faster more modern version of pipenv.

### The checklist


Pat242567 (approved):

pipenv/utils/resolver.py
```diff
@@ -29,8 +29,7 @@
 from pipenv.patched.pip._vendor import pkg_resources, rich
 from pipenv.project import Project
 from pipenv.vendor import click
-from pipenv.vendor.requirementslib import Requirement
-from pipenv.vendor.requirementslib.models.requirements import Line
+from pipenv.vendor.requirementslib.models.requirements import Line, Requirement
```
Pat242567, line 32:
```suggestion
from pipenv.vendor.requirementslib.models.requirements import Line, Requirement
```

pipenv/utils/resolver.py
```diff
@@ -347,6 +346,13 @@ def get_deps_from_req(
                 )
             ):
                 requirements = [v for v in getattr(setup_info, "requires", {}).values()]
+                if req.extras:
+                    for extra in req.extras:
+                        requirements.extend(
+                            v
```
Pat242567, line 352: pipenv/utils/resolver.py

matteius:
Replaced by:  https://github.com/pypa/pipenv/pull/5670

PR #5672, opened 2023-04-29
Requirementslib==2.3.0

matteius:
Thanks to @oz123 requirementslib==2.3.0 drops usage of vistir.

This PR brings pipenv up todate with requirementslib==2.3.0.  It does also address the remaining usage of vistir which I was going to defer to @oz123, but I needed to get these changes in place to update my refactor of requirementslib PR (targeting 3.0.0) and so he can help review this instead :-D

For some reason when I vendor'd this in, I got some random import errors and failures in building so I also ripped the bandaid off that was modifying the sysconfig for the import paths -- we finally have correctly imported everything so those patches/hacks are no longer needed.

### The checklist


pipenv/environment.py
```diff
@@ -24,7 +24,10 @@
 from pipenv.utils.indexes import prepare_pip_source_args
 from pipenv.utils.processes import subprocess_run
 from pipenv.utils.shell import make_posix, normalize_path
-from pipenv.vendor import click, vistir
+from pipenv.vendor import click
```
oz123, line 24: We now have normalize path twice.
```
from pipenv.utils.shell import make_posix, normalize_path
``` 
Should be dropped.

pipenv/utils/shell.py
```diff
@@ -466,3 +461,37 @@ def shorten_path(location, bold=False):
...
+def isatty(stream):
+    try:
+        is_a_tty = stream.isatty()
+    except Exception:  # pragma: no cover
+        is_a_tty = False
+    return is_a_tty
+
+
+@contextmanager
+def cd(path) -> Iterator[None]:
```
oz123, line 475: You can find `cd` in https://github.com/sarugaku/requirementslib/blob/400781f91d354f9f614f4d0b1e3c4187b3f3225c/src/requirementslib/fileutils.py#L24

oz123: It is already included in reqlib 2.3.0 no need to define it here.

oz123 (approved):

oz123 (changes requested):
Sorry, I meant to click request changes.

oz123 (approved):

PR #5688, opened 2023-05-16
Explore hashed URL fragments PEP 503

matteius:
### The issue

PEP 503 https://peps.python.org/pep-0503/

> The URL SHOULD include a hash in the form of a URL fragment with the following syntax: #<hashname>=<hashvalue>, where <hashname> is the lowercase name of the hash function (such as sha256) and <hashvalue> is the hex encoded digest.

### The fix

This should enable non pypi.org sources that are following PEP 503 to be able to leverage those package hashes without downloading the package.   This may also lead to a wider inclusion of package hashes in the lock file.

Note:  This branch is based on my fix-pyenv-asdf branch because I wanted to test it locally, and therefore should not be merged until that work is wrapped up first.

### The checklist

### DIFF
--- a/pipenv/project.py
+++ b/pipenv/project.py
@@ -850,14 +850,21 @@ class Project:
 
     def pipfile_sources(self, expand_vars=True):
         if self.pipfile_is_empty or "source" not in self.parsed_pipfile:
-            return [self.default_source]
+            sources = [self.default_source]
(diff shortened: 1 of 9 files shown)


matteius:
replaced by:  https://github.com/pypa/pipenv/pull/5701

PR #5699, opened 2023-05-20
Drop cerberus, it seems it's not used anywhere

oz123:


matteius (approved):

oz123:
I added 2 more commits, hence the requerst to review again.

tests/unit/test_utils_windows_executable.py
```diff
@@ -14,6 +14,7 @@
 
 
 @pytest.mark.utils
+@pytest.mark.skip_windows
```
matteius, line 17: Why are we skipping windows tests on windows, what is the back story here?

oz123: Ah. Good catch, I probably meant this:

```
@pytest.mark.skipif(os.name != "nt", reason="Windows file paths tested")
```

matteius (approved):
The test failure on one of the runners was network/github related.

PR #5721, opened 2023-06-06
Fix formatting in /docs

plgruener:
Fix formatting in docs, mainly links and codeblocks that were not adjusted in the rst->md transition, plus remove some duplication in the install section.

Feel free to change anything you don't like.


oz123 (approved):
Thanks!

PR #5730, opened 2023-06-11
Drop the --keep-outdated flag and --selective-upgrade flags

matteius:
Drop the --keep-outdated flag and --selective-upgrade flags that have been deprecated in favor of update/upgrade commands.

Note:
Let's plan on doing this at the same time we bring in the new requirementslib 3.x series.   
I want the next release before that to focus on bug fixes first.

### The issue & Fix

These flags were deprecated and more correct resolution phases were added to the update and upgrade command.

Use instead: 

* update -- calls upgrade under the hood, upgrades the relevant packages in the lock file and installs them.
* upgrade -- Only upgrade the relevant packages in the lock file (no install phase).


oz123 (approved):
YAY! removing all older code :-)

oz123 (changes requested):
News item is missing.

thisisrachelramos:
For those googling on what to do now that `--keep-outdated` is deprecated:

> The flag --keep-outdated has been deprecated for removal...Consider using the new `pipenv upgrade` command to selectively upgrade packages.

[source](https://github.com/pypa/pipenv/blob/8c8d3d1f8e0acee6b22a09f9ae9d80b093d4e7a1/pipenv/cli/options.py#L160)

PR #5769, opened 2023-07-01
Upgrade pydantic to 1.10.10

matteius:


cclauss:
No rush on this but it is good to see that v2 is shipping now...
* https://github.com/pydantic/pydantic/releases

matteius:
@cclauss pydantic v2 will require pipenv to look deep within itself at how its bundling dependencies because of the rust binaries aspect.

Issue #5794, opened 2023-07-17
Some Pipfiles trigger AttributeError: 'OutOfOrderTableProxy' object has no attribute '_body'


flichtenheld:
Be sure to check the existing issues (both open and closed!), and make sure you are running the latest version of Pipenv.

Check the [diagnose documentation](https://pipenv.pypa.io/en/latest/diagnose/) for common issues before posting! We may close your issue if it is very similar to one of them. Please be considerate, or be on your way.

Make sure to mention your debugging experience if the documented solution failed.

### Issue description

When the Pipfile has a certain TOML layout, specifically a `packages` section and later a `packages.<package>` section,
which are not directly adjacent to each other, an exception is thrown when trying to write to the Pipfile.

### Expected result

Pipfile is updated successfully.

### Actual result

```
Loading .env environment variables...
Installing distro...
Resolving distro...
Adding distro to Pipfile's [packages] ...
Traceback (most recent call last):
  File "/home/flichtenheld/.local/bin/pipenv", line 8, in <module>
    sys.exit(cli())
  File "/home/flichtenheld/.local/lib/python3.10/site-packages/pipenv/vendor/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/home/flichtenheld/.local/lib/python3.10/site-packages/pipenv/cli/options.py", line 58, in main
    return super().main(*args, **kwargs, windows_expand_args=False)
  File "/home/flichtenheld/.local/lib/python3.10/site-packages/pipenv/vendor/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/home/flichtenheld/.local/lib/python3.10/site-packages/pipenv/vendor/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/flichtenheld/.local/lib/python3.10/site-packages/pipenv/vendor/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/flichtenheld/.local/lib/python3.10/site-packages/pipenv/vendor/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/home/flichtenheld/.local/lib/python3.10/site-packages/pipenv/vendor/click/decorators.py", line 84, in new_func
    return ctx.invoke(f, obj, *args, **kwargs)
[...]

PR #5797, opened 2023-07-18
utils.toml: Handle tomlkit OutOfOrderTableProxy

flichtenheld:
We get this when we have subtables that do not
directly follow their parent table.

Fixes: #5794

### The fix

I think this issue and the fix show that we're way too deep into the tomlkit internals. I think
in the long run it would be much better to try to use a higher level interface... But I really don't
have time to go that deep into the code to find out whether that is even feasible. So for now
the minimal fix.

### The checklist

* [X] Associated issue


matteius:
Thanks for your contribution @flichtenheld !

PR #5804, opened 2023-07-21
 Cleanup secho

oz123:
Working on reusing rich from pipenv, so we can reuse vendored libs from it.

### DIFF
--- a/pipenv/routines/clean.py
+++ b/pipenv/routines/clean.py
@@ -2,6 +2,7 @@ import sys
 
 from pipenv.patched.pip._internal.build_env import get_runnable_pip
 from pipenv.routines.lock import do_lock
+from pipenv.utils import console, err
 from pipenv.utils.processes import run_command
 from pipenv.utils.project import ensure_project
 from pipenv.utils.requirements import BAD_PACKAGES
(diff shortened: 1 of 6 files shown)


matteius:
Does rich already handle the NO_COLOR env var?

oz123:
Yes. It does:

https://rich.readthedocs.io/en/stable/console.html?highlight=NO_COLOR#environment-variables

pipenv/utils/pip.py
```diff
@@ -174,20 +165,12 @@ def pip_install_deps(
         pip_command.extend(["-r", normalize_path(file.name)])
         if project.s.is_verbose():
             msg = f"Install Phase: {'Standard Requirements' if file == standard_requirements else 'Editable Requirements'}"
-            click.echo(
-                click.style(msg, bold=True),
-                err=True,
-            )
+            err.print(msg, style="bold")
```
matteius, line 168: Is this conversion 1:1?

oz123: Yes. The API of rich is easy to get.

pipenv/routines/clean.py
```diff
@@ -2,6 +2,7 @@
 
 from pipenv.patched.pip._internal.build_env import get_runnable_pip
 from pipenv.routines.lock import do_lock
+from pipenv.utils import console, err
```
matteius, line 5: But where does this import come from?

oz123: I initialized them here:
https://github.com/pypa/pipenv/blob/8240ea73d2e348960f97ca01e7ed0f169876fb0c/pipenv/utils/__init__.py

matteius (approved):

PR #5806, opened 2023-07-22
Remove --skip-lock flag

### LINKED ISSUE #5805: Finish dropping --skip-lock so we can move forward with migrating away from Requirementslib
### Issue description

A simplified architecture for pipenv would be:

image

`--skip-lock` is problematic because it requires also being able to generate the install lines from the Pipfile.  Additionally given the other bugs around this flag, the recommendation is if you want to skip the lock and install something for testing just run `pipenv run pip install`  If you want to do this from the Pipfile run `pipenv requirements` to generate a requirements file and then run `pipenv run pip install -r requirements.txt`   

This will also improve the security of pipenv because some users may be using the `--skip-lock` flag in their CI builds.

matteius:
Fixes #5805 

### The checklist

* [X] Associated issue


oz123 (approved):
This change is more than welcome :-)

PR #5807, opened 2023-07-22
Fix ruff-complaints

matteius:
Fix all ruff warnings.

@matteius should be OK now.


oz123 (approved):
@matteius since you opened the PR I can't add you as a reviewer.

PR #5825, opened 2023-08-06
Some docs cleanups

jeffwidman:
1. `pypi.python.org` now redirects to `pypi.org`, so let's point folks towards that to save them the redirect.
2. Travis CI has struggled mightily ever since they were acquired by a PE firm. Most open source projects have moved away from them.
3. Update the python version in the example so it doesn't look quite so dated.
4. Use the `name` key for `sources` as that's now [required](https://github.com/pypa/pipenv/discussions/5370#discussioncomment-3701061).

I didn't include a news fragment as I thought the changes were too trivial to list.

I also didn't take the time to regenerate the lockfiles after changing the `Pipfile`... that means the hashes are going to be wrong, but for the purposes of docs I think that's fine... no human will notice, and if they do run `pipenv` against the example and are surprised by the outcome they're welcome to submit a PR fixing that too. But for now I think tis fine. Certainly this PR leaves things better then they were.

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -77,9 +77,9 @@ file very easily.
 Let's take this ``Pipfile``::
 
     [[source]]
-    url = "https://pypi.python.org/simple"
-    verify_ssl = true
     name = "pypi"
+    url = "https://pypi.org/simple"
+    verify_ssl = true
 
     [packages]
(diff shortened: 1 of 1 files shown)


matteius:
Thanks Jeff!

PR #5836, opened 2023-08-19
Update build to use exclusively pyproject.toml

### LINKED ISSUE #5837: Update build to use exclusively pyproject.toml
### The goal: modernize build

- consolidates all of `setuptools` metadata and configuration as well as `coverage` configuration within `pyproject.toml`.
- removes deprecated `setup.cfg` and `setup.py`.
- updates `run-tests.sh` to use  `pyproject.toml` instead of `setup.cfg` for `coverage` configuration.

kalebmckale:

### The goal: modernize build

- consolidates all of `setuptools` metadata and configuration as well as `coverage` configuration within `pyproject.toml`.
- removes deprecated `setup.cfg` and `setup.py`.
- updates `run-tests.sh` to use 	`pyproject.toml` instead of `setup.cfg` for `coverage` configuration.

### The checklist

* [X] Associated issue: #5837 

Resolves #5837


pyproject.toml
```diff
@@ -1,11 +1,109 @@
...
+
+
+[tool.setuptools.packages.find]
+where = ["."]
+exclude = ["tests*", "tests.*", "tasks*", "tasks.*"]
+
+
+[tool.setuptools.package-data]
+pipenv = ["LICENSE", "NOTICES"]
+pipenv.patched.safety = ["VERSION", "safety-policy-template.yml"]
```
kalebmckale, line 75: It seems I will need to do more research into how console scripts formatted like this are implemented in `pyproject.toml` format.

kalebmckale:
Now it’s failing for things unrelated to my changes. @matteius ?

kalebmckale:
Seems like everything is working except the versioning. I expect this is because `setuptools_scm` cannot see the `git tag`. I figured this might come up since there’s already an in-place script that handles updating version, but thought it might still work. Nevertheless, using `setuptools_scm` isn’t a requirement to make this work and really unnecessary since there seems to be an existing solution that doesn’t require human intervention. So, if it ain’t broke…. I’ll make the changes to pull the version from the `__version__.py` file instead.

news/5837.behavior.rst
```diff
@@ -0,0 +1,47 @@
+{% if top_line %}
```
matteius, line 1: I don't understand the conditional logic in this news fragment ...

kalebmckale: Oh! I just took the template and replaced the text. I was a bit confused, too, since I didn't recall there being so much text last time. I suspect now that is not what I should have done. I can fix it.

oz123:
@kalebmckale please pull the changes from the master branch. This will fix the ruff errors.

oz123 (changes requested):
It's not a behaviour change. Please add this to feature. 
(e.g. use XXXX.feature.rst).

pyproject.toml
```diff
@@ -1,7 +1,93 @@
...
+where = ["."]
+exclude = ["tests*", "tests.*", "tasks*", "tasks.*"]
+
+
+[tool.setuptools.package-data]
+"*" = ["LICENSE", "NOTICES"]
+"pipenv.patched.safety" = ["VERSION", "safety-policy-template.yml"]
+"pipenv.patched.pip._vendor.certifi" = ["*.pem"]
+"pipenv.patched.pip._vendor.requests" = ["*.pem"]
+"pipenv.patched.pip._vendor.distlib._backport" = ["sysconfig.cfg"]
```
oz123, line 75: We can remove this line. distlib   no longer includes _backport.

oz123 (changes requested):

news/5837.behavior.rst
kalebmckale: Just to be clear. What's the "standard" format for news file? I don't have my old one to refer to, and I assume using the template was incorrect.

P.S. Thank you for your patience with me!

cclauss:
Please rebase.

kalebmckale:
> Please rebase.

Sorry I couldn't get to it until tonight.

pyproject.toml
kalebmckale: @oz123 You okay with changes?

pyproject.toml
```diff
@@ -1,7 +1,92 @@
...
+  "t64.exe",
+  "w32.exe",
+  "w64.exe",
+]
+"pipenv.vendor.ruamel" = ["yaml"]
+
+
+[tool.setuptools.dynamic]
+version = {attr = "pipenv.__version__"}
+
```
oz123, line 86: Please add __version_ to  gitgnore.

oz123 (changes requested):

pyproject.toml
kalebmckale: Sure. I mean `__version__.py` is a file that belongs to the repo, but I can see how you might not want someone to accidentally overwrite and commit it.

kalebmckale: @oz123 I just want to make sure this isn't going to break
```python
if commit:
    ctx.run(f"git add {version_file.as_posix()}")
    log("Committing...")
    ctx.run('git commit -s -m "Bumped version."')
```
in `release.py`.

I'm not certain if ignored files have to be force added the first time or every time after.

matteius:
Hey @kalebmckale if you want to join the unofficial Python Developers slack group https://pyslackers.com/web  we can chat more in there -- I am thinking we could bounce ideas off each-other about the currently broken editable/local file installs.

kalebmckale:
Ok @matteius... um never used slack so how do I find you on there? 🙂

pyproject.toml
cclauss: Perhaps I am confused about `pipenv.__version__` vs. `pipenv.__version__.__version__`...

The `pip` project stores `pip.__version__` in a `pip.__init__.py` file while this project stores the version in a separate `pipenv.__version__.py` file.
* https://github.com/pypa/pip/blob/main/src/pip/__init__.py#L3 vs.
* https://github.com/pypa/pipenv/blob/main/pipenv/__version__.py#L5

Will the following lines create the variable `pipenv.__version__` or `pipenv.__version__.__version__`?
```python
dynamic = ["version"]
# and
version = {attr = "pipenv.__version__"}
```
Was this pull request created with a tool like https://pypi.org/project/pyproject-migrator or was it done manually?

kalebmckale: It was done manually. `__version__.py` files are pretty standard for a number of projects. The import of `__version__` from `__version__.py` is already a part of the package. What the lines referenced do is read the version from the file for the purpose of the build. If you review the build/test artifacts, you can see this is doing exactly what is expected.

kalebmckale: @oz123 Do you know if the script will still work properly if the file is added to `.gitignore`?

oz123: I don't know if it will break. Need to test it, of course. I asked to remove it, because I thought that file is auto generation by setuptools_scm if your configuration.

kalebmckale: Latest version doesn't use `setuptools_scm`. Thought I'd use it, but when investigated the current release process, I realized it would require much more work than what I intended as the scope of this issue.

kalebmckale: @oz123 We can consider later to move to `setuptools_scm`, but it would require changing the entire automated-release process.

Right now, `__version__.py` is used to tell both the package (via `__init__.py`) and the build what the current version is. It's a file the automated-release process doesn't create but rather updates each time.

kalebmckale: Adding to `.gitignore` could be innocuous if `git add` without the `--force` option will update the existing file. Otherwise, one would also have to change the automated-release file with this option. Since this package actually includes `__version__.py` in its repo, my opinion would be to leave as-is until such a time that one changes the release process.

kalebmckale:
Rebased on v2023.8.26.

kalebmckale:
Added a new commit to test if adding `__version__.py` to `.gitignore` will break the release process. I suspect this will not be tested until after the merge has taken place and `release.py` is called since I could find no reference in the repo files themselves to call this script (and assume it’s a setting in the administrative CI/CD process).

oz123:
@kalebmckale I am sorry. I thought we need to exclude `__version__` as that is what one should do with https://github.com/pypa/setuptools_scm. However, you are using https://setuptools.pypa.io/en/latest/build_meta.html.
build_meta does not care about storing the version in git. Once again, than you for taking the time for contributing and apologies for the confusion (and extra work required from you).

.gitignore
```diff
@@ -63,6 +63,7 @@ wheels/
 .installed.cfg
 *.egg
 MANIFEST
+__version__.py
```
matteius, line 66: This needs to be removed

matteius (approved):

kalebmckale:
Just noticed this in `release.py`
```python
@invoke.task
def build_dists(ctx):
    drop_dist_dirs(ctx)
    py_version = ".".join(str(v) for v in sys.version_info[:2])
    env = {"PIPENV_PYTHON": py_version}
    with ctx.cd(ROOT.as_posix()), temp_environ():
        executable = ctx.run(
            "python -c 'import sys; print(sys.executable)'", hide=True
        ).stdout.strip()
        log("Building sdist using %s ...." % executable)
        os.environ["PIPENV_PYTHON"] = py_version
        ctx.run("pipenv install --dev", env=env)
        ctx.run("pipenv run pip install -e . --upgrade --upgrade-strategy=eager", env=env)
        log("Building wheel using python %s ...." % py_version)
        ctx.run("pipenv run python setup.py sdist bdist_wheel", env=env)
```

I know during testing the package is being built and installed, or it at least appears to do so, but here it looks to still use `setup.py` for build.

kalebmckale:
Found in `Makefile`, too.
```Makefile
.PHONY: build
build: install-virtualenvs.stamp install.stamp
	PIPENV_PYTHON=3.7 pipenv run python setup.py sdist bdist_wheel
```

matteius:
That is what I was afraid of -- if we merge this, we will break the pypi upload of the built wheel capabilities, I suspect.

cclauss:
> Successfully created virtual environment!
Virtualenv location: /Users/runner/.local/share/virtualenvs/pipenv-XC1sj3ZQ
Your Pipfile.lock (4d9c54) is out of date.  Expected: (60755f).
Usage: pipenv install [OPTIONS] [PACKAGES]...

What would happen if you did `git rm Pipfile.lock`?  Would the GitHub Action test process recreate that file?

matteius:
I don't want to remove the Pipfile.lock from the checked in code though-- there will be a path forwards on this, even if it means merging into a different pipenv branch so I can run the lock phase.

kalebmckale:
I was able to lock on my iPad using Python 11, then I realized that would probably not be good since the minimum Python version is 3.7. I’ve been working for days to get other options. I can get Python 3.9 and looking to see if I can get Python 3.7. If there’s another solution sooner, we should go with that. However, I may be able to eventually get it done.

kalebmckale:
@matteius Another option is to include an updated locked Pipfile in another merge/release. Adding the two dev packages wouldn’t interfere with other changes.

cclauss:
If the `minimum Python version is 3.7` then Python 3.11 is acceptable.

cclauss:
* #5901 (merged) has added those two dependencies to `Pipfile` and `Pipfile.lock` so you should be able to rebase this PR to get those changes.

pyproject.toml
```diff
@@ -1,7 +1,92 @@
 [build-system]
-requires = ["setuptools", "wheel"]
+requires = ["setuptools>=67.0.0"]
 build-backend = "setuptools.build_meta"
 
+
+[project]
+name = "pipenv"
+authors = [
+  {name = "Pipenv maintainer team", email = "distutils-sig@python.org"},
```
cclauss, line 9: https://github.com/pypa/pipenv/pull/5900#pullrequestreview-1606631997

kalebmckale: I think this should be a new issue. At some point, issue drift starts to occur and pull requests never get merged. The focus of this issue was to merely move content from `setup.py` and `setup.cfg` to `pyproject.toml`. That required a few additional changes where the files were referenced and due to the change in associated build processes. I don't believe, however, we should hold up the PR or require another run of CI/CD for this at this time, given how long the PR has been open and how many times it has been rebased. Additionally, I don't believe there is a known value to substitute at this time.

kalebmckale: #5906

matteius: I wasn't calling out we had to fix it to merge this work, just I noticed it in reviewing it.

pyproject.toml
```diff
@@ -1,7 +1,92 @@
 [build-system]
-requires = ["setuptools", "wheel"]
```
matteius, line 2: Are you sure we don't need to require wheel?

kalebmckale: Yes. Looking for documentation, but setuptools after (I want to say) version 61.0.0 includes wheel.

kalebmckale: As a side note, but not something I ever do with projects because of readability, if using `setuptools` as used here, one doesn’t technically need the `build-system` section since `pyproject` uses it by default. Of course, it’s much easier to just include it and more explicit to know what version of `setuptools` you’re using.

kalebmckale: > The setuptools package implements the build_sdist command and the wheel package implements the build_wheel command; **the latter is a dependency of the former exposed via [PEP 517](https://peps.python.org/pep-0517/)** hooks.

Ref: https://setuptools.pypa.io/en/latest/build_meta.html

cclauss: For a follow-up PR... Add [pyproject-fmt](https://pypi.org/project/pyproject-fmt/) and [validate-pyproject](https://pypi.org/project/validate-pyproject/) to [.pre-commit-config.yaml](https://github.com/pypa/pipenv/blob/main/.pre-commit-config.yaml).

kalebmckale: @cclauss Please add this to issue #5906 since this also deals with a follow-up to this issue.

pyproject.toml
kalebmckale: @oz123 Your review is still pending.

pyproject.toml
oz123, line 9: The email is wrong. Please correct the email address. @matteius should our email adresses be here?

kalebmckale: Alright. I've been working on this PR for two weeks now and rebased multiple times. Someone else is welcome to take this on, but I've surpassed the amount of work I'm willing to put in at this time.

matteius: We will update the email address outside this PR -- thanks for your effort on this @kalebmckale

matteius (approved):

PR #5863, opened 2023-08-23
Caching properties to deduplicate sources

kalebmckale:
Without caching, the Resolver duplicates the sources listed in the `Pipfile` multiple times (since the property re-adds them every time they're called) in the arguments sent to the `pip` parser.

Removing unused command-line option `--debug`

### The issue

While diagnosing another issue, I found through `print` debugging that the multiple calls to these properties, `pip_args` and `pip_options`, ended up duplicating the sources listed in the `Pipfile` multiple times for the `pip` parser.

Related to the resolver, the `--debug` command-line option for `pipenv-resolver` is unused.

### The fix

To fix this, I replaced `property` decorator with `cached_property`, which was already imported in the module. As best as I can tell, these properties (and a number of other ones) only need to be evaluated once instead of every single time they are called. Caching not only fixes this duplication but will also lead to faster, more efficient code.

If a debug option is later implemented, the command-line option can be added at that time.

### The checklist

* [ ] Associated issue

### DIFF
--- a/pipenv/resolver.py
+++ b/pipenv/resolver.py
@@ -40,7 +40,6 @@ def get_parser():
         action="store",
         default=None,
     )
-    parser.add_argument("--debug", action="store_true", default=False)
     parser.add_argument("--system", action="store_true", default=False)
     parser.add_argument("--parse-only", action="store_true", default=False)
     parser.add_argument(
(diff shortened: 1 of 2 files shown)


matteius (approved):

kalebmckale:
@matteius What type of change would this be considered for my news file?

matteius:
@kalebmckale I'd be fine with `bugfix`.

matteius:
I am going to include a news fragment on your behalf so we can include this fix in the release.

kalebmckale:
Thanks!

Issue #5858, opened 2023-08-22
2023.8.19 through to 2023.8.22 break editable package installation


stewartmiles:
### Issue description

Editable package installation appears to be broken from pipenv 2023.8.19 onwards.
For example, given a package in the directory `test-hello` under the pipenv root installing the package with `pipenv install -e ./test-hello` fails.

### Expected result

pipenv 2023.7.23 and prior versions would install editable packages in Pipfile and in the venv.

### Actual result

Here's a log of package installation across recent versions of pipenv.
[try_pipenv_versions.log](https://github.com/pypa/pipenv/files/12411124/try_pipenv_versions.log)

For example:
```
+ pipx install pipenv==2023.8.19
creating virtual environment...
installing pipenv from spec 'pipenv==2023.8.19'...
done! ✨ 🌟 ✨
  installed package pipenv 2023.8.19, installed using Python 3.8.10
  These apps are now globally available
    - pipenv
    - pipenv-resolver
+ git clean -dfx
Removing .venv/
Removing applications/test-hello/test_hello.egg-info/
Removing test-hello/test_hello.egg-info/
+ echo '=== Using pipenv version 2023.8.19 ==='
=== Using pipenv version 2023.8.19 ===
+ pipenv sync
Creating a virtualenv for this project...
Pipfile: /home/stewart/src/pipenv_editable_install_repro/Pipfile
Using /usr/bin/python3.8 (3.8.10) to create virtualenv...
created virtual environment CPython3.8.10.final.0-64 in 48ms
  creator CPython3Posix(dest=/home/stewart/src/pipenv_editable_install_repro/.venv, clear=False, no_vcs_ignore=False, global=False)
  seeder FromAppData(download=False, pip=bundle, setuptools=bundle, wheel=bundle, via=copy, app_data_dir=/home/stewart/.local/share/virtualenv)
    added seed packages: pip==23.2.1, setuptools==68.0.0, wheel==0.41.1
  activators BashActivator,CShellActivator,FishActivator,NushellActivator,PowerShellActivator,PythonActivator

✔ Successfully created virtual environment!
Virtualenv location: /home/stewart/src/pipenv_editable_install_repro/.venv
Installing dependencies from Pipfile.lock (db4242)...
To activate this project's virtualenv, run pipenv shell.
Alternatively, run a command inside the virtualenv with pipenv run.
All dependencies are now up-to-date!
+ install_test_and_uninstall_package ./test-hello
+ path=./test-hello
[...]

PR #5870, opened 2023-08-26
Attempt to fix the local/editable file installs

matteius:
Attempt to fix the local/editable file installs which broke when we refactored away from requirementslib.

### The issue

Fixes (??) Issue #5858

### The fix

### The checklist

* [ ] Associated issue

### DIFF
--- a/pipenv/utils/funktools.py
+++ b/pipenv/utils/funktools.py
@@ -9,7 +9,7 @@ import subprocess
 import time
 import warnings
 from functools import partial
-from itertools import count, islice, tee
+from itertools import count, islice
 from typing import Any, Iterable
 
 DIRECTORY_CLEANUP_TIMEOUT = 1.0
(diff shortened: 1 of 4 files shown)


pipenv/utils/funktools.py
```diff
@@ -62,18 +62,13 @@ def unnest(elem: Iterable) -> Any:
...
-        target = elem
-    if not target or not _is_iterable(target):
-        yield target
-    else:
-        for el in target:
+        for el in elem:
             if isinstance(el, Iterable) and not isinstance(el, str):
-                el, el_copy = tee(el, 2)
-                yield from unnest(el_copy)
+                yield from unnest(el)
```
oz123, line 67: This makes total sense to use recursion here. I wonder how it was not done before.

oz123 (approved):
LGTM.

PR #5873, opened 2023-08-27
ruff rules SIM

cclauss:

Some code simplifications made by `ruff --select=SIM --fix .`
* https://beta.ruff.rs/docs/rules/#flake8-simplify-sim

### The issue

### The fix

### The checklist

* [ ] Associated issue

### DIFF
--- a/pipenv/utils/pipfile.py
+++ b/pipenv/utils/pipfile.py
@@ -1,3 +1,4 @@
+import contextlib
 import io
 import itertools
 import os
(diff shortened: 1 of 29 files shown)


matteius (approved):

matteius:
@cclauss I think things are in a much more stable state now from my recent refactor, but I did introduce a conflict fixing things with this PR.  If that can be resolved I think this can be merged.

pipenv/utils/pipfile.py
```diff
@@ -55,11 +56,10 @@ def find_pipfile(max_depth=3):
     for c, _, _ in walk_up(os.getcwd()):
         i += 1
 
-        if i < max_depth:
-            if "Pipfile":
-                p = os.path.join(c, "Pipfile")
-                if os.path.isfile(p):
-                    return p
+        if i < max_depth and "Pipfile":
```
cclauss, line 59: This one bugs me in both the current code and in this PR... `if "Pipfile":` is ALWAYS True.

matteius: We should reduce stuff like that -- I've been trying to when I come across it.

PR #5877, opened 2023-08-28
Add PIPENV_REQUESTS_TIMEOUT and use it for requests session

birdhackor:

### The issue

When I was using pipelinev in an environment with network issues, I needed to increase the timeout setting for the connection to complete properly, but I found that the default behavior was set to 10 seconds and could not be changed, so I increased it.

### The fix

Add a new ENV config for timeout, so that user can change it.

### The checklist

* [x] Associated issue


matteius (approved):

matteius:
@birdhackor Please also update corresponding documentation for this new env var.

birdhackor:
> @birdhackor Please also update corresponding documentation for this new env var.

Seems `docs/configuration.md` use autoclass to generate documentation. I think it will automaticly use docstrings as documentation.

Am I missing anything that needs to be modified?

matteius:
@birdhackor Just a new fragment then, and there is a conflict to resolve -- also there is a second spot now in that method where there is a conflict where you'll want to pass the configured var instead of 10.

birdhackor:
@matteius I create a new PR #5907 (Re-send due to force push).

PR #5892, opened 2023-08-30
Top level Pipfile sys_platform markers should be transitive

matteius:
### The issue

I want transitive platform locking to work better.

### The fix

By following the resolution chain (where the package came from) we can get back to the Pipfile entry and apply the same sys_platform markers the user intended.

Additionally add a top level platform_machine specifier that works the same way.

Note:  Adding directly to the markers will behave different with the pip resolver than adding them as distinct specifier keys in the Pipfile.

### The checklist

* [ ] Associated issue

### DIFF
--- a/pipenv/utils/resolver.py
+++ b/pipenv/utils/resolver.py
@@ -450,9 +450,10 @@ class Resolver:
     def _fold_markers(self, dependency_tree, install_req):
         comes_from = dependency_tree[install_req.name]
         if comes_from == "Pipfile":
-            if isinstance(self.pipfile_entries[install_req.name], dict):
-                sys_platform = self.pipfile_entries[install_req.name].get("sys_platform")
-                markers = self.pipfile_entries[install_req.name].get("markers")
+            pipfile_entry = self.pipfile_entries.get(install_req.name)
+            if pipfile_entry and isinstance(pipfile_entry, dict):
+                sys_platform = pipfile_entry.get("sys_platform")
+                markers = pipfile_entry.get("markers")
                 if sys_platform:
                     sys_platform = f"sys_platform {sys_platform}"
                 if markers and sys_platform:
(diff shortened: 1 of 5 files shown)


pipenv/utils/resolver.py
```diff
@@ -447,9 +447,49 @@ def resolve(self):
...
+        markers = pipfile_entry.get("markers")
+
+        if sys_platform:
+            sys_platform = f"sys_platform {sys_platform}"
+        if platform_machine:
+            platform_machine = f"platform_machine {platform_machine}"
+
+        combined_markers = [
+            f"({marker})"
+            for marker in [sys_platform, markers, platform_machine]
```
cclauss, line 462: Nit but tuples take less RAM than lists.
```suggestion
            for marker in (sys_platform, markers, platform_machine)
```

pipenv/utils/resolver.py
```diff
@@ -447,9 +447,49 @@ def resolve(self):
...
+    def _fold_markers(self, dependency_tree, install_req):
+        comes_from = dependency_tree[install_req.name]
+
+        if comes_from == "Pipfile":
+            pipfile_entry = self.pipfile_entries.get(install_req.name)
+            if pipfile_entry and isinstance(pipfile_entry, dict):
+                return self._get_pipfile_markers(pipfile_entry)
+        else:
+            markers = self._fold_markers(dependency_tree, comes_from)
+            if markers:
```
cclauss, line 477:
```suggestion
            if markers := self._fold_markers(dependency_tree, comes_from):
```

cclauss (approved):
LGTM.  A few minor ideas...

pipenv/utils/resolver.py
matteius: I guess we can use the walrus operator, but it won't work in 3.7 so we essentially can hold this back until we drop 3.7, but I have mixed feelings about it with the open reports about hash collection against some private indexes.  🤔

cclauss: I am fine to remain Walrus-less.  My two freelance clients are both Python 3.11-only so I forgot about 3.7 in this case.

PR #5895, opened 2023-08-30
Apply patch for install_search_all_sources = True

### LINKED ISSUE #5743: `install_search_all_sources = true` is not being respected in latest version of pipenv
### Issue description

The latest version of pipenv `v2023.6.18` has caused our repos to fail install. This is a regression from the solution we found in [#5543](https://github.com/pypa/pipenv/issues/5543) - see that issue for more detailed information.

Downgrading to `v2023.6.12` makes the repos install correctly. 

The issue appears to be caused by this change in [PR #5737](https://github.com/pypa/pipenv/pull/5737/files#)

image

The change effectively nullifies the `install_search_all_sources = true` setting in the `Pipfile` and results in transitive dependencies published on a private registry not being found

### Expected result

package installs should be able to locate sub-dependencies when they are not defined in the `Pipfile` but `install_search_all_sources = true` is set.

### Actual result

Dependency Installs are failing to find sub-dependencies that are not defined in the `Pipfile` and are published on a private registry despite the inclusion of `install_search_all_sources = true` in the `Pipfile`

### Steps to replicate

See issue [#5543](https://github.com/pypa/pipenv/issues/5543)  for more details on reproducing

### LINKED ISSUE #5781: Another possible `install_search_all_sources = true` regression
There is this associated issue: https://github.com/pypa/pipenv/issues/5743
It was closed, but I think this is a little different so figured I'd raise a new issue for it...

### Issue description

In pipenv version 2023.6.18 (and onwards), attempting to run `pipenv sync --dev` (`--dev` flag likely unimportant) raises an error only when the following is in the pipfile:
```
[pipenv]
install_search_all_sources = true
```
I have a number of dependencies in private repos and all have their repos explictly specified, as well as any sub-dependencies which are also in our private repo. The issue is NOT with locking. This works perfectly fine. The issue is only when calling `sync`.

### Expected result

I'd not expect this flag to affect syncing, least of all cause it to fail.

### Actual result

```
Installing initially failed dependencies...
Looking in indexes: https://pypi.org/simple, <private_index_url_1>, <private_index_url_2>, <private_index_url_1>, <private_index_url_2>
...
[pipenv.exceptions.InstallError]: ERROR: Could not find a version that satisfies the requirement <private_repo>==0.1.dev19+g92d125f (from versions: none)
[pipenv.exceptions.InstallError]: ERROR: No matching distribution found for <private_repo>==0.1.dev19+g92d125f
ERROR: Couldn't install package: [Requirement(_name='alembic', vcs=None, req=NamedRequirement(name='alembic', version='==1.11.1', req=Requirement.parse('alembic==1.11.1'), extras=[], editable=False, _parsed_line=<Line (editable=False, name=alembic, path=None, uri=None, extras=(), markers=None, vcs=None, specifier===1.11.1, ...
```

(yes, the private repo urls are listed twice like that... Looks a bit weird...)

### Steps to replicate

See description.
Locking the file works fine in all versions. All the dependencies are resolved correctly and the lock file looks correct (private dependencies have the correct index urls in the lock file).
Running `pipenv sync --dev` in version 2023.6.18 onward fails when `install_search_all_sources = true` is present, but passes when it isn't, whereas 2023.6.12 has no issues.

This isn't the biggest issue as it's easy to remove and we don't need it since we have all the dependencies specified, but it did take a while to figure out why it was happening, and I'm mostly confused as to why this is even happening...

matteius:

### The issue

A while back I broke install_search_all_sources = True and I've realized this patch.

### The fix

Search scopes gets used for resolver and installer, but install we invoke pip interface.   This gives a way to ensure index restricted packages for the resolver, but bypass that during install.

Fixes #5743 
Fixes #5781 

### The checklist

* [X] Associated issue

### DIFF
--- a/news/5895.bugfix.rst
+++ b/news/5895.bugfix.rst
@@ -1,0 +1,1 @@
+Apply patch for install_search_all_sources = True functionality.
--- a/pipenv/patched/pip/_internal/models/search_scope.py
+++ b/pipenv/patched/pip/_internal/models/search_scope.py
@@ -20,7 +20,7 @@ class SearchScope:
     Encapsulates the locations that pip is configured to search.
     """
 
-    __slots__ = ["find_links", "index_urls", "no_index", "index_lookup"]
+    __slots__ = ["find_links", "index_urls", "no_index", "index_lookup", "index_restricted"]
 
     @classmethod
     def create(
@@ -29,6 +29,7 @@ class SearchScope:
         index_urls: List[str],
         no_index: bool,
         index_lookup: Optional[Dict[str, List[str]]] = None,
+        index_restricted: bool = False,
     ) -> "SearchScope":
         """
         Create a SearchScope object after normalizing the `find_links`.
@@ -64,6 +65,7 @@ class SearchScope:
             index_urls=index_urls,
             no_index=no_index,
             index_lookup=index_lookup,
+            index_restricted=index_restricted,
         )
 
     def __init__(
@@ -72,11 +74,13 @@ class SearchScope:
         index_urls: List[str],
         no_index: bool,
         index_lookup: Optional[Dict[str, List[str]]] = None,
+        index_restricted: bool = False,
     ) -> None:
         self.find_links = find_links
         self.index_urls = index_urls
         self.no_index = no_index
         self.index_lookup = index_lookup if index_lookup else {}
+        self.index_restricted = index_restricted
 
     def get_formatted_locations(self) -> str:
         lines = []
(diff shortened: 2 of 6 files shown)


monkeyman192:
I pulled this branch and built it locally using `python -m build --wheel`. I then installed the current latest released pipenv build and added
```
[pipenv]
install_search_all_sources = true
```
to the bottom of my Pipfile, and then ran `pipenv sync --dev`, and it failed.
I then uninstalled the latest release and installed pipenv from the wheel I just built and it sync'd successfully, so I'd say this looks good, at least for this particular case.

PR #5900, opened 2023-08-31
Migrate metadata from setup.py to setup.cfg

cclauss:
Perhaps a step towards #5836

This pull request was created using https://pypi.org/project/setuptools-py2cfg plus manual modifications.

https://setuptools.pypa.io/en/latest/userguide/quickstart.html#basic-use
> We recommend users expose as much as possible configuration in a more declarative way via the [pyproject.toml](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html) or [setup.cfg](https://setuptools.pypa.io/en/latest/userguide/declarative_config.html), and keep the setup.py minimal with only the dynamic parts (or even omit it completely if applicable).

---
___After___ a pull request like this is tested and merged, I usually do a follow-up PR like...

### PEP 621: Migrate settings from setup.cfg into pyproject.toml
Migrate settings from `setup.cfg` into `pyproject.toml` using [ini2toml](https://pypi.org/project/ini2toml) to do the file conversion and running [pyproject-fmt](https://pypi.org/project/pyproject-fmt) and then [validate-pyproject](https://github.com/abravalheri/validate-pyproject) in pre-commit to validate the results.
* Currently, flake8 is not yet compatible with `pyproject.toml` so move its settings into [`.flake8`](https://flake8.pycqa.org/en/latest/user/configuration.html#configuration-locations).

https://setuptools.pypa.io/en/latest/userguide/quickstart.html#basic-use
> We recommend users expose as much as possible configuration in a more declarative way via the [pyproject.toml](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html) or [setup.cfg](https://setuptools.pypa.io/en/latest/userguide/declarative_config.html), and keep the setup.py minimal with only the dynamic parts (or even omit it completely if applicable).

Ran [`validate-pyproject`](https://github.com/abravalheri/validate-pyproject) to validate the results.

From `validate-pyproject` README:
> With the approval of [PEP 517](https://www.python.org/dev/peps/pep-0517/) and [PEP 518](https://www.python.org/dev/peps/pep-0518/), the Python community shifted towards a strong focus on standardization for packaging software, which allows more freedom when choosing tools during development and make sure packages created using different technologies can interoperate without the need for custom installation procedures.

> This shift became even more clear when [PEP 621](https://www.python.org/dev/peps/pep-0621/) was also approved, as a standardized way of specifying project metadata and dependencies.

> `validate-pyproject` was born in this context, with the mission of validating `pyproject.toml` files, and making sure they are compliant with the standards and PEPs.


matteius:
@cclauss I think this effort will conflict with:  https://github.com/pypa/pipenv/pull/5836

cclauss:
Did you read the first line of the commit message?

I am highly supportive of 5836 but it is failing its tests.  I thought we could advance step-by-step until we can rebase 5836 in a way that passes all tests.  Please feel free to close this PR with unmerged commits.

matteius:
We can keep it open -- I am not sure what path would be better and I am not opinionated in this case.  CC @kalebmckale

setup.cfg
```diff
@@ -1,7 +1,73 @@
 [metadata]
-long_description = file: README.md
+name = pipenv
+author = Pipenv maintainer team
+author_email = distutils-sig@python.org
```
matteius, line 4: We should figure out something to change this to at some point since that email list is gone.

cclauss:
Closing in favor of #5836

PR #5920, opened 2023-09-06
include markers when skipping lock

### LINKED ISSUE #5912: If --skip-lock is used that python_version marker is not respected

Check the [diagnose documentation](https://pipenv.pypa.io/en/latest/diagnose/) for common issues before posting! We may close your issue if it is very similar to one of them. Please be considerate, or be on your way.

Make sure to mention your debugging experience if the documented solution failed.

### Issue description

I know `--skip-lock` is not recommended. I still use it in several repos to test against several python versions, I noticed strange behaviour that using `--skip-lock` also ignores marker at least for python_version.

`sphinx = {version = "==7.2.4", markers="python_version > '3.8'"}`

Installation works fine if I do not use the `--skip-lock` flag.

### Expected result

sphinx dependency is safely ignored

### Actual result

```
[pipenv.exceptions.InstallError]: ERROR: No matching distribution found for sphinx==7.2.4
ERROR: Couldn't install package: {}
```

### Steps to replicate

Add this line to the Pipfile:

`sphinx = {version = "==7.2.4", markers="python_version > '3.8'"}`

Attempt to install it on python 3.8 using `--skip-lock` flag.

matteius:
### The issue

Fixes #5912 

### The fix

Include the Pipfile markers in the install phase when using `--skip-lock`

### The checklist

* [X] Associated issue

### DIFF
--- a/news/5920.bugfix.rst
+++ b/news/5920.bugfix.rst
@@ -1,0 +1,1 @@
+Include the Pipfile markers in the install phase when using ``--skip-lock``.
--- a/pipenv/routines/install.py
+++ b/pipenv/routines/install.py
@@ -402,6 +402,7 @@ def do_install_dependencies(
                 install_req, markers, req_line = install_req_from_pipfile(
                     req_name, pipfile_entry
                 )
+                req_line = f"{req_line}; {markers}" if markers else f"{req_line}"
                 deps_list.append(
                     (
                         install_req,


matteius:
Ignore the python 3.7 failures at this point -- we will drop it soon, and its too much work to pin everything down that has already dropped it.

PR #5931, opened 2023-09-11
Expand the pipenv resolver to be more customizable for what python version is specified and how markers should evaluate.

### LINKED ISSUE #5908: Issue resolving of python version with point pyton version check
### Issue description

Requires is not correctly resolved for minor python version (or when specifiying full version).

```
[requires]
python_version = "3.11"

[packages]
redis = "==5.0.0"
# async-timeout = "==4.0.3"  # uncomment after pipenv bugfix needed by redis for 3.11.2 and less
```

I then I generate lock file using python 3.11.2 and 3.11.5.
Now 3.11.5 does not lock async-timeout when 3.11.2 does.
Ideally python version that I run pipenv with should be irrelevant.

But I set that I require python version 3.11 so I should always get async-timeout and shouldn't be related to python version that I run pipenv with.
 
https://github.com/redis/redis-py/blob/master/setup.py

```
    install_requires=[
        'importlib-metadata >= 1.0; python_version < "3.8"',
        'typing-extensions; python_version<"3.8"',
        'async-timeout>=4.0.2; python_full_version<="3.11.2"',
    ],
```

I tried various changes like:
python_full_version = "3.11.0"

### Expected result

Get requirements based on python version not version that I run pipenv with. So for 3.11 we should have also this dependency locked with correct marker.

```
        "async-timeout": {
            "hashes": [
                "sha256:4640d96be84d82d02ed59ea2b7105a0f7b33abe8703703cd0ab0bf87c427522f",
                "sha256:7405140ff1230c310e51dc27b3145b9092d659ce68ff733fb0cefe3ee42be028"
            ],
            "index": "pypi",
            "markers": "python_full_version<='3.11.2'",
            "version": "==4.0.3"
        },
```

### Actual result

pipenv thinks async-timeout is not needed for python 3.11.

### Steps to replicate

```
[[source]]
 name = "pypi"
 verify_ssl = true

[requires]
python_full_version = "3.11.0"
# python_version = "3.10"

[packages]
redis = "==5.0.0"
# async-timeout = "==4.0.3"  # uncomment after pipenv bugfix needed by redis for 3.11.2 and less
```

PIPENV_IGNORE_VIRTUALENVS=1 pipenv lock

matteius:
I am exploring how the resolver determines what python version its using and what markers to evaluate against.

### The issue

Fixes #5908

### The fix

Adds another reserved named category for pipenv and a patch to packaging, that allows overriding any of the default environment keys.   Additionally, support ranges for `python_version` and `python_full_version` and a `python` key to specify the base package finder python version to use for the resolver.

In my example, I've added these constraints to our Pipfile to show that I can resolve the requirements that require python 3.7 using a python 3.11 version.

Also explore allowing multiple operating system markers to resolve.

Note:  This branch started as me removing some unused code based on a coverage report and some extra analysis.

### The checklist

* [X] Associated issue


pipenv/patched/pip/_vendor/packaging/markers.py
```diff
@@ -236,7 +236,6 @@ def _evaluate_markers(markers: List[Any], environment: Dict[str, str]) -> bool:
             else:
                 lhs_value = lhs.value
                 rhs_value = _get_env(environment, rhs.value)
-
```
matteius, line 239: I should undo this non-change.

pipenv/utils/indexes.py
```diff
@@ -99,8 +54,6 @@ def get_source_list(
 
 
 def parse_indexes(line, strict=False):
-    from argparse import ArgumentParser
```
oz123, line 102: I don't agree with the linter here.
```
from argparse import ArgumentParser
```

Can be used locally inside a function if it's imported to be used in a different place.
Imports can be expensive (argparse isn't giant, but still...).
The linter rule here should be relaxed.

pipenv/utils/indexes.py
```diff
@@ -1,18 +1,11 @@
 from __future__ import annotations
 
 import re
-from collections.abc import Mapping
+from argparse import ArgumentParser
```
oz123, line 4: See comment below.

Pipfile
```diff
@@ -41,3 +41,9 @@ test = "pytest -vvs"
 
 [pipenv]
 allow_prereleases = true
+
+[resolver]
```
oz123, line 45: This is going to be optional for usrers?

Pipfile
```diff
@@ -41,3 +41,9 @@ test = "pytest -vvs"
...
+python_full_version = "3.7.13"
+python_version = "3.7"
+finder_python = "3.7"
+os_name = "win32"
```
oz123, line 49: How do teams that work on multiple OS can specify multiple OSs here?

oz123 (changes requested):

Pipfile
matteius: Yup -- the code still gets the default platform dictionary, this just provide a mechanism to override.

Pipfile
matteius: This is a minimal patch, so it doesn't provide a way to target multiple versions -- I went down that path initially but it involves changes in the packaging markers code, and it got too complicated for a first pass.

matteius:
Might revisit later on.

Issue #5934, opened 2023-09-15
lock/syncing from git source branch not working anymore


thehesiod:
This used to work a few months ago and broke I think in a september release.

### Issue description

The latest pipenv version broke locking/syncing combo of modules that come from a specific branch from a git repo.

### Expected result

Should be able to sync after locking

### Actual result

fails to sync.

### Steps to replicate

<details>
<summary>Pipfile.lock</summary>

```
[packages]
google-api-python-client = {git = "https://github.com/thehesiod/google-api-python-client.git@thehesiod/batch-retries2"}
```

</details>

<details>
<summary>Pipfile.lock</summary>

```
{
    "_meta": {
        "hash": {
            "sha256": "eb7271cabdeff4e0d8e90d6c590dfc06d896d685e95e6178fe9b2e8f13f9ad9b"
        },
        "pipfile-spec": 6,
        "requires": {},
        "sources": [
            {
                "name": "pip_conf_index_global",
                "url": "https://pypi.fbn.org/simple",
                "verify_ssl": false
            }
        ]
    },
    "default": {
        "cachetools": {
            "hashes": [
                "sha256:95ef631eeaea14ba2e36f06437f36463aac3a096799e876ee55e5cdccb102590",
                "sha256:dce83f2d9b4e1f732a8cd44af8e8fab2dbe46201467fc98b3ef8f269092bf62b"
            ],
            "markers": "python_version >= '3.7'",
            "version": "==5.3.1"
        },
        "certifi": {
            "hashes": [
                "sha256:539cc1d13202e33ca466e88b2807e29f4c13049d6d87031a3c110744495cb082",
                "sha256:92d6037539857d8206b8f6ae472e8b77db8058fec5937a1ef3f54304089edbb9"
            ],
            "markers": "python_version >= '3.6'",
            "version": "==2023.7.22"
        },
        "charset-normalizer": {
            "hashes": [
                "sha256:04e57ab9fbf9607b77f7d057974694b4f6b142da9ed4a199859d9d4d5c63fe96",
                "sha256:09393e1b2a9461950b1c9a45d5fd251dc7c6f228acab64da1c9c0165d9c7765c",
                "sha256:0b87549028f680ca955556e3bd57013ab47474c3124dc069faa0b6545b6c9710",
                "sha256:1000fba1057b92a65daec275aec30586c3de2401ccdcd41f8a5c1e2c87078706",
[...]

PR #5935, opened 2023-09-16
Solving slashes or other less usual symbols in vcs refs were not being interpreted correctly.

matteius:

### The issue

Fixes issue #5934

### The fix

urllib urlparse has to be better at separating netlocs from the remaining path than I can do.

### The checklist

* [X] Associated issue

### DIFF
--- a/tests/integration/test_lockfile.py
+++ b/tests/integration/test_lockfile.py
@@ -1,0 +1,63 @@
+from collections import defaultdict
+import json
+import pytest
+
+from pipenv.project import Project
+from pipenv.utils import requirements
+
+
+@pytest.fixture
+def pypi_lockfile():
+    lockfile = defaultdict(dict)
+    lockfile["_meta"] = {
+        "sources": [
+            {
+                "name": "pypi",
+                "url": "https://pypi.org/simple",
+                "verify_ssl": True,
+            }
+        ]
+    }
+    yield lockfile
+
+
+def test_git_branch_contains_slashes(
+        pipenv_instance_pypi, pypi_lockfile
+):
+    pypi_lockfile["default"]["google-api-python-client"] = {
+        "git": "https://github.com/thehesiod/google-api-python-client.git@thehesiod/batch-retries2",
+        "markers": "python_version >= '3.7'",
+        "ref": "03803c21fc13a345e978f32775b2f2fa23c8e706"
+    }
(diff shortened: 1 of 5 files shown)


tests/integration/test_lockfile.py
```diff
@@ -0,0 +1,63 @@
...
+    }
+    yield lockfile
+
+
+def test_git_branch_contains_slashes(
+        pipenv_instance_pypi, pypi_lockfile
+):
+    pypi_lockfile["default"]["google-api-python-client"] = {
+        "git": "https://github.com/thehesiod/google-api-python-client.git@thehesiod/batch-retries2",
+        "markers": "python_version >= '3.7'",
```
oz123, line 29: Will this test break if the branch is removed?
We should have an isolated test that does not depend on other people's repository.

oz123 (commented):
In general, looks ok. Just the tests look scary. 
Would like to hear if @thehesiod approve that these changes work.
Can this branch upstream be protected from removal?

tests/integration/test_lockfile.py
matteius: These are unit level tests but I had to do them as integration to leverage the pipfile/lockfile stuff we get from those fixtures, but I am starting to write tests that are more unit level in nature -- so we are really just testing the string conversions that happen through these constructs -- no actual lock resolution or installation happening in these tests.  More like is it giving intelligible things to pip resolver or installer based on the inputs.  For example, that commit hash was from a different test and not actually part of that repo.

oz123 (approved):

tests/integration/test_lockfile.py
thehesiod: I don't care using mine, however I'm guessing you'll want to keep your own long term ;)

oz123: > These are unit level tests but I had to do them as integration to leverage the pipfile/lockfile stuff we get from those fixtures, but I am starting to write tests that are more unit level in nature -- so we are really just testing the string conversions that happen through these constructs -- no actual lock resolution or installation happening in these tests. More like is it giving intelligible things to pip resolver or installer based on the inputs. For example, that commit hash was from a different test and not actually part of that repo.

You also submit a PR with the example files to plette repo. I started a collection of test cases there.

Issue #5791, opened 2023-07-13
Private packages not updated unless environment is removed


joaomcarlos:
### Issue description

Private packages defined with an object like `mylib = {ref = "dev" git = "ssh://git.mycompany.com/my-lib.git"}` are not being updated on my virtual env, even after the pipenv update creates a new Pipfile.lock with the correct commit id.

### Expected result

After calling `pipenv update`, I expect the Pipfile.lock ref to change and also the code in my environment to change, such that I have access to the new code within my project.

### Actual result

I see the Pipfile.lock ref change, but the actual code for my dependency does not change.

If I do `pipenv --rm` to remove the environment and then re-create the environment, I can get the actual code change to replicate on disk.

### Steps to replicate

Update `my-lib` with some code change, commit and push. On my project, do `pipenv update` and wait for the process to finish.

Check that the `Pipfile.lock` has changed and has the correct ref for the new commit in `my-lib`.

Check that the code for `my-lib` hasnt actually changed within my original projects virtual env.

edva-lson:
good

matteius:
Could you help me verify if this is still an issue on this overhaul branch:  https://github.com/pypa/pipenv/pull/5793

joaomcarlos:
Hi,

I installed this branch of pipenv and my `pipenv update`s started failing.

If I revert back to normal pipenv it works perfectly.

Error output: 

```pipenv update
Loading .env environment variables...
Running $ pipenv lock then $ pipenv sync.
Locking [packages] dependencies...
Building requirements...
Resolving dependencies...
✔ Success!
Locking [dev-packages] dependencies...
Building requirements...
Resolving dependencies...
✘ Locking Failed!
⠼ Locking...
[ResolutionFailure]:   File "/usr/local/lib/python3.10/site-packages/pipenv/resolver.py", line 665, in _main
[ResolutionFailure]:       resolve_packages(
[ResolutionFailure]:   File "/usr/local/lib/python3.10/site-packages/pipenv/resolver.py", line 632, in resolve_packages
[ResolutionFailure]:       results, resolver = resolve(
[ResolutionFailure]:       ^^^^^^^^
[ResolutionFailure]:   File "/usr/local/lib/python3.10/site-packages/pipenv/resolver.py", line 612, in resolve
[ResolutionFailure]:       return resolve_deps(
[ResolutionFailure]:       ^^^^^^^^^^^^^
[...]

PR #5936, opened 2023-09-16
Assume that vcs and direct URL installs need to be reinstalled

matteius:

### The issue

Fixes issue #5791 

### The fix

Since skip-lock is still in play, and the existing method is operating off only the install req, we cannot with certainty know if the remote requirement has been updated, so queue it for install by pip.

### The checklist

* [X] Associated issue


oz123 (approved):

oz123:
@matteius The change is OK. But does it really fix the issue? Reading the issue (briefly) it seems there is still work going on? Can this be merged?

fmssn:
Possibly 
```python
return(
    if req.link is not None and
    ...
)
```

Would also be ok. The change is pretty important, e.g. it atm crashes the current prefect installation at python 3.11.

matteius:
@fmssn Sorry but I don't understand what you mean by its crashing on a perfectly good install of 3.11 or how this change could affect that.

@oz123 Yeah I am not sure yet why this change wouldn't have solved their issue but probably needs more discovery and possibly this change isn't quite it yet.

fmssn:
@matteius 

For reproduction, you can open a dev container with e.g. "prefecthq/prefect:2.13-python3.11". 

With Pipfile

```
[[source]]
verify_ssl = true
name = "pypi"

[packages]
prefect = "*"

[dev-packages]
pytest = "*"
```
Executing
```sh
pipenv lock 
pipenv install --system
```

yields 
`AttributeError: 'NoneType' object has no attribute 'scheme'`

This is because req.link is optional and of Type `InstallRequirement`, with the default being None. Therefore accessing the attribute scheme if it is `None` raises an Attribute Error.

Issue #5950, opened 2023-09-27
Attribute Error in environment.py


fmssn:
### Issue description

On specific python installations for specific packages, installation fails. The especially occured when installing prefect from Pipfile when using the Docker image  `prefecthq/prefect:2.13-python3.11`. However this was also the case for other 3.11 installations.

### Expected result

Installing normally.

### Actual result

```
root@06bb7bbc7617:/workspaces/prefect/prefect# pipenv install --system
Installing dependencies from Pipfile.lock (4f58fc)...
Traceback (most recent call last):
  File "/usr/local/bin/pipenv", line 8, in <module>
    sys.exit(cli())
             ^^^^^
  File "/usr/local/lib/python3.11/site-packages/pipenv/vendor/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/pipenv/cli/options.py", line 58, in main
    return super().main(*args, **kwargs, windows_expand_args=False)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/pipenv/vendor/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/pipenv/vendor/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/pipenv/vendor/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/pipenv/vendor/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/pipenv/vendor/click/decorators.py", line 84, in new_func
    return ctx.invoke(f, obj, *args, **kwargs)
[...]

PR #5951, opened 2023-09-27
Catch req.link = None

fmssn:
req.link is of type InstallRequirements which can be None. Return False (requirement not met) if this is the case.

### The issue

Fix issue #5950. 

`req` is of type `InstallRequirements`, which has an optional attribute `link` which can be `None`. However, attributes of `req.link` are accessed directly.

### The fix

Catch if `req.link` is `None` and return `False` (requirement not met) before attributes of `req.link` are accessed.

### The checklist

* [x] Associated issue

### DIFF
--- a/pipenv/environment.py
+++ b/pipenv/environment.py
@@ -770,8 +770,8 @@ class Environment:
         )
         if match is not None:
             if req.link is None:
-                return False
-            if req.editable and req.link and req.link.is_file:
+                return True
+            if req.editable and req.link.is_file:
                 requested_path = req.link.file_path
                 if os.path.exists(requested_path):
                     local_path = requested_path
@@ -796,7 +796,7 @@ class Environment:
                     and vcs_ref == requested_revision
                     and direct_url_metadata["url"] == pipfile_part[req.link.scheme]
                 )
-            elif req.link and req.link.is_vcs:
+            elif req.link.is_vcs:
                 return False
             elif req.specifier is not None:
                 return SpecifierSet(str(req.specifier)).contains(
--- a/pipenv/environment.py
+++ b/pipenv/environment.py
@@ -769,9 +769,13 @@ class Environment:
             None,
         )
         if match is not None:
+            if req.specifier is not None:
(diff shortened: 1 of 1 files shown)


pipenv/environment.py
```diff
@@ -769,6 +769,8 @@ def is_satisfied(self, req: InstallRequirement):
             None,
         )
         if match is not None:
+            if req.link is None:
```
matteius, line 772: Initial thoughts, I am wondering if this should actually return True since there was a match -- alternatively, we might handle this on line 780/782:  `elif req.link and match.has_metadata("direct_url.json")`

fmssn: Then we check on 772, 780 and 797 the boolean value of `req.link`, which could be done once before and the cut out from there. Maybe that is a style choice.

My idea was that if we can not properly compary e.g. that from 
```python
return (
    vcs_type == req.link.scheme
    and vcs_ref == requested_revision
    and direct_url_metadata["url"]
    == pipfile_part[req.link.scheme]
)
```

I would assume that expected behavour for not being able to make the comparison to `req.link.scheme`  should default to False.

But yes, thinking about it, it makes more sense to return True, as after the if statements True is also returned by default.

matteius (approved):

fmssn:
This should be the right order to check against. Tests passed locally, had an environment fault earlier which is why they passed. Sorry for the hickup.

oz123 (approved):

myii:
@fmssn I believe this PR resulted in a regression for VCS-based package upgrades, specifically by commit 85c1ba5234ff7228cc64cb18c08f8b8fd9fd3e23.  I've added my findings on the bug report here:

* https://github.com/pypa/pipenv/issues/5791#issuecomment-2336817238

The main query I have is if the proposed reordering of the conditional blocks will actually trigger #5950 again, which this PR resolved.  If you have any feedback to share, please reply to the comment I've linked above, on issue #5791 -- that would be greatly appreciated.

Issue #5939, opened 2023-09-19
Noisy output in pipenv check


BorgPrincess:
### Issue description

Running `pipenv check` throws up messages probably meant for debugging.
Adding `-q` does not have any effect.

### Expected result

[pipenv_check.expected.txt](https://github.com/pypa/pipenv/files/12662237/pipenv_check.expected.txt)

### Actual result

[pipenv_check.actual.txt](https://github.com/pypa/pipenv/files/12662242/pipenv_check.actual.txt)

### Steps to replicate

`pipenv check`

-------------------------------------------------------------------------------

<details><summary>$ pipenv --support</summary>

Pipenv version: `'2023.9.8'`

Pipenv location: `'/home/michaela/.local/lib/python3.11/site-packages/pipenv'`

Python location: `'/usr/bin/python'`

OS Name: `'posix'`

User pip version: `'23.2.1'`

user Python installations found:

PEP 508 Information:

```
{'implementation_name': 'cpython',
 'implementation_version': '3.11.3',
 'os_name': 'posix',
 'platform_machine': 'x86_64',
 'platform_python_implementation': 'CPython',
 'platform_release': '5.4.0-163-generic',
 'platform_system': 'Linux',
 'platform_version': '#180~18.04.1-Ubuntu SMP Fri Sep 8 14:17:56 UTC 2023',
 'python_full_version': '3.11.3',
 'python_version': '3.11',
 'sys_platform': 'linux'}
```

System environment variables:

  - `CLUTTER_IM_MODULE`
  - `LS_COLORS`
  - `LC_MEASUREMENT`
  - `LESSCLOSE`
  - `LC_PAPER`
  - `LC_MONETARY`
  - `XDG_MENU_PREFIX`
  - `LANG`
  - `DISPLAY`
  - `GNOME_SHELL_SESSION_MODE`
  - `COLORTERM`
  - `DESKTOP_AUTOSTART_ID`
  - `USERNAME`
  - `XDG_VTNR`
  - `SSH_AUTH_SOCK`
  - `MANDATORY_PATH`
  - `FUNCTIONS_CORE_TOOLS_TELEMETRY_OPTOUT`
  - `LC_NAME`
  - `XDG_SESSION_ID`
  - `USER`
  - `DESKTOP_SESSION`
  - `QT4_IM_MODULE`
  - `TEXTDOMAINDIR`
  - `GNOME_TERMINAL_SCREEN`
  - `DEFAULTS_PATH`
  - `PWD`
[...]

PR #5793, opened 2023-07-16
stop using requirementslib models

matteius:
The requirementslib models are a nested mess.  I fixed the issue of wheel extras not being installed and broke a bunch of other stuff in the process the other night, but it dawned on me that there is a lot of translating between pip lines and data models and  back and forth and it really is messy and doesn't make much sense.

This is the start of re-establishing what interfaces are actually needed to get between Pipfile, Pipfile.lock and pip lines.   The preferred data model to use is pip's InstallRequirement since requirementslib was using it under the hood anwyay as `ireq`.

There are still a number of edge cases to work out, but I am opening  this draft PR now because I want to see if I actually broke the test runner or just my local environment.

At a high level this is what I am thinking but it requires dropping `--skip-lock` the rest of the way first:
image

Fixes #5817
Fixes #5789
Fixes #5696
Fixes #5626
Fixes #5595 
Fixes #5490
Fixes #5716 (probably)

### The checklist

* [X] Associated issue
* [X] A news fragment in the `news/` directory to describe this fix with the extension `.bugfix.rst`, `.feature.rst`, `.behavior.rst`, `.doc.rst`. `.vendor.rst`. or `.trivial.rst` (this will appear in the release changelog). Use semantic line breaks and name the file after the issue number or the PR #.

PR #5957, opened 2023-09-30
set regular logging level to ERROR

fmssn:
### The issue

Fix noisy output in pipenv check as described in issue #5939. Prior to release v2023.8.19 in `pipenv/utils/__init__.py` logging basic config was set to ERROR, then to INFO. I am not sure whether that was on purpose, was changed in a large commit #5793, maybe it was just used for debugging and then not changed back. @matteius 

### The fix

Change back to `logging.ERROR`.

### The checklist

* [x] Associated issue

### DIFF
--- a/pipenv/utils/__init__.py
+++ b/pipenv/utils/__init__.py
@@ -2,7 +2,7 @@ import logging
 
 from pipenv.patched.pip._vendor.rich.console import Console
 
-logging.basicConfig(level=logging.INFO)
+logging.basicConfig(level=logging.ERROR)
 
 console = Console()
 err = Console(stderr=True)


matteius:
The change to INFO was intentional -- queit/bare should be fixed similar to how it was addressed here: 
 https://github.com/pypa/pipenv/pull/5897

Otherwise we will not be able to debug resolution internal to pip -- there were plenty of other issue reports related, which is why I made the change you referenced initially.    My stance is yes we want to better support quiet/bare, but its better to have too much information than not enough, and so I would rather leave the default as INFO and override specific to the quiet/bare flags.

fmssn:
@matteius I adjusted the code fitting to your recent pull request, setting to warn if not verbose. 

I would like to have some thought based on how to pass the verbosity arg to  the `check` function in `pipenv/cli/command.py`: There is the possibility passing it via 'direct' argument, inserting `verbose=False` into the args, and the possibility of getting it via the `State `object. Both worked in my tests. I decided to go for the latter choice, however I am not 100% sure on that, as for example the quiet argument could, afaik, also be passed via the `State`, yet this is not the case. I would be grateful for further feedback.

matteius (approved):

matteius:
I think the click args would all get passed via the state, the way you are doing it -- this change looks reasonable to me.

Issue #5531, opened 2022-12-09
`pipenv install <package>` with a CLI flag to install the most recent version and have it save to the Pipfile


abarbet:
### Is your feature request related to a problem? Please describe.

When running `pipenv install <package>` without specifying a version, the Pipfile stores the reference to the package as `<package> = "*"`. While this is perfectly functional, anytime a new lockfile is generated, the most recent version of that package will be pulled into the new dependency tree. A version update like this could potentially be a breaking change for the project.

Of course, I could manually go to PyPI at the time of install and manually specify the version, but if I know it works at that version number, it can be a lot of extra work to find that most recent version number and specify it in the command line (especially if I'm bulk installing a lot of packages at once).

### Describe the solution you'd like

Something like an additional flag when running `pipenv install` would be ideal, i.e., `pipenv install --lock-recent <package>` would fetch and install the most recent package, but ALSO save the explicit reference to the most recent version in the Pipfile. 

### Describe alternatives you've considered

Alternative would be the manual lookup in PyPI.

-------------------------------------------------------------------------------

lroberts7:
stack overflow question with the same request:

https://stackoverflow.com/questions/70725867/pipenv-install-package-with-exact-latest-version

indicates this feature request would be used by a large portion of the `pipenv` users.

Is this a contribution that the `pipenv` maintainers would welcome as a CR?

mmisiewicz:
I would like this feature as well!

oz123:
I think it's a reasonable behavior and I welcome a PR for always pinning the latest version.
However,  it is a behavioral change so this needs to be added with a new flag as suggested.

ryanovas:
I'm a little surprised this isn't the default behaviour.... This way I might just run "pipenv install" and accidentally install a new major version. It's super annoying to have to pip freeze every time and manually set the version.

oz123:
I agree.. It's a weird behavior of pipenv.. unfortunately it's something we inherited. As I said, PR welcome.

tadasajon:
I'm honestly flabbergasted that this isn't the default behavior

jatinb92:
Hi, @matteius @oz123 I am trying to work on it, but I have a doubt and need your confirmation before proceeding ahead.

If we want to use this flag "--lock-recent",  which means we should only install the latest version of the new package but all the rest of the packages should not get updated and only install the version mentioned in the pipfile. 

[...]

PR #5963, opened 2023-10-03
feat(install): add 'exact' option

cansin:
### The issue

Tries to very crudely address https://github.com/pypa/pipenv/issues/5531 .

### The fix

Introduces an `--exact` argument to `pipenv install` that locks the exact package version to `Pipfile` instead of `*`. It is a very rudimentary implementation and I am sure calling `resolve_deps` that early in the pipeline might introduce some unintended behavior. But I wanted to open up the PR to get some conversation going, and perhaps inspire someone else to implement a proper fix. Also to just say hi to @matteius I guess 👋 

### To-do

* [ ] Write tests.
* [ ] Add docs.
* [ ] Update `news/`.

### The checklist

* [X] Associated issue

### DIFF
--- a/pipenv/project.py
+++ b/pipenv/project.py
@@ -1197,12 +1197,22 @@ class Project:
         else:
             return name, normalized_name, entry
 
-    def add_package_to_pipfile(self, package, pip_line, dev=False, category=None):
+    def add_package_to_pipfile(
+        self, package, pip_line, dev=False, category=None, exact=False
+    ):
         category = category if category else "dev-packages" if dev else "packages"
 
         name, normalized_name, entry = self.generate_package_pipfile_entry(
             package, pip_line, category=category
         )
+
+        if exact:
+            from pipenv.utils.resolver import resolve_deps
+
+            resolved_packages, resolver = resolve_deps(
+                {normalized_name: package.name}, None, self, category=category
+            )
+            entry = str(resolver.install_reqs[normalized_name].specifier)
 
         return self.add_pipfile_entry_to_pipfile(
             name, normalized_name, entry, category=category
--- a/pipenv/cli/command.py
+++ b/pipenv/cli/command.py
@@ -212,6 +212,7 @@ def install(state, **kwargs):
         python=state.python,
         pypi_mirror=state.pypi_mirror,
         system=state.system,
+        exact=state.installstate.exact,
         ignore_pipfile=state.installstate.ignore_pipfile,
         requirementstxt=state.installstate.requirementstxt,
         pre=state.installstate.pre,
--- a/pipenv/cli/options.py
+++ b/pipenv/cli/options.py
@@ -77,6 +77,7 @@ class InstallState:
     def __init__(self):
         self.dev = False
         self.pre = False
+        self.exact = False
         self.ignore_pipfile = False
         self.code = False
         self.requirementstxt = None
@@ -127,6 +128,24 @@ def editable_option(f):
         callback=callback,
         type=click_types.Path(file_okay=False),
         help="An editable Python package URL or path, often to a VCS repository.",
+    )(f)
+
+
+def exact_option(f):
+    def callback(ctx, param, value):
+        state = ctx.ensure_object(State)
+        state.installstate.exact = value
+        return value
+
+    return option(
+        "--exact",
+        is_flag=True,
+        default=False,
+        expose_value=False,
+        help="Add exact package version to Pipfile when installing, instead of *.",
+        callback=callback,
+        type=click_types.BOOL,
+        show_envvar=True,
     )(f)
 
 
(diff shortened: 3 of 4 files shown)


pipenv/project.py
```diff
@@ -1197,13 +1197,23 @@ def generate_package_pipfile_entry(self, package, pip_line, category=None):
...
+    def add_package_to_pipfile(
+        self, package, pip_line, dev=False, category=None, exact=False
+    ):
         category = category if category else "dev-packages" if dev else "packages"
 
         name, normalized_name, entry = self.generate_package_pipfile_entry(
             package, pip_line, category=category
         )
 
+        if exact:
```
matteius, line 1209: I wonder if we could get this from the actual full resolver call rather than do a 1-off resolution of just this package, because the version may end up being different if another package constraints it.   Just from looking a bit at the code again, I'll note the lock resolution and updating the lock file already happened prior to this point inside of `do_init` -- We could just read the version of that package from the lockfile for the happy path, and then perhaps handle the `--skip-lock` case the way you are describing below (since that flag bypasses the lock phase and lock file).

matteius:
@cansin Really great hearing from you!  I hope you have been happy and healthy.  I did a double take when I saw the PR opened and thought well this is really cool! 😃

pipenv/project.py
cansin: Hmm, that is a valid concern. Let me see if I can figure it out.

cansin:
@matteius hey hey! I've been happy and most definitely healthier! Taking some sweet summertime back in Europe. Hope things are going great back home for you as well 👋

pipenv/project.py
cansin: After a quick look, I "think" `do_init` (`pipenv.routines.install:286`) happens _after_ `add_package_to_pipfile` (`pipenv.routines.install:246&260`) call. Am I missing something?

cansin: > ... get this from the actual full resolver call rather than do a 1-off resolution of just this package, because the version may end up being different if another package constraints it. ...

^ But I think you are right about the above concern, so this PR might be a no-go, as is.

matteius: Hmm, I think you are right now -- its a catch-22 because you need to be able to resolve against _something_ to get the result to know what specifier you want to _exactly_ pin to -- its probably a case of:
1.) Leave the Pipfile entry alone at this step
2.) Let the lock resolution and lock file update happen
3.) Amend the Pipfile entry after that, likely reading from the lock file for the specifier.   (This is the hard part, it would need to happen post-lock but somewhere that has knowledge of the newly added package from the CLI args).

cansin: Or perhaps separate pip list update, pipfile update, lock resolution and lock file update so that the order could become:

1. Update the list of pip packages,
2. Update the lock resolution,
3. Update the `Pipfile`,
4. Update the `Pipfile.lock`.

I think the main challenge right now is 1 and 3, as well as 2 and 4 is coupled so that it is not easy to introduce an intermediary step. If they are to be separated one could:

1. Update the list of pip packages,
2. Update the lock resolution,
3. Update the list of pip packages yet again to fixate the version for the new package,
4. Update the `Pipfile`,
5. Update the `Pipfile.lock`.

But I do not know how easy of a refactor that would be.

cansin: I guess then again, writing to file is not "that" heavy, and we might not need to worry about it. My worry is, when the version of the package at the `Pipfile` changes, would that somehow affect the `Pipfile.lock` content? For instance, I believe it'd on `npm` as afaik they do type out both the requested version and the resolved version to their lock file.

cansin: If that is the case the order would become:

1. Update the list of pip packages,
2. Update the lock resolution,
3. Update the list of pip packages yet again to fixate the version for the new package,
4. Update the lock resolution again,
5. Update the `Pipfile`,
6. Update the `Pipfile.lock`.

matteius: > when the version of the package at the Pipfile changes, would that somehow affect the Pipfile.lock content?

Ah that is a good call out -- the meta hash in the lock file is basically a hash of the Pipfile content -- if the Pipfile content changes after the lockfile is updated, the meta-hash would need to be re-set as well.   That would be about it though, the * specifier would get used for resolution, and then set to the exact version from the resolution results.

Just going to restate the steps you outlined with some notes:

1. Update the list of pip packages, (update the list of pacakges to resolve based on Pipfile specifiers)
2. Perform the lock resolution using the complete list of specifiers (new package from CLI defaults to * as today)
3, Update the list of pip packages yet again to fixate the version for the new package, (This really just needs to be done in the Pipfile at some point using the lock resolution results)
4. Update the lock resolution again, (I don't think this is necessary because the first lock resolution would have had everything specified and the result of that is what we know to be the exact version of the CLI package)
5. Update the Pipfile,  (Maybe not necessary if we already had done this).
6. Update the Pipfile.lock.    Yes -- if we update the Pipfile before we update the lock with the resolution results, then there is no mucking around with re-updating the meta hash.   I still content the original resolution phase though has all the information we need for normal install and no reason to try two resolution phases.

Issue #5964, opened 2023-10-05
Feature Request: Sort packages alphabetically inside each category


daveschaefer:
Hello - Thank you for creating pipenv. It is really interesting and useful. Thanks for putting all of the package and environment management into one place.

What do you think about the idea of sorting package names alphabetically inside each category?

I tried to collect some examples of where else this is done in the python community:

* I see that pipenv itself added alphabetical sorting for the hashes inside the lock file in March 2018 - https://github.com/pypa/pipenv/issues/1529 .
* Both `pip freeze` and `pip list` output their list of packages in a case-insensitive sorted order, and the help docs for each command specifically mention this.
* Tools like `isort` are built specifically to sort and organize packages - https://pycqa.github.io/isort/
* Tutorials like RealPython mention sorting as a best practice - "I like to sort them alphabetically because that makes it easier to find what I am looking for." - https://realpython.com/lessons/requirements-files-best-practices/

I have encountered many python projects and requirements files where the package list is sorted alphabetically for easier readability.
Personally I find package lists easier to read and diff when the names are sorted. It might be nice if `pipenv` took care of this automatically.

From reading through pipenv code, issues, and discussions - my understanding is that this would be only a cosmetic change, and not have any change on the functionality. It appears that pipenv does smart calculations to figure out the correct package installation sequence, and always installs packages in a good working order. Thus sorting the contents of the Pipfile package list should only affect how humans read and use the Pipfile. It would not affect the tool behaviour in any way. Is this correct?

I have a very small working prototype, which I can send a PR for shortly. Simply to start a discussion. (edit: here - https://github.com/pypa/pipenv/pull/5965 )
I realize this might be a drastic change for existing projects that use pipenv, so I thought I would start by filing this Issue to ask.

Thank you for your time.

..

From the Issue template:

> Existing issues 

* pipenv added alphabetical sorting for the hashes inside the lock file in March 2018 - https://github.com/pypa/pipenv/issues/1529 
* Some discussion on ordering here, but closed in Aug 2022 as "may have improved" - https://github.com/pypa/pipenv/issues/2274

> make sure you are running the latest version of Pipenv.

I just cloned the repo today to write a test PR:

```
$ pipenv --version
pipenv, version 2023.10.4.dev0
```

### Is your feature request related to a problem? Please describe.

I find it faster and easier to read a list of python packages if that list is sorted alphabetically by package name.
[...]

PR #5965, opened 2023-10-05
Sort packages alphabetically inside each category

daveschaefer:
Add a new `[pipenv]` directive to sort package names alphabetically inside the category, for easier reading.

- [x] sort on `install`
- [x] sort on `uninstall`
- [ ] sort on `upgrade`
- [ ] Does this need to run on `update`?

--

Unsure if this is the best place or way to implement. Small prototype to add to discussion in https://github.com/pypa/pipenv/issues/5964 .

Tests:

before patch:

```
AssertionError: assert ['atomicwrite...ama', 'build'] == ['atomicwrite...', 'colorama']
  At index 1 diff: 'colorama' != 'build'
  Full diff:
  - ['atomicwrites', 'build', 'colorama']
  ?                  ---------
  + ['atomicwrites', 'colorama', 'build']
  ?                            +++++++++
```

after patch: pass.

------

### The checklist

* [x] Associated issue

### DIFF
--- a/pipenv/project.py
+++ b/pipenv/project.py
@@ -1117,6 +1117,8 @@ class Project:
         p = self.parsed_pipfile
         if name:
             del p[category][name]
+            if self.settings.get("sort_pipfile"):
+                p[category] = dict(sorted(p[category].items()))
             self.write_toml(p)
             return True
         return False
--- a/pipenv/project.py
+++ b/pipenv/project.py
@@ -1227,7 +1227,7 @@ class Project:
 
         p[category][normalized_name] = entry
 
-        if self.settings.get("sort_alphabetical"):
+        if self.settings.get("sort_pipfile"):
             p[category] = dict(sorted(p[category].items()))
 
         # Write Pipfile.
--- a/pipenv/project.py
+++ b/pipenv/project.py
@@ -1111,6 +1111,9 @@ class Project:
                 return name
         return None
 
+    def _sort_category(self, category):
+        return dict(sorted(category.items()))
+
     def remove_package_from_pipfile(self, package_name, category):
         # Read and append Pipfile.
         name = self.get_package_name_in_pipfile(package_name, category=category)
@@ -1118,7 +1121,7 @@ class Project:
         if name:
             del p[category][name]
             if self.settings.get("sort_pipfile"):
-                p[category] = dict(sorted(p[category].items()))
+                p[category] = self._sort_category(p[category])
             self.write_toml(p)
             return True
         return False
@@ -1230,7 +1233,7 @@ class Project:
         p[category][normalized_name] = entry
 
         if self.settings.get("sort_pipfile"):
-            p[category] = dict(sorted(p[category].items()))
+            p[category] = self._sort_category(p[category])
 
         # Write Pipfile.
         self.write_toml(p)
--- a/pipenv/project.py
+++ b/pipenv/project.py
@@ -1226,7 +1226,9 @@ class Project:
             newly_added = True
 
         p[category][normalized_name] = entry
-        p[category] = dict(sorted(p[category].items()))
+
+        if self.settings.get("sort_alphabetical"):
+            p[category] = dict(sorted(p[category].items()))
 
         # Write Pipfile.
         self.write_toml(p)
--- a/pipenv/project.py
+++ b/pipenv/project.py
@@ -1112,7 +1112,14 @@ class Project:
         return None
 
     def _sort_category(self, category):
-        return dict(sorted(category.items()))
+        # toml tables won't maintain sorted dictionary order
+        # so construct the table in the order that we need
+        sorted_category = dict(sorted(category.items()))
+        table = tomlkit.table()
+        for key, value in sorted_category.items():
+            table.add(key, value)
+
+        return table
 
     def remove_package_from_pipfile(self, package_name, category):
         # Read and append Pipfile.
--- a/pipenv/project.py
+++ b/pipenv/project.py
@@ -1226,6 +1226,7 @@ class Project:
             newly_added = True
 
         p[category][normalized_name] = entry
+        p[category] = dict(sorted(p[category].items()))
 
         # Write Pipfile.
         self.write_toml(p)
--- a/tests/integration/test_install_basic.py
+++ b/tests/integration/test_install_basic.py
@@ -557,9 +557,30 @@ dataclasses-json = {file = "https://files.pythonhosted.org/packages/85/94/1b3021
         c = p.pipenv("""run python -c "from dataclasses_json import dataclass_json" """)
         assert c.returncode == 0
 
+
 @pytest.mark.basic
 @pytest.mark.install
-def test_packages_sorted_alphabetically_in_category(pipenv_instance_private_pypi):
+def test_category_sorted_alphabetically_with_directive(pipenv_instance_private_pypi):
+    with pipenv_instance_private_pypi() as p:
+        with open(p.pipfile_path, "w") as f:
+            contents = """
+[pipenv]
+sort_alphabetical = true
+
+[packages]
+atomicwrites = "*"
+colorama = "*"
+            """.strip()
+            f.write(contents)
+        c = p.pipenv("install build")
+        assert c.returncode == 0
+        assert "build" in p.pipfile["packages"]
+        assert list(p.pipfile["packages"].keys()) == ["atomicwrites", "build", "colorama"]
+
+
+@pytest.mark.basic
+@pytest.mark.install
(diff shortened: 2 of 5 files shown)


tests/integration/test_install_basic.py
```diff
@@ -556,3 +556,20 @@ def test_install_tarball_is_actually_installed(pipenv_instance_pypi):
...
+[packages]
+atomicwrites = "*"
+colorama = "*"
+            """.strip()
+            f.write(contents)
+        c = p.pipenv("install build")
+        assert c.returncode == 0
+        assert "build" in p.pipfile["packages"]
+        assert list(p.pipfile["packages"].keys()) == ["atomicwrites", "build", "colorama"]
+
```
daveschaefer, line 575: For this test I just chose three arbitrary packages that are already used inside pipenv's own Pipfile.

matteius:
I would recommend making this a behavior driven by a [pipenv] directive in the Pipfile -- the reason why is because then its opt-in/opt-out at the project level if the directive is in the Pipfile; it will sort the packages alphabetically for the project whenever a change is made, install/uninstall/update/upgrade -- but otherwise it will continue to behave as is for folks that don't want this behavior.

daveschaefer:
> I would recommend making this a behavior driven by a [pipenv] directive in the Pipfile -- the reason why is because then its opt-in/opt-out at the project level if the directive is in the Pipfile; it will sort the packages alphabetically for the project whenever a change is made, install/uninstall/update/upgrade -- but otherwise it will continue to behave as is for folks that don't want this behavior.

Hello, thank you for the feedback. Implementing this as a config setting/directive sounds quite reasonable.

The [pipenv] section appears to be the same place that stores settings such as `allow_prereleases`, `disable_pip_input`, `install_search_all_sources`, and `keep_outdated`. Is that correct?
At first glance, implementing a new setting like this seems like it could simply be an `if` statement to check for the setting. i.e. in `add_pipfile_entry_to_pipfile()`, something like:

```
if self.settings.get("sort_alphabetical"):
    p[category] = dict(sorted(p[category].items())
```

Does that sound like a good way to do it?

Let me know if there is a particular setting name you want. Perhaps `sort_alphabetical`?

daveschaefer:
Which commands should we make this run on? Perhaps `install`, `uninstall`, and `upgrade` - since those are the commands that normally make changes to the `Pipfile`?

matteius:
Thanks @daveschaefer -- some thoughts:
Perhaps `sort_alphabetical` --> `sort_pipfile`

> Which commands should we make this run on? Perhaps install, uninstall, and upgrade

That makes sense, just calling out: upgrade and update (update uses upgrade, but I forget if it touches Pipfile directly so worth taking a look).

daveschaefer:
> Thanks @daveschaefer -- some thoughts: Perhaps `sort_alphabetical` --> `sort_pipfile`

Thank you. I have updated the code to use `sort_pipfile`, and added code and tests for `uninstall`.

Once this is implemented, is there a place in the docs that I can or should update to mention this new directive?
I don't think I see a specific place in the documentation that discusses the `[pipfile]` section. Should I add one? I can make that a separate PR if that's best.

tests/integration/test_uninstall.py
```diff
@@ -243,3 +243,54 @@ def test_uninstall_multiple_categories(pipenv_instance_private_pypi):
 
         assert "six" not in p.lockfile.get("prereq", {})
         assert "six" not in p.lockfile["default"]
+
+
+@pytest.mark.uninstall
```
daveschaefer, line 248: I see some of the uninstall tests are marked with `@pytest.mark.install`, and some are not. Is this based on whether the test calls `install` during the test? Something else? Let me know if you want me to mark these two tests with `@pytest.mark.install` also.

matteius: >  Is this based on whether the test calls install during the test? 
Yeah, the test markers needs some separate TLC -- but that is the idea.

matteius:
> I don't think I see a specific place in the documentation that discusses the [pipfile] section. Should I add one? I can make that a separate PR if that's best.

We should include the documentation in this PR -- let's see ... probably this is the closest section that documents the Pipfile:  https://pipenv.pypa.io/en/latest/pipfile/

Any TLC you put into the docs would be appreciated.

daveschaefer:
> We should include the documentation in this PR -- let's see ... probably this is the closest section that documents the Pipfile: https://pipenv.pypa.io/en/latest/pipfile/
> 
> Any TLC you put into the docs would be appreciated.

Thank you. I added a section to `pipfile.md` based on the notes I could find from the codebase, docs, and release history.
I also added a note for the `sort_pipfile` setting that would be created with this patch.
Since it looks like the `keep_outdated` flag has been discontinued, I left it out and did not write anything for it.

Feedback on any of the docs or wording is welcome.

I may not have time immediately, but I will see if I can add tests and code for sorting on `upgrade` or `update` as well.

tests/integration/test_uninstall.py
daveschaefer: Thank you. I have marked these tests with `@pytest.mark.install`.

I saw there are some other tests in `test_uninstall.py` that call `install`, but don't have the mark.
Since I am here editing the file, I marked those tests as well.
I can drop the commit if that's not helpful.

daveschaefer:
So, the good news: the `upgrade` command already calls `add_pipfile_entry_to_pipfile()`, where this code already does the sorting. So we may be able to keep all of the sorting code in one place.

The bad news: sorting categories and passing a dict to `write_toml()` works when all Pipfile entries have the same value type, but not if the category values have both strings and dictionary values.

What I mean is: If all of the package entries use strings like `parse = "*", sphinx = "==4.*"`, the sorting works fine.
But some package entries use other descriptions, such as a dictionary like `gunicorn = {version = "*"}`.
In this case the toml sorting and printing will default to sorting the data so that all of the string packages like `= "*"` are sorted first, as a group, followed by all of the dictionary packages like `= {version = "*"}` after that.
The Pipfile will look like:

```
aa = "*"
bb = "*"
cc = "*"
aaa = {version = "*"}
bbb = {version = "*"}
ccc = {version = "*"}
```

I have added test cases to `install` and `uninstall` for this to enforce dealing with it.

Currently I have solved this by implementing a private `_sort_category()` function in `project.py` that takes a category `dict` and builds a `tomlkit.table` with data in the correct sorted order.
This succeeds, and sorts the Pipfile correctly. But feels like a bit of a kludge.
Technically the `p[category]` data is already of type `tomlkit.table` right where `_sort_category()` is called anyway. It gets called right before the data is written with `write_toml()`. But it feels less clean.

Let me know what you think. If there is a better way to solve this, I am all ears.
I could also move `_sort_category()` into `toml.py` if that's a better place to put the toml doc formatting code and keep it all in one place.

daveschaefer:
Is it helpful if I annotate functions with argument types and return types?

tests/integration/test_upgrade.py
```diff
@@ -0,0 +1,52 @@
+import pytest
```
daveschaefer, line 1: I didn't see any existing integration tests for the `upgrade` command, so I created this file.
I am attempting to match the pattern of test format from other tests.
Let me know if you want me to put these somewhere else.

pipenv/project.py
```diff
@@ -1111,12 +1111,24 @@ def get_package_name_in_pipfile(self, package_name, category):
                 return name
         return None
 
+    def _sort_category(self, category):
+        # toml tables won't maintain sorted dictionary order
+        # so construct the table in the order that we need
+        sorted_category = dict(sorted(category.items()))
```
matteius, line 1117: I think this will definitely work, but I am wondering if it would be more efficient to move into the loop and  not cast to a dict so like:
`for key, value in sorted(category.items():`

Correct me if that won't work or there is an issue with it, but thats my last bit of feedback.   I'll see if @oz123 can review this too but I want to get this merged into the release for this month.

daveschaefer: Thank you very much! That is very kind.

Are there other test cases I should consider or other tests I can add?

I noticed some projects that have `# comments` on some lines of their Pipfiles. When I add a comment on both string-type lines and dictionary-type lines those are both sorted correctly.

Are there other values or types that can go on a line?

daveschaefer: > I am wondering if it would be more efficient to move into the loop and not cast to a dict so like: `for key, value in sorted(category.items():`

Thanks! I created a PR to do that - https://github.com/pypa/pipenv/pull/5990/files

Is it helpful if I create a news fragment `.rst` ? Or is that no longer useful since the PR has been merged?

matteius: the new change is trivial and the docs change you did is good -- we just need to figure out why the docs don't seem to be getting published at some point during the new releases.

PR #5970, opened 2023-10-10
Drop pep517

oz123:
All tests pass on Linux without it.


matteius:
Good find, this is no longer required now that we dropped requirementslib 🎉

matteius (approved):

Issue #5969, opened 2023-10-10
Newlines in output of `--py`


notEvil:
### Issue description

`pipenv --py` adds newlines. For instance
`python -c "import subprocess; print(subprocess.Popen(['pipenv', '--py'], stdout=subprocess.PIPE).stdout.read())"`

### Expected result

should print something like `b'/home/user/.local/share/virtualenvs/name-hash/bin/python\n'`

### Actual result

but actually prints things like `b'/home/user/.local/share/virtualenvs/name-hash/bin/p\nython\n'`. This is due to
https://github.com/pypa/pipenv/blob/e45cc7a9ab7e7d805ef6008db80a5469f038806c/pipenv/cli/command.py#L771

### Steps to replicate

- make the console small enough so the path doesn't fit on one line
- run `python -c "import subprocess; print(subprocess.Popen(['pipenv', '--py'], stdout=subprocess.PIPE).stdout.read())"`

This is an issue if run as part of a script or application, for instance [here](https://github.com/fannheyward/coc-pyright/blob/8789fcc25cd055a0cc1e70fc83bfcca90720c384/src/configSettings.ts#L77). Maybe add the argument `bare` similar to https://github.com/pypa/pipenv/blob/e45cc7a9ab7e7d805ef6008db80a5469f038806c/pipenv/utils/virtualenv.py#L360

notEvil:
Proposal:

<details>

```diff
diff --git a/pipenv/cli/command.py b/pipenv/cli/command.py
index fb4ee05..a61b9a6 100644
--- a/pipenv/cli/command.py
+++ b/pipenv/cli/command.py
@@ -27,6 +27,7 @@ from pipenv.utils.processes import subprocess_run
 from pipenv.vendor.click import (
     Choice,
     argument,
+    echo,
     edit,
     group,
     option,
@@ -125,7 +126,7 @@ def cli(
             do_where(state.project, bare=True)
             return 0
         elif py:
-            do_py(state.project, ctx=ctx)
+            do_py(state.project, ctx=ctx, bare=True)
             return 0
         # --support was passed...
         elif support:
@@ -758,7 +759,7 @@ if __name__ == "__main__":
     cli()
 
 
-def do_py(project, ctx=None, system=False):
+def do_py(project, ctx=None, system=False, bare=False):
     if not project.virtualenv_exists:
         err.print(
             "[red]No virtualenv has been created for this project[/red] "
[...]

PR #5971, opened 2023-10-11
Change `--py` to use `print` instead of `console.print`

notEvil:
### The issue

#5969 

### The fix

Similar to `do_where(bare=True)`, the PR adds the argument `bare=False` to `do_py` and calls it with `bare=True`.
Like `--venv`, it will use `print` instead of `console.print`. (`do_where` uses `click.echo`)

### The checklist

* [x] Associated issue

### DIFF
--- a/news/5969.behavior.rst
+++ b/news/5969.behavior.rst
@@ -1,0 +1,1 @@
+Change ``--py`` to use ``print`` preventing insertion of newline characters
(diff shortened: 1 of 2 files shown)


matteius:
Hey @notEvil -- looks like this just needs a news fragment.

matteius (approved):

PR #5985, opened 2023-10-18
Update advanced.md to restore Automatic Python Installation header

iamed2:
### The issue

The docs header "Automatic Python Installation" was inadvertently removed in https://github.com/pypa/pipenv/pull/5909. 

### The fix

Added it back in, where it was before.

### The checklist

N/A, this just corrects a docs typo.

P.S.: I did notice the docs search still identified "Automatic Python Installation" as a heading when e.g. searching for "pyenv", even though it no longer appeared. Is there some search index that should be regenerated?


matteius:
> P.S.: I did notice the docs search still identified "Automatic Python Installation" as a heading when e.g. searching for "pyenv", even though it no longer appeared. Is there some search index that should be regenerated?

I think there is an issue with the docs build deploy, because it seems that it does not get updated on every release anymore -- noticed this recently that the new release changelogs don't show up for the published build.  Needs more investigation.

PR #5987, opened 2023-10-20
Python 3.12 support (passing)

matteius:
+ enables Python 3.12 in the CI

### The issue

Try getting python 3.12 support enabled.

### The checklist

* [ ] Associated issue


pipenv/vendor/vendor.txt
```diff
@@ -4,7 +4,7 @@ colorama==0.4.6
 dparse==0.6.3
 markupsafe==2.1.3
 pexpect==4.8.0
-pipdeptree==2.13.0
+pipdeptree==2.8.0
```
oz123, line 7: Just curious, why is that you reverted pipdeptree?

matteius: We have a custom patch that needs to be re-worked, it couldn't be applied  and I didn't have time to dig deeper.

PR #5993, opened 2023-10-24
New branch, removes empty flag in SVN install command. Fixes SVN inst…

rsilver23:
…alls

### The issue

This branch fixes pipenv install of SVN repos. Pipenv was adding and extra empty argument when the verbosity flag was <= 0. This empty string was causing an error with SVN checkout commands.

### The fix

This pull request removes the empty flag from the command arguments when the verbosity is not set.

### DIFF
--- a/pipenv/patched/pip/_internal/vcs/subversion.py
+++ b/pipenv/patched/pip/_internal/vcs/subversion.py
@@ -288,17 +288,22 @@ class Subversion(VersionControl):
             display_path(dest),
         )
         if verbosity <= 0:
-            flag = "--quiet"
+            cmd_args = make_command(
+                "checkout",
+                "--quiet",
+                self.get_remote_call_options(),
+                rev_options.to_args(),
+                url,
+                dest,
+            )
         else:
-            flag = ""
-        cmd_args = make_command(
-            "checkout",
-            flag,
-            self.get_remote_call_options(),
-            rev_options.to_args(),
-            url,
-            dest,
-        )
+            cmd_args = make_command(
+                "checkout",
+                self.get_remote_call_options(),
+                rev_options.to_args(),
+                url,
+                dest,
+            )
         self.run_command(cmd_args)
 
     def switch(self, dest: str, url: HiddenText, rev_options: RevOptions) -> None:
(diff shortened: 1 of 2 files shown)


matteius:
Actually I am pushing a fix to the ruff lint to main -- has nothing to do with this PR, but a news fragment would be good.

rsilver23:
Hey Matt,

Thanks for looking at this so quickly, is there documentation for how to do the news fragment? I'm unfamiliar with this method.

Thanks,
Ryne
________________________________

matteius:
@rsilver23 Basically create a file in the news directory `5993.bugfix.rst` and explain the correction.   Helpful to first run `pre-commit install` so that the linter will check your news file when you check it in.

matteius:
@rjsilver oh hey -- there is a problem I just noticed, which is you are modifying the pip internals.  We don't want to do that, if there was truly a reason to do so, we would provide a patch, but I think in this case you mentioned it is something to do with click?

matteius:
@rjsilver I wonder if https://github.com/pypa/pip/issues/11050 is related?

rsilver23:
@matteius [https://github.com/pypa/pip/issues/11050](https://github.com/pypa/pip/issues/11050) definitely looks related

rsilver23:
@matteius looks like this code already exists in a patch. Do you have a better recommendation where to put this?

matteius:
@rsilver23 So we vendor in pip to pipenv and we have some patches that we apply after the fact.  This feels like a case where pip itself would accept this patch, and the original PR that fixes it was tagged with needs a rebase.   Perhaps you could open a new PR against pip -- the problem is, if I merge this as-is, the next time we vendor in a new version of pip it would get overwritten.

pipenv/patched/pip/_internal/vcs/subversion.py
```diff
@@ -288,17 +288,22 @@ def fetch_new(
             display_path(dest),
         )
         if verbosity <= 0:
-            flag = "--quiet"
+            cmd_args = make_command(
```
oz123, line 291: This definitely looks like something that would not encounter  much resistance in upstream.

devsagul:
Hi! FYI the original PR opened into pip repo and mentioned here has been merged.

oz123:
You mean this one? https://github.com/pypa/pip/pull/11051?
As soon as there is new pip release, this will lend in pipenv too!

Issue #6022, opened 2023-11-21
Additional instllation method


y-vectorfield:
In the README, I think the following installation method should be added.

```bash
python -m pip install pipenv
```

y-vectorfield:
Hello, @kennethreitz, what do you think this PR?
If there is no problem, please approve this.

y-vectorfield:
Hello, @matteius,
Thank you very much for your message.
> I will take a look soon.
Oh, I see. Thank you.

PR #6023, opened 2023-11-21
Fix the issue(#6022): Add additional installation method in README

y-vectorfield:
Fit the issue #6022 
Add additional installation method in README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,6 +98,10 @@ Alternatively, some users prefer to use [Pipx](https://pypi.org/p/pipx):
 
     pipx install pipenv
 
+Or, some users prefer to use Python pip module
+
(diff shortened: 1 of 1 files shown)


matteius:
@y-vectorfield Can you explain more why this should be a preferred/supported installation method?  I mean the change itself seems simple, but I wonder what is the difference  between that and invoking pip directly?

y-vectorfield:
Hello, @matteius, thank you very much for your comment.
Actually, we often cannot install pip command using package manager as sudors.
This methdod(python -m pip ~) can install python libraries if we cannot install pip commad specific environment.
Hence, I think we should add this installation method.

PR #6055, opened 2024-01-04
Vendor bump pipdeptree

oz123:
Bump version of  pipdeptree. Some changes required to move to this version on our side too.


oz123:
@matteius can we please include this in the last 2023.XX version? It does not have to wait to the next semver release.

Issue #6056, opened 2024-01-05
Pipenv selects wrong PyPi repository for private packages with non canonical names


mangin:
### Issue description

Let's assume that we have a package 'some-cool-package-name'. Pip allows non-canonical names using underscores. In this particular case it would look like 'some_cool_package_name`

if the package is located in private PyPi index than pipenv would select default index during creating a lock. That would create problems with locking.

### The reason of the problem
In this file we normalize the names of packages:
https://github.com/pypa/pipenv/blob/main/pipenv/patched/pip/_internal/resolution/resolvelib/factory.py#L250

And we pass normalized name to finder:
https://github.com/pypa/pipenv/blob/main/pipenv/patched/pip/_internal/resolution/resolvelib/factory.py#L284

Here we get the repository of packages by package name:
https://github.com/pypa/pipenv/blob/8e0608593a1c65439ada455a7980050c9e4efbe0/pipenv/patched/pip/_internal/models/search_scope.py#L142

Here you can see how we create mapping between package name and repository:
https://github.com/pypa/pipenv/blob/8e0608593a1c65439ada455a7980050c9e4efbe0/pipenv/utils/resolver.py#L319

But there we create mapping between package name and repository. (Pay attention that here we don't use normalized package name)

### Expected result

Pipenv should be able to work with packages that have non-canonical names.

### Actual result

Pipenv has a problem with locking non-canonical names

### Steps to replicate
```
pipenv install -d 
```

```
[[source]]
url = "https://pypi.org/simple"
verify_ssl = true
name = "pypi"

[[source]]
url = "https://test.pypi.org/simple"
verify_ssl = true
name = "testpypi"

[packages]
pipenv_test_private_package = {version = "*", index = "testpypi"}
```

Stack trace:
```
Pipfile.lock not found, creating...
Locking [packages] dependencies...
Building requirements...
Resolving dependencies...
✘ Locking Failed!
⠸ Locking...False
CRITICAL:pipenv.patched.pip._internal.resolution.resolvelib.factory:Could not find a version that satisfies the requirement pipenv_test_private_package (from versions: none)
[ResolutionFailure]:   File "~/.venv/lib/python3.10/site-packages/pipenv/resolver.py", line 645, in _main
[ResolutionFailure]:       resolve_packages(
[ResolutionFailure]:   File "~/.venv/lib/python3.10/site-packages/pipenv/resolver.py", line 612, in resolve_packages
[ResolutionFailure]:       results, resolver = resolve(
[...]

PR #6057, opened 2024-01-05
Fixed a bug with locking packages with non canonical names

mangin:
### The issue

Fix a bug with locking projects that contains packages with non canonical names from private indexes
More details here:
https://github.com/pypa/pipenv/issues/6056

### The checklist

* [ X] Associated issue
* [X ] A news fragment in the `news/` directory to describe this fix with the extension `.bugfix.rst`, `.feature.rst`, `.behavior.rst`, `.doc.rst`. `.vendor.rst`. or `.trivial.rst` (this will appear in the release changelog). Use semantic line breaks and name the file after the issue number or the PR #.
* [] Send a fix to pip ?

### DIFF
--- a/pipenv/utils/resolver.py
+++ b/pipenv/utils/resolver.py
@@ -25,6 +25,7 @@ from pipenv.patched.pip._internal.req.req_file import parse_requirements
 from pipenv.patched.pip._internal.req.req_install import InstallRequirement
 from pipenv.patched.pip._internal.utils.temp_dir import global_tempdir_manager
 from pipenv.patched.pip._vendor import pkg_resources, rich
+from pipenv.patched.pip._vendor.packaging.utils import canonicalize_name
 from pipenv.project import Project
 from pipenv.utils.fileutils import create_tracked_tempdir
 from pipenv.utils.requirements import normalize_name
@@ -200,6 +201,7 @@ class Resolver:
         for package_name, dep in deps.items():  # Build up the index and markers lookups
             if not dep:
                 continue
+            canonical_package_name = canonicalize_name(package_name)
             is_constraint = True
             install_req, _ = expansive_install_req_from_line(dep, expand_env=True)
             original_deps[package_name] = dep
@@ -210,14 +212,18 @@ class Resolver:
                 pipfile_entries[package_name] = pipfile_entry
                 if isinstance(pipfile_entry, dict):
                     if packages[package_name].get("index"):
-                        index_lookup[package_name] = packages[package_name].get("index")
+                        index_lookup[canonical_package_name] = packages[package_name].get(
+                            "index"
+                        )
                     if packages[package_name].get("skip_resolver"):
                         is_constraint = False
                         skipped[package_name] = dep
                 elif index:
-                    index_lookup[package_name] = index
+                    index_lookup[canonical_package_name] = index
                 else:
-                    index_lookup[package_name] = project.get_default_index()["name"]
+                    index_lookup[canonical_package_name] = project.get_default_index()[
+                        "name"
+                    ]
             if install_req.markers:
                 markers_lookup[package_name] = install_req.markers
             if is_constraint:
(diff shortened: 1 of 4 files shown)


mangin:
From my point of view the patch can't be send to pip.
Because they don't handle mapping between packages and indexes.
They just try to install package from ALL indexes:
https://github.com/pypa/pip/blob/e88d39ae49ab11a8b80609a018e5a36ea4ccad89/src/pip/_internal/models/search_scope.py#L112

oz123:
This seems like a legitimate fix, yet ... patching pip is a cause of trouble. 
Your patch should be here:
```
tasks/vendoring/patches/patched
```

like other pip patches.

If @matteius is OK with this patch I can move it there.

mangin:
No problem i would move the fix there.
I would like to finish the job

oz123:
Please go a head and move the patch there. You will also need to apply it to check that our vendoring system understands it.

matteius:
I am convinced this must be fixable in the code we use from pipenv without patching pip at all.    I will try to help look more into this tonight, assuming we don't loose power in this storm.

matteius:
Does this have to do with dash vs underscore in the package name?   I do think I broke something there when we factored away requirementslib.

pipenv/utils/resolver.py
```diff
@@ -210,14 +212,14 @@ def create(
                 pipfile_entries[package_name] = pipfile_entry
                 if isinstance(pipfile_entry, dict):
                     if packages[package_name].get("index"):
-                        index_lookup[package_name] = packages[package_name].get("index")
+                        index_lookup[canonical_package_name] = packages[package_name].get("index")
```
mangin, line 215: Let's create mapping with canonical package names instead of raw package names

pipenv/utils/resolver.py
```diff
@@ -25,6 +25,7 @@
 from pipenv.patched.pip._internal.req.req_install import InstallRequirement
 from pipenv.patched.pip._internal.utils.temp_dir import global_tempdir_manager
 from pipenv.patched.pip._vendor import pkg_resources, rich
+from pipenv.patched.pip._vendor.packaging.utils import canonicalize_name
 from pipenv.project import Project
 from pipenv.utils.fileutils import create_tracked_tempdir
 from pipenv.utils.requirements import normalize_name
```
mangin, line 31: From my point of vie we should use canonicalize_name funiction from pip instead of normalize_name function everywhere.
WDYT?

pipenv/utils/resolver.py
```diff
@@ -210,14 +212,14 @@ def create(
...
                         is_constraint = False
                         skipped[package_name] = dep
                 elif index:
-                    index_lookup[package_name] = index
+                    index_lookup[canonical_package_name] = index
                 else:
-                    index_lookup[package_name] = project.get_default_index()["name"]
+                    index_lookup[canonical_package_name] = project.get_default_index()["name"]
             if install_req.markers:
                 markers_lookup[package_name] = install_req.markers
```
mangin, line 224: Maybe we should use everywhere canonical_package_name instead of raw_package_name.
But i didn't start to do it because it would make affect on locking files => i'm a little bit afraid of it.

mangin:
@matteius can you please take a look one more time?
This change wouldn't require patching pip.

pipenv/utils/resolver.py
matteius: Yeah you are right, I wish this is something I hadn't broke, but I think your patch makes sense as an iteration in the right direction.  I will think more about it.

mangin: Can we apply my patch?

P.S. I sent it because at my team we have some issues with pipenv. And we can't migrate to the freshest version (we still use pretty old one)

mangin:
Fixed code style... i forgot to set up precommit hook. Sorry.

matteius (approved):

mangin:
Hey @matteius,
Do we have any timeline about releasing the fix?

matteius:
I think we will get another 2023 release in (minior or patch) release within the next week and then start looking at our more major 2024 version release after that.   At least that's my plan, I have to travel for work next week so if I can get it in before then,.  Our goal is to adopt semver going forward.

matteius:
@mangin Your patch is in the release from 15 minutes ago -- I hope that it is helpful.

mangin:
Thank you for the release ^_^

Issue #6054, opened 2024-01-04
As of v2023.8.19 local file dependencies default to being installed as editable


edmorley:
### Issue description

Between Pipenv `2023.7.23` and `2023.8.19` there was an undocumented breaking change in default installation behaviour for local `file` directory installs that don't specify an explicit `editable` property, such as:

```toml
[packages]
mypackage = {file = "packages/mypackage"}
```

Before (in `2023.7.23` and earlier), such a dependency specifier was installed as a standard non-editable install.

Now (in `2023.8.19` and later), the install is performed as an editable install - ie: as though `editable = "true"` had been specified in the `Pipfile` / `Pipfile.lock`.

I presume this might be an unintended side-effect of #5793?

### Expected result

If this behaviour change was intentional:
* For it to have been documented in the changelog

If this wasn't intentional:
* For the behaviour to have not changed
* For Pipenv to have test coverage of this

And for bonus points either way:
* For the intended default behaviour (ie: the behaviour when `editable = ....` is not explicitly specified in `Pipfile`) to be documented, eg at https://pipenv.pypa.io/en/latest/specifiers.html#editable-dependencies-e or similar

### Steps to replicate

Using https://github.com/edmorley/testcase-pipenv-file-editable:

Behaviour before (with `2023.7.23`):

```
$ docker run --rm -it python:3.12 bash -c 'git clone -q https://github.com/edmorley/testcase-pipenv-file-editable && cd testcase-pipenv-file-editable && pip install -q --disable-pip-version-check --root-user-action=ignore pipenv==2023.7.23 && pipenv install && pipenv run pip show mypackage | grep -i location'
...
Location: /root/.local/share/virtualenvs/testcase-pipenv-file-editable-YkSiblPt/lib/python3.12/site-packages
```

Behaviour now (with `2023.8.19`):

```
$ docker run --rm -it python:3.12 bash -c 'git clone -q https://github.com/edmorley/testcase-pipenv-file-editable && cd testcase-pipenv-file-editable && pip install -q --disable-pip-version-check --root-user-action=ignore pipenv==2023.8.19 && pipenv install && pipenv run pip show mypackage | grep -i location'
...
Location: /root/.local/share/virtualenvs/testcase-pipenv-file-editable-YkSiblPt/lib/python3.12/site-packages
Editable project location: /testcase-pipenv-file-editable/packages/mypackage
```

Note in the second output, the `Editable project location` shows that the package was installed in editable mode, when it was not before. 

[...]

PR #6069, opened 2024-01-28
Only editable entry should trigger editable installs

matteius:
Exploration of #6054 

### The checklist

* [X] Associated issue

### DIFF
--- a/news/6069.bugfix.rst
+++ b/news/6069.bugfix.rst
@@ -1,0 +1,1 @@
+Adjusted logic which assumed any file, path or VCS install should be considered editable.  Instead relies on the user specified editable flag to mark requirement as editable install.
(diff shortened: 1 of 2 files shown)


oz123 (approved):

matteius:
@edmorley if you look back on the issue that is still marked open, no one has responded to me when I ask for more details like what the difference even is between a editable and non editable local file install--

https://github.com/pypa/pipenv/issues/6054#issuecomment-1913572665

I'm not opposed to adding test cases, however I think if you look you'll see I've added the most new test cases in the last two years, and the test runner used to take 45 minutes plus, now takes a fraction of the time, wasn't without a lot of hard effort.  Yes the test runner needs more cases and more TLC, no I cannot be the one to ensure all those tests exist.  I've refactored and simplified a lot of code in pipenv, and I often get caught by the lack of tests that already existed around original functionality folks expects.   It is easy to ask for more tests, harder to implement them.   I've spent hours at points trying to add tests for some things only to back them out for reasons due to the nature of trying to test some of this stuff.

matteius:
Like I am not even convinced it is fixed, that is why I left the original issue open.   But I thought this is a step on the path.

Issue #6107, opened 2024-03-08
`pipenv shell` not activating env on PowerShell for mac (possibly *nix)


samcarswell:
### Issue description

Using `pipenv shell` on Powershell for mac does not activate the environment. Instead it displays the source command for bash and does not attempt to execute it. I can add `.ps1` and hit enter to activate the environment correctly.

### Expected result

`pipenv shell` to automatically activate the environment, like it does with fish or bash.

### Actual result

```shell
psh ❯ pipenv shell
Creating a virtualenv for this project...
Pipfile: /Users/scar/tmp/test-pipenv/Pipfile
Using default python from /Users/scar/.pyenv/versions/3.11.8/bin/python3.11 (3.11.8) to create virtualenv...
⠙ Creating virtual environment...created virtual environment CPython3.11.8.final.0-64 in 162ms
  creator CPython3Posix(dest=/Users/scar/.local/share/virtualenvs/test-pipenv-W2MgHCvQ, clear=False, no_vcs_ignore=False, global=False)
  seeder FromAppData(download=False, pip=bundle, setuptools=bundle, wheel=bundle, via=copy, app_data_dir=/Users/scar/Library/Application Support/virtualenv)
    added seed packages: pip==24.0, setuptools==69.0.3, wheel==0.42.0
  activators BashActivator,CShellActivator,FishActivator,NushellActivator,PowerShellActivator,PythonActivator

✔ Successfully created virtual environment!
Virtualenv location: /Users/scar/.local/share/virtualenvs/test-pipenv-W2MgHCvQ
Creating a Pipfile for this project...
Launching subshell in virtual environment...
 . /Users/scar/.local/share/virtualenvs/test-pipenv-W2MgHCvQ/bin/activate
PowerShell 7.4.1

~/tmp/test-pipenv 🐍 v3.11.8
psh ❯  . /Users/scar/.local/share/virtualenvs/test-pipenv-W2MgHCvQ/bin/activate
```

The shell is left here, with the bash source command not executed.

### Steps to replicate

Displayed above in actual result

-------------------------------------------------------------------------------

<details><summary>$ pipenv --support</summary>

Pipenv version: `'2023.12.1'`

Pipenv location: `'/Users/scar/.local/lib/python3.11/site-packages/pipenv'`

Python location: `'/Users/scar/.pyenv/versions/3.11.8/bin/python3.11'`

OS Name: `'posix'`

User pip version: `'23.3.2'`

user Python installations found:

[...]

PR #6108, opened 2024-03-08
Use fancy mode for pwsh on *nix (draft)

samcarswell:
Ran into this issue when attempting to activate pipenv environments using pwsh on mac. I'm not 100% sure this is the correct solution as the [documentation](https://docs.pipenv.org/basics/#about-shell-configuration) for fancy mode is sparse, but it seems the code enforces fancy mode if running on windows (because it's assumed it's pwsh?)

Definitely looking for some feedback. Thanks.

### The issue

Issue is described in https://github.com/pypa/pipenv/issues/6107

### The fix

Defaulting to fancy mode for pwsh on *nix

### The checklist

* [X] Associated issue

### DIFF
--- a/pipenv/cli/command.py
+++ b/pipenv/cli/command.py
@@ -388,8 +388,12 @@ def shell(state, fancy=False, shell_args=None, anyway=False, quiet=False):
             sys.exit(1)
     # Load .env file.
     load_dot_env(state.project)
-    # Use fancy mode for Windows.
-    if os.name == "nt":
+    # Use fancy mode for Windows or pwsh on *nix.
+    if (
+        os.name == "nt"
+        or os.environ["PIPENV_SHELL"].split(os.path.sep)[-1] == "pwsh"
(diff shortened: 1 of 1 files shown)


pipenv/cli/command.py
```diff
@@ -388,8 +388,12 @@ def shell(state, fancy=False, shell_args=None, anyway=False, quiet=False):
             sys.exit(1)
     # Load .env file.
     load_dot_env(state.project)
-    # Use fancy mode for Windows.
-    if os.name == "nt":
+    # Use fancy mode for Windows or pwsh on *nix.
+    if (
+        os.name == "nt"
+        or os.environ["PIPENV_SHELL"].split(os.path.sep)[-1] == "pwsh"
```
matteius, line 394: I am not  sure if these lookups from os.environ should actually be gets in case they are not defined?

matteius (approved):

Issue #6126, opened 2024-04-11
pipenv check command crashes with FileNotFoundError if the Pipfile.lock does not exist.


jclopes:
Be sure to check the existing issues (both open and closed!), and make sure you are running the latest version of Pipenv.

Check the [diagnose documentation](https://pipenv.pypa.io/en/latest/diagnose/) for common issues before posting! We may close your issue if it is very similar to one of them. Please be considerate, or be on your way.

Make sure to mention your debugging experience if the documented solution failed.

### Issue description

The check command crashes on a freshly new project.
pipenv, version 2023.12.1

### Expected result

Check completed and passed since there is nothing there.

### Actual result

The check command crashes with the exception FileNotFoundError and refers to the 'Pipfile.lock' missing.

  File "/usr/lib/python3.11/site-packages/pipenv/project.py", line 1292, in load_lockfile
    with open(self.lockfile_location, encoding="utf-8") as lock:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/home/uuuuu/xxxxxx/Pipfile.lock'

### Steps to replicate

Created a new environment with the shell command:
    $ pipenv shell

Executed a check command immediately after:
    $ pipenv check

-------------------------------------------------------------------------------

<details><summary>$ pipenv --support</summary>

Pipenv version: `'2023.12.1'`

Pipenv location: `'/usr/lib/python3.11/site-packages/pipenv'`

Python location: `'/usr/bin/python'`

OS Name: `'posix'`

User pip version: `'23.3.2'`

user Python installations found:

PEP 508 Information:

```
{'implementation_name': 'cpython',
 'implementation_version': '3.11.8',
 'os_name': 'posix',
 'platform_machine': 'x86_64',
 'platform_python_implementation': 'CPython',
 'platform_release': '6.8.4-arch1-1',
 'platform_system': 'Linux',
 'platform_version': '#1 SMP PREEMPT_DYNAMIC Fri, 05 Apr 2024 00:14:23 +0000',
 'python_full_version': '3.11.8',
 'python_version': '3.11',
 'sys_platform': 'linux'}
```

System environment variables:

  - `SHELL`
  - `COLORTERM`
  - `XDG_SESSION_PATH`
  - `PIPENV_ACTIVE`
  - `I3SOCK`
  - `LC_ADDRESS`
  - `LC_NAME`
[...]

PR #6135, opened 2024-04-21
Fix the issue(#6126): add lockfile_path presence in pipenv check command

nitintecg:

### The issue

This PR fixes issue #6126 

### The fix

I have added additional check of presence of lockfile in pipenv check command so that it doesn't fails if Pipfile.lock is not present.

### The checklist

* [X] Associated issue


matteius (approved):
Please add a bugfix news fragment with the PR #

PR #6140, opened 2024-04-25
prototyping pythonfinder pep514 support

matteius:
Prototyping bringing back a level of pep514 support to pythonfinder (would back-vendor it to pythonfinder if successful).

### The checklist

* [ ] Associated issue


pipenv/vendor/pythonfinder/models/python.py
```diff
@@ -302,6 +317,39 @@ def which(self, name) -> PathEntry | None:
...
+                        except FileNotFoundError:
+                            continue
+
+                        version = Version(version_key_name)
+                        launcher_entry = WindowsLauncherEntry(
+                            version=version,
+                            install_path=install_path,
+                            executable_path=executable_path,
+                            company=company,
+                        )
```
TBBle, line 346: I think the current code is inconsistent with the docs and examples in [PEP-514](https://peps.python.org/pep-0514), and also doesn't match what I see in my local registry. I think it should look something like this, untested code:

```suggestion

                    with winreg.OpenKey(key, version_key_name) as version_key:
                        try:
                            install_path_key = winreg.OpenKey(version_key, "InstallPath")
                            try:
                                executable_path = winreg.QueryValue(install_path_key, "ExecutablePath")
                            except FileNotFoundError:
                                # TODO: Only a valid default for PythonCore, otherwise skip.
                                executable_path = winreg.QueryValue(install_path_key, None)+"\\python.exe"
                        except FileNotFoundError:
                            continue

                        try:
                            version = Version(winreg.QueryValue(key, "SysVersion"))
                        except FileNotFoundError:
                            # TODO: Only a valid default for PythonCore, otherwise unknown so will need to be probed later.
                            version = Version(version_key_name)

                        try:
                            architecture = winreg.QueryValue(key, "SysArchitecture")
                        except FileNotFoundError:
                            # TODO: Implement PEP-514 defaults for architecture for PythonCore based on key and OS architecture.
                            architecture = None
                        
                        launcher_entry = WindowsLauncherEntry(
                            version=version,
                            executable_path=executable_path,
                            company="PythonCore",
                        )
```

I also extracted SysArchitecture while I was there, so that can be plumbed through into `PythonVersion.from_windows_launcher`.

pipenv/vendor/pythonfinder/models/python.py
```diff
@@ -302,6 +317,39 @@ def which(self, name) -> PathEntry | None:
         non_empty_match = next(iter(m for m in matches if m is not None), None)
         return non_empty_match
 
+    def find_python_versions_from_windows_launcher(self):
+        # Open the registry key for Python launcher
+        key_path = r"Software\Python\PythonCore"
+        try:
+            import winreg
+            with winreg.OpenKey(winreg.HKEY_LOCAL_MACHINE, key_path) as key:
```
TBBle, line 325: There's two other registry locations we need to apply the same process to, to find user-land installations and 32-bit installs on 64-bit Windows.

pipenv/vendor/pythonfinder/models/python.py
```diff
@@ -302,6 +317,39 @@ def which(self, name) -> PathEntry | None:
         non_empty_match = next(iter(m for m in matches if m is not None), None)
         return non_empty_match
 
+    def find_python_versions_from_windows_launcher(self):
+        # Open the registry key for Python launcher
+        key_path = r"Software\Python\PythonCore"
```
TBBle, line 322: It'd be nice to support non-PythonCore environments at some point, e.g., conda.

pipenv/vendor/pythonfinder/models/python.py
```diff
@@ -39,6 +39,14 @@
 logger = logging.getLogger(__name__)
 
 
+class WindowsLauncherEntry:
+    def __init__(self, version: Version, install_path: str, executable_path: str, company: Optional[str]):
+        self.version = version
+        self.install_path = install_path
+        self.executable_path = executable_path
+        self.company = company
+
```
TBBle, line 48:
```suggestion
        self.company = company
        self.architecture = architecture
```

pipenv/vendor/pythonfinder/models/python.py
```diff
@@ -39,6 +39,14 @@
 logger = logging.getLogger(__name__)
 
 
+class WindowsLauncherEntry:
+    def __init__(self, version: Version, install_path: str, executable_path: str, company: Optional[str]):
```
TBBle, line 43:
```suggestion
    def __init__(self, version: Version, install_path: str, executable_path: str, company: str, architecture: Optional[str]):
```

pipenv/vendor/pythonfinder/models/python.py
```diff
@@ -302,6 +317,51 @@ def which(self, name) -> PathEntry | None:
...
+                            architecture = winreg.QueryValue(key, "SysArchitecture")
+                        except FileNotFoundError:
+                            # TODO: Implement PEP-514 defaults for architecture for PythonCore based on key and OS architecture.
+                            architecture = None
+                        
+                        launcher_entry = WindowsLauncherEntry(
+                            version=version,
+                            executable_path=executable_path,
+                            company="PythonCore",
+                        )
```
TBBle, line 358:
```suggestion
                            company="PythonCore",
                            architecture=architecture,
                        )
```

pipenv/vendor/pythonfinder/models/python.py
```diff
@@ -577,39 +637,28 @@ def parse_executable(cls, path) -> dict[str, str | int | Version | None]:
...
-                "company": company,
-            }
+        py_version = cls.create(
+            major=launcher_entry.version.major,
+            minor=launcher_entry.version.minor,
+            patch=launcher_entry.version.micro,
+            is_prerelease=launcher_entry.version.is_prerelease,
+            is_devrelease=launcher_entry.version.is_devrelease,
+            is_debug=False,  # Assuming debug information is not available from the registry
+            architecture=None,  # Assuming architecture information is not available from the registry
```
TBBle, line 655:
```suggestion
            architecture=launcher_entry.architecture,
```

TBBle (changes requested):
Quickly plumbed through the architecture, since we should have that available for Python 3.5 onwards.

pipenv/vendor/pythonfinder/models/python.py
matteius: Thanks for your help with all this -- greatly appreciated.   Do you know what the other paths would be?  I can get back to looking at this soon.

TBBle: Per [PEP-0514](https://peps.python.org/pep-0514/#structure), the three registry paths that need to be checked are:

```
HKEY_CURRENT_USER\Software\Python\<Company>\<Tag>
HKEY_LOCAL_MACHINE\Software\Python\<Company>\<Tag>
HKEY_LOCAL_MACHINE\Software\Wow6432Node\Python\<Company>\<Tag>
```

You can check them all the same way, there's some notes there about what to do if you see the same Company-Tag combination in different registry paths, and specific notes about how to disambiguate 32-bit and 64-bit Python older than 3.5, which did not include architecture in its Tag, and so has the same Company-Tag for 32-bit and 64-bit installs.

matteius: Just took a stab at the refactor, not really sure if it's 100% correct though.

pipenv/vendor/pythonfinder/models/python.py
```diff
@@ -318,50 +318,80 @@ def which(self, name) -> PathEntry | None:
...
+                                    except FileNotFoundError:
+                                        continue
+
+            except FileNotFoundError:
+                pass
+
+    def _get_registry_value(self, key, value_name, default=None):
+        try:
+            import winreg
+            return winreg.QueryValue(key, value_name)
```
TBBle, line 392: Although this matches the code from PEP-514, it doesn't actually work. (I just checked, and the example from PEP-514 also fails. With the PEP-514 example code, my system-wide 64-bit Python 3.10 install reports as "Python 3.1 (Unknown)". Because of the default logic, at the time of PEP-514's writing, the output was mostly correct, so maybe no one noticed?)

It should be something like:

```suggestion
            import winreg
            # TODO: Not specified in PEP-514, but probably should reject any value that isn't `winreg.REG_SZ`.
            # PEP-514's (faulty) example code uses `winreg.QueryValue`, which can only return `winreg.REG_SZ` values.
            # They document "string value" which could match `winreg.REG_SZ` or `winreg.REG_EXPAND_SZ`.
            return winreg.QueryValueEx(key, value_name)[0]
```

This is because `winreg.QueryValue`'s second parameter is actually a sub-key to query for its default value, not a value name. `winreg.QueryValueEx` expects you to have opened the target key already (as we have) and so its second parameter is the value name, or `None` for the default value.

I note that pep514tools doesn't use `winreg.QueryValue` either, in favour of dumping all the values into a dictionary using `winreg.EnumValue`, so it wouldn't have seen this issue. It also only accepts `winreg.REG_SZ` values.

matteius:
After spending a few hours on it tonight and trying to test it and seeing differences in behaviors between git bash and powershell, and also it finding my pyenv versions and maybe not even my windows versions -- I feel this still needs a ton of work.

matteius:
@TBBle -- I rewrote python finder over my lunch break -- it seems to work with pyenv on linux + windows store python installs from registry.   https://github.com/pypa/pipenv/pull/6360 replaces this PR.

PR #6146, opened 2024-04-27
Fix windows CI for 3.8 and the outdated command that was refactored to remove pkg_resources.

matteius:
Fix windows CI for 3.8 and the outdated command that was refactored to remove pkg_resources.


matteius:
The ruff issue is fixed in the install refactor PR.

matteius:
May as well go with the newer importlib-metadata since it works the same as the older one.  I commited skipping those windows py 3.8 tests to the install refactor branch after spending 2 hours trying to get them to pass and seeing they actually could work locally outside of pytest.

Issue #6144, opened 2024-04-27
Specify Python Version Range in Pipfile


Semnodime:
### Issue description

```
[requires]
python_version = ">=3.12"
```

### Expected result

Choose the latest python version available on the system to satisfy the constraints or raise the constraint error.

### Actual result
```
Warning: Python >=3.12 was not found on your system...
Neither 'pyenv' nor 'asdf' could be found to install Python.
```

### Steps to replicate
```
pipenv --python 3.12
# Change last line in Pipfile to `python_version = ">=3.12"`
pipenv --rm
pipenv --install
```

matteius:
Only absolute values are supported, and when upgrading python versions virtualenvs need to be recreated.

Semnodime:
> Only absolute values are supported, and when upgrading python versions virtualenvs need to be recreated.

Correct. That's what I observed. This is a bug since the rest of the Pipfile uses the same version specifiers and `pipenv --install` does not communicate that only absolute values are supported and a range has been specified.
I acknowledge that one could conclude that from the pipenv error, but IMHO it'd be better to signal this parsing constraint explicitly along the error message.

sanspareilsmyn:
@Semnodime @matteius I'd love to handle this issue. I'll send a PR to fix this in near future.

matteius:
@sanspareilsmyn I think I have it solved in my open PR that fixes a number of issues -- if either of you or @Semnodime can check out this branch:  https://github.com/pypa/pipenv/pull/6282 and give it a whirl.

PR #6164, opened 2024-05-24
Handle unsupported python versioning on Pipfile

sanspareilsmyn:
### The issue

This is a bug fix of Pipfile [requires] regards to https://github.com/pypa/pipenv/issues/6144
Pipenv only supports absolute python version. If the user specifies a Python version with inequality signs like >=3.12, <3.12 in the [requires] field, the code has been modified to explicitly express in an error log that absolute versioning must be used.

### The fix
- Add protection logic to parse python verison on Pipfile.
- Enhance error log that absolute versioning must be used.
- Add test codes.

### The checklist

* [x] Associated issue

### DIFF
--- a/pipenv/utils/virtualenv.py
+++ b/pipenv/utils/virtualenv.py
@@ -241,18 +241,11 @@ def ensure_python(project, python=None):
         if python:
             range_pattern = r"^[<>]=?|!="
             if re.search(range_pattern, python):
-                click.echo(
-                    "{}: Python version range specifier '{}' is not supported. {}".format(
-                        click.style("Error", fg="red", bold=True),
-                        click.style(python, fg="cyan"),
-                        click.style(
-                            "Please use an absolute version number or specify the path to the Python executable on Pipfile.",
-                            fg="yellow",
-                        ),
-                    ),
-                    err=True,
+                err.print(
+                    f"[bold red]Error[/bold red]: Python version range specifier '[cyan]{python}[/cyan]' is not supported. "
+                    "[yellow]Please use an absolute version number or specify the path to the Python executable on Pipfile.[/yellow]"
                 )
-                abort()
+                sys.exit(1)
 
     if not python:
         python = project.s.PIPENV_DEFAULT_PYTHON_VERSION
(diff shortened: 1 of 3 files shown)


pipenv/utils/virtualenv.py
```diff
@@ -237,6 +238,22 @@ def abort(msg=""):
     # Find out which python is desired.
     if not python:
         python = project.required_python_version
+        if python:
+            range_pattern = r"^[<>]=?|!="
+            if re.search(range_pattern, python):
+                click.echo(
```
matteius, line 244: We have been migrating away from click.echo to use the rich console -- you can find other examples in the code of how to print the equivalent thing using that.

pipenv/utils/virtualenv.py
```diff
@@ -237,6 +238,22 @@ def abort(msg=""):
...
+                        click.style("Error", fg="red", bold=True),
+                        click.style(python, fg="cyan"),
+                        click.style(
+                            "Please use an absolute version number or specify the path to the Python executable on Pipfile.",
+                            fg="yellow",
+                        ),
+                    ),
+                    err=True,
+                )
+                abort()
```
matteius, line 255: Maybe `abort` is fine, not sure how it behaves different then the other runtime errors we raise--just calling out this is an uncommon pattern in the project so deserves additional thought to standardize on a best approach.

pipenv/utils/virtualenv.py
sanspareilsmyn: @matteius Thanks for guidance. I'll check and modify it to the convention.

pipenv/utils/virtualenv.py
sanspareilsmyn: @matteius I'll think about it more about how to handle the runtime errors in another ways.

matteius: How about `sys.exit(1)`?   That is what I am using in my PR to clean up and not render a stack error trace when resolver fails.

matteius (approved):

PR #6177, opened 2024-06-13
Fix bump version in CI

oz123:
Automatic bump of the version in CI after release was broken mistakenly.
This should fix the issue.


matteius:
This is still broken fwiw.

PR #6178, opened 2024-06-16
refactor: click styling-> rich

oz123:
Replace in shell.py and virtualenv.py.
Also, replace in some exceptions.

This is continuing my previous work in attempt to remove click in favor of rich.

@aidencullo

### DIFF
--- a/pipenv/exceptions.py
+++ b/pipenv/exceptions.py
@@ -58,15 +58,14 @@ def handle_exception(exc_type, exception, traceback, hook=sys.excepthook):
 sys.excepthook = handle_exception
 
 
-class PipenvException(ClickException):
-    message = "{}: {{}}".format(click.style("ERROR", fg="red", bold=True))
+class PipenvException:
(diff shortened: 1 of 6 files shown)


pipenv/exceptions.py
```diff
@@ -62,13 +72,14 @@ def __init__(self, message=None, **kwargs):
...
+        console = Console(file=file)
         if self.extra:
             if isinstance(self.extra, str):
                 self.extra = [self.extra]
             for extra in self.extra:
                 extra = f"[pipenv.exceptions.{self.__class__.__name__}]: {extra}"
-                click.echo(extra, file=file)
-        click.echo(f"{self.message}", file=file)
+                console.print(extra)
+        console.print(f"{self.message}")
```
aidencullo, line 82: i think this show function is unreachable, were you able to hit it?

oz123: I have not seen an exception raised with extra in the code. But we might just want to leave it, as it should be used.

PR #6179, opened 2024-06-17
Swap old_version and new_version in `pipenv update --outdated` output

BarrensZeppelin:
The old and new versions of packages reported by `pipenv update --outdated` are swapped.


oz123 (changes requested):
This looks fine to me, I still need to really test it.
Can you please add a news file under news ?

BarrensZeppelin:
Do you need more from my side, @oz123 ?

matteius (approved):

PR #6182, opened 2024-06-21
Remove click.echo from pipenv/cli

aidencullo:
### The issue

We want to remove vendored click library as it is sizable and much of its functionality can be done with rich already.

### The fix

replace `click.echo` calls in:

pipenv/cli/command.py
pipenv/cli/options.py

replace `click.style` calls in:

pipenv/utils/display.py

### Notes

This will not remove all uses of click library in pipenv/cli, this is part of an incremental transition from click to rich.

@oz123

### DIFF
--- a/news/6182.vendor.rst
+++ b/news/6182.vendor.rst
@@ -1,0 +1,1 @@
+Remove click.echo from pipenv/cli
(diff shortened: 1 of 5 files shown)


oz123 (changes requested):
Please add a news file. I forgot this in your other PR.

PR #6192, opened 2024-07-06
Remove all instances of click from project.py

aidencullo:
The issue
migrate from click to rich

The fix
replace click calls in:

pipenv/project.py

@oz123


aidencullo:
need to rename the branch

PR #6199, opened 2024-07-10
Fix incorrect command reminder after installs

nmagee:
### Issue

When installing new packages using `pipenv install <package>` the following reminder may be returned when the `Pipfile` is out of date:

    Pipfile.lock (<hash>) out of date: run `pipfile lock` to update to (<new-hash>)...

However, no such command `pipfile lock` exists. The next returned line proceeds to run the correct command for the user:

    Running $ pipenv lock then $ pipenv sync.

So it appears the author actually meant the reminder message to read `pipenv lock` since that is what is subsequently run.

### Fix

Updated line of `pipenv/routines/install.py` to return correct command. See [here](https://github.com/nmagee/pipenv/commit/7a8f12d5507d7166bc0f4255f7e31ae9d1711c68#diff-d8d49f5505a98880a70cb2fd653aa0dc25eed5a0f3da3e74eab6a1157a8d3014L617)


oz123 (changes requested):
Looks OK. Please add a news file under news/

nmagee:
@oz123 this is all set now. Thanks!

PR #6029, opened 2023-12-01
Smarter uninstall

matteius:
Ran into this at work recently, uninstall does a complete lock so not nearly as useful as upgrade path.   This makes the uninstall smarter like the upgrade path is (but the inverse).

### The issue (before this change)
`$ pipenv  uninstall tomli --categories=develop`

```
$ git diff
diff --git a/Pipfile b/Pipfile
index fa27de08..21799786 100644
--- a/Pipfile
+++ b/Pipfile
@@ -24,7 +24,6 @@ colorama= {version = "*", sys_platform = "== 'win32'"}
 myst-parser = {extras = ["linkify"], version = "*"}
 invoke = "==2.0.0"
 exceptiongroup = "==1.1.0"
-tomli = "*"
 pyyaml = "==6.0.1"
 build = "*"
 twine = "*"
diff --git a/Pipfile.lock b/Pipfile.lock
index 861d9132..39d53450 100644
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "4d1a9f3599cadad8b7884cd1143a964290191e9ee6a2ebf7acec9e6e27f2a69e"
+            "sha256": "d7cdc853e6e34c10abbcd19a0c79905a61d12d50233e5b13259ecf3331ddde04"
         },
         "pipfile-spec": 6,
         "requires": {},
@@ -55,11 +55,11 @@
         },
         "babel": {
             "hashes": [
-                "sha256:04c3e2d28d2b7681644508f836be388ae49e0cfe91465095340395b60d00f210",
-                "sha256:fbfcae1575ff78e26c7449136f1abbefc3c13ce542eeb13d43d50d8b047216ec"
+                "sha256:33e0952d7dd6374af8dbf6768cc4ddf3ccfefc244f9986d4074704f2fbd18900",
+                "sha256:7077a4984b02b6727ac10f1f7294484f737443d7e2e66c5e4380e41a3ae0b4ed"
             ],
             "markers": "python_version >= '3.7'",
-            "version": "==2.13.0"
+            "version": "==2.13.1"
         },
         "beautifulsoup4": {
             "hashes": [
@@ -109,11 +109,11 @@
         },
         "certifi": {
             "hashes": [
-                "sha256:539cc1d13202e33ca466e88b2807e29f4c13049d6d87031a3c110744495cb082",
[...]

PR #6209, opened 2024-07-29
Corrected behavior of uninstall --all

### LINKED ISSUE #6200: Major pipenv uninstall --all behavior change
### Issue description

Running `pipenv uninstall --all` (the command) with the latest version of pipenv uninstalls all packages in the `[packages]` group and removes them from the Pipfile. It doesn't uninstall packages from any other group.

### Expected result

I had been relying on the behavior last seen in Release v2023.12.1, where running the command produced output like:
```
Un-installing all [dev-packages] and [packages]...
Found 47 installed package(s), purging...
Found existing installation: six 1.16.0
Uninstalling six-1.16.0:
  Successfully uninstalled six-1.16.0
Found existing installation: Jinja2 3.1.4
Uninstalling Jinja2-3.1.4:
  Successfully uninstalled Jinja2-3.1.4
Found existing installation: cryptography 42.0.8
Uninstalling cryptography-42.0.8:
  Successfully uninstalled cryptography-42.0.8
Found existing installation: packaging 24.1
...
Environment now purged and fresh!
```

There's a minor bug in that output in that it only lists that it's uninstalling from the `[dev-packages]` and `[packages]` group, but that's just a display issue. It does actually uninstall all packages in all groups from the virtual environment and leaves the Pipfile untouched. This is consistent with the [documentation](https://pipenv.pypa.io/en/latest/commands.html#uninstall).

### Actual result

```
Un-installing all [packages]...
Uninstalling src...
Found existing installation: src 1.0
Uninstalling src-1.0:
  Successfully uninstalled src-1.0
```

For this project, the `[packages]` group only contains one package, `src-1.0`. All the other packages are in either `[dev-packages]` or `[layer-packages]`. The current version of pipenv only uninstalls packages from the `[packages]` group and removes them from the Pipfile as well.

### Steps to replicate

Create a Pipfile with several groups:
```toml
[[source]]
name = "pypi"
verify_ssl = true

# For local development, run
# pipenv install --categories="packages layer-packages dev-packages"

[packages]
src = {editable = true, path = "./src"}

[requires]
python_version = "3.11"

[dev-packages]
moto = ">=5.0.0"
pytest = "*"
pytest-monitor = "*"
mypy = "*"
boto3-stubs = "*"
pystache = "*"
types-requests = "*"
pytest-mock = "*"

[layer-packages]
boto3 = "*"
urllib3 = "*"
botocore = "*"
appsync-client = {path = "./../../../../../packages/appsync_client"}
pandas = "2.2.2"
openpyxl = "3.1.5"
aws-lambda-powertools = "*"
```
Install all of the packages:

```bash
pipenv install --categories="packages layer-packages dev-packages"
```

Now run the command:
```bash
pipenv uninstall --all
```

Examine the Pipfile, and see that the `[packages]` group is empty, but the other groups are still populated. Examine the site-packages in the virtualenv and note that the all group packages are still there.

### Notes on the code

It appears that the [`do_purge`](https://github.com/pypa/pipenv/blob/5329af8d914639a692e5ac8200fe2dffebc1115c/pipenv/routines/uninstall.py#L169) function is now an orphan. There's no active code that will display [`Environment now purged and fresh!`](https://github.com/pypa/pipenv/blob/5329af8d914639a692e5ac8200fe2dffebc1115c/README.md?plain=1#L321).

matteius:

### The issue

When I recently refactored the uninstall command behavior to be smarter, I broke the expected behavior of `--all` flag.
Specifically there should not be removal from the Pipfile and the do_purge logic should be invoked (it wasn't after my change).
Fixes #6200 

### The checklist

* [X] Associated issue

### DIFF
--- a/pipenv/routines/uninstall.py
+++ b/pipenv/routines/uninstall.py
@@ -15,6 +15,7 @@ from pipenv.utils.requirements import BAD_PACKAGES
 from pipenv.utils.resolver import venv_resolve_deps
 from pipenv.utils.shell import cmd_list_to_shell, project_python
 from pipenv.vendor import click
+from pipenv.vendor.importlib_metadata.compat.py39 import normalized_name
 
 
 def _uninstall_from_environment(project, package, system=False):
@@ -82,21 +83,28 @@ def do_uninstall(
     if all:
         click.secho(
             click.style(
-                "Un-installing all {}...".format(click.style("[packages]", fg="yellow")),
+                "Un-installing all packages...",
                 bold=True,
             )
         )
-        # Uninstall all dev-packages from environment
-        for package in project.get_pipfile_section("packages"):
-            _uninstall_from_environment(project, package, system=system)
-        # Remove the package from the Pipfile
-        if project.reset_category_in_pipfile(category="packages"):
-            click.echo("Removed [packages] from Pipfile.")
+        # Uninstall all packages from all groups
+        for category in project.get_package_categories():
+            if category in ["source", "requires"]:
+                continue
+            for package in project.get_pipfile_section(category):
+                _uninstall_from_environment(project, package, system=system)
 
-        # Finalize changes to lockfile
-        lockfile_content["default"] = {}
+        # Clear all categories in the lockfile
+        for category in list(lockfile_content.keys()):
+            if category != "_meta":
+                lockfile_content[category] = {}
+
         lockfile_content.update({"_meta": project.get_lockfile_meta()})
         project.write_lockfile(lockfile_content)
+
+        # Call do_purge to remove all packages from the environment
+        do_purge(project, bare=False, downloads=False, allow_global=system)
+        return
 
     package_args = list(packages) + [f"-e {pkg}" for pkg in editable_packages]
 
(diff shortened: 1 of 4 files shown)


pipenv/routines/uninstall.py
```diff
@@ -82,22 +82,32 @@ def do_uninstall(
...
-        # Remove the package from the Pipfile
-        if project.reset_category_in_pipfile(category="packages"):
-            click.echo("Removed [packages] from Pipfile.")
+        # Uninstall all packages from all groups
+        for category in project.pipfile_sources:
+            if category in ["source", "requires"]:
+                continue
+            for package in project.get_pipfile_section(category):
+                _uninstall_from_environment(project, package, system=system)
+            # Remove the packages from the Pipfile
```
dan-hook, line 95: I can't verify that this comment is accurate for what the code is doing, but removing the packages from the Pipfile is exactly what we don't want to do during uninstall --all to be consistent with the documented behavior.  

There was a test changed in a previous PR that should be reverted:  https://github.com/pypa/pipenv/pull/6029#discussion_r1674707464
If the test as it currently stands is passing, then uninstall --all isn't working like it used to.

dan-hook (changes requested):
Thanks for looking into this.

dan-hook (approved):

tests/integration/test_uninstall.py
```diff
@@ -127,7 +130,6 @@ def test_uninstall_all_dev(pipenv_instance_private_pypi):
 
         c = p.pipenv("uninstall --all-dev")
         assert c.returncode == 0
-        assert p.pipfile["dev-packages"] == {}
```
matteius, line 130: I think I need to revert this part of the PR -- while it doesn't make sense to me in terms of the code paths that --all-dev does something different than --all, I guess it does.

Issue #6197, opened 2024-07-09
Linuxbrew shells are detected as `nu` shell


JoshStern:
### Issue description

When running `pipenv shell` using a brew-installed shell the `cmd` will include `linuxbrew` which trips the [nu branch](https://github.com/pypa/pipenv/blob/5329af8d914639a692e5ac8200fe2dffebc1115c/pipenv/shells.py#L50)  when identifying the activate file and command to run.

### Expected result

Shells with `nu` elsewhere in the path are not detected as `nu` shell.

### Actual result

```
$ pipenv shell
Launching subshell in virtual environment...
 overlay use <venvpath>/bin/activate.nu
```
Followed by (I am using `zsh`):
```
$  overlay use <venvpath>/bin/activate.nu
zsh: command not found: overlay
```

### Steps to replicate

This can be replicated without a Pipfile:
1. Install brew from [guide](https://brew.sh/).
2. Install `pipenv`, and `zsh`:
    ```sh
    brew install pipenv zsh
    ```
3. Start `zsh`.
4. Attempt to start a pipenv shell:
    ```sh
    pipenv shell
    ```

### Proposed fix
This may require some conversation. Switching the branches to check the end of the path would be a breaking change for anyone who depends on the current behavior. It may be worth adding a `zsh` branch which does the same check before `nu`. I'm happy to make the change with some guidance from a contributor.

-------------------------------------------------------------------------------
<details><summary>$ pipenv --support</summary>

Pipenv version: `'2024.0.1'`

Pipenv location: `'/home/linuxbrew/.linuxbrew/Cellar/pipenv/2024.0.1/libexec/lib/python3.12/site-packages/pipenv'`

Python location: `'/home/linuxbrew/.linuxbrew/Cellar/pipenv/2024.0.1/libexec/bin/python'`

OS Name: `'posix'`

User pip version: `'24.0'`

user Python installations found:

PEP 508 Information:

```
{'implementation_name': 'cpython',
 'implementation_version': '3.12.3',
 'os_name': 'posix',
 'platform_machine': 'x86_64',
 'platform_python_implementation': 'CPython',
 'platform_release': '6.2.0-33-generic',
 'platform_system': 'Linux',
 'platform_version': '#33~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Thu Sep  7 '
                     '10:33:52 UTC 2',
 'python_full_version': '3.12.3',
 'python_version': '3.12',
 'sys_platform': 'linux'}
```

[...]

PR #6212, opened 2024-07-29
Check end of path for shell detection

JoshStern:

### The issue
Shell detection checks the entire path when looking for shell substrings. This can cause the wrong branch to get tripped if shells are installed in a path with another shell in the tree. In my case a `li[nu]xbrew`-installed `zsh` goes down the `nu` shell path.

### The fix

This updates the check to only look at the end of `cmd` when detecting shells.

### The checklist

* [x] Associated issue: #6197 

### DIFF
--- a/news/6197.bugfix.rst
+++ b/news/6197.bugfix.rst
@@ -1,0 +1,1 @@
+Update shell detection to only check the end of the command used.
--- a/pipenv/shells.py
+++ b/pipenv/shells.py
@@ -37,7 +37,10 @@ def _get_activate_script(cmd, venv):
     """
     # Suffix and source command for other shells.
     # Support for fish shell.
-    if cmd.endswith("/fish"):
+    if cmd.endswith("/sh", "/bash", "/zsh"):
+        suffix = ""
+        command = "."
+    elif cmd.endswith("/fish"):
         suffix = ".fish"
         command = "source"
     # Support for csh shell.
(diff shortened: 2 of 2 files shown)


oz123 (changes requested):
Actually, I prefer this over the split over `.linuxbrew`.
This solution will cause future maintainers to raise eyebrows and wrinkle their foreheads trying to figure it out without reading the original issue.

This approach is much more maintainable. However, I have a small suggestion:

Change:

```
cmd.endswith("<shell name>")
```
to:
```
cmd.endswith("/<shell name>"):
```

This will allow to for shortcutting the if ...elif at the head in favor of the most common shells (bash, zsh, sh):

```
if cmd.endswith("/sh","/bash","/zsh")":
   suffix = ""
  command = "."
elif cmd.endswith("/csh"):
  ...
  
else:
     raise ValueError("unknown shell ...") ....
```

This works for all three:
```
>>> cmd="/bin/sh"
>>> cmd.endswith(("/bash","/zsh", "/sh"))
True
>>> cmd="/bin/zsh"
>>> cmd.endswith(("/bash","/zsh", "/sh"))
True
>>> cmd="/bin/bash"
>>> cmd.endswith(("/bash","/zsh", "/sh"))
True
>>> 
```

oz123:
@JoshStern thank you for your effort so far (immediately adding news file, is great! ).

I gave your PR some consideration and I think that your solution is better than what I initially suggested.
Adding a shortcut for the most common shells is only possible with your approach.

JoshStern:
@oz123 updated! Thank you for the feedback and quick turnaround.

oz123:
Yes. Please. The zsh, bash and sh bash should be at the top. Unknown shells should not fall silently to imaginary default. They should raise exceptions.

JoshStern:
Updated, I'm a little concerned about other `sh` or `bash` compatible shells out there that will get sent down the ValueError branch. `ksh` comes to mind.

oz123:
Thanks! I merged your commits with one on top!

JoshStern:
Thank you for the help with this! Glad I could contribute 😄

PR #6213, opened 2024-07-29
Remove linuxbrew from path when detecting shell

JoshStern:
### The issue

Installing shells through linuxbrew are always detected as `nu` shells because of the substring check.

### The fix

Split `cmd` by `.linuxbrew` to avoid having it in the check.

This only fixes the specific case of installing things via brew. #6212 offers a broader fix if `cmd` contains other shell names anywhere in the path. It comes with a higher risk of breaking setups though (maybe worth saving for a major release).

### The checklist

* [x] Associated issue: pypa#6197

### DIFF
--- a/news/6197.bugfix.rst
+++ b/news/6197.bugfix.rst
@@ -1,0 +1,1 @@
+Remove linuxbrew from path when detecting shells.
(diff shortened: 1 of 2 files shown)


oz123:
Closing in favour of #6212

PR #6219, opened 2024-08-02
remove all click.style calls from exceptions.py

aidencullo:
### The issue
Click library is large and redundant.

### The fix
replace click.style calls in:

pipenv/exceptions.py

@oz123

### DIFF
--- a/pipenv/exceptions.py
+++ b/pipenv/exceptions.py
@@ -163,12 +163,8 @@ class PipfileNotFound(PipenvFileError):
     def __init__(self, filename="Pipfile", extra=None, **kwargs):
         extra = kwargs.pop("extra", [])
         message = "{} {}".format(
-            click.style("Aborting!", bold=True, fg="red"),
-            click.style(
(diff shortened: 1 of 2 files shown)


matteius (approved):

pipenv/exceptions.py
```diff
@@ -137,14 +137,14 @@ def show(self, file=None):
 
 
 class PipenvFileError(FileError):
-    formatted_message = "{} {{}} {{}}".format(click.style("ERROR:", fg="red", bold=True))
+    formatted_message = "{} {{}} {{}}".format("[bold red]ERROR:[/bold red]")
```
matteius, line 140: I am not sure -- does this work or does it have to be `[bold][red]`?

aidencullo: yes it works like that, it can be either

matteius: Ah ok thanks --sorry I accidentally marked this as ready for review -- feel free to change back.

PR #6215, opened 2024-07-31
Linuxbrew shell fix

oz123:
Add some cosmetic change to the comments.

Basically, all the work was done by @JoshStern.

See #6212 .

PR #6230, opened 2024-09-01
Fix shell detection

oz123:
I introduced a silly bug - with some major consequeces - in #6215.

First, I forgot to include brackets in `endswith`. Second, I added a "/" as if the command should be `/bin/bash` or `/usr/bin/bash`, which in Windows will be '/'. This is actually not needed and detecting the shell based on the last word (the command) without the path, should be fine on all OSs.

### DIFF
--- a/pipenv/shells.py
+++ b/pipenv/shells.py
@@ -40,8 +40,6 @@ def _get_activate_script(cmd, venv):
 
     if cmd.endswith("fish"):
         suffix = ".fish"
-    elif cmd.endswith(("sh", "bash", "zsh")):
-        suffix = ""
     elif cmd.endswith("csh"):
         suffix = ".csh"
     elif cmd.endswith("xonsh"):
@@ -52,6 +50,8 @@ def _get_activate_script(cmd, venv):
     elif cmd.endswith(("pwsh", "powershell")):
         suffix = ".ps1"
         command = "."
+    elif cmd.endswith(("sh", "bash", "zsh")):
+        suffix = ""
     else:
         sys.exit(f"unknown shell {cmd}")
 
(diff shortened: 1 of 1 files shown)


pipenv/shells.py
```diff
@@ -36,19 +36,19 @@ def _get_activate_script(cmd, venv):
     does not work elsewhere anyway.
     """
     # Suffix and source command for various shells.
-    if cmd.endswith("/sh", "/bash", "/zsh"):
+    if cmd.endswith(("sh", "bash", "zsh")):
         suffix = ""
         command = "."
-    elif cmd.endswith("/fish"):
+    elif cmd.endswith("fish"):
```
matteius, line 42: Shouldn't this come before the first conditional since fish ends in sh?

oz123: True! Good catch. The order is a little crazy. Maybe `endswith` is the wrong way .
Pattern matching or Regex might be better?

pipenv/shells.py
```diff
@@ -36,26 +36,34 @@ def _get_activate_script(cmd, venv):
...
-    elif cmd.endswith("/fish"):
+    command = "source"
+
+    if cmd.endswith("fish"):
         suffix = ".fish"
-        command = "source"
-    elif cmd.endswith("/csh"):
+    elif cmd.endswith(("sh", "bash", "zsh")):
+        suffix = ""
+    elif cmd.endswith("csh"):
```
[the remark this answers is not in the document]
matteius: Same for pwsh -- basically I think the check with just "sh" should come  at end.

matteius (approved):

PR #6242, opened 2024-09-14
refinements to make installing vcs from CLI work smoother

matteius:
I ran into this earlier where installing a vcs from CLI yields the wrong  Pipfile entry such that it cannot lock; I iterated on this bug here -- I believe there are reports of this in the backlog as well.

### The checklist

* [ ] Associated issue


matteius:
Unfortunately these code paths are a bit complex and will require more effort to not cause a regression.

oz123 (approved):
LOOKS GOOD. Missing news.

Issue #6185, opened 2024-07-02
[Regression] Running `pipenv uninstall --all` fails when no venv exists


monkeyman192:
### Issue description

We have been running `pipenv uninstall --all` as part of our CI to ensure that the venv is clean before installing and this was working fine until [this PR](https://github.com/pypa/pipenv/pull/6029).
The venv didn't always exist, but the command never failed, so I assumed it was fine.

### Expected result

I would expect `pipenv uninstall --all` to not fail even if there is no venv present (ie. there is nothing to uninstall so all good, as was the previous behaviour)

### Actual result

```
pipenv uninstall --all
Un-installing all [packages]...
Uninstalling six...
Traceback (most recent call last):
  File "/usr/local/bin/pipenv", line 8, in <module>
    sys.exit(cli())
  File "/usr/local/lib/python3.9/dist-packages/pipenv/vendor/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.9/dist-packages/pipenv/cli/options.py", line 58, in main
    return super().main(*args, **kwargs, windows_expand_args=False)
  File "/usr/local/lib/python3.9/dist-packages/pipenv/vendor/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.9/dist-packages/pipenv/vendor/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.9/dist-packages/pipenv/vendor/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.9/dist-packages/pipenv/vendor/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
  File "/usr/local/lib/python3.9/dist-packages/pipenv/vendor/click/decorators.py", line 92, in new_func
    return ctx.invoke(f, obj, *args, **kwargs)
  File "/usr/local/lib/python3.9/dist-packages/pipenv/vendor/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
  File "/usr/local/lib/python3.9/dist-packages/pipenv/vendor/click/decorators.py", line 33, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/usr/local/lib/python3.9/dist-packages/pipenv/cli/command.py", line 290, in uninstall
    retcode = do_uninstall(
[...]

PR #6250, opened 2024-09-22
Fix uninstall --all fails when venv is deleted

Naofal-Helal:
### The issue

This is a regression where `pipenv uninstall --all` fails when the venv does not exist.

### The solution

Add a check to make sure the environment is valid (venv exists).

ref #6185 

* [x] Associated issue

### DIFF
--- a/pipenv/environment.py
+++ b/pipenv/environment.py
@@ -826,4 +826,3 @@ class Environment:
             finally:
                 sys.path = original_path
                 sys.prefix = original_prefix
-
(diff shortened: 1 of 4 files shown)


pipenv/environment.py
```diff
@@ -761,6 +768,8 @@ def run_activate_this(self):
                 code = compile(f.read(), activate_this, "exec")
                 exec(code, {"__file__": activate_this})
 
+    EnvironmentActivationResult = namedtuple("EnvironmentActivationResult", ["ok"])
```
matteius, line 771: I think `namedtuple` may be a bit much here considering  we are really just leveraging it to store a boolean--could we just use the boolean value directly?

matteius (approved):

PR #6253, opened 2024-09-23
Vendor in Pip 24.1.2

matteius:
It will take more work to get to pip 24.2, but I had done a bunch of the patch work in that branch that allows vendoring  in `24.1.2` which brings a lot of improvements with it, so it would be worth bumping this version in a release until the issues I encountered vendoring 24.2 can be sorted.

### The checklist

* [X] Associated issue


oz123 (approved):
As mentioned in our slack there was no need to include pyparsing and I'm glad to see you addressed it.

PR #6275, opened 2024-10-12
Update pipdeptree

oz123:
update pipdeptree to a the latest version.

Somehow these tests passed locally on my systems but on github they failed. The reaon is that on my tests env:
```
openpyxl==2.5.4 [requires: et-xmlfile]
```
and in github:
```
openpyxl==2.5.4 [requires: et_xmlfile]
```
I modified the assertion to catch both, since it's not a blocker for the functionality of graphing the dependencies.


matteius:
Looks good, I'd recommend adding back the LICENSE file and add a news fragment.

oz123:
@matteius News and license added.

matteius (approved):

PR #6311, opened 2024-11-06
Performance minded with some bug fixes

matteius:
### The issue

Testing against the sentry requirements file profiling from a fresh virtualenv creation with no Pipfile/Pipfile.lock:  https://raw.githubusercontent.com/getsentry/sentry/refs/heads/master/requirements-base.txt

Main branch:

image

This branch:

image

### The fix

* Reintroduce Pipfile caching -- carefully and with extra time checks
* Refactor resolver methods to reduce context managers
* cache the paths environment property.
* Solve edge case with markers not be passed through when installing from CLI
* Solve a different markers edge case
* Eliminate some dead (unused) methods

### The checklist

* [ ] Associated issue

### DIFF
--- a/pipenv/environment.py
+++ b/pipenv/environment.py
@@ -78,7 +78,7 @@ class Environment:
         self.prefix = Path(prefix if prefix else sys.prefix)
         self._base_paths = {}
         if self.is_venv:
-            self._base_paths = self.get_paths()
+            self._base_paths = self.get_paths
         self.sys_paths = get_paths()
 
     def safe_import(self, name: str) -> ModuleType:
@@ -180,7 +180,7 @@ class Environment:
             paths = self._base_paths.copy()
         else:
             try:
-                paths = self.get_paths()
+                paths = self.get_paths
             except Exception:
                 paths = get_paths(
                     self.install_scheme,
@@ -257,12 +257,6 @@ class Environment:
 
     @cached_property
     def sys_path(self) -> list[str]:
-        """
-        The system path inside the environment
-
-        :return: The :data:`sys.path` from the environment
-        :rtype: list
-        """
         import json
 
         current_executable = Path(sys.executable).as_posix()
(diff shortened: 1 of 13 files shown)


pipenv/environment.py
```diff
@@ -257,12 +257,6 @@ def python(self) -> str:
 
     @cached_property
     def sys_path(self) -> list[str]:
-        """
```
oz123, line 260: We made a terrible mistake in the python community remove docstrings in favor of annotations.

At list leave us with "The system path inside the environment" so we know that attribute is.

Also, if something is a list, one can consider calling it "sys_paths".

pipenv/project.py
```diff
@@ -689,27 +750,6 @@ def _parse_pipfile(
             # Fallback to toml parser, for large files.
             return toml.loads(contents)
 
-    def _read_pyproject(self) -> None:
```
oz123, line 692: I am not sure why this was introduced, but is it OK to simply remove it?

pipenv/project.py
```diff
@@ -663,11 +667,68 @@ def requirements_location(self) -> str | None:
...
+                # Use non-blocking to prevent deadlocks
+                fcntl.flock(file_obj.fileno(), fcntl.LOCK_EX | fcntl.LOCK_NB)
+                return True
+            except OSError:
+                return False
+
+    def _release_file_lock(self, file_obj):
+        """Release lock on an existing file object"""
+        if sys.platform == "win32":
+            import msvcrt
```
oz123, line 697: This import block is repeated in _acquire_file_lock and in _relase lock. Consider moving to the top.

pipenv/project.py
```diff
@@ -795,25 +835,6 @@ def get_editable_packages(self, category):
...
         return packages
 
-    def _get_vcs_packages(self, dev=False):
-        from pipenv.utils.requirementslib import is_vcs
-
-        section = "dev-packages" if dev else "packages"
-        packages = {
-            k: v
-            for k, v in self.parsed_pipfile.get(section, {}).items()
-            if is_vcs(v) or is_vcs(k)
```
oz123, line 805: How are VCS packages handeled now?

oz123 (changes requested):

PR #6324, opened 2024-12-10
Completely remove click echo

oz123:
Continuing the work to get rid of click.echo ...


matteius:
This has been done essentially in other PRs so I think we can close/delete this branch.

PR #6334, opened 2025-01-09
ci: bump ruff in pre-commit

oz123:
New version of ruff released.

### DIFF
--- a/pipenv/routines/outdated.py
+++ b/pipenv/routines/outdated.py
@@ -60,9 +60,7 @@ def do_outdated(project, pypi_mirror=None, pre=False, clear=False):
                     package_info(package, str(version), str(updated_packages[norm_name]))
                 )
             elif canonicalize_name(package) in outdated_packages:
-                skipped.append(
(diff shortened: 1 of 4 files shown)


pipenv/routines/outdated.py
```diff
@@ -60,7 +60,9 @@ def do_outdated(project, pypi_mirror=None, pre=False, clear=False):
                     package_info(package, str(version), str(updated_packages[norm_name]))
                 )
             elif canonicalize_name(package) in outdated_packages:
-                skipped.append(outdated_packages[canonicalize_name(package)])
+                skipped.append(
+                    outdated_packages[canonicalize_name(package)]  # noqa: PLC0206
```
matteius, line 64: I am confused -- why are the noqa markers being added?   Can we not do both?

matteius (approved):

pipenv/routines/outdated.py
oz123: We can specify this just once. The error reporting is misleading, unfortunately.

PR #6362, opened 2025-03-20
Issue 6348

### LINKED ISSUE #6348: PEP660 editable VCS dependencies not reinstalled correctly
### Issue description

For editable VCS (eg Git) dependencies, re-running `pipenv install` does not reinstall the dependency if the linked source repository checkout is missing (eg if it has been moved or otherwise removed).

### Expected result

For `pipenv install` to reinstall an editable VCS dependency, if the `src/` directory containing the repository checkout has been removed since the last time `pipenv install` was run.

For example, this works with pip:

```Dockerfile
FROM python:3.13
WORKDIR /testcase
RUN pip install -e git+https://github.com/benoitc/gunicorn@23.0.0#egg=gunicorn
RUN python -c 'import gunicorn'
RUN rm -rf src/
RUN pip install -e git+https://github.com/benoitc/gunicorn@23.0.0#egg=gunicorn
RUN python -c 'import gunicorn'
```

### Actual result

Pipenv doesn't reinstall the editable VCS dependency - the `pipenv install` is a no-op:

```
#14 [ 9/10] RUN pipenv install --system --verbose
#14 0.282 Installing dependencies from Pipfile.lock (13dee2)...
#14 DONE 0.3s
```

Which then results in a `ModuleNotFoundError`:

```
#15 [10/10] RUN python -c 'import gunicorn'
#15 0.130 Traceback (most recent call last):
#15 0.132   File "<string>", line 1, in <module>
#15 0.132     import gunicorn
#15 0.132 ModuleNotFoundError: No module named 'gunicorn'
```

### Steps to replicate

Create a `Dockerfile`, with contents:

```Dockerfile
FROM python:3.13
WORKDIR /testcase

COPY <<EOF Pipfile
[[source]]
verify_ssl = true
name = "pypi"

[packages]
gunicorn = {git = "git+https://github.com/benoitc/gunicorn", ref = "23.0.0", editable = true}
EOF

COPY <<EOF Pipfile.lock
{
    "_meta": {
        "hash": {
            "sha256": "0979fd8702b00f90e3e2e93d14919cf23f76b992c80662d6e610dc850d13dee2"
        },
        "pipfile-spec": 6,
        "requires": {},
        "sources": [
            {
                "name": "pypi",
                "verify_ssl": true
            }
        ]
    },
    "default": {
        "gunicorn": {
            "editable": true,
            "git": "git+https://github.com/benoitc/gunicorn",
            "markers": "python_version >= '3.7'",
            "ref": "411986d6191114dd1d1bbb9c72c948dbf0ef0425"
        },
        "packaging": {
            "hashes": [
                "sha256:09abb1bccd265c01f4a3aa3f7a7db064b36514d2cba19a2f694fe6150451a759",
                "sha256:c228a6dc5e932d346bc5739379109d49e8853dd8223571c7c5b55260edc0b97f"
            ],
            "markers": "python_version >= '3.8'",
            "version": "==24.2"
        }
    },
    "develop": {}
}
EOF

RUN pip install pipenv==v2024.4.1
RUN pipenv install --system --verbose

# This works
RUN python -c 'import gunicorn'

# Remove the gunicorn source checkout
RUN rm -rf src/

# Reinstall
RUN pipenv install --system --verbose

# The src dir is missing here since pipenv didn't reinstall the editable VCS dependency
RUN ls -al

# And so this fails with a `ModuleNotFoundError`
RUN python -c 'import gunicorn'
```

Then run it with: `docker build . --progress plain --no-cache`

### Additional findings

If I downgrade the gunicorn version, to one that uses `setup.py` instead of `pyproject.toml` (by changing `ref = "23.0.0"` to `ref = "21.2.0"`) and set `ENV PIP_USE_PEP517=0` to force a legacy `setup.py develop` editable install then this issue no longer reproduces. With the legacy editable install, `site-packages` contains:

```
easy-install.pth
gunicorn.egg-link
```

Whereas when using the PEP517 / PEP660 install (as per the `Dockerfile` above), `site-packages` instead contains:

```
gunicorn-23.0.0.dist-info/
__editable__.gunicorn-23.0.0.pth
__editable___gunicorn_23_0_0_finder.py
```

As such, I've presuming this might be an issue with how Pipenv handles editable VCS dependencies specifically when they use the new PEP660 editable interface?

Is the issue somewhere here?
https://github.com/pypa/pipenv/blob/46bc4ef3b482f1a4fbffee14056780c9c9a8133a/pipenv/environment.py#L730-L761

matteius:
### The issue

Fixes #6348 

### The checklist

* [X] Associated issue

### DIFF
--- a/pipenv/environment.py
+++ b/pipenv/environment.py
@@ -744,6 +744,20 @@ class Environment:
                 )
             if req.link is None:
                 return True
+            elif req.editable and req.link.is_vcs:
+                # For editable VCS dependencies, check if the source directory exists
+                # This ensures we reinstall if the source checkout is missing
+                src_dir = os.path.join(os.getcwd(), "src")
+                if not os.path.exists(src_dir):
+                    return False
+
+                # If we have a specific package directory, check that too
+                if req.name:
+                    pkg_dir = os.path.join(src_dir, req.name)
+                    if not os.path.exists(pkg_dir):
+                        return False
+
+                return True
             elif req.editable and req.link.is_file:
                 requested_path = req.link.file_path
                 if os.path.exists(requested_path):
--- a/tests/integration/test_editable_vcs.py
+++ b/tests/integration/test_editable_vcs.py
@@ -1,0 +1,56 @@
+import os
+import shutil
+from pathlib import Path
+
+import pytest
+
+from pipenv.utils.processes import subprocess_run
+
+
+@pytest.mark.integration
+@pytest.mark.install
+@pytest.mark.editable
+@pytest.mark.vcs
+def test_editable_vcs_reinstall(pipenv_instance_private_pypi):
+    """Test that editable VCS dependencies are reinstalled when the source checkout is missing."""
+    with pipenv_instance_private_pypi() as p:
+        # Create a Pipfile with an editable VCS dependency
+        with open(p.pipfile_path, "w") as f:
+            f.write("""
+[[source]]
+url = "https://pypi.org/simple"
+verify_ssl = true
+name = "pypi"
+
+[packages]
+gunicorn = {git = "https://github.com/benoitc/gunicorn", ref = "23.0.0", editable = true}
+            """.strip())
+
+        # Install the dependency
+        c = p.pipenv("install")
+        assert c.returncode == 0, f"Failed to install: {c.stderr}"
+
+        # Verify the src directory was created
+        src_dir = Path(p.path) / "src"
+        assert src_dir.exists(), "src directory was not created"
+        assert any(src_dir.iterdir()), "src directory is empty"
+
+        # Import the package to verify it's installed correctly
+        c = p.pipenv("run python -c 'import gunicorn'")
+        assert c.returncode == 0, f"Failed to import gunicorn: {c.stderr}"
+
+        # Remove the src directory to simulate the issue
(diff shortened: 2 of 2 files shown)


pipenv/environment.py
```diff
@@ -743,6 +743,20 @@ def is_satisfied(self, req: InstallRequirement):
                     match.version, prereleases=True
                 )
             if req.link is None:
+                return True
+            elif req.editable and req.link.is_vcs:
+                # For editable VCS dependencies, check if the source directory exists
+                # This ensures we reinstall if the source checkout is missing
+                src_dir = os.path.join(os.getcwd(), "src")
```
edmorley, line 750: Should this use `get_src_prefix` so that it works for when running under a venv too?
https://github.com/pypa/pipenv/blob/f1ec23da75c2e03f8109b0147fc647f92de52b8d/pipenv/patched/pip/_internal/locations/base.py#L55-L61

edmorley (commented):
Thank you for taking a look at this!

pipenv/environment.py
matteius: Yes that seems reasonable.

matteius:
Unfortunately none of what I have tried so far has fixed this particular issue.

PR #6365, opened 2025-03-21
Check in work to make safety a module.

matteius:

### The issue

If your pull request makes a significant change to Pipenv, such as the user interface or intended functionality, please open a discussion or issue report first.

### The fix

### The checklist

* [ ] Associated issue


pipenv/routines/check.py
```diff
@@ -109,32 +104,100 @@ def get_requirements(project, use_installed, categories):
 
 
 def create_temp_requirements(project, requirements):
-    temp_requirements = tempfile.NamedTemporaryFile(
-        mode="w+",
-        prefix=f"{project.virtualenv_name}",
-        suffix="_requirements.txt",
-        delete=False,
+    """Create a temporary requirements file that safety can access."""
```
oz123, line 107: This step is not necessary as safety can directly work with pipfiles.
We can remove it in the future.

pyproject.toml
```diff
@@ -43,6 +43,10 @@ optional-dependencies.dev = [
   "sphinx",
   "towncrier",
 ]
+optional-dependencies.safety = [
+  "safety>=3.0.0",
+  "typer>=0.9.0",
```
oz123, line 48: isnt' typer pulled automatically with safety?

tests/integration/test_cli.py
```diff
@@ -180,18 +180,6 @@ def test_pipenv_check(pipenv_instance_private_pypi):
         assert "Ignoring" in c.stderr
 
 
-@pytest.mark.cli
-@pytest.mark.needs_internet(reason="required by check")
-@pytest.mark.parametrize("category", ["CVE", "packages"])
-def test_pipenv_check_check_lockfile_categories(pipenv_instance_pypi, category):
-    with pipenv_instance_pypi() as p:
-        c = p.pipenv(f"install wheel==0.37.1 --categories={category}")
```
oz123, line 188: why remove this test?

oz123 (changes requested):

tests/integration/test_cli.py
matteius: Because its a deprecated command and we don't have an API key to test the --scan with, plus it seems like we are just testing safety internals at this point, but the real reason it was failing ...

oz123 (approved):

PR #6369, opened 2025-03-30
Binary safety

oz123:
Since safety now requires a login and it's breaking the API we used, we can move it to a "plugin" we install.
This PR isn't adding a new scan command. Instead for pipenv users it keeps the "check" interface, but under the hood,
it will call the new scan command.
The new scan command understands Pipefile.lock, so there is no need to export to `requirements.txt` before scanning.
Hence, it directly reads it. Also, the scan command will only work if the user will configure `~/.safety/auth.ini`.
Also, it does not parse safety output, instead it delivers it as is. 

The PR is quite radical and is breaking a lot of things at once,


matteius:
I recommend keeping the check command intact as it was since it still works with the existing database, and support the scan command as an optional flag or via `pipenv scan`.

Issue #5978, opened 2023-10-15
Fresh creation of pipenv creates and fails immediately in Windows


gigatesseract:
Be sure to check the existing issues (both open and closed!), and make sure you are running the latest version of Pipenv.

Check the [diagnose documentation](https://pipenv.pypa.io/en/latest/diagnose/) for common issues before posting! We may close your issue if it is very similar to one of them. Please be considerate, or be on your way.

Make sure to mention your debugging experience if the documented solution failed.

### Issue description

Unable to create a virtualenv using pipenv

### Expected result

To create a virtualenv successfully

### Actual result

RuntimeError

### Steps to replicate

* Create an empty directory
* Install latest pipenv
* Try creating an environment (Eg: pipenv --python 3.10)
* No Pipfile

```
PS C:\Users\nprav\repro-pipenv> pipenv --version       
pipenv, version 2023.10.3
PS C:\Users\nprav\repro-pipenv> virtualenv --version   
virtualenv 20.24.5 from C:/msys64/mingw64/lib/python3.10/site-packages/virtualenv/__init__.py
PS C:\Users\nprav\repro-pipenv> pipenv --python 3.10
Creating a virtualenv for this project...
Pipfile: C:/Users/nprav/repro-pipenv/Pipfile
Using C:/msys64/mingw64/bin/python3.exe (3.10.11) to create virtualenv...
[=   ] Creating virtual environment...created virtual environment CPython3.10.11.final.0-64 in 735ms
  creator CPython3Windows(dest=C:/Users/nprav/.virtualenvs/repro-pipenv-rk4kLvXU, clear=False, no_vcs_ignore=False, global=False)
  seeder FromAppData(download=False, pip=bundle, setuptools=bundle, wheel=bundle, via=copy, app_data_dir=C:/Users/nprav/AppData/Local/pypa/virtualenv)
    added seed packages: pip==23.2.1, setuptools==68.2.2, wheel==0.41.2
  activators BashActivator,BatchActivator,FishActivator,NushellActivator,PowerShellActivator,PythonActivator

Successfully created virtual environment!
Virtualenv location: C:/Users/nprav/.virtualenvs/repro-pipenv-rk4kLvXU
Traceback (most recent call last):
  File "C:/msys64/mingw64/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:/msys64/mingw64/lib/python3.10/runpy.py", line 86, in _run_code
[...]

Issue #5109, opened 2022-05-22
Pipenv searches virtual environment's python binary in wrong path


elegos:
Hello,

I'm having trouble using Pipenv lately. In particular, after I updated my OS (Fedora 36), pipenv during the `install` command tries to search the python binary inside `{local_virtualenv}/local/bin/python`, while the symlink to python is in `{local_virtualenv}/bin/python`. I tried to remove and reinstall the virtual environment (`pipenv --rm` and `pipenv install`), but the problem persists.

Application versions
```sh
➜  etl git:(develop) pip --version
pip 22.1.1 from /home/gfurlan/.local/lib/python3.10/site-packages/pip (python 3.10)
➜  etl git:(develop) pipenv --version
pipenv, version 2022.5.2
➜  etl git:(develop) python --version
Python 3.10.4
➜  etl git:(develop) which python
/usr/bin/python
```

Pipenv install
```sh
➜  etl git:(develop) pipenv install -d
Creating a virtualenv for this project...
Pipfile: /home/gfurlan/src/project/glue/etl/Pipfile
Using /usr/bin/python3.7m (3.7.13) to create virtualenv...
⠙ Creating virtual environment...created virtual environment CPython3.7.13.final.0-64 in 89ms
  creator CPython3Posix(dest=/home/gfurlan/.local/share/virtualenvs/etl-XCf49Mm9, clear=False, no_vcs_ignore=False, global=False)
  seeder FromAppData(download=False, pip=bundle, setuptools=bundle, wheel=bundle, via=copy, app_data_dir=/home/gfurlan/.local/share/virtualenv)
    added seed packages: pip==22.0.4, setuptools==62.1.0, wheel==0.37.1
  activators BashActivator,CShellActivator,FishActivator,NushellActivator,PowerShellActivator,PythonActivator

✔ Successfully created virtual environment!
Traceback (most recent call last):
  File "/home/gfurlan/.local/bin/pipenv", line 8, in <module>
    sys.exit(cli())
  File "/home/gfurlan/.local/lib/python3.10/site-packages/pipenv/vendor/click/core.py", line 1128, in __call__
    return self.main(*args, **kwargs)
  File "/home/gfurlan/.local/lib/python3.10/site-packages/pipenv/cli/options.py", line 56, in main
    return super().main(*args, **kwargs, windows_expand_args=False)
  File "/home/gfurlan/.local/lib/python3.10/site-packages/pipenv/vendor/click/core.py", line 1053, in main
[...]

PR #6373, opened 2025-04-10
Refactor virtualenv `bin/` / `Scripts/` path resolution using `sysconfig` mechanism

ByteB4rb1e:
This pull request refactors the logic for determining the venv's `scripts/`. or `bin/` directory location by leveraging `sysconfig`’s built-in mechanism for retrieving the complete global install directory path of scripts and retrieving its basename. This removes any platform-dependent logic and does not require an additional dependency.

Changes:

    feat: Introduced a new approach for venv scripts path lookup, eliminating reliance on platform identifiers and improving compatibility with POSIX-like environments such as Cygwin and MinGW.

    refactor: Updated both project-wide and routine-specific implementations to use the virtualenv mechanism for dynamic script location lookup, streamlining execution.

These changes mitigate execution inconsistencies in non-native Windows environments and improve portability across platforms.

Fixes:  https://github.com/pypa/pipenv/issues/5978

### DIFF
--- a/pipenv/environment.py
+++ b/pipenv/environment.py
@@ -26,6 +26,7 @@ from pipenv.utils.funktools import chunked, unnest
 from pipenv.utils.indexes import prepare_pip_source_args
 from pipenv.utils.processes import subprocess_run
 from pipenv.utils.shell import temp_environ
+from pipenv.utils.virtualenv import virtualenv_scripts_dir
 from pipenv.vendor.importlib_metadata.compat.py39 import normalized_name
 from pipenv.vendor.pythonfinder.utils import is_in_path
 
@@ -246,16 +247,12 @@ class Environment:
     @property
     def python(self) -> str:
         """Path to the environment python"""
-        if self._python is not None:
-            return self._python
-        if os.name == "nt" and not self.is_venv:
-            py = Path(self.prefix).joinpath("python").absolute().as_posix()
-        else:
-            py = Path(self.script_basedir).joinpath("python").absolute().as_posix()
-        if not py:
-            py = Path(sys.executable).as_posix()
-        self._python = py
-        return py
+        if self._python is None:
+            self._python = (
+                (virtualenv_scripts_dir(self.prefix) / "python").absolute().as_posix()
+            )
+
+        return self._python
 
     @cached_property
     def sys_path(self) -> list[str]:
--- a/news/6737.bugfix.rst
+++ b/news/6737.bugfix.rst
@@ -1,0 +1,22 @@
+# Improved virtualenv scripts path resolution
+
+## Summary
+
+This PR refactors the logic for determining virtual environment script paths
+by leveraging ``sysconfig``'s built-in mechanisms. By removing
+platform-dependent logic, ``pipenv`` now offers enhanced compatibility with
+POSIX-like environments, including Cygwin and MinGW. The fix also mitigates
+execution inconsistencies in non-native Windows environments, improving
+portability across platforms.
+
+## Motivation
+
+The original logic for determining the scripts path was unable to handle the
+deviations of MSYS2 MinGW CPython identifying as ``nt`` platform, yet using a
+POSIX ``{base}/bin`` path, instead of ``{base}/Scripts``.
+
+## Changes
+
+Removed custom logic for determining virtualenv scripts path in favor of
+retrieving the basename of the path string returned by
+``sysconfig.get_path('scripts')```.
--- a/pipenv/project.py
+++ b/pipenv/project.py
@@ -69,6 +69,7 @@ from pipenv.utils.shell import (
     system_which,
 )
 from pipenv.utils.toml import cleanup_toml, convert_toml_outline_tables
+from pipenv.utils.virtualenv import virtualenv_scripts_dir
 from pipenv.vendor import plette, tomlkit
 
 try:
@@ -411,11 +412,19 @@ class Project:
     @property
     def virtualenv_exists(self) -> bool:
         venv_path = Path(self.virtualenv_location)
+
+        scripts_dir = self.virtualenv_scripts_location
+
         if venv_path.exists():
-            if os.name == "nt":
-                activate_path = venv_path / "Scripts" / "activate.bat"
+
+            # existence of active.bat is dependent on the platform path prefix
+            # scheme, not platform itself. This handles special cases such as
+            # Cygwin/MinGW identifying as 'nt' platform, yet preferring a
+            # 'posix' path prefix scheme.
+            if scripts_dir.name == "Scripts":
+                activate_path = scripts_dir / "activate.bat"
             else:
-                activate_path = venv_path / "bin" / "activate"
+                activate_path = scripts_dir / "activate"
             return activate_path.is_file()
 
         return False
@@ -611,6 +620,10 @@ class Project:
             loc = Path(self.project_directory) / "src"
         loc.mkdir(parents=True, exist_ok=True)
         return loc
+
+    @property
+    def virtualenv_scripts_location(self) -> Path:
+        return virtualenv_scripts_dir(self.virtualenv_location)
 
     @property
     def download_location(self) -> Path:
@@ -1422,10 +1435,10 @@ class Project:
     def finders(self):
         from .vendor.pythonfinder import Finder
 
-        scripts_dirname = "Scripts" if os.name == "nt" else "bin"
-        scripts_dir = Path(self.virtualenv_location) / scripts_dirname
         finders = [
-            Finder(path=str(scripts_dir), global_search=gs, system=False)
+            Finder(
+                path=str(self.virtualenv_scripts_location), global_search=gs, system=False
+            )
             for gs in (False, True)
         ]
         return finders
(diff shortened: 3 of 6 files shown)


matteius:
@ByteB4rb1e please `pip install pre-commit` and then run `pre-commit install` in the pipenv dir, and check the lint changes for this PR so the tests can run.

ByteB4rb1e:
Oh, that is unfortunate, that `virtualenv` is not a dependency. My bad, I didn't check the requirements and just assumed `virtualenv` to already be a dependency. I don't want to introduce a new dependency for these small changes. So I'm reconsidering my approach, instead focusing on the built-in `venv`, which was introduced in v3.3, so would pass the *Vendoring* test.

I've tested `venv` in CPython 3.9 locally and am running into a similar issue as `pipenv` currently does, only that the wrong directories are being created. The issue was seemingly [fixed in CPython v3.11](https://github.com/python/cpython/commit/48d926269963cfe7a49c0a4f34af4fe9b832399b), interestingly the [linked issue report](https://github.com/python/cpython/issues/89576) references [`pipenv` Issue #5109 (closed)](https://github.com/pypa/pipenv/issues/5109). I don't yet know what the implications of that are.

A quick workaround that comes to mind would be:

```
>>> import venv
>>> from pathlib import Path
>>> Path(venv.sysconfig.get_path('scripts')).name
'bin'
```

So I will see how the internals of that work and compare it to how `venv::__main__` does the bootstrapping.

@matteius, @oz123, is this approach fine with you?

matteius (approved):

matteius:
@ByteB4rb1e Would be great if we can get a new test to cover this.  Also everything else looked good--please add a news fragment too when you get a chance.

ByteB4rb1e:
@matteius, will do! It'll probably have to wait for a couple of days though as I have to take a brief detour. I'm currently setting up another proposal for refactoring the test suite to work locally without the `run-tests.sh` wrapper script as I currently need a VM and also had to adapt the script to work. Doing that I ran into the same path issues with the `pre-commit` package, so I will file an issue report and create a PR there first.

ByteB4rb1e:
[0017-sysconfig-treat-MINGW-builds-as-POSIX-builds.patch](https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-python/0017-sysconfig-treat-MINGW-builds-as-POSIX-builds.patch) is the MSYS2 specific patch of [CPython's sysconfig](https://github.com/python/cpython/blob/main/Lib/sysconfig/__init__.py) I aim to have test coverage for, as this is the strongest deviation from conventional CPython distributions. The patch is indeed a little quirky, but doesn't provoke any unintended behavior for what we're achieving with this refactoring, as we're utilizing a higher-level facility. Actually quite the opposite: The refactoring ensures, that the logic stays intact, even for other unconventional CPython builds that may not exist yet. Just stumbled upon this to figure out what's best to mock...

Sadly though, there was a [patch](https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-python/0023-sysconfig-mingw-sysconfig-like-posix.patch) introduced which requires special handling in the test case. I've filed a [bug report](https://github.com/msys2/MINGW-packages/issues/23992) and am hopeful that this can be resolved soon!

I will now add parameterization for other platform combinations, then add the news fragment, as requested by @matteius.

matteius:
looks like the tests are failing for a variety of reasons depending on platform and version.

ByteB4rb1e:
Ok, wow... I didn't expect it to be this construed. What I'm taking from this is, while the `sysconfig::get_path` interface is constant, each distributor has different ways of patching things further down along the call tree. I've tested it locally with the embedded python distribution for Windows, MSYS2 MinGW distribution and Ubuntu distribution and they all passed.

It would be possible to define separate test cases for each platform and just ignore non-applicable ones, but that doesn't sit quite right with me as this wouldn't address testing any future deviations by distributors.

I think I will push through with this, trying to analyze the deviations and make the test case work as I intended. It will take some time though as I first have to identify the distributions in use by the CI test and come up with a faster testing setup for debugging.

Bear with me, please. I'm hoping to get it done before the next release.

matteius:
> which actually do require the separation of test cases

We have this pattern with `pytest` markers, so you can specific which OS a test should run on.   That being said, as you pointed out, I've run into it too--the vendor implementations of sysconfig are always different, so relying on private `_` methods seems a bit risky.

ByteB4rb1e:
You're right. I got carried away. I'm just trying to grasp the history of how this issue arose in the first place. But this is the wrong circumstance to do that. Mocking private API functions is a testament to that. Pardon...
There are now two separate test cases, authored like all the others, with the exception that I've added a secondary condition for skipping the test, so that they properly execute within MinGW CPython as well.

I'll create the news fragment now and then make sure the changes actually fixed the original problem.

ByteB4rb1e:
I've tested all interactive commands manually, they're working fine. In addition, fixed another fault during the uninstall routine, and fixed the faulty path in the Windows test case. All is seemingly working now. Don't know though why ruff didn't trigger locally...

```
$> cat /proc/version
MSYS_NT-10.0-26100 version 3.5.7-2644508f.x86_64 (runneradmin@fv-az1853-986) (gcc version 13.3.0 (GCC) ) 2025-01-30 09:08 UTC
$> which python3 && python3 -V && pacman -Q mingw-w64-x86_64-python
/mingw64/bin/python3
Python 3.12.9
mingw-w64-x86_64-python 3.12.9-1
$> test -d .venv || echo $?
1
$> python3 -m venv .venv
$> .venv/bin/pip install git+https://github.com/ByteB4rb1e/pipenv.git@bugfix/5978 2>&1 | tee pip-install-pipenv.txt
...
Successfully installed certifi-2025.1.31 distlib-0.3.9 filelock-3.18.0 packaging-24.2 pipenv-2024.4.1 platformdirs-4.3.7 setuptools-78.1.0 virtualenv-20.30.0

[notice] A new release of pip is available: 24.3.1 -> 25.0.1
[notice] To update, run: C:/Users/me/Repositories/local/pipenv-debug/.venv/bin/python3.exe -m pip install --upgrade pip
$> .venv/bin/pipenv install cowsay && .venv/bin/cowsay -t works
Installing cowsay...
Installation Succeeded
Installing dependencies from Pipfile.lock (b54bd8)...
Upgrading cowsay in  dependencies.
Building requirements...
Resolving dependencies...
Success!
Building requirements...
Resolving dependencies...
Success!
Installing dependencies from Pipfile.lock (c93707)...
Installing dependencies from Pipfile.lock (c93707)...
  _____
| works |
  =====
     \
      \
        ^__^
        (oo)\_______
        (__)\       )\/\
            ||----w |
            ||     ||
$> .venv/bin/pipenv uninstall cowsay && .venv/bin/cowsay || echo 'removed'
Removed cowsay from Pipfile.
Uninstalling cowsay...
Found existing installation: cowsay 6.1
Uninstalling cowsay-6.1:
  Successfully uninstalled cowsay-6.1

-bash: .venv/bin/cowsay: No such file or directory
removed
```

[pip-install-pipenv.txt](https://github.com/user-attachments/files/19782790/pip-install-pipenv.txt)

pipenv/environment.py
```diff
@@ -248,14 +249,8 @@ def python(self) -> str:
...
-        if os.name == "nt" and not self.is_venv:
-            py = Path(self.prefix).joinpath("python").absolute().as_posix()
-        else:
-            py = Path(self.script_basedir).joinpath("python").absolute().as_posix()
-        if not py:
-            py = Path(sys.executable).as_posix()
-        self._python = py
-        return py
+
+        return (virtualenv_scripts_dir(self.prefix) / "python").absolute().as_posix()
```
matteius, line 253: I am actually surprised this doesn't break the `--system` flag tests because my understanding is this property would also be used for `--system` based installs which wouldn't be using a virtualenv.   Have you considered that angle at all or could do some testing of the `--system` flag on cli?

ByteB4rb1e: https://github.com/pypa/pipenv/blob/29b4197693ad6d02bfe6c5b8034189ae345d35de/pipenv/environment.py#L247-L257

As far as I comprehended, the condition in line 255 would never be met, as the `joinpath('python')` in line 254 guarantees that `py` will never become a false-equivalent value, since it's definition is in the else statement, hence catches the unmatched conditional in line 251. That's why I disregarded it altogether.
I assumed `self.script_basedir` and `self.prefix` to be synonyms? They returned the same value in a Ubuntu VM. That's why I applied `virtualenv_scripts_dir` there.

I will see if I can execute with the `--system` flag in an embedded Python distribution for Windows, but honestly didn't consider it. In MSYS2 Python package management is handled by `pacman` and on Ubuntu by `apt-get`/`aptitude`, so I assumed that any sort of global options wouldn't work anyway with my setup.

ByteB4rb1e: Rebased the commit, assigning to `self._python` now and also inverted the first condition as to not require an intermediate assignment to `py`. Honestly, I hope this will break any test case. Otherwise I think there's a test case missing for checking the `--system` case @matteius mentioned. I couldn't provoke an error using the `--system` flag.

matteius (approved):
Thanks for your hard work on this change!

Issue #6409, opened 2025-05-21
Installing an editable package can result in using the wrong canonical package name in Pipfile


stewartmiles:
### Issue description

Create a local Python package called `mypackage` with a `mypackage/mypackage/tests/test_setup.py` file (see [mypackage.zip](https://github.com/user-attachments/files/20375752/mypackage.zip)):

```python
import unittest

class TestNothing(unittest.TestCase):
    def setup(**kwargs):
        pass

    def test_nada(self):
        _ = setup(name='foo')
```
 
* Create a pipenv virtual environment using say
  `( export PIPENV_NO_INHERIT=1; pipenv run echo )`
* Install the package as editable using `pipenv install -e ./mypackage`
* See the package installed in the Pipfile uses the name in inferred from test file _not_ the top level `setup.py`, i.e: `grep mypackage Pipfile` shows `foo = {file = "mypackage"}` not `mypackage = {file = "mypackage"}` as expected.

### Expected result

The package should be named correctly in Pipfile.

### Actual result

The package name is incorrectly parsed from a test file.

### Steps to replicate

See above.

-------------------------------------------------------------------------------

Please run `$ pipenv --support`, and paste the results here. Don't put backticks (`` ` ``) around it! The output already contains Markdown formatting.

<details><summary>$ pipenv --support</summary>

Pipenv version: `'2025.0.2'`

Pipenv location: `'/home/stewart/.local/share/pipx/venvs/pipenv/lib/python3.10/site-packages/pipenv'`

Python location: `'/home/stewart/.local/share/pipx/venvs/pipenv/bin/python'`

OS Name: `'posix'`

User pip version: `'25.0.1'`

user Python installations found:

  - `3.10.12`: `/usr/bin/python3`
  - `3.10.12`: `/usr/bin/python`
  - `3.10.12`: `/bin/python3`
  - `3.10.12`: `/bin/python`
  - `3.8.20`: `/usr/bin/python3.8`
  - `3.8.20`: `/bin/python3.8`

PEP 508 Information:

```
{'implementation_name': 'cpython',
 'implementation_version': '3.10.12',
 'os_name': 'posix',
 'platform_machine': 'x86_64',
 'platform_python_implementation': 'CPython',
 'platform_release': '6.8.0-59-generic',
 'platform_system': 'Linux',
 'platform_version': '#61~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Tue Apr 15 '
                     '17:03:15 UTC 2',
[...]

PR #6411, opened 2025-05-22
Fix derivation of local package names in Pipfile

stewartmiles:
Fixes https://github.com/pypa/pipenv/issues/6409

### The issue

Fixes https://github.com/pypa/pipenv/issues/6409

### The fix

Iterates parent directories searching for package metadata files (e.g setup.py etc.) before leaf directories within a package.

### DIFF
--- a/pipenv/utils/dependencies.py
+++ b/pipenv/utils/dependencies.py
@@ -684,9 +684,10 @@ def find_package_name_from_directory(directory):
             directory_path = Path(directory_str[1:])
 
     try:
-        # Sort contents - directories first, then files
+        # Sort contents - files first, then directories to search parent
+        # directories before leaf directories.
         directory_contents = sorted(
-            directory_path.iterdir(), key=lambda x: (x.is_file(), x.name)
+            directory_path.iterdir(), key=lambda x: (x.is_dir(), x.name)
         )
 
         for path in directory_contents:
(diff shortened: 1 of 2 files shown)


stewartmiles:
It looks like a load of tests failed in https://github.com/pypa/pipenv/actions/runs/15193270143/job/42735598641 . I wasn't able to run all tests successfully locally either with or without this patch. Is this expected?

matteius:
Thank you @stewartmiles  for taking a look at this issue -- the determine package name logic was some one off stuff that was partly my work around and partly early Claude/Chat GPT help when I was working the conversion off requirementslib -- the prior code in requirements lib was very hard to follow how it arrived at the package name for these local file cases.   There has to be a better way.

stewartmiles:
@matteius AFAIK when installing a local package from a folder it's possible to get the name using the same approach as pip / requirementslib, i.e using the new [PEP-517](https://peps.python.org/pep-0517/#prepare-metadata-for-build-wheel) equivalent of the  `egg_info` setuptools command to build the package metadata and the read the name from that but unfortunately the hook to get package metadata is optional.

A more expensive, though simple alternative is to build a wheel with pip and extract the package information using the build wheel hook in PEP-517. Here's an example using pip's command line interface, using the package attached to https://github.com/pypa/pipenv/issues/6409 doing something like this:

```bash
# Build a wheel into mypackage_build
pip wheel --no-deps ./mypackage -w mypackage_build
# Read the name from the package metadata.
unzip -p mypackage_build/* *METADATA | awk '$1 ~ /Name:/ { print $2 }'
```

which prints `mypackage`

matteius:
@stewartmiles Can you take a look at:  https://github.com/pypa/pipenv/pull/6413/files

I basically fed the patch file of this PR, the issue report and comments, and the test failures into Augment to correct for the edge cases, and I appear to be getting a passing build.

PR #6416, opened 2025-06-04
Fix editable installation from relative paths on Windows

stewartmiles:
When running `pipenv install -e .\..\mypackage` the pip requirement line `-e .\..\mypackage` is written to a temporary requirements file which is passed to pip to install the package. pip fails to install the package from the requirements file as it incorrectly interprets the relative path, treating the slashes as escapes. To work around this issue replace all back slashes (windows directory separators) with forward slashes (POSIX directory separators) in pip lines for editable package installs.

### The issue

If your pull request makes a significant change to Pipenv, such as the user interface or intended functionality, please open a discussion or issue report first.

### The fix

### The checklist

* [ ] Associated issue


stewartmiles:
Oh things still break when locking as the Windows style path is written to `Pipfile`, I'll fix that instead / as well. Until then hold off on reviewing this.

stewartmiles:
> Oh things still break when locking as the Windows style path is written to `Pipfile`, I'll fix that instead / as well. Until then hold off on reviewing this.

I had to fiddle with a few more code paths to get this to work when locking after installing a package from the parent directory of the configured pipenv venv directory.

PR #6417, opened 2025-06-09
chore: remove tomli from vendor

oz123:
Instead, use the tomli from bundeled pip.

Pretty small and quick to review ;-)


matteius:
I seem to recall there needing to be a change to the vendor script to support this better during revendoring.

oz123:
> I seem to recall there needing to be a change to the vendor script to support this better during revendoring.

Indeed, tomli search and replace was part of the vendor script, but it is no longer needed as we removed the library from out vendor directory.
As I already touched it now, I pushed a few more commits doing some more cleanup of this script.

matteius (approved):

PR #6418, opened 2025-06-09
chore: replace importlib python3.9 compact

oz123:
If this works, we can remove the vendored importlib backport.
On my local machine, 3 tests fail seemingly randomly, so testing CI.


oz123:
Needs revisiting when we remove python3.9 at the end of the year.

PR #6425, opened 2025-06-25
Bugfix: Make sure ensure_python() installs Python when conditions are met

EpiX-1:
### The issue

The issue arises when a user tries to create a new venv with a version of Python that hasn't been installed PREVIOUSLY with pyenv. Normally, pipenv should handle the installation of the desired Python version thanks to an installer (pyenv) before creating the new venv. However, this is not the case, and pipenv creates a new venv with the same Python version as the system one. 

This behavior happens whether or not the conditions for the installation are met:
- `PIPENV_YES` is set
- The user replied `y` when prompted: `"Would you like us to install {version} with {installer}?"`

### Expected behavior

The Python version of the .venv should match the version specified in the Pipfile.

### Actual behavior

The Python version of the .venv is the same as the one from the system.

### Reproduce the issue

1. Install pipenv with `pip install pipenv`. For the rest of the steps, it is assumed that the system Python version is 3.13.
2. Create the following Pipfile:
```
[[source]]
verify_ssl = true
name = "pypi"

[packages]

[dev-packages]

[requires]
python_version = "3.8"

```
3. Create the venv with the command: `pipenv install`
4. Notice that the venv created uses the system Python version (3.13) rather than the one specified in the Pipfile (3.8)

### The fix

Remove a useless else statement that would cause the program to exit abruptly.


matteius:
@EpiX-1 could you install pre-commit and run `pre-commit install` or otherwise fix the ruff linter issue(s)?  Ill be taking a look in more detail soon.

matteius:
Actually the ruff issues aren't related to your change -- I am investigating them.

matteius (approved):

matteius:
@EpiX-1 would you be able to add a news fragment though?

EpiX-1:
> @EpiX-1 would you be able to add a news fragment though?

Sure, I'll do it tomorrow.

oz123 (approved):

PR #6432, opened 2025-08-01
Add exclude_versions support to Pipfile and resolver

ghost:
This pull request adds support for the exclude_versions field in the Pipfile and updates the resolver logic accordingly.

With this feature, users can specify versions of a package that should be excluded from installation. This is useful for skipping problematic or broken releases without locking dependencies to an old version.

**Example Pipfile usage:**
[packages]
tracerite = {version = "*", exclude_versions = ["1.1.2", "1.1.3"]}

- The resolver will now automatically add '!=<version>' constraints for any versions listed in exclude_versions.
- This helps keep dependencies up-to-date while avoiding specific problematic versions.

Let me know if you need any changes or improvements!

### DIFF
--- a/pipenv/exclude.py
+++ b/pipenv/exclude.py
@@ -1,0 +1,35 @@
+import toml # pyright: ignore[reportMissingModuleSource]
+
+def read_pipfile(path="pipenv/Pipfile"):
+    with open(path, "r", encoding="utf-8") as f:
+        return toml.load(f)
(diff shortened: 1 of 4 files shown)


matteius:
One problem here is your modifying packaging which is in the vendor dir -- so any changes to those packages get wiped out when we upgrade vendor'ed packages.

PR #6441, opened 2025-10-15
Swap pip-args with  extra-pip-args

agalazis:
pip-args should be extra-pip-args

### The issue

I followed  the instructions to speedup my build.  pip-args is not accepted with a cryptic error `{}`

### The fix

I have found through another doc page that extra-pip-args is the flag:
https://pipenv.pypa.io/en/latest/advanced.html
  
### The checklist

* [ ] Associated issue


matteius:
@agalazis Please add a news fragment, and thank you.

PR #6506, opened 2025-12-19
Fix MD5 usage to support FIPS-enabled Python environments

Aditevil:
In FIPS mode, md5 might be available and approved. It might also be
available, but only on opt-in basis for unapproved usage. And very
strict systems might not have md5 even on opt-in basis.

Python has API to expose this as "usedforsecurity=False" argument, see
python documentation.

The rolling document checksum is not used for cryptograpic protection,
but rather used out of convenience. Hence allow using MD5 on more FIPS
systems.

This is no effective change for regular non-fips python builds.

### DIFF
--- a/pipenv/utils/resolver.py
+++ b/pipenv/utils/resolver.py
@@ -751,7 +751,7 @@ def _generate_resolution_cache_key(
     ]
 
     key_string = "|".join(key_components)
-    return hashlib.md5(key_string.encode()).hexdigest()
+    return hashlib.md5(key_string.encode(), usedforsecurity=False).hexdigest()
 
 
 def _should_use_resolution_cache(cache_key, clear):


matteius:
Sorry for my ignorance, what is a FIPS enabled system and why should it prefer md5 over something modern and secure?

Aditevil:
> Sorry for my ignorance, what is a FIPS enabled system and why should it prefer md5 over something modern and secure?

@matteius 

[FIPS](https://www.nist.gov/standardsgov/compliance-faqs-federal-information-processing-standards-fips) is a U.S. government standard that defines which cryptographic modules and algorithms are allowed to be used. On systems running in FIPS mode, only approved crypto algorithms are permitted. 
Importantly, FIPS does not “prefer” MD5 it actually restricts it, because MD5 is considered cryptographically broken.
Pipenv uses MD5, but only for a non-security purpose and Python provides a `usedforsecurity=False` flag to handle exactly this case.

PR #6511, opened 2026-01-23
clarify path dependency default install behavior

Darshangouda55:

### The issue

The documentation example for local path dependencies suggested editable = true, which can be misleading since path installs are non-editable by default unless explicitly marked editable.

### The fix

Updates the example to remove editable = true so the default behavior is clear.

### The checklist

* [ ] Associated issue


oz123:
@Darshangouda55 thank you for you contribution. Can you please fix the linting issue?

Darshangouda55:
Thanks! I’ve fixed the linting issue by adding the missing newline. Please let me know if anything else is needed.

Issue #6568, opened 2026-03-24
Virtualenv creation fails when current directory contains a file that shadows a module used by virtualenv


FlavioAmurrioCS:
### Issue description

If user has a file in the current directory that shadows a module used by virtualenv it will cause the virtualenv creation to fail.

I did some digging and it seems that because pipenv calls virtualenv with `python -m virtualenv ...` that python's default behavior is to add the current directory to the PYTHONPATH
https://github.com/pypa/pipenv/blob/main/pipenv/utils/virtualenv.py#L122-L126
After some googling it seems we could just add `-P` to the command and that should fix it. (Will try and if it works will raise a PR)

### Expected result

virtualenv should be created regardless of what files may be present in current directory

### Actual result

virtualenv creation fails
Check next section for details.

### Steps to replicate

```bash
[flavio@Mac ~]
$ mkdir -p /tmp/foo

[flavio@Mac ~]
$ cd /tmp/foo

[flavio@Mac /tmp/foo]
$ echo 'import requests' > secrets.py

[flavio@Mac /tmp/foo]
$ pipenv install certifi
Creating a virtualenv for this project
Pipfile: /private/tmp/foo/Pipfile
Using /Users/flavio/Applications/brew/bin/python33.14.2 to create virtualenv...
⠙ Creating virtual environment...ModuleNotFoundError: No module named 'requests'

✘ Failed creating virtual environment

Failed to create virtual environment.
```
-------------------------------------------------------------------------------

Please run `$ pipenv --support`, and paste the results here. Don't put backticks (`` ` ``) around it! The output already contains Markdown formatting.

If you're on macOS, run the following:

    $ pipenv --support | pbcopy

If you're on Windows, run the following:

    > pipenv --support | clip

If you're on Linux, run the following:

    $ pipenv --support | xclip

FlavioAmurrioCS:
btw, issues was alway there but I just happen to notice because multiple of out lambdas have a secrets.py file
Virtualenv recently extracted a module into a self contained package https://github.com/pypa/virtualenv/commit/f89d46c0883334ca9f2f7eb3e2339457e7a44673 
Which just happens to to import secrets https://github.com/tox-dev/python-discovery/blob/main/src/python_discovery/_cached_py_info.py#L10 which before it being extracted it didnt do.

I was able to fix secrets.py issue by installing pipenv with a virtualenv before this python-discovery module but decided to contribute a proper fix.

PR #6569, opened 2026-03-24
Run virtualenv from a empty directory to prevent current directory from polluting PYTHONPATH

FlavioAmurrioCS:
### The issue

Issue: https://github.com/pypa/pipenv/issues/6568

TLDR: If a user has a file in the current directory that shadows a module used by virtualenv it will make the virtualenv creation fail

### The fix

When calling a module with 'python -m <module>' python's default behavior is to prepend the current directory to the PYTHONPATH which if there is a file with the same name as library/module use by the module we are trying to invoke, it will cause an error. To avoid this issue python has a `-P` flag that prevent this behavior.

```
-P     : don't prepend a potentially unsafe path to sys.path; also PYTHONSAFEPATH
```

 Example of the solution working
```bash
[flavio@Mac ~]
$ mkdir -p /tmp/foo

[flavio@Mac ~]
$ cd /tmp/foo

[flavio@Mac /tmp/foo]
$ echo 'import requests' > secrets.py

[flavio@Mac /tmp/foo]
$ pipenv install certifi
Creating a virtualenv for this project
Pipfile: /private/tmp/foo/Pipfile
Using /Users/flavio/Applications/brew/bin/python33.14.2 to create virtualenv...
⠦ Creating virtual environment...created virtual environment CPython3.14.2.final.0-64-arm64 in 488ms
  creator CPython3Posix(dest=/Users/flavio/.virtualenvs/foo-mzLZXOua, clear=False, no_vcs_ignore=False,
global=False)
  seeder FromAppData(download=False, pip=bundle, via=copy,
app_data_dir=/Users/flavio/Library/Caches/virtualenv)
    added seed packages: pip==26.0.1
  activators
BashActivator,CShellActivator,FishActivator,NushellActivator,PowerShellActivator,PythonActivator

Virtualenv location: /Users/flavio/.virtualenvs/foo-mzLZXOua
Locking  dependencies...
Locking  dependencies...
Updated Pipfile.lock (be43225d7d56471fd9b60269c82252cc19448eb3578abebecdcd914e2213171e)!
Installing certifi...
✔ Installation Succeeded
Installing dependencies from Pipfile.lock (13171e)...
Upgrading certifi in  dependencies.
Building requirements...
Resolving dependencies...
✔ Success!
Installing dependencies from Pipfile.lock (c5a106)...
Installing dependencies from Pipfile.lock (c5a106)...
```

### The checklist

* [x] Associated issue

### DIFF
--- a/news/6569.bugfix.rst
+++ b/news/6569.bugfix.rst
@@ -1,0 +1,1 @@
+Run python -m virtualenv from empty directory to prevent PYTHONPATH pollution. Fixes #6568.
--- a/pipenv/utils/virtualenv.py
+++ b/pipenv/utils/virtualenv.py
@@ -10,6 +10,7 @@ from pipenv import environments, exceptions
 from pipenv.utils import Confirm, console, err
 from pipenv.utils.dependencies import python_version
 from pipenv.utils.environment import ensure_environment
+from pipenv.utils.fileutils import create_tracked_tempdir
 from pipenv.utils.processes import subprocess_run
 from pipenv.utils.shell import find_python, shorten_path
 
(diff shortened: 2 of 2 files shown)


matteius:
Hmmm, we need to do something differently for 3.10, maybe upgrade virtualenv version in pyproject.tomo?  I'm not sure

FlavioAmurrioCS:
Oh, I see `-P` was added in 3.11
https://docs.python.org/3/using/cmdline.html#cmdoption-P

The only other option I see is to run the subcommand from a cwd that is known to be empty, we can create a temporary directory. Let me try that really quick.

matteius:
or do a check on the os python version maybe?

FlavioAmurrioCS:
I have updated the PR. I am now using a temporary directory which will work for all versions without any conditional logic.

The other approach I had in mind was to call virtualenv directly instead of using `python -m virtualenv` but that will required to find the executable from the virtualenv where pipenv is installed in which is not as simple as the approach in this PR.

FlavioAmurrioCS:
tempfile.TemporaryDirectory was having issues on windows, but it seems that using the existing utils function `create_tracked_tempdir` works


Here is a run on my fork (should have done this sooner) https://github.com/FlavioAmurrioCS/pipenv/actions/runs/23472575258/job/68298369586?pr=6

FlavioAmurrioCS:
Could you re run the fail job? wondering if it a flaky test.

PR #6620, opened 2026-03-28
ci: consolidate linting and optimize test matrix

oz123:
CI optimizations - reduce waste of compute resources....

 Linting (pre-commit)                                                                                                                                                           
  - Black is replaced by ruff-format — one less tool, same formatting enforcement                                                                                                
  - The redundant standalone ruff CI job is dropped — ruff was already running via pre-commit                                                                                                                                                                   
                                                                                                                                                                                 
  Test execution order                                                                                                                                                           
                                                                                                                                                                                 
  Before: all 15 jobs (5 versions × 3 OSes) launched in parallel immediately after lint.                                                                                         
                                                                                                                                                                                 
  After:                                                                                                                                                                         
  1. tests-smoke — 3 jobs (3.12 × all OSes) run first
  2. Only if all 3 pass, the remaining 12 jobs (4 versions × 3 OSes) start                                                                                                       
                                                                          
  The win: if a bug is version-agnostic (the common case), it surfaces in 3 jobs instead of 15. The other 12 jobs never run, saving ~80% of compute for those failures. If the   
  bug is truly version-specific, the smoke still passes and the full matrix catches it.                                                                                          
                                                                                                                                                                                 
  The trade-off: when everything passes, total wall-clock time is slightly longer because the 12-job matrix now waits for smoke to finish before starting. But resource usage is 
  the same in the green path.
  
Windows Defender real-time scanning intercepts every file read/write during Python installation, package downloads, virtualenv creation, and test execution. Excluding these   
  paths means the antivirus scanner no longer blocks on each file operation, which can cut Windows job times significantly — typically 20-40% faster depending on how many       
  packages are installed and how many files tests touch. No security risk in CI since the runner is ephemeral and discarded after each job.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -76,6 +76,14 @@ jobs:
         os: [MacOS, Ubuntu, Windows]
     steps:
       - uses: actions/checkout@v4
+      - name: Disable Windows Defender for Python paths
+        if: contains(matrix.os, 'Windows')
+        shell: powershell
+        run: |
+          Add-MpPreference -ExclusionPath "C:\hostedtoolcache\windows\Python"
+          Add-MpPreference -ExclusionPath "$env:GITHUB_WORKSPACE"
+          Add-MpPreference -ExclusionPath "$env:USERPROFILE\.virtualenvs"
+          Add-MpPreference -ExclusionPath "$env:LOCALAPPDATA\pip"
       - name: Set up Python 3.12
         uses: actions/setup-python@v5
         with:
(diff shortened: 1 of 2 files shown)


matteius:
I'm not convinced that we have parity between the rulesets -- I found many times last night where the ruff and/or black had different errors, one passed and not the other.

oz123:
So the ruff hook is a legacy alias for ruff check — it runs ruff check --force-exclude under the hood. Since we removed the --fix and --exit-non-zero-on-fix args, our         
  pre-commit config now runs:                                                                                                                                                    
                                                                                                                                                                                 
  ruff check --force-exclude                                                                                                                                                     
                                                                                                                                                                                 
  The removed CI job ran:                                                                                                                                                        
  ruff check --output-format=github .                                                                                                                                            
                                                                                                                                                                                 
  Effectively the same — both just check and report errors without touching files. The only differences are:                                                                     
  - --force-exclude (pre-commit) vs no flag: pre-commit respects the exclude patterns in pyproject.toml even for explicitly passed files                                         
  - --output-format=github was CI-specific to render annotations in the GitHub PR UI — pre-commit doesn't need that    
  
So the behavior is the same. 

Earlier the pre-commit also ran `ruff-format` which indeed had a different behavior. I think that explains what you saw.

PR #6680, opened 2026-06-09
Pip 26.1.2

matteius:

### The issue

Routine vendoring refresh for patched `pip` and Pipenv's vendored helper libraries.

### The fix

This PR updates patched `pip` to `26.1.2` and refreshes the vendored source for the related version bump.

It also updates Pipenv's vendored libraries:

* `pythonfinder` to `3.0.4`
* `packaging` to `26.2`
* `pipdeptree` to `3.1.0`
* `tomlkit` to `0.15.0`

The vendoring patches were adjusted for the refreshed upstream sources, including the patched `pip` tarfile-link-safety context and the `pipdeptree` `__main__.py` import/path bootstrap patch for the `3.1.0` layout.

### The checklist

* [ ] Associated issue: N/A, routine vendoring update
* [ ] News fragment: not added


pipenv/vendor/vendor.txt
```diff
@@ -1,9 +1,9 @@
-packaging==26.0
+packaging==26.2
```
oz123, line 1: packaging is a dependecy of new pipdeptree, and should be declared as such.

```
pipdeptree==3.1.0
    packaging==26.2
```

But there is a bigger issue with vendorng packaging here. We already ship a copy of packaging in :

`patched/pip/_vendor`

And it is the exact same version. Can we use that instead?

oz123 (changes requested):
Not sure about what copilot is saying, but the packaging vendor needs a change.

matteius (commented):
Thanks for the review, @oz123!

** fix:** Updated to use  (same pattern as  below it) in both the live  and the vendoring patch, so future pip refreshes carry the same fix.

** deduplication:** I investigated whether  can share with . The two copies differ *only* in namespace references — intra-package imports and a few docstrings get rewritten by the vendoring process ( vs ). Merging them would require redirecting all pipdeptree's packaging imports to pip's internal vendor path, which couples unrelated vendoring namespaces and would be fragile to maintain.

In the interest of keeping this a clean routine-vendoring PR, I've:

  1. Added a  annotation in  to document the dependency relationship.
  2. Left deduplication as a potential follow-up (it would need a dedicated approach, e.g. shim re-exports or a vendoring rewrite rule).
  
  Let me know if you'd prefer a different approach on the deduplication front.

pipenv/vendor/vendor.txt
matteius: Done.

oz123:
I still feel like we should not ship two copies of packaging. The import inside patched pip from our vendor is especially awkward.

But it is the master branch. Not a release, so as a stop gap solution we can accept it for now.

matteius:
Well after trying twice to resolve your comment, and claude and codex being unsuccesful at understanding it I guess, I decided to just merge it to get the security fixes in, the duplication can definitely be revisited.  CC @oz123

